PR #3560, opened 2023-03-11
`<mdspan>`: Require concepts

JMazurkiewicz:
This PR makes concepts mandatory for `<mdspan>`.

* Replace SFINAE with concepts in `std::extents` class,
* Address @StephanTLavavej's comments from my previous PR (#3535):
  * `extends` -> `extents` (obvious typo)
  * Improve `static_assert` messages (https://github.com/microsoft/STL/pull/3535#discussion_r1125485604).

I'm going to replace the rest of uses of SFINAE with concepts in follow-up PR.

### DIFF
--- a/stl/inc/mdspan
+++ b/stl/inc/mdspan
@@ -8,9 +8,9 @@
 #define _MDSPAN_
 #include <yvals.h>
 #if _STL_COMPILER_PREPROCESSOR
-#if !_HAS_CXX23
+#if !_HAS_CXX23 || !defined(__cpp_lib_concepts) // TRANSITION, GH-395
 _EMIT_STL_WARNING(STL4038, "The contents of <mdspan> are available only with C++23 or later.");
-#else // ^^^ !_HAS_CXX23 / _HAS_CXX23 vvv
+#else // ^^^ not supported / supported language mode vvv
 #include <algorithm>
 #include <array>
 #include <span>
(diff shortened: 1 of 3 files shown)


stl/inc/mdspan
```diff
@@ -8,9 +8,9 @@
 #define _MDSPAN_
 #include <yvals.h>
 #if _STL_COMPILER_PREPROCESSOR
-#if !_HAS_CXX23
+#if !_HAS_CXX23 || !defined(__cpp_lib_concepts) // TRANSITION, GH-395
 _EMIT_STL_WARNING(STL4038, "The contents of <mdspan> are available only with C++23 or later.");
```
SuperWig, line 12: Shouldn't this also be updated to say "with concepts support" like `<concepts>` and similarly `<source_location>`

SuperWig (commented):
(urgh GitHub app can't leave a single comment)

stl/inc/mdspan
StephanTLavavej: Good catch, but perhaps surprisingly, the variation about mentioning concepts doesn't really matter. This is because we have enabled concepts when we specifically detect IntelliSense in the IDE (as opposed to our command-line only use of `/BE`) - this is by special agreement with the IDE team that they won't complain about any bugs encountered, because IntelliSense doesn't need to fully instantiate templates in order to provide most of its functionality. So in practice, users don't experience "C++23 without concepts" mode, only we do in the test suite.

StephanTLavavej: Exact precedent: https://github.com/microsoft/STL/blob/daa994bfc41c36196c536f2b68388f859d6bd656/stl/inc/expected#L11-L13

StephanTLavavej (approved):

PR #3566, opened 2023-03-13
`_Weakly_unwrappable` should use `_Allow_inheriting_unwrap_v`

CaseyCarter:
... so as not to require `_Prevent_inheriting_unwrap`; the intent of the unwrapping design is that `_Prevent_inheriting_unwrap` is optional.

I'm willing to be convinced that `_Prevent_inheriting_unwrap` should be made mandatory, in which case we should still accept this change and also make changes to `_Allow_inheriting_unwrap_v` to maintain consistency between Ranges and non-Ranges unwrapping machinery.

### DIFF
--- a/stl/inc/xutility
+++ b/stl/inc/xutility
@@ -2287,9 +2287,8 @@ namespace ranges {
     using sentinel_t = decltype(_RANGES end(_STD declval<_Rng&>()));
 
     template <class _Wrapped>
-    concept _Weakly_unwrappable =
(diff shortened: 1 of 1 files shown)


StephanTLavavej (approved):

StephanTLavavej:
I prefer the typedef to be optional, as this asks user-defined iterators to do less work. Inheriting from an iterator is squirrely and most code doesn't need to worry about that scenario.

barcharcraz (approved):
I kinda like it being optional, as well.

StephanTLavavej:

StephanTLavavej:
Thanks for fixing this bug! :lady_beetle: :black_cat: :heart_eyes_cat:

PR #3574, opened 2023-03-15
Mark constructors that do not throw exceptions as noexcept

AtariDreams:
By ensuring we strength each move constructor with noexcept as we can, we can avoid needless exception checking and handling.

Had to open a new PR because GitHub wouldn't let me reopen the last one because it turns out editing commit history when the PR is closed is not good.

### DIFF
--- a/stl/inc/expected
+++ b/stl/inc/expected
@@ -233,7 +233,7 @@ public:
     }
 
     // clang-format off
-    expected(expected&&) requires
+    expected(expected&&) noexcept requires
         is_trivially_move_constructible_v<_Ty> && is_trivially_move_constructible_v<_Err> = default;
     // clang-format on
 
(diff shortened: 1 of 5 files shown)


StephanTLavavej:
We talked about this at the weekly maintainer meeting and decided to close this PR without merging. We've repeatedly asked to avoid force-pushing changes, and multiple people have noted that many of these functions are physically allocating memory (e.g. `hash_map` is backed by `list` which dynamically allocates a sentinel node in all modes). We're sorry, but simply resubmitting the same PR (#3278, #3565) and gradually trimming down the list of edits as people raise issues, rather than verifying each carefully, is not a productive use of the maintainer team's very limited review capacity.

AtariDreams:
> We talked about this at the weekly maintainer meeting and decided to close this PR without merging. We've repeatedly asked to avoid force-pushing changes, and multiple people have noted that many of these functions are physically allocating memory (e.g. `hash_map` is backed by `list` which dynamically allocates a sentinel node in all modes). We're sorry, but simply resubmitting the same PR (#3278, #3565) and gradually trimming down the list of edits as people raise issues, rather than verifying each carefully, is not a productive use of the maintainer team's very limited review capacity.

That isn't what happened. What happened was I closed the original PR to do some research on the code I was modifying and then made those changes. Once those changes were made, GitHub would not allow me to reopen. I should have marked it as a draft. There is nothing I am trying to hide. It was an honest Git mistake.

PR #3578, opened 2023-03-16
Prefer fill_n over fill where applicable

AtariDreams:
We can avoid doing the addition ourselves when we do this.

### DIFF
--- a/stl/inc/deque
+++ b/stl/inc/deque
@@ -1488,7 +1488,7 @@ private:
                 }
 
                 _Mid = begin() + static_cast<difference_type>(_Count);
-                _STD fill(_Mid, _Mid + static_cast<difference_type>(_Off), _Val); // fill in rest of values
+                _STD fill_n(_Mid, _Off, _Val); // fill in rest of values
             } else { // insert not longer than prefix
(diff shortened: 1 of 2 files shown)


CaseyCarter (approved):

fsb4000 (approved):

StephanTLavavej (approved):

StephanTLavavej:
This is a good cleanup, thanks for finding it.

Separately, I will comment: Please avoid force-pushing branches after creating PRs, as GitHub doesn't make it easy to see what's changed. (In this case, nobody had started reviewing so it wasn't an issue, but the general principle still holds.)

Rewriting history *before* you create a PR is fine (either to squash all changes into a single commit, or to structure complicated changes as a series of comprehensible commits). After creating a PR, you should push ordinary commits to fix things like typos, which is what I see happened here.

We've asked you to avoid force pushes in the past, and you've acknowledged this, so I am confused as to why it keeps happening. If you have a tool that eagerly force-pushes, I recommend exploring how to configure it.

StephanTLavavej:

StephanTLavavej:
`fill_n(back_inserter(comments), 100, "Thanks! 😻🎉🚀");`

PR #3579, opened 2023-03-17
`<chrono>`: Optimize `to_sys` and `to_local`

cpplearner:
... by only retrieving necessary information.

This makes `to_local` 3x faster on my machine (although it's still hundreds times slower than `localtime`).

### DIFF
--- a/stl/src/msvcp_atomic_wait.src
+++ b/stl/src/msvcp_atomic_wait.src
@@ -35,7 +35,6 @@ EXPORTS
     __std_tzdb_delete_time_zones
     __std_tzdb_get_current_zone
     __std_tzdb_get_leap_seconds
-    __std_tzdb_get_sys_info_v2
     __std_tzdb_get_sys_info
     __std_tzdb_get_time_zones
     __std_wait_for_threadpool_work_callbacks
--- a/stl/src/msvcp_atomic_wait.src
+++ b/stl/src/msvcp_atomic_wait.src
@@ -35,6 +35,7 @@ EXPORTS
     __std_tzdb_delete_time_zones
     __std_tzdb_get_current_zone
     __std_tzdb_get_leap_seconds
+    __std_tzdb_get_sys_info_v2
     __std_tzdb_get_sys_info
     __std_tzdb_get_time_zones
     __std_wait_for_threadpool_work_callbacks
--- a/stl/inc/__msvc_tzdb.hpp
+++ b/stl/inc/__msvc_tzdb.hpp
@@ -67,7 +67,11 @@ struct __std_tzdb_sys_info {
 };
 
 enum class __std_tzdb_sys_info_type : char {
-    _Full,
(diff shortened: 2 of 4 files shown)


stl/src/msvcp_atomic_wait.src
```diff
@@ -35,6 +35,7 @@ EXPORTS
     __std_tzdb_delete_time_zones
     __std_tzdb_get_current_zone
     __std_tzdb_get_leap_seconds
+    __std_tzdb_get_sys_info_v2
```
StephanTLavavej, line 38: This is an ABI break - we can't change the export surface of DLLs, including adding any exports (for tedious reasons). That's why we had to add the (now-misnamed) atomic_wait satellite DLL in the first place, because we couldn't add exports directly to msvcp140.dll.

Yet another satellite DLL independently loading ICU would be an extremely logistically complicated change for a performance improvement. (Our policy is essentially to avoid satellite DLLs unless they are absolutely required to implement features; a perf improvement doesn't meet that bar.) Having the import lib independently load ICU would be less logistically complicated, but might not be worth it.

Can you think of any other way to achieve this without breaking ABI? If not, we'll have to close this PR and record the possible improvement as a vNext issue. Crying cat emoji.

StephanTLavavej (changes requested):

stl/src/msvcp_atomic_wait.src
cpplearner: It seems possible to encode the type in the `_Tz` string, so this optimization (hopefully) doesn't need to wait for vNext.

StephanTLavavej: Thanks, after a brief glance this appears to be ABI-compatible (I'll need to think about it more, of course).

StephanTLavavej (approved):

StephanTLavavej:
I believe that this is correct and ABI-safe.

In particular, I believe that it doesn't need to wait for the 17.8 Preview 1 unlocked redist. That's because both forms of mixing are compatible:

Code | Redist | Result
---|---|:---
Old | Old | Status quo, full info, slow but correct
Old | New | Bincompat scenario, old code's null terminator is interpreted as `_Full`, slow but correct
New | Old | This scenario is usually forbidden and will persist until 17.8p1!<br>New code's "type terminator" is ignored, old redist strictly avoids reading beyond the given length, full info returned, slow but correct
New | New | Desired state, fast and correct

strega-nil-ms (changes requested):
Minor comments requested.

strega-nil-ms:
@StephanTLavavej pushed changes after you approved.

strega-nil-ms (approved):

StephanTLavavej (approved):

StephanTLavavej:
Thanks, great comment! :heart_eyes_cat:

StephanTLavavej:

StephanTLavavej:
Thanks - must go faster!

:car: :t-rex: :watch:
===

PR #3580, opened 2023-03-17
`<mdspan>`: Add separate tests for `extents` and `default_accessor`

JMazurkiewicz:
Short motivation:
* Separate tests are easier to read and maintain,
* We avoid files with 1000 lines of code.

That said, this PR adds:
* Add separate tests for `extents` and `default_accessor` (I'll introduce the rest of the tests in other PR),
* Drive-by: restore `<type_traits>` include in `<mdspan>` (see https://github.com/microsoft/STL/pull/3564#discussion_r1134606987).

`P0009R18_mdspan_extents/test.cpp` tests have two `FIXME` comment, as there *might* be a bug in `extents` implementation

### DIFF
--- a/tests/std/tests/P0009R18_mdspan_default_accessor/test.cpp
+++ b/tests/std/tests/P0009R18_mdspan_default_accessor/test.cpp
@@ -50,7 +50,7 @@ constexpr void test_one(array<ElementType, 3> elems) {
 constexpr bool test() {
     test_one<char>({'a', 'b', 'c'});
     test_one<long>({1, 2, 3});
-    test_one<float>({1.1, 2.2, 3.3});
+    test_one<double>({1.1, 2.2, 3.3});
     test_one<wstring>({L"1", L"2", L"3"});
     test_one<intmax_t>({3, 2, 1});
     return true;
--- a/tests/std/tests/P0009R18_mdspan_default_accessor/test.cpp
+++ b/tests/std/tests/P0009R18_mdspan_default_accessor/test.cpp
@@ -4,6 +4,7 @@
 #include <array>
 #include <cassert>
 #include <concepts>
+#include <cstdint>
 #include <mdspan>
 #include <string>
 #include <type_traits>
@@ -35,8 +36,8 @@ constexpr void test_one(array<ElementType, 3> elems) {
     }
 
     { // Check 'access' member function
-        same_as<ElementType&> decltype(auto) ref = accessor.access(elems.data(), 1);
-        assert(ref == elems[1]);
+        same_as<ElementType&> decltype(auto) accessed_elem = accessor.access(elems.data(), 1);
+        assert(accessed_elem == elems[1]);
         static_assert(noexcept(accessor.access(elems.data(), 0)));
     }
 
--- a/tests/std/tests/P0009R18_mdspan_default_accessor/test.cpp
+++ b/tests/std/tests/P0009R18_mdspan_default_accessor/test.cpp
@@ -1,0 +1,62 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <array>
+#include <cassert>
+#include <concepts>
+#include <mdspan>
+#include <string>
+#include <type_traits>
+
+using namespace std;
+
+template <class ElementType>
+constexpr void test_one(array<ElementType, 3> elems) {
+    using DefaultAccessor = default_accessor<ElementType>;
+
+    // Check modeled concepts
+    static_assert(is_trivially_copyable_v<DefaultAccessor>);
+    static_assert(semiregular<DefaultAccessor>);
+
+    // Check nested types
+    static_assert(same_as<typename DefaultAccessor::offset_policy, DefaultAccessor>);
+    static_assert(same_as<typename DefaultAccessor::element_type, ElementType>);
+    static_assert(same_as<typename DefaultAccessor::reference, ElementType&>);
+    static_assert(same_as<typename DefaultAccessor::data_handle_type, ElementType*>);
+
+    // Check default constructor
+    DefaultAccessor accessor;
+    static_assert(is_nothrow_default_constructible_v<DefaultAccessor>);
+
+    { // Check converting constructor from other accessor
+        [[maybe_unused]] default_accessor<const ElementType> const_accessor = accessor;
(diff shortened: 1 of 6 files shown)


tests/std/tests/P0009R18_mdspan_default_accessor/test.cpp
```diff
@@ -0,0 +1,62 @@
...
+        static_assert(noexcept(accessor.offset(elems.data(), 0)));
+    }
+}
+
+constexpr bool test() {
+    test_one<char>({'a', 'b', 'c'});
+    test_one<long>({1, 2, 3});
+    test_one<double>({1.1, 2.2, 3.3});
+    test_one<wstring>({L"1", L"2", L"3"});
+    test_one<intmax_t>({3, 2, 1});
```
StephanTLavavej, line 55: Technically should include `<cstdint>`.

tests/std/tests/P0009R18_mdspan_default_accessor/test.cpp
```diff
@@ -0,0 +1,62 @@
...
+    static_assert(is_nothrow_default_constructible_v<DefaultAccessor>);
+
+    { // Check converting constructor from other accessor
+        [[maybe_unused]] default_accessor<const ElementType> const_accessor = accessor;
+        static_assert(is_nothrow_constructible_v<default_accessor<const ElementType>, DefaultAccessor>);
+        static_assert(!is_constructible_v<DefaultAccessor, default_accessor<const ElementType>>);
+    }
+
+    { // Check 'access' member function
+        same_as<ElementType&> decltype(auto) ref = accessor.access(elems.data(), 1);
```
StephanTLavavej, line 38: This potentially shadows `<functional>`'s `std::ref`. I recommend choosing any other name.

tests/std/tests/P0009R18_mdspan_extents/test.cpp
```diff
@@ -0,0 +1,294 @@
...
+};
+
+struct NonConvertibleToAnything {};
+
+template <class T>
+constexpr void check_implicit_conversion(T); // not defined
+
+template <class T, class... Args>
+concept NotImplicitlyConstructibleFrom =
+    constructible_from<T, Args...> && !requires(Args... args) { check_implicit_conversion<T>({args...}); };
```
StephanTLavavej, line 30: It might not make a difference for this test, but I think that this should either use `_STD declval<Args>()` or `_STD forward<Args>(args)`, to match `constructible_from` and avoid always asking the question for lvalues.

tests/std/tests/P0009R18_mdspan_extents/test.cpp
```diff
@@ -0,0 +1,294 @@
...
+    Ext ext;
+    static_assert(is_nothrow_default_constructible_v<Ext>);
+
+    // Check 'extent' observer
+    assert((((ext.extent(Indices) == Extents && Extents != dynamic_extent) || ext.extent(Indices) == 0) && ...));
+
+    using OtherIndexType = conditional_t<is_signed_v<IndexType>, long long, unsigned long long>;
+
+    { // Check construction from other extents
+        using Ext2 = extents<OtherIndexType, Extents...>;
```
StephanTLavavej, line 65: `Ext2` is exactly repeated 3 times - I think it would be safe to lift up.

tests/std/tests/P0009R18_mdspan_extents/test.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+        auto arr = to_array<IndexType>({ext.extent(Indices)...});
+        Ext ext2a{arr};
+        assert(((ext.extent(Indices) == ext2a.extent(Indices)) && ...));
+        assert(ext == ext2a);
+        static_assert(is_nothrow_constructible_v<Ext2, decltype(arr)>);
+
+        span s{arr};
+        Ext ext2b{s};
```
StephanTLavavej, line 92: :warning: These are constructing `Ext`, but it appears that `Ext2` was intended, based on their names, and the `is_nothrow_constructible_v<Ext2, MEOW>` lines.

tests/std/tests/P0009R18_mdspan_extents/test.cpp
```diff
@@ -0,0 +1,294 @@
...
+        assert(((ext.extent(Indices) == ext2b.extent(Indices)) && ...));
+        assert(ext == ext2b);
+        static_assert(is_nothrow_constructible_v<Ext2, decltype(s)>);
+        // Other tests are defined in 'check_construction_from_array_and_span' function
+    }
+}
+
+template <class IndexType, size_t... Extents>
+constexpr void check_members() {
+    do_check_members<IndexType, Extents...>(make_index_sequence<sizeof...(Extents)>{});
```
StephanTLavavej, line 102: This could use `index_sequence_for`.

tests/std/tests/P0009R18_mdspan_extents/test.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+template <class T, size_t ExpectedRank>
+constexpr bool all_extents_dynamic = false;
+
+template <class IndexType, size_t... Extents, size_t ExpectedRank>
+constexpr bool all_extents_dynamic<extents<IndexType, Extents...>, ExpectedRank> =
+    ((Extents == dynamic_extent) && ...) && (sizeof...(Extents) == ExpectedRank);
+
+template <class... Args>
+concept CanDeduceExtents = requires(Args... args) { extents{args...}; };
```
StephanTLavavej, line 278: This also seems like it should follow the `declval/forward` convention.

StephanTLavavej (changes requested):

tests/std/tests/P0009R18_mdspan_extents/test.cpp
JMazurkiewicz: `index_sequence_for<Extents...>` would not work, since it expects types (not `size_t`s)

StephanTLavavej: Ah, duh! Thanks, never mind then :joy_cat:

tests/std/tests/P0009R18_mdspan_extents/test.cpp
StephanTLavavej: I forgot we were in test code, thanks for omitting `_STD`. My insufficiently caffeinated brain... :scream_cat:

StephanTLavavej (approved):

StephanTLavavej:
Thanks for *extending* this test coverage! :joy_cat:

PR #3581, opened 2023-03-19
Don't warn when using operator<=> with 0

fsb4000:
LLVM has an issue: https://github.com/llvm/llvm-project/issues/43670

That issue marked as `libc++` but actually every C++ standard library affected , and MSVC STL and libstdc++.

```c++
#include <compare>

auto b = 1 <=> 2 < 0;

int main() {}
```

```
clang-cl /std:c++20 -Wzero-as-null-pointer-constant main.cpp
main.cpp(3,20): warning: zero as null pointer constant [-Wzero-as-null-pointer-constant]
auto b = 1 <=> 2 < 0;
                   ^
                   nullptr
1 warning generated.
```

изображение

I also created: https://reviews.llvm.org/D146372

Yes, after that patch we accept 
```c++
auto b = 1 <=> 2 < (1-1);
```
but probably this is fine because the behavior is undefined. At least we don't warn on valid code.

### DIFF
--- a/stl/inc/compare
+++ b/stl/inc/compare
@@ -27,24 +27,22 @@ _STL_DISABLE_CLANG_WARNINGS
 #undef new
 
 _STD_BEGIN
-struct partial_ordering;
-struct weak_ordering;
-struct strong_ordering;
+_EXPORT_STD struct partial_ordering;
+_EXPORT_STD struct weak_ordering;
+_EXPORT_STD struct strong_ordering;
 
 void _Literal_zero_is_expected();
 
 struct _Literal_zero {
-
-    consteval _Literal_zero(int _Zero) noexcept {
+    template <class _Ty, enable_if_t<is_same_v<_Ty, int>, int> = 0>
+    consteval _Literal_zero(_Ty _Zero) noexcept {
         // Can't use _STL_VERIFY because this is a core header
         if (_Zero != 0) {
             _Literal_zero_is_expected();
         }
     }
-
-    template <class _Ty, class = enable_if_t<!_Is_any_of_v<_Ty, int, partial_ordering, weak_ordering, strong_ordering>>>
-    _Literal_zero(_Ty) = delete;
 };
+
 using _Compare_t = signed char;
 
 // These "pretty" enumerator names are safe since they reuse names of user-facing entities.
--- a/stl/inc/compare
+++ b/stl/inc/compare
@@ -27,8 +27,20 @@ _STL_DISABLE_CLANG_WARNINGS
 #undef new
 
 _STD_BEGIN
-using _Literal_zero = decltype(nullptr);
-using _Compare_t    = signed char;
+struct partial_ordering;
+struct weak_ordering;
+struct strong_ordering;
+
+struct _Literal_zero {
+
+    consteval _Literal_zero(int _Zero) noexcept {
+        _STL_VERIFY(_Zero == 0, "A literal zero is expected");
+    }
+
+    template <class _Ty, class = enable_if_t<!_Is_any_of_v<_Ty, int, partial_ordering, weak_ordering, strong_ordering>>>
+    _Literal_zero(_Ty) = delete;
+};
+using _Compare_t = signed char;
 
 // These "pretty" enumerator names are safe since they reuse names of user-facing entities.
 enum class _Compare_eq : _Compare_t { equal = 0, equivalent = equal };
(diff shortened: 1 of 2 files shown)


stl/inc/compare
```diff
@@ -27,8 +27,25 @@ _STL_DISABLE_CLANG_WARNINGS
 #undef new
 
 _STD_BEGIN
-using _Literal_zero = decltype(nullptr);
-using _Compare_t    = signed char;
+struct partial_ordering;
+struct weak_ordering;
+struct strong_ordering;
```
StephanTLavavej, line 32: These forward declarations of public types should be `_EXPORT_STD`. (We follow a convention of marking all declarations and definitions `_EXPORT_STD` so we don't need to worry about forgetting it.)

stl/inc/compare
```diff
@@ -27,8 +27,25 @@ _STL_DISABLE_CLANG_WARNINGS
...
-using _Literal_zero = decltype(nullptr);
-using _Compare_t    = signed char;
+struct partial_ordering;
+struct weak_ordering;
+struct strong_ordering;
+
+void _Literal_zero_is_expected();
+
+struct _Literal_zero {
+
```
StephanTLavavej, line 37: Nitpick: Unnecessary newline.

stl/inc/compare
```diff
@@ -27,8 +27,25 @@ _STL_DISABLE_CLANG_WARNINGS
...
+    consteval _Literal_zero(int _Zero) noexcept {
+        // Can't use _STL_VERIFY because this is a core header
+        if (_Zero != 0) {
+            _Literal_zero_is_expected();
+        }
+    }
+
+    template <class _Ty, class = enable_if_t<!_Is_any_of_v<_Ty, int, partial_ordering, weak_ordering, strong_ordering>>>
```
StephanTLavavej, line 45: This should use our conventional `enable_if_t<CONDITION, int> = 0` SFINAE.

stl/inc/compare
```diff
@@ -27,8 +27,25 @@ _STL_DISABLE_CLANG_WARNINGS
...
+    _Literal_zero(_Ty) = delete;
+};
+using _Compare_t = signed char;
```
StephanTLavavej, line 48: Suggestion: Now that `_Literal_zero` is a struct, adding a newline before `_Compare_t` would be nice.

tests/std/tests/P0768R1_spaceship_operator/test.cpp
```diff
@@ -7,6 +7,10 @@
 #include <compare>
 #include <type_traits>
 
+#ifdef __clang__
+#pragma clang diagnostic error "-Wzero-as-null-pointer-constant"
+#endif // __clang__
```
StephanTLavavej, line 12: I recommend citing your PR here, since otherwise it could be mysterious why we're enabling this warning as an error. (For similar reasons, we cite bugs when testing things.)

stl/inc/compare
StephanTLavavej, line 45: Why can't the first constructor be the only one, constrained to exist only when `is_same_v<_Ty, int>`?

StephanTLavavej (changes requested):

stl/inc/compare
```diff
@@ -27,8 +27,23 @@ _STL_DISABLE_CLANG_WARNINGS
 #undef new
 
 _STD_BEGIN
-using _Literal_zero = decltype(nullptr);
-using _Compare_t    = signed char;
+_EXPORT_STD struct partial_ordering;
+_EXPORT_STD struct weak_ordering;
+_EXPORT_STD struct strong_ordering;
```
StephanTLavavej, line 32: These forward declarations are no longer necessary after the other refactoring.

tests/std/tests/P0768R1_spaceship_operator/test.cpp
```diff
@@ -7,6 +7,11 @@
 #include <compare>
 #include <type_traits>
 
+// See https://github.com/microsoft/STL/pull/3581 for details
```
StephanTLavavej, line 10: We can just cite the PR with `GH-`. (The GitHub Pull Requests & Issues extension will provide a hover card in VSCode.)

StephanTLavavej (approved):

StephanTLavavej:
Thanks! FYI @strega-nil-ms I pushed very minor changes after you approved.

StephanTLavavej:

StephanTLavavej:
Thanks for restoring the usability of the zero-as-null-pointer-constant warning! :warning: :smile_cat: :tada:

PR #3586, opened 2023-03-23
`<mdspan>`: Completely rework `std::extents`

JMazurkiewicz:
Changes to `std::extents`:

* Get rid of `_Mdspan_extent_type` base type,
* Prefer named functions instead of internal names (e.g. `sizeof...(_Extents)` -> `rank()`, `_Static_extents[_Idx]` -> `static_extent(_Idx)`),
* Fix `_EXPORT_STD` placement (only for `extents` and `dextents`),
* Implement preconditions:
  * Those preconditions are unconditional for now (Should I guard them with `_CONTAINER_DEBUG_LEVEL` macro?),
  * Some preconditions are checked only if type of inspected value satisfies `integral`,
  * Preconditions use `std::cmp_` functions from C++20 to avoid signed/unsigned comparison warnings,
  * Add `_death` tests,
* Reimplement deduction guide with workaround for not-yet reported MSVC ICE,
* Reimplement `dextents` - get rid of IILE,
* Fix some bugs and enable previously failing tests added by #3580,
* Fix some tests from old suite,
* Remove dead code - `_Fill_extents` function.

### DIFF
--- a/stl/inc/mdspan
+++ b/stl/inc/mdspan
@@ -111,7 +111,8 @@ public:
         extents(const extents<_OtherIndexType, _OtherExtents...>& _Other) noexcept {
         auto _It = _Dynamic_extents.begin();
         for (rank_type _Idx = 0; _Idx < rank(); ++_Idx) {
-            _STL_VERIFY(static_extent(_Idx) == dynamic_extent || static_extent(_Idx) == _Other.extent(_Idx),
+            _STL_VERIFY(
+                static_extent(_Idx) == dynamic_extent || _STD cmp_equal(static_extent(_Idx), _Other.extent(_Idx)),
                 "Value of other.extent(r) must be equal to extent(r) for each r for which extent(r) is a static extent "
                 "(N4928 [mdspan.extents.cons]/2.1)");
             _STL_VERIFY(_STD in_range<index_type>(_Other.extent(_Idx)),
@@ -176,12 +177,11 @@ public:
               && is_nothrow_constructible_v<index_type, const _OtherIndexType&> && (_Size == rank())
     constexpr explicit extents(span<_OtherIndexType, _Size> _Exts, index_sequence<_Indices...>) noexcept
         : _Dynamic_extents{static_cast<index_type>(_STD as_const(_Exts[_Dynamic_index_inv(_Indices)]))...} {
-        for (rank_type _Idx = 0; _Idx < rank(); ++_Idx) {
-            _STL_VERIFY(static_extent(_Idx) == dynamic_extent || static_extent(_Idx) == _Exts[_Idx],
-                "Value of exts[r] must be equal to extent(r) for each r for which extent(r) is a static extent (N4928 "
-                "[mdspan.extents.cons]/10.1)");
-
-            if constexpr (_Is_standard_integer<_OtherIndexType> && _Size != 0) {
+        if constexpr (_Is_standard_integer<_OtherIndexType>) {
+            for (rank_type _Idx = 0; _Idx < rank(); ++_Idx) {
+                _STL_VERIFY(static_extent(_Idx) == dynamic_extent || _STD cmp_equal(static_extent(_Idx), _Exts[_Idx]),
+                    "Value of exts[r] must be equal to extent(r) for each r for which extent(r) is a static extent "
+                    "(N4928 [mdspan.extents.cons]/10.1)");
                 _STL_VERIFY(_Exts[_Idx] >= 0 && _STD in_range<index_type>(_Exts[_Idx]),
                     "Either N must be zero or exts[r] must be nonnegative and must be representable as value of type "
                     "index_type for every rank index r (N4928 [mdspan.extents.cons]/10.2)");
--- a/stl/inc/mdspan
+++ b/stl/inc/mdspan
@@ -55,7 +55,7 @@ private:
         }
         _Result[rank()] = _Counter;
         return _Result;
-#pragma warning(pop)
+#pragma warning(pop) // TRANSITION, "/analyze:only" BUG?
     }
 
     static constexpr array<rank_type, rank()> _Static_extents      = {_Extents...};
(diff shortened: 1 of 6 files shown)


cpplearner:
> - Get rid of `_Mdspan_extent_type` base type,

IIUC this makes `extents` unconditionally store a `std::array<index_type, rank_dynamic()>` member. Since `std::array<T, 0>` has a non-zero size, `extents` would have a space overhead when all extents are statically known. I believe that this is undesirable.

tests/std/tests/P0009R18_mdspan/test.cpp
```diff
@@ -178,8 +173,8 @@ void extent_tests_copy_ctor_other() {
     static_assert(!is_constructible_v<extents<size_t, 2, 3>, extents<size_t, 3, 2>>);
 
     // Static extents are constructible, but not convertible, from dynamic extents.
-    // static_assert(is_constructible_v<extents<size_t, 2, 3>, extents<size_t, 2, dynamic_extent>>);
-    constexpr extents<size_t, 2, 3> ex0{extents<size_t, 2, dynamic_extent>{}};
+    static_assert(is_constructible_v<extents<size_t, 2, 3>, extents<size_t, 2, dynamic_extent>>);
+    constexpr extents<size_t, 2, 3> ex0{extents<size_t, 2, dynamic_extent>{3}};
     (void) ex0;
```
StephanTLavavej, line 178: No change requested in this PR (followup OK): Should this test consistently use `[[maybe_unused]]` now instead of `(void)`?

tests/std/tests/P0009R18_mdspan_extents/test.cpp
```diff
@@ -160,15 +162,26 @@ constexpr void check_construction_from_extents_pack() {
...
+        Ext ext1a{4, ConvertibleToInt<int>{}, 4};
+        Ext ext1b{4, 1, 4};
+        assert(ext1a == ext1b);
+
+        Ext ext2a{4, ConvertibleToInt<int>{}};
+        Ext ext2b{4, 1};
+        assert(ext2a == ext2b);
+    }
+
+    { // Check construciton with integers with mismatched signs
```
StephanTLavavej, line 176: Comment typo here, followup OK. Occurs below.

tests/std/tests/P0009R18_mdspan_extents/test.cpp
```diff
@@ -201,6 +214,49 @@ constexpr void check_construction_from_array_and_span() {
         static_assert(!is_constructible_v<Ext, span<NonConvertibleToAnything, 2>>);
     }
 
+    { // // Check construction from arrays/spans where [array/span].size() is equal to rank_dynamic()
```
StephanTLavavej, line 217: Doubled-up comment here, followup OK.

stl/inc/mdspan
```diff
@@ -24,151 +24,126 @@ _STL_DISABLE_CLANG_WARNINGS
...
               && ((_OtherExtents == dynamic_extent || _Extents == dynamic_extent || _OtherExtents == _Extents) && ...)
     constexpr explicit((((_Extents != dynamic_extent) && (_OtherExtents == dynamic_extent)) || ...)
                        || (numeric_limits<index_type>::max)() < (numeric_limits<_OtherIndexType>::max)())
         extents(const extents<_OtherIndexType, _OtherExtents...>& _Other) noexcept {
-        auto _Dynamic_it = _Mybase::_Begin_dynamic_extents();
-        for (rank_type _Dim = 0; _Dim < sizeof...(_Extents); ++_Dim) {
-            if (_Mybase::_Static_extents[_Dim] == dynamic_extent) {
-                *_Dynamic_it++ = _Other.extent(_Dim);
+        auto _It = _Dynamic_extents.begin();
+        for (rank_type _Idx = 0; _Idx < rank(); ++_Idx) {
```
StephanTLavavej, line 135: No change requested but you can followup if you want: Because `rank()` is an ordinary `constexpr` function (not `consteval`), I note that usage of it outside constant expressions will result in real function calls being emitted in non-optimized debug mode. This is more of an implementer's choice than anything critical, so I'm inclined to defer to your judgement about what's clearest. However, if I were writing this, I would probably introduce a `static constexpr _Rank` data member, so I could bake it into all of these functions without real calls, while keeping the clarity of saying "rank".

stl/inc/mdspan
```diff
@@ -24,151 +24,126 @@ _STL_DISABLE_CLANG_WARNINGS
...
-    constexpr index_type* _Begin_dynamic_extents() noexcept {
-        return _Dynamic_extents;
-    }
-
-    constexpr const index_type* _Begin_dynamic_extents() const noexcept {
-        return _Dynamic_extents;
+    _NODISCARD static _CONSTEVAL auto _Make_dynamic_indices_inv() noexcept {
+        array<rank_type, rank()> _Result{};
+        for (rank_type _Idx = 0; _Idx < rank(); ++_Idx) {
+            for (rank_type _Rdx = 0; _Rdx < rank(); ++_Rdx) {
```
StephanTLavavej, line 71: Suggestion: `_Idx` and `_Rdx` differ by a single character, making this somewhat difficult to read and mistake-prone. Could a clearer name be found for `_Rdx`?

stl/inc/mdspan
```diff
@@ -24,151 +24,126 @@ _STL_DISABLE_CLANG_WARNINGS
...
-    }
-
     _NODISCARD static constexpr rank_type rank_dynamic() noexcept {
-        return ((_Extents == dynamic_extent) + ... + 0);
+        return _Rank_dynamic;
     }
 
     _NODISCARD static constexpr size_t static_extent(const rank_type _Idx) noexcept {
-        return _Mybase::_Static_extents[_Idx];
+        _STL_VERIFY(_Idx < rank(), "Index must be less than rank() (N4928 [mdspan.extents.obs]/1)");
```
StephanTLavavej, line 107: `_STL_VERIFY` performs the check all of the time, and these occurrences aren't guarded by IDL or CDL. (We haven't exactly put a lot of design effort into CDL, but the general idea seems to be that it should perform cheap checks that don't require ABI-changing bookkeeping.) Except in highly unusual cases, we prefer release mode with IDL=0 and CDL=0 to perform no precondition checks.

`_STL_ASSERT` is controlled by `_DEBUG` and we use that occasionally, although it's probably inconsistent that elsewhere we use `_STL_VERIFY` guarded by IDL or CDL.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! I have some comments for followup but I think we can go ahead and merge this now. I'll also merge with `main` in a PR after this.

stl/inc/mdspan
JMazurkiewicz: `_Idx_inv` might be better name (I'm going to include it in my next PR).

tests/std/tests/P0009R18_mdspan/test.cpp
JMazurkiewicz: > Should this test consistently use `[[maybe_unused]]` now instead of `(void)`?

Yes, but I'm going to remove this test file anyway (after I create separate tests for all classes from this paper).

PR #3596, opened 2023-03-28
Allow zero sigma for normal_distribution

torsten-pf:
For simulation purposes a constant output is sometimes desired (setting sigma to zero, but be able to change it to other values if desired without recompilation). The implementation is able to realize this: `return _Res * _Par0._Sigma + _Par0._Mean;`

### DIFF
--- a/stl/inc/random
+++ b/stl/inc/random
@@ -2996,7 +2996,7 @@ public:
         }
 
         void _Init(_Ty _Mean0, _Ty _Sigma0) { // set internal state
-            _STL_ASSERT(0.0 < _Sigma0, "invalid sigma argument for normal_distribution");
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
Thanks for the PR! This is a non-Standard extension, which is an explicit [non-goal](https://github.com/microsoft/STL#non-goals) of our project. We'll talk about this at the weekly maintainer meeting on Wednesday.

StephanTLavavej:
We talked about this at the weekly maintainer meeting and have decided not to accept this PR, but thanks again for submitting it. If you're interested in pursuing this change, we recommend submitting a Library Working Group issue to the C++ Standardization Committee for them to consider changing this precondition. That would have the advantage of making such usage Standard and portable across all implementations. See https://isocpp.org/std/submit-issue for more info.

PR #3598, opened 2023-03-29
Python cleanups

StephanTLavavej:
I used Pylance to find these issues.

* Use `_` when ignoring part of a destructured tuple.
  + We've done this elsewhere, so now it's consistent.
* Drop unused imports.
* Drop unused `filename` variable.
* Use `name` and `value` variables.
* Non-Python: Detect 17.6 Preview 2 in `test_header_units_and_modules.hpp`.
  + We don't usually test `_MSC_FULL_VER` (as it's too volatile between MSVC-internal branches) but this value should be safe as it's released. The PR/CI system is currently using Preview 1, and we won't be able to update to Preview 2 due to a blocking bug in the VS Installer, so this is needed for local test runs to pass, until we can update to Preview 3 and entirely clear out this workaround.

### DIFF
--- a/tests/std/include/test_header_units_and_modules.hpp
+++ b/tests/std/include/test_header_units_and_modules.hpp
@@ -661,7 +661,7 @@ constexpr bool impl_test_source_location() {
     using namespace std;
     const auto sl = source_location::current();
     assert(sl.line() == __LINE__ - 1);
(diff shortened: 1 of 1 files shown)


StephanTLavavej:

strega-nil-ms (approved):

PR #3593, opened 2023-03-25
`<mdspan>`: Various cleanups (mostly renames)

JMazurkiewicz:
* Fix feature test macro - it should be guarded with `__cpp_lib_concepts`,
* Fix various typos,
* Cleanup `_EXPORT_STD` - it should be on the same line as the class,
* Update working draft revision to N4944 and add missing citations,
* Prefer type aliases instead of _Ugly names (e.g. `_Extents` -> `extents_type`),
* Replace `_Mysomething` names with `_Something`.

PR #3603, opened 2023-03-30
`<mdspan>`: `layout_left` improvements

JMazurkiewicz:
`<mdspan>`:
* Implement `layout_left::mapping`'s preconditions,
* Fix `is_(unique/exhaustive/strided)` functions - they should be `static`,
* Add exposition-only public member *`_Fwd_prod_of_extents`* to `extents`,
* Fix some issues with signed-unsigned conversions,
* Fix `mdspan`'s constructor constraints,
* Address https://github.com/microsoft/STL/pull/3593#discussion_r1148923347.

Testing:
* Introduce `test_mdspan_support.hpp` header,
* Add separate tests for `layout_left`: regular tests and death tests.

### DIFF
--- a/stl/inc/mdspan
+++ b/stl/inc/mdspan
@@ -587,7 +587,7 @@ public:
               && is_nothrow_constructible_v<index_type, const _OtherIndexType&>
 #endif // ^^^ no workaround ^^^
     constexpr mapping(const extents_type& _Exts_, const span<_OtherIndexType, extents_type::rank()> _Strides_) noexcept
-        : _Exts{_Exts_} {
+        : _Exts(_Exts_) {
         for (rank_type _Idx = 0; _Idx < extents_type::rank(); ++_Idx) {
             _Strides[_Idx] = _Strides_[_Idx];
         }
@@ -604,7 +604,7 @@ public:
 #endif // ^^^ no workaround ^^^
     constexpr mapping(
         const extents_type& _Exts_, const array<_OtherIndexType, extents_type::rank()>& _Strides_) noexcept
-        : _Exts{_Exts_} {
+        : _Exts(_Exts_) {
         for (rank_type _Idx = 0; _Idx < extents_type::rank(); ++_Idx) {
             _Strides[_Idx] = _Strides_[_Idx];
         }
--- a/stl/inc/mdspan
+++ b/stl/inc/mdspan
@@ -240,6 +240,15 @@ public:
         }
     }
 
+    // TRANSITION, LWG ISSUE? I believe that this function should return 'index_type'
+    _NODISCARD constexpr index_type _Fwd_prod_of_extents(const rank_type _Idx) const noexcept {
+        index_type _Result = 1;
+        for (rank_type _Dim = 0; _Dim < _Idx; ++_Dim) {
+            _Result *= extent(_Dim);
+        }
+        return _Result;
+    }
+
     _NODISCARD static _CONSTEVAL bool _Is_index_space_size_representable() {
         if constexpr (rank_dynamic() == 0 && rank() > 0) {
             return _STD in_range<index_type>((_Extents * ...));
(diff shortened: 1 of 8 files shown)


tests/std/tests/P0009R18_mdspan_layout_left/test.cpp
```diff
@@ -0,0 +1,370 @@
...
+    check_construction_from_other_left_mapping();
+    check_construction_from_other_right_mapping();
+    check_construction_from_other_stride_mapping();
+    check_call_operator();
+    check_comparisons();
+    check_correctness();
+    return true;
+}
+
+
```
JMazurkiewicz, line 366: TODO: remove extra newline in follow-up

stl/inc/mdspan
```diff
@@ -314,24 +323,52 @@ public:
...
+        // TRANSITION, CHECK [mdspan.layout.left.cons]/1 (REQUIRES '_Multiply_with_overflow_check' FROM #3561)
+    }
 
     template <class _OtherExtents>
         requires is_constructible_v<extents_type, _OtherExtents>
     constexpr explicit(!is_convertible_v<_OtherExtents, extents_type>)
         mapping(const mapping<_OtherExtents>& _Other) noexcept
-        : _Exts(_Other.extents()) {}
+        : _Exts(_Other.extents()) {
+        _STL_VERIFY(_STD in_range<index_type>(_Other.required_span_size()),
```
MattStephanson, line 335: I don't think we usually check for precondition violations, at least outside of debug mode. Maybe `_STL_ASSERT` would be better?

stl/inc/mdspan
```diff
@@ -314,24 +323,52 @@ public:
...
+                index_type _Prod = 1;
+                return (
+                    (_Other.stride(_Indices)
+                        == (_Indices + 1 == extents_type::rank()
+                                ? _Prod
+                                : _STD exchange(_Prod, static_cast<index_type>(_Prod * _Exts.extent(_Indices + 1)))))
+                    && ...);
+            }
+            (make_index_sequence<extents_type::rank()>{});
+            _STL_VERIFY(_Verify, "For all r in the range [0, extents_type::rank()), other.stride(r) must be equal to "
```
MattStephanson, line 365: Even if this were changed to `_STL_ASSERT`, the computation above it would still be done in release mode, unless the optimizer notices that it's a dead store. As with the other instances of `_STL_VERIFY`, I'm not sure that's a worthwhile tradeoff.

tests/std/tests/P0009R18_mdspan_layout_left_death/test.cpp
```diff
@@ -0,0 +1,61 @@
...
+    layout_stride::mapping<dextents<int, 1>> mapping1{dextents<int, 1>{256}, array{1}};
+    // Value of other.required_span_size() must be representable as a value of type index_type
+    layout_left::mapping<dextents<unsigned char, 1>> mapping2{mapping1};
+}
+#pragma warning(pop) // TRANSITION, "/analyze:only" BUG?
+
+void test_stride_function() {
+    layout_left::mapping<extents<int, 3>> mapping;
+    // Value of i must be less than extents_type::rank()
+    (void) mapping.stride(1);
```
StephanTLavavej, line 47: Suggestion: Although this isn't shadowing, I find that it's best to name variables differently from any Standard identifiers that are involved. Here, I'd like to be able to glance at `mapping` and know that it's always a Standard nested type, not a test variable.

tests/std/tests/P0009R18_mdspan_layout_left/test.cpp
```diff
@@ -0,0 +1,370 @@
...
+#include <cstddef>
+#include <mdspan>
+#include <span>
+#include <type_traits>
+
+#include "test_mdspan_support.hpp"
+
+template <class Mapping, class... Indices>
+concept CanInvokeCallOperatorOfMapping = requires(Mapping mapping, Indices... i) {
+                                             { mapping(i...) } -> same_as<typename Mapping::index_type>;
```
StephanTLavavej, line 16: Similar comment to the other test, it would be nice to avoid the `mapping` name here, even though it is not shadowing. Occurs below, and there's a variable named `copy` that raises my eyebrows.

StephanTLavavej (approved):

stl/inc/mdspan
```diff
@@ -846,8 +873,8 @@ public:
         requires (is_convertible_v<_OtherIndexTypes, index_type> && ...)
               && (is_nothrow_constructible_v<index_type, _OtherIndexTypes> && ...)
               && (sizeof...(_OtherIndexTypes) == rank())
-    _NODISCARD constexpr reference operator()(const _OtherIndexTypes... _Indices) const {
-        return _Acc.access(_Ptr, _Map(static_cast<index_type>(_STD move(_Indices))...));
+    _NODISCARD constexpr reference operator()(_OtherIndexTypes... _Indices) const {
+        return _Acc.access(_Ptr, static_cast<size_t>(_Map(static_cast<index_type>(_STD move(_Indices))...)));
```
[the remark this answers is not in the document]
StephanTLavavej: Our usual policy is that we avoid emitting warnings that are produced entirely within our code, but we will emit warnings if the user has asked us to perform an operation on their behalf (i.e. involving types they've specified) that would warn if they wrote it themselves.

stl/inc/mdspan
```diff
@@ -367,44 +399,40 @@ public:
...
+        _STL_VERIFY(_Idx < extents_type::rank(),
+            "Value of i must be less than extents_type::rank() (N4944 [mdspan.layout.left.obs]/6).");
+        return _Exts._Fwd_prod_of_extents(_Idx);
     }
 
     template <class _OtherExtents>
         requires (extents_type::rank() == _OtherExtents::rank())
     _NODISCARD_FRIEND constexpr bool operator==(const mapping& _Left, const mapping<_OtherExtents>& _Right) noexcept {
-        return _Left.extents() == _Right.extents();
+        return _Left._Exts == _Right.extents();
```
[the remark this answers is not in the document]
JMazurkiewicz: `_Right._Exts` always exists, there is totally different reason why I left "`_Right.extents()`" here - this `operator==` is not always a friend of `_Right`, therefore sometimes (specifically when `same_as<extents_type, _OtherExtents>` is false) it cannot access `_Exts`.

stl/inc/mdspan
```diff
@@ -340,19 +377,14 @@ public:
...
-        requires (sizeof...(_Indices) == extents_type::rank()) && (is_convertible_v<_Indices, index_type> && ...)
-              && (is_nothrow_constructible_v<index_type, _Indices> && ...)
-    _NODISCARD constexpr index_type operator()(_Indices... _Idx) const noexcept {
-        return _Index_impl<conditional_t<true, index_type, _Indices>...>(
-            static_cast<index_type>(_Idx)..., make_index_sequence<extents_type::rank()>{});
+    template <class... _IndexTypes>
+        requires (sizeof...(_IndexTypes) == extents_type::rank()) && (is_convertible_v<_IndexTypes, index_type> && ...)
+              && (is_nothrow_constructible_v<index_type, _IndexTypes> && ...)
+    _NODISCARD constexpr index_type operator()(_IndexTypes... _Indices) const noexcept {
+        return _Index_impl(make_index_sequence<extents_type::rank()>{}, _Indices...);
```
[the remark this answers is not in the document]
JMazurkiewicz: OK, I'm going to bring it back.

stl/inc/mdspan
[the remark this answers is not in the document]
StephanTLavavej: This is a `_NODISCARD_FRIEND` so we don't have access to the guts of `mapping<_OtherExtents>`. (I always, always forget this rule :joy_cat:)

stl/inc/mdspan
StephanTLavavej: Correct. As mentioned in a previous review, we should find a pattern for precondition checks and follow it consistently. It seemed that the pattern here was IDL=2 for iterator stuff and CDL for checks outside the iterators. (`_STL_ASSERT` is directly controlled by `_DEBUG`. Yeah, we've developed inconsistency here.)

stl/inc/mdspan
StephanTLavavej: Agreed, this is a case where the two parts should be wrapped in an `#if`.

StephanTLavavej:
Happy to merge this now and followup later, or you can iterate on this PR, your pick.

JMazurkiewicz:
> Happy to merge this now and followup later, or you can iterate on this PR, your pick.

I'd like to iterate on this PR, since I've already made some changes.

stl/inc/mdspan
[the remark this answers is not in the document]
JMazurkiewicz: > There also seems to be a tension in the design that `index_type` can be signed but the second parameter to `access`/`offset` is `size_t`. Maybe they should take a `ptrdiff_t` or the layout mapping should return a `size_type`.

The result of `_Map(indices...)` must be non-negative and less or equal to `limits<size_t>::max` ([[mdspan.layout.reqmts]/8](http://eel.is/c++draft/views.multidim#mdspan.layout.reqmts-8)), so changing parameters/return types looks like overkill to me. IMHO `static_cast`ing to `size_t` is enough and maybe standard should explicitly say it.

> LWG issue needed?

I'm not sure if it is necessary. Maybe we could submit PR to https://github.com/cplusplus/draft with explicit `static_cast`.

StephanTLavavej (approved):

StephanTLavavej:
LGTM. Want to merge this now?

JMazurkiewicz:
Yes, please

StephanTLavavej:
Thanks! :heart_eyes_cat: :rocket: :tada:

Issue #3606, opened 2023-03-31
Investigate ppltasks dependency for vNext


StephanTLavavej:
I'm going to need to revert #2654 and #3255, so here are vNext todos that were mentioned in the latter PR's added comment:

When we can break ABI, we should:

* In `<ppltasks.h>`, reconsider whether `_Task_impl_base::_Wait()` should throw `invalid_operation`; it's questionable whether that's conforming, and if users want to block their UI threads, we should let them.
* Investigate whether we can avoid the ppltasks dependency entirely, making all of these issues irrelevant.

AlexGuteniev:
> * Investigate whether we can avoid the ppltasks dependency entirely, making all of these issues irrelevant.

I think that `<ppltasks.h>` is only used to support asynchronous (non-deferred) `std::async`, to spawn separate workers.

Instead of using `<ppltasks.h>` for that, we should use thread pool directly, like we currently do in `<execution>` /  `parallel_algorithms.cpp`.

PR #3255, opened 2022-12-01
ppltasks.cpp: Fix bincompat for VS 2015

StephanTLavavej:
# Scenario

This fixes internal VSO-1684985, a high-priority bug reported by 4 customers through support cases. The scenario happens when a program:

* Was built with VS 2015
* Uses a "Single-Threaded Apartment"
  + DirectX reportedly requires this, so it's common
  + In the provided repro, this is activated by calling `CoInitialize(nullptr)`
* Calls `get()` on a `future` returned by `std::async()`
* Runs on an end user's machine with the VS 2022 17.4 redist installed
  + This is why we're hearing about it now

When these stars align, the program crashes due to ppltasks machinery throwing an `invalid_operation` exception that slams into `noexcept`, so no recovery is possible.

# Analysis

`future::get()` eventually calls into this code in `ppltasks.h`:

```cpp
    struct _Task_impl_base
    {
// ...
        task_status _Wait()
        {
            bool _DoWait = true;

#if (_PPL_TASK_CONTEXT_CONTROL_ENABLED)
            if (_IsNonBlockingThread())
            {
                // In order to prevent Windows Runtime STA threads from blocking the UI, calling task.wait() task.get() is illegal
                // if task has not been completed.
                if (!_IsCompleted() && !_IsCanceled())
                {
                    _THROW(invalid_operation("Illegal to wait on a task in a Windows Runtime STA"));
                }
```

This calls `_Task_impl_base::_IsNonBlockingThread()`, which is separately compiled in `ppltasks.cpp`. Both the header and the source file have changed over time, which is how we broke bincompat without breaking consistent fresh builds.

Header change: The usage of `_IsNonBlockingThread()` is guarded by `_PPL_TASK_CONTEXT_CONTROL_ENABLED`, which (near the top of the file on lines 50-54) appears to be opt-in (or enabled by C++/CX, which these customers apparently aren't using). This was introduced by TFS changeset 1685976 "Sync Windows changes back to DevDiv (second round)." on 2018-01-12. I believe this is why this hasn't been reported before - anyone building with VS 2019/2022 (unless they opt-in or are C++/CX) doesn't have this codepath that can throw an `invalid_operation` exception. (I'd need to do more source archaeology to be certain, but I suspect that this change also flowed into some updates of VS 2017, leaving only VS 2015 customers affected.)

[...]

PR #2654, opened 2022-04-13
Coordinated fix for capturing continuation context for tasks in Desktop apps

Scottj1s:
With the adoption of WinUI3 (via Windows App SDK), Desktop Xaml apps are now supported. However, ppltasks client code in a Desktop app would fail to capture the apartment context for continuation callbacks, by default (globally, without needing to pass an explicit task_continuation_context to every continuation). This PR fixes that bug.

For C++/CX client code (which may be in process of migration to Desktop C++/WinRT code), no other changes are needed.

For ISO C++ client code, this fix builds on work to recover ppltasks functionality lost during the UCRT migration (see https://www.osgwiki.com/wiki/Universal_CRT#PPL_Tasks  for details), which requires adding
`#define PPL_TASK_CONTEXT_ERROR_CONTROL 1 `
to opt into using context callbacks and error reporting (UWP behavior).

Note that C++/CLI does not support ppltasks.h: fatal error C1189: #error: <condition_variable> is not supported when compiling with /clr or /clr:pure

If previous behavior is desired (callbacks on an arbitrary continuation thread), code can add
`#define PPL_TASK_CONTEXT_DEFAULT_ARBITRARY 1`
C++/CX code can also explicitly pass concurrency::task_continuation_context::use_arbitrary() to then().

The fix requires CoGetApartmentType and CoGetObjectContext, in turn requiring Windows 7 or later. This dependency is handled by a #pragma comment ( lib, "ole32" ) in ppltasks.cpp. In addition, msvcp140.dll is built with delay-loaded ole32, so that applications can continue to be deployed to Vista, if this code is not used. Support for OneCore SKUs is assumed to be handled by ApiSet reverse-forwarding from the ole32.dll linkage. No compile time checks against WINVER/_WIN32_WINNT have been added, as the docs officially state Windows 7 as the base supported version for Visual Studio.

PR #3607, opened 2023-03-31
Revert the ppltasks change that introduced an `ole32.dll` dependency

### LINKED ISSUE #3257: ppltasks.cpp: Breaking change - New OLE32.dll import prevents process mitigations
Hello,

Commit https://github.com/microsoft/STL/commit/80ebe871c7af413e5f3adced95e84ec238a884a2 introduced a new import for ole32.dll in shipped msvcp140.dll. 

This is a problem for us, because this import of ole32 cascades into eventually loading user32, immediately at process startup.
This is indirectly a breaking change, because our program needs to enable process mitigations (hardening) on itself, and the system will refuse to enable anti-win32k mitigations if user32 is already loaded in the process.

Workarounds possibly exist for our specific usecase (such as not enabling the mitigation on ourselves and instead having a parent process enable them on a child at the time of CreateProcess), but it remains a breaking change - [SetProcessMitigationPolicy](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setprocessmitigationpolicy) is explicitly designed to set mitigations on yourself.

Could ole32 be delay-loaded only when ppltasks are used? 

Edit: the commit I linked does mention the intent was that ole32 should be delay-loaded. This does not appear to be the case in msvcp140 14.34.31931.0 (what I currently have on latest VS) and beyond?

This is also DevCom-10217462.

StephanTLavavej:
This is a clean revert (no manual changes) of #2654 and the following #3255 (in reverse order). #3255 was a targeted fix (partial revert) for an incredibly disruptive bincompat bug, but #2654 has caused even more trouble. This full revert:

* Fixes #3257, where the newly added dependency on `ole32.dll` was disruptive to certain exotic scenarios, and
* Fixes a "sandboxed process" scenario for the XStore team (reported by @vaboca), also disrupted by `ole32.dll`.
  + Don't ask me to explain this in any more detail, I don't know what's going on here. :exploding_head: That's the whole problem with this change - nobody fully understood the implications - and that's why we need to fully revert it, not mess around with it more.

By performing a full revert, we no longer need the partial revert. This returns us to the state that we successfully lived with in VS 2015 through VS 2022 17.2 inclusive.

@Scottj1s, the original author of #2654, agrees with the full revert here, and has confirmed that the original affected customer who needed that change has the ability to use a workaround in perpetuity.

:mirror: The mirror of this PR, MSVC-PR-462260, additionally reverts the non-GitHub change in MSVC-PR-339372 affecting `<ppltasks.h>`.

:grey_exclamation: By changing `ppltasks.cpp` again, this affects the VCRedist. For the bincompat scenario involving mixing VS 2015, this should be unquestionably safe (the partial revert and the full revert result in the same behavior for the function in question, that's why the partial revert worked). For the question of when we can ship this fix, because it is urgently needed by the XStore team, we are planning to port this fix to VS 2022 17.6 before it reaches General Availability, which will require unlocking its redist, and then that unlocked redist will flow into VS 2022 17.7, so there should be no remaining mix-and-match nightmares (I hope).

I am not performing any other changes or cleanups at this time (I may come back to `ppltasks.cpp` in the future to add some `#endif` comments, but no logic changes, I have had a lifetime of being burned here).

I've filed #3606 to capture the vNext comments that are being reverted here.

:warning: Note to self: Remember to start the 17.6 backport process after merging this.

### DIFF
--- a/stl/CMakeLists.txt
+++ b/stl/CMakeLists.txt
@@ -576,7 +576,7 @@ function(add_stl_dlls D_SUFFIX REL_OR_DBG)
     target_stl_compile_options(msvcp${D_SUFFIX}_eha_objects ${REL_OR_DBG})
 
     add_library(msvcp${D_SUFFIX} SHARED)
-    target_link_libraries(msvcp${D_SUFFIX} PRIVATE msvcp${D_SUFFIX}_eha_objects msvcp${D_SUFFIX}_objects msvcp${D_SUFFIX}_init_objects "${TOOLSET_LIB}/vcruntime${D_SUFFIX}.lib" "${TOOLSET_LIB}/msvcrt${D_SUFFIX}.lib" "ucrt${D_SUFFIX}.lib" "ole32.lib")
+    target_link_libraries(msvcp${D_SUFFIX} PRIVATE msvcp${D_SUFFIX}_eha_objects msvcp${D_SUFFIX}_objects msvcp${D_SUFFIX}_init_objects "${TOOLSET_LIB}/vcruntime${D_SUFFIX}.lib" "${TOOLSET_LIB}/msvcrt${D_SUFFIX}.lib" "ucrt${D_SUFFIX}.lib")
     set_target_properties(msvcp${D_SUFFIX} PROPERTIES ARCHIVE_OUTPUT_NAME "msvcp140_base${D_SUFFIX}${VCLIBS_SUFFIX}")
     set_target_properties(msvcp${D_SUFFIX} PROPERTIES ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}")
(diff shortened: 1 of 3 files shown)


Scottj1s:
Thanks STL.  Gives me a case of the Moody Blues to see it go, but agree with the call.

MahmoudGSaleh (approved):

tringi:
Hey. I wanted to note that our SCADA software is too affected by this issue.
Luckily the vast majority of users don't upgrade vcruntime that often (or at all).
Can we expect 17.6 soon?

StephanTLavavej:
@tringi As a general rule, we can't talk about release dates before they've been officially announced.

In this case, I think I can get away with saying that as 17.6 Preview 6 is the current Preview release, you can look at our historical patterns (which are fairly consistent although not absolute clockwork) and guess that the production release is coming soon (whereas if we were currently something like Preview 2, it would not be "soon"). I can't be any more specific than that, sorry.

PR #3615, opened 2023-04-03
`<mdspan>`: `layout_right` improvements

JMazurkiewicz:
`layout_right`:
* Implement preconditions (checked unconditionally for now),
* Fix `is_(unique/exhaustive/strided)` functions - all three should be `static`,
* Implement *`rev-prod-of-extents`* function in `extents`,
* Add separate regular tests and death tests.

`layout_left`:
* Fix incorrect preconditions in constructor from `layout_stride`,
* Fix tests and add extra coverage.

Testing:
* Get rid of `CanInvokeCallOperatorOfMapping` concept and use already defined in helper header `CheckCallOperatorOfLayoutMapping` concept.

### DIFF
--- a/tests/std/tests/P0009R18_mdspan_layout_right/test.cpp
+++ b/tests/std/tests/P0009R18_mdspan_layout_right/test.cpp
@@ -1,0 +1,390 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <array>
+#include <cassert>
+#include <concepts>
+#include <cstddef>
+#include <mdspan>
+#include <span>
+#include <type_traits>
+
+#include "test_mdspan_support.hpp"
+
+using namespace std;
+
+template <class IndexType, size_t... Extents, size_t... Indices>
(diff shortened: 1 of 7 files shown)


tests/std/tests/P0009R18_mdspan_layout_right/test.cpp
```diff
@@ -0,0 +1,390 @@
...
+    }
+
+    { // Check implicit conversions
+        static_assert(!NotImplicitlyConstructibleFrom<layout_right::mapping<extents<int, 3>>,
+                      layout_left::mapping<extents<int, 3>>>);
+        static_assert(NotImplicitlyConstructibleFrom<layout_right::mapping<extents<int, 3>>,
+            layout_left::mapping<extents<long long, 3>>>);
+        static_assert(NotImplicitlyConstructibleFrom<layout_right::mapping<extents<int, 3>>,
+            layout_left::mapping<extents<int, dynamic_extent>>>);
+    }
```
StephanTLavavej, line 192: Suggestion: `layout_right::mapping<extents<int, 3>>` is repeated 3 times here, so it would be clearer to extract a `Mapping` typedef, like the block immediately above. This would also avoid `static_assert` wrapping, making the variation on the RHS even more clear.

StephanTLavavej (approved):

PR #3619, opened 2023-04-03
Fix silent bad codegen for vectorized `meow_element()` above 4 GB

### LINKED ISSUE #3617: `<algorithm>`: Silent bad codegen for vectorized `meow_element()` above 4 GB
Reported by Johan Lundberg and Felix Ungman. This is a regression in VS 2022 17.4 after the PR series #2447 #2821 #2825.

This repro is specific to x64 because it requires a range over 4 gigabytes. The size in bytes, not the size in elements, is the threshold, so we can repro it with slightly over 1 billion `int`s.

```
```
```cpp
#include <algorithm>
#include <iostream>
#include <vector>
using namespace std;

int main() {
    for (const auto n : {0x3FFF'FFFFull, 0x4000'0010ull}) {
        vector<int> v(n, 25);
        v[n - 2] = 24;
        v[n - 1] = 26;

        cout << "n: " << n << "\n";
        cout << "min_element: " << *min_element(v.begin(), v.end()) << "\n";
        cout << "max_element: " << *max_element(v.begin(), v.end()) << "\n";
        const auto p = minmax_element(v.begin(), v.end());
        cout << "minmax_element: " << *p.first << ", " << *p.second << "\n";
        cout << "\n";
    }
}
```
```
D:\GitHub\STL\out\x64>cl /EHsc /nologo /W4 /std:c++latest /MT /O2 meow.cpp && meow
meow.cpp
n: 1073741823
min_element: 24
max_element: 26
minmax_element: 24, 26

n: 1073741840
min_element: 25
max_element: 25
minmax_element: 25, 25
```

The escape hatch restores correctness:

```
D:\GitHub\STL\out\x64>cl /EHsc /nologo /W4 /std:c++latest /MT /O2 /D_USE_STD_VECTOR_ALGORITHMS=0 meow.cpp && meow
meow.cpp
n: 1073741823
min_element: 24
max_element: 26
minmax_element: 24, 26

n: 1073741840
min_element: 24
max_element: 26
minmax_element: 24, 26
```

I found the root cause and am working on a fix. The code isn't ignoring data beyond 4 GB - it actually finds the right answers, but then has a 32-bit truncation mistake immediately before returning the result.

I verified that no other vectorized algorithms are affected, as they don't perform these advanced index computations.

StephanTLavavej:
Fixes #3617.

* Fix truncation by adding `static_cast<size_t>`.
  + The bug was lurking in the expression `_Base + _V_pos * 16 + _H_pos`, where `_Base` was `const char *`, `_V_pos` was `unsigned int`, and `_H_pos` was `unsigned long`.
* Pre-existing: `test_various_containers` as we disable instructions.
  + This was an oversight in #804 on 2020-07-27.
* Extract `test_min_max_element_support.hpp`.
* Add `GH_003617_vectorized_meow_element`.
  + This uses the `fast_matrix.lst` (one configuration, optimized) because it takes a while and consumes a ton of memory if the usual matrix is used. (I originally tried to fuse it into `VSO_0000000_vector_algorithms` but that was infeasible.)

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -884,8 +884,7 @@ namespace {
                             _BitScanForward(&_H_pos, _Mask); // lgtm [cpp/conditionallyuninitializedvariable]
 
                             const auto _V_pos = _Traits::_Get_v_pos(_Cur_idx_min, _H_pos); // Extract its vertical index
-                            _Res._Min =
-                                _Base + static_cast<size_t>(_V_pos) * 16 + _H_pos; // Finally, compute the pointer
+                            _Res._Min         = _Base + size_t{_V_pos} * 16 + _H_pos; // Finally, compute the pointer
                         }
                     }
 
@@ -931,8 +930,7 @@ namespace {
                             }
 
                             const auto _V_pos = _Traits::_Get_v_pos(_Cur_idx_max, _H_pos); // Extract its vertical index
-                            _Res._Max =
-                                _Base + static_cast<size_t>(_V_pos) * 16 + _H_pos; // Finally, compute the pointer
+                            _Res._Max         = _Base + size_t{_V_pos} * 16 + _H_pos; // Finally, compute the pointer
                         }
                     }
                     // Horizontal part done, results are saved, now need to see if there is another portion to process
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -884,7 +884,8 @@ namespace {
                             _BitScanForward(&_H_pos, _Mask); // lgtm [cpp/conditionallyuninitializedvariable]
 
                             const auto _V_pos = _Traits::_Get_v_pos(_Cur_idx_min, _H_pos); // Extract its vertical index
-                            _Res._Min         = _Base + _V_pos * 16 + _H_pos; // Finally, compute the pointer
+                            _Res._Min =
+                                _Base + static_cast<size_t>(_V_pos) * 16 + _H_pos; // Finally, compute the pointer
                         }
                     }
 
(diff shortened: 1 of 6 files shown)


stl/src/vector_algorithms.cpp
```diff
@@ -884,7 +884,8 @@ namespace {
                             _BitScanForward(&_H_pos, _Mask); // lgtm [cpp/conditionallyuninitializedvariable]
 
                             const auto _V_pos = _Traits::_Get_v_pos(_Cur_idx_min, _H_pos); // Extract its vertical index
```
AlexGuteniev, line 886: Wouldn't it be nicer to have `size_t` type here and avoid `static_cast` below?

StephanTLavavej: We could do that if you like, but then I think it would merit a comment that the implicit conversion is absolutely critical, since someone might be tempted to "improve" it back to `auto`. In contrast, the `static_cast` makes it clear that we need the conversion, and the immediate context of the multiplication and additions explains why, without the need for a comment.

AlexGuteniev: ok, agree

AlexGuteniev: Can we make `size_t{_V_pos}` then instead?

StephanTLavavej: Sure! It's sign-preserving, so there's no concern about narrowing conversions, and I think that's sufficiently explicit.

StephanTLavavej: Sorry, I forgot about x86, where this can truncate from `unsigned long long` to `size_t`. Have to go back to `static_cast`.

stl/src/vector_algorithms.cpp
```diff
@@ -884,7 +884,8 @@ namespace {
                             _BitScanForward(&_H_pos, _Mask); // lgtm [cpp/conditionallyuninitializedvariable]
 
                             const auto _V_pos = _Traits::_Get_v_pos(_Cur_idx_min, _H_pos); // Extract its vertical index
-                            _Res._Min         = _Base + _V_pos * 16 + _H_pos; // Finally, compute the pointer
+                            _Res._Min =
+                                _Base + static_cast<size_t>(_V_pos) * 16 + _H_pos; // Finally, compute the pointer
```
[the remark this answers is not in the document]
AlexGuteniev: https://github.com/microsoft/STL/pull/3619#discussion_r1156912405

> Sorry, I forgot about x86, where this can truncate from `unsigned long long` to `size_t`. Have to go back to `static_cast`.

StephanTLavavej:

strega-nil-ms (approved):

PR #3621, opened 2023-04-04
Build the import lib with `_ENFORCE_ONLY_CORE_HEADERS`

StephanTLavavej:
Related to (and inspired by) #2141.

This will need some MSVC-internal changes, but the impact is surprisingly small.

* In `stl/CMakeLists.txt`, we need to define `_ENFORCE_ONLY_CORE_HEADERS` when building the `msvcp${D_SUFFIX}_implib_objects`.
  + We don't want to modify the other place that the `IMPLIB_SOURCES` are mentioned - `libcpmt${FLAVOR_SUFFIX}` is the static lib.
* `nothrow.cpp` was including `<new>`, a non-core header, but it can just include `<vcruntime_new.h>` which is where `nothrow_t` is actually defined.
  + We can keep `_STD_BEGIN`/`_STD_END` if we include `<yvals_core.h>`.
  + I'm dropping the useless `// define nothrow` comment.
* `locale0.cpp` and `<xfacet>` are too difficult to untangle at this time. I'm just exempting them from this enforcement.


stl/src/locale0.cpp
```diff
@@ -3,6 +3,8 @@
 
 // class locale basic member functions
 
+#undef _ENFORCE_ONLY_CORE_HEADERS // TRANSITION, <xfacet> should be a core header
```
AlexGuteniev, line 6: Can/should we do it more precisely, like this:
```c++
#pragma push_macro(_ENFORCE_ONLY_CORE_HEADERS)
#undef _ENFORCE_ONLY_CORE_HEADERS // TRANSITION, <xfacet> should be a core header
#include <xfacet>
#pragma pop_macro(_ENFORCE_ONLY_CORE_HEADERS)
```

StephanTLavavej: I don't think that would achieve anything useful. This whole file is the scariest single file in the entire STL. We shouldn't be messing with it, so making the exemption fine-grained seems like unnecessary complexity.

strega-nil-ms (approved):

StephanTLavavej:

StephanTLavavej:
I've resolved a trivial adjacent-add conflict in `stl/src/locale0.cpp` by taking #2141's comment before this PR's `#undef`.

PR #3630, opened 2023-04-06
`vector_algorithms.cpp`: Add `vzeroupper`, so that it is there even in `/Od`

### LINKED ISSUE #3601: <algorithm>: vectorized algorithms hurt performance.
**Describe the bug**

Some STL algorithms will automatically use the AVX version if supported,  without calling `_mm256_zeroupper()` at the end of the function, which will slow down subsequent SSE code on some CPUs (e.g. Skylake).

[Why is this SSE code 6 times slower without VZEROUPPER on Skylake?](https://stackoverflow.com/questions/41303780/why-is-this-sse-code-6-times-slower-without-vzeroupper-on-skylake)

In following code, the second `calculate()` is much slower than the first `calculate()`. 

**Command-line test case**
```
C:\Temp>type repro.cpp

#include <Windows.h>
#include <vector>
#include <algorithm>

void calculate() {
    int t = ::GetTickCount();
    float a = 1.0f, b = 1.0f, m = 0.0f;
    for (int i = 0; i < 30000000; i++) {
        m = a * b;
        m -= a * 0.1f;
        m += b * 0.2f;
    }
    printf("time cost %d\n", ::GetTickCount() - t);
}

int main() {
    std::vector<float> v1;
    v1.resize(10);
    for (auto it = v1.begin(); it != v1.end(); it++)
        *it = (float)rand();
    std::vector<float> tmp(10);

    std::reverse_copy(v1.end() - 7, v1.end(), tmp.begin());
    calculate();

    std::reverse_copy(v1.end() - 8, v1.end(), tmp.begin());  // AVX code path is selected.
    calculate();
}

C:\Temp>cl /EHsc /W4 /WX .\repro.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.34.31937 for x86

repro.cpp
Microsoft (R) Incremental Linker Version 14.34.31937.0

/out:repro.exe
repro.obj

C:\Temp>.\repro.exe
time cost 172
time cost 296
```

AlexGuteniev:
Fix #3601

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -66,6 +66,14 @@ namespace {
     void _Advance_bytes(const void*& _Target, ptrdiff_t _Offset) noexcept {
         _Target = static_cast<const unsigned char*>(_Target) + _Offset;
     }
+
+    // TRANSITION, DevCom-10331414
+    struct _Zeroupper_on_exit {
+        ~_Zeroupper_on_exit() {
+            _mm256_zeroupper();
+        }
+    };
+
 } // unnamed namespace
 
 extern "C" {
@@ -89,6 +97,8 @@ __declspec(noalias) void __cdecl __std_swap_ranges_trivially_swappable_noalias(
             _Advance_bytes(_First1, 32);
             _Advance_bytes(_First2, 32);
         } while (_First1 != _Stop_at);
+
+        _mm256_zeroupper(); // TRANSITION, DevCom-10331414
     }
 
     constexpr size_t _Mask_16 = ~((static_cast<size_t>(1) << 4) - 1);
@@ -174,6 +184,8 @@ __declspec(noalias) void __cdecl __std_reverse_trivially_swappable_1(void* _Firs
             _mm256_storeu_si256(static_cast<__m256i*>(_Last), _Left_reversed);
             _Advance_bytes(_First, 32);
         } while (_First != _Stop_at);
+
+        _mm256_zeroupper(); // TRANSITION, DevCom-10331414
     }
 
     if (_Byte_length(_First, _Last) >= 32 && _Use_sse42()) {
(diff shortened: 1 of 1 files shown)


strega-nil-ms (approved):

stl/src/vector_algorithms.cpp
```diff
@@ -66,6 +66,14 @@ namespace {
     void _Advance_bytes(const void*& _Target, ptrdiff_t _Offset) noexcept {
         _Target = static_cast<const unsigned char*>(_Target) + _Offset;
     }
+
+    // TRANSITION, DevCom-10331414
+    struct _Zeroupper_on_exit {
```
StephanTLavavej, line 71: As this is a guard type, we should mark it as `[[nodiscard]]` to prevent classic misuse in the future.

stl/src/vector_algorithms.cpp
```diff
@@ -66,6 +66,14 @@ namespace {
     void _Advance_bytes(const void*& _Target, ptrdiff_t _Offset) noexcept {
         _Target = static_cast<const unsigned char*>(_Target) + _Offset;
     }
+
+    // TRANSITION, DevCom-10331414
+    struct _Zeroupper_on_exit {
+        ~_Zeroupper_on_exit() {
```
StephanTLavavej, line 72: We should also make this noncopyable.

StephanTLavavej (approved):

strega-nil-ms (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: I verified that all occurrences of `_Use_avx2()` have been updated, and that the guard type is used if and only if an early `return` is possible. FYI @strega-nil-ms I pushed a small commit to enhance the guard type after you approved (no functional change, just resists future misuse).

StephanTLavavej:

PR #3635, opened 2023-04-07
`<system_error>`: Optimize internal class `_System_error`

diamante0018:
Hello, from a conversation that occurred on the STL Discord server, I noted that a call to ctor of the internal class `_System_error` appeared to take some extra unnecessary steps when the error message was empty.
As suggested by Stephan I opened a pull request in an attempt to simplify the code.

The goal of this pull request is to skip entirely the call to the internal function `_Makestr` since it would have no effect on an empty string. I think my pull request does not have any unintended consequences. This is also my first contribution so please let me know if I messed something up. I have been lurking on the activity on this repo for a while and on the Discord server so I tried my best to follow the code conventions 😄

### DIFF
--- a/stl/inc/system_error
+++ b/stl/inc/system_error
@@ -473,6 +473,8 @@ private:
     }
 
 protected:
+    _System_error(error_code _Errcode) : runtime_error(_Errcode.message()), _Mycode(_Errcode) {}
+
     _System_error(error_code _Errcode, const string& _Message)
         : runtime_error(_Makestr(_Errcode, _Message)), _Mycode(_Errcode) {}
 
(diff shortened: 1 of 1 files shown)


StephanTLavavej (approved):

StephanTLavavej:
Looks perfect, thank you!

One note - in the future, we recommend avoiding force-pushing branches after creating PRs, because GitHub makes incremental reviewing difficult when this happens.

diamante0018:
> Looks perfect, thank you!
> 
> One note - in the future, we recommend avoiding force-pushing branches after creating PRs, because GitHub makes incremental reviewing difficult when this happens.

Thanks for the TIPs here and on Discord

stl/inc/system_error
```diff
@@ -473,6 +473,8 @@ private:
     }
 
 protected:
+    _System_error(error_code _Errcode) : runtime_error(_Errcode.message()), _Mycode(_Errcode) {}
```
AlexGuteniev, line 476: Maybe we should add `explicit` to the unary constructor to prevent (internal) misuse, even though `system_error` constructors are not `explicit`?

StephanTLavavej: I wouldn't object to that, but I don't think there's a strong need for it, as this type isn't widely used and doesn't appear as a parameter anywhere.

strega-nil-ms (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for improving the performance of these constructors, and congratulations on your first microsoft/STL commit! :rocket: :tada: :heart_eyes_cat:

This will ship in VS 2022 17.7 Preview 2.

PR #3645, opened 2023-04-10
Status Chart: Weekly updates, drop es-module-shims, hide Combined Wait

StephanTLavavej:
* Weekly updates.
* Dependency updates.
* `index.html`: Drop es-module-shims.
  + Import maps are now supported by the latest versions of all major browsers (and the Status Chart makes no attempt to support older versions).
* Hide 'Combined Wait' by default.
  + This line was an interesting idea, but it's more distracting than useful (in my opinion).
* Update `usernames_maintainers.txt`.

As usual, this contains a "DROP BEFORE MERGING: Regen files." commit to make the live preview work.

:chart_with_downwards_trend: Live preview: [stephantlavavej.github.io/STL/](https://stephantlavavej.github.io/STL/)
===

### DIFF
--- a/usernames_maintainers.txt
+++ b/usernames_maintainers.txt
@@ -10,4 +10,5 @@ cbezault
 MahmoudGSaleh
 mnatsuhara
 StephanTLavavej
+strega-nil
 strega-nil-ms
--- a/index.html
+++ b/index.html
@@ -8,8 +8,6 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>STL Status Chart</title>
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@primer/css@20.8.3/dist/primer.min.css" />
-    <script async src="https://cdn.jsdelivr.net/npm/es-module-shims@1.7.0/dist/es-module-shims.min.js"
-        crossorigin="anonymous"></script>
     <script type="importmap">
         { "imports": {
             "@kurkle/color": "https://cdn.jsdelivr.net/npm/@kurkle/color@0.3.2/dist/color.esm.min.js",
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,9 +26,9 @@
       }
     },
     "node_modules/@esbuild/android-arm": {
-      "version": "0.17.12",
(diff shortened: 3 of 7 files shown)


AlexGuteniev:
> Hide 'Combined Wait' by default.

huh? to me average charts are hidden (both)

StephanTLavavej:
Yes, the Average Age and Average Wait lines are already hidden by default (they were the first uses of the hide-by-default feature). Now this PR is also hiding Combined Wait by default, leaving only Combined Age.

I hid the averages long ago because they wildly bounce around and aren't terribly useful. (Notably, if we merge a bunch of new PRs, which is generally a good thing, the average age goes up.)

usernames_maintainers.txt
```diff
@@ -10,4 +10,5 @@ cbezault
 MahmoudGSaleh
 mnatsuhara
 StephanTLavavej
+strega-nil
```
strega-nil-ms, line 13: I'm trying to understand why this is being done?

StephanTLavavej: It's because you requested changes with that account, and the script needs to categorize everyone who has ever requested changes as either a maintainer (to affect the Wait lines) or a contributor.

StephanTLavavej: I could drop the Wait lines entirely, if we no longer think that this information is worth displaying.

AlexGuteniev: Maybe only the remaining shown by default one is worth keeping.

AlexGuteniev: Also mouthy produced PRs would possibly be curious

strega-nil-ms: Ah, fair enough; I don't have a problem with it, then, seems reasonable.

strega-nil-ms (approved):

usernames_maintainers.txt
StephanTLavavej: We talked about this at the weekly maintainer meeting - I'll remove the Wait line machinery in the next Status Chart PR. (In the future I could add monthly created PRs or other metrics; right now I'm not sure how useful they'd be.)

PR #3651, opened 2023-04-12
Toolset update: VS 2022 17.6 Preview 3

StephanTLavavej:
* Use PowerShell 7.3.3.
* Drop the Spectre libs.
  + We don't need them for testing.
* Drop unused `InstallPath` and `Nickname` parameters.
* Drop unused `$ReleaseInPath` and `$Sku` variables.
* Use `Update-AzConfig` to silence breaking change warnings.
  + The previous method stopped working.
* Enable trusted launch (secure boot), supported as of 2023-03-20.
* Python 3.11.3.
* New pool.
* VS 2022 17.6 Preview 3.
* Require CMake 3.26.0.
* Pass my new `/quiet` option to `ml[64]`.
  + This makes the STL's build output pretty again.
* Remove workarounds for DevCom-10265322 "Equality comparison with reversed order of `operator==` is sometimes buggy".
* Remove workarounds for VSO-1734935 "Copy elision fails in permissive mode".
  + Now we don't need the `is_permissive` detection machinery.
* Remove workaround for DevCom-10247495 "MSVC forces us to export some internal enumeration functions for working with C++ modules".
* Remove workaround for VSO-1582358 "Standard Library Modules: chrono formatting emits bogus error C3861: `'_Fill_tm'`: identifier not found".
* Remove workarounds for VSO-1592329 "Standard Library Modules: Parallel `sort()` emits bogus error C2065: `'_Atomic_counter_t'`: undeclared identifier".
* Remove workaround for VSO-1705654 "Standard Library Modules: `/analyze` ICE".
* Remove workaround for VSO-1726729 "REPORTED: EDG rejects `constexpr vector insert()` in debug mode with 'attempt to access run-time storage' errors".
* Remove workarounds for VSO-1161663 "REPORTED: \[Feedback\] EDG `constexpr` rejects `left <= right` when null pointers are passed by reference".
* Remove numerous workarounds for VSO-1726722 "REPORTED: EDG rejects `constexpr` `string`/`vector` in debug mode with 'attempt to access expired storage' errors, part 2".
* Fully remove workaround for VSO-1592409 "Standard Library Modules: `source_location::column()` misbehaves in user headers".
* Simplify `<atomic>`: `intrin0.inl.h` has been updated to always define `_HAS_ARM64_LOAD_ACQUIRE`.
* Keep 5 workarounds for VSO-1799670 "EDG rejects `constexpr vector` in debug mode with 'attempt to access expired storage' errors, part 3".
* Fix signed/unsigned mismatch warnings for ARM64 `__stlr`.
  + They're declared as:
    ```cpp
    __MACHINEARM64(void __stlr8(volatile unsigned __int8 * _Target, unsigned __int8 _Value))
    __MACHINEARM64(void __stlr16(volatile unsigned __int16 * _Target, unsigned __int16 _Value))
    __MACHINEARM64(void __stlr32(volatile unsigned __int32 * _Target, unsigned __int32 _Value))
    __MACHINEARM64(void __stlr64(volatile unsigned __int64 * _Target, unsigned __int64 _Value))
    ```
    So we need to `static_cast` the `_Desired` value.
* Fix signed/unsigned mismatch warnings for ARM64 `__load_acquire`.
  + They're declared as:
    ```cpp
    __MACHINEARM64(unsigned __int8 __load_acquire8(const volatile unsigned __int8 * _Target))
    __MACHINEARM64(unsigned __int16 __load_acquire16(const volatile unsigned __int16 * _Target))
    __MACHINEARM64(unsigned __int32 __load_acquire32(const volatile unsigned __int32 * _Target))
    __MACHINEARM64(unsigned __int64 __load_acquire64(const volatile unsigned __int64 * _Target))
    ```
    We need to `static_cast` the return value to match `__iso_volatile_load`:
    ```cpp
    __MACHINE(__int8 __iso_volatile_load8(const volatile __int8 *))
    __MACHINE(__int16 __iso_volatile_load16(const volatile __int16 *))
    __MACHINE(__int32 __iso_volatile_load32(const volatile __int32 *))
    __MACHINE(__int64 __iso_volatile_load64(const volatile __int64 *))
    ```
* `<atomic>`: Clang ARM64 lacks `__load_acquire`/`__stlr`.
  + I've filed llvm/llvm-project#62103 to request support.
* Improve `bug-report.md` Markdown.
  + Use headers instead of bold text.
    - This creates clearer sections, and prevents the headers from flowing together with following text.
  + Use `+` for nested bullet points.
    - This makes the structure a bit easier to see.
  + Remove a newline between "Option 1" and "Option 2" so the bulleted list is formatted densely.
* Update `bug-report.md` for VS 2022, add `/std:c++latest`.
  + This adds `/std:c++latest` to the example repro. Although it displays a lengthy notice, I believe it's valuable to include here. (Many bug reports will need it and it's easy for people new to the command line to forget it.)
  + This also updates the example versions to VS 2022 17.6 Preview 3, since we hadn't updated them since open-sourcing.
  + Finally, this shows how the Copy Info button can be used, and mentions that "we need only the first two lines".


StephanTLavavej:
I don't think it's necessary to update the Bug Report issue template for *every* toolset update, as it clearly depicts example output, but I agree that an update right now would be reasonable, since it's still showing VS 2019. I'll add those changes, thanks.

strega-nil-ms (approved):

PR #3655, opened 2023-04-15
`std::atomic<std::shared_ptr>::wait` should compare control blocks

### LINKED ISSUE #3602: `std::atomic<std::shared_ptr>::wait` does not seem to care about control block difference. Is this a bug?
### Discussed in https://github.com/microsoft/STL/discussions/3555

<div type='discussions-op-text'>

<sup>Originally posted by **zwhfly** March 10, 2023</sup>
This code seems to block forever under MSVC, but returns successfully under <del>gcc/clang</del>[libstdc++](https://godbolt.org/z/hG8e6oYKn).

```
#include <memory>
#include <atomic>


int main()
{
    auto sp1 = std::make_shared<char>();
    auto holder = [sp1] {};
    auto sp2 = std::make_shared<decltype(holder)>(holder);
    std::shared_ptr<char> sp3{ sp2, sp1.get() };

    std::atomic<std::shared_ptr<char>> asp{ sp1 };
    asp.wait(sp3);

    return 0;
}
```

It is said [here](https://en.cppreference.com/w/cpp/memory/shared_ptr/atomic2#:~:text=atomic%3Cshared_ptr%3CT%3E%3E%3A%3Await) that

> (atomic<shared_ptr<T>>::wait) compares load(order) with old and if they are equivalent then blocks ...

> Notes: two shared_ptrs are equivalent if they store the same pointer and either share ownership or are both empty.

I thought that `sp1` and `sp3` do not have the same control block, so they do not share ownership. Right? So the `wait` call should return immediately.

But by checking the source code, it seems that the STL implementation for `std::atomic<std::shared_ptr>::wait` compares the object pointer (the `char *`) only. It does not compare the control block pointer at all.

Is this a bug in STL? Or what am I missing here?
</div>

fsb4000:
Fixes #3602


stl/inc/memory
```diff
@@ -3870,10 +3870,10 @@ protected:
     _Atomic_ptr_base(remove_extent_t<_Ty>* const _Px, _Ref_count_base* const _Ref) noexcept
         : _Ptr(_Px), _Repptr(_Ref) {}
 
-    void _Wait(remove_extent_t<_Ty>* _Old, memory_order) const noexcept {
+    void _Wait(remove_extent_t<_Ty>* _Old, _Ref_count_base* const _Old_Rep, memory_order) const noexcept {
         for (;;) {
             auto _Rep   = _Repptr._Lock_and_load();
-            bool _Equal = _Ptr.load(memory_order_relaxed) == _Old;
+            bool _Equal = _Ptr.load(memory_order_relaxed) == _Old && (_Old == nullptr || _Rep == _Old_Rep);
```
AlexGuteniev, line 3876: I don't think this is a correct and complete fix.
It handles different control blocks in this check, but not in the `__std_atomic_wait_direct` below.

AlexGuteniev (approved):

AlexGuteniev:
I'm also afraid that this may not be possible to fix without an ABI break.

stl/inc/memory
fsb4000: Ok. If I understand correctly then we don't handle this situation:

1) `_Ptr == _Old` and `_Rep` == `_Old_Rep`
2) Call `__std_atomic_wait_direct(&_Ptr, &_Old);`
3) Change `_Old_Rep`, but `_Ptr == _Old` and we won't notice that  `_Rep` != `_Old_Rep` and still waiting...

AlexGuteniev: Actually I've looked into the implementation in more detail, and it is more complicated

We are waiting for `_Rep` to change, not `_Ptr` in `_Atomic_ptr_base`

So this does handle different control blocks, but does not seem to handle aliased shared pointers with one control blocks.

I think then your fix is a good thing, and regarding aliased pointers with single control blocks we can have a separate issue. This situation is hard to create anyway, most of the time aliased pointers have different types, so don't compare.

AlexGuteniev: Hm, no, we are waiting on `_Ptr`...

StephanTLavavej:
@fsb4000 @AlexGuteniev It looks like this PR is stalled, needing further investigation. @strega-nil-ms suggests that we could check libc++'s implementation to see what they do (however, if it extends to copying actual code, we'll need the usual dance of clearly separating that code and marking it with the LLVM banner); she'll chime in with additional suggestions here. If we can't make progress here, should we close this PR without merging, and update the bug with what we've learned so far?

AlexGuteniev:
> we could check libc++'s implementation to see what they do

They do nothing. They don't support `std::atomic<std::shared_ptr<T>>`

https://godbolt.org/z/TbTYGq4Mn

AlexGuteniev:
I think it is not fixable until `vNext`. 
I may be wrong though.

I suggest that in addition to the provided repro we need a modified repro like this

```c++
#include <atomic>
#include <chrono>
#include <memory>
#include <thread>

int main()
{
    auto sp1 = std::make_shared<char>();
    auto holder = [sp1] {};
    auto sp2 = std::make_shared<decltype(holder)>(holder);
    std::shared_ptr<char> sp3{ sp2, sp1.get() };

    std::atomic<std::shared_ptr<char>> asp{ sp3 };

    std::thread t([&] 
    {
        std::this_thread::sleep_for(std::chrono::milliseconds(100));
        asp = sp1; 
        asp.notify_one();  
    } );
    
    asp.wait(sp3);

    t.join();

    return 0;
}
```

This way we test for what I think currently is missed from this PR.

StephanTLavavej:
If the current implementation is broken, we might be able to fix it in v14, under the theory that usage is still low, and people shouldn't have been depending on the broken impl.

AlexGuteniev:
> If the current implementation is broken

I think it is fine for the most of the usages.
The repro looks like a contrived scenario. 
One would not normally use same object pointer with different deleters,
So a breaking fix may do more harm than the bug.

zwhfly:
Here's a non-contrived scenario, but it's about `std::atomic<std::weak_ptr>` instead of `std::atomic<std::shared_ptr>`.

When user code waits on an expired `std::atomic<std::weak_ptr<T>>`, the current implementation waits on the `atomic<T *>`, which is a dangling pointer to `T`. After that, if the atomic was updated with a new weak_ptr but reused the old address for `T`, the waiting side would not unblock. I believe conforming implementations should unblock in such scenario.

Reproducing code:
```c++
std::atomic<std::weak_ptr<int>> awp;

int get_int()
{
    for (;;)
    {
        auto wp = awp.load();
        if (auto sp = wp.lock())
            return *sp;
        awp.wait(wp);
    }
}

int main()
{
    std::shared_ptr<int> sp{ std::make_unique<int>() };
    awp.store(std::weak_ptr<int>{ sp });

    auto addr = reinterpret_cast<std::uintptr_t>(sp.get());
    std::printf("%" PRIxPTR "\n", addr);

    sp.reset();

    std::unique_ptr<int> up;

    //demand the old address
    for (;;)
    {
        up = std::make_unique<int>();
        auto addr2 = reinterpret_cast<std::uintptr_t>(up.get());
        std::printf("%" PRIxPTR "\n", addr2);
        if (addr2 == addr)
            break;
    }

    std::future<int> future = std::async(std::launch::async, get_int);

    using namespace std::chrono_literals;
    std::this_thread::sleep_for(2s);

    std::shared_ptr<int> sp2{ std::move(up) };
    awp.store(std::weak_ptr<int>{ sp2 });
    awp.notify_one();

    return future.get();
}
```

StephanTLavavej (approved):

StephanTLavavej:
After hiding from this PR for a year :joy_cat:, I've finally looked at it and I believe I'm comfortable with the change (after pushing a fix/test for "owning the null pointer"). I'm definitely happy that @AlexGuteniev's suggested test case has been added, I don't think that this poses significant regression risks, and I don't think that there's any vNext interaction (this is just a behavioral change; mix-and-match scenarios shouldn't be harmed beyond maybe not getting the fix).

AlexGuteniev:
I'm afraid the PR doesn't fully fix hang in any of the scenario - it only makes hangs unlikely, but still possible, thus introducing inconsistency.

AlexGuteniev (approved):

StephanTLavavej (approved):

StephanTLavavej:

stl/inc/memory
```diff
@@ -4000,15 +4000,18 @@ protected:
...
         : _Ptr(_Px), _Repptr(_Ref) {}
 
-    void _Wait(remove_extent_t<_Ty>* _Old, memory_order) const noexcept {
+    void _Wait(remove_extent_t<_Ty>* _Old_ptr, _Ref_count_base* const _Old_rep, memory_order) const noexcept {
+        unsigned long _Remaining_timeout = 16; // milliseconds
+        const unsigned long _Max_timeout = 1048576; // milliseconds, ~17.5 minutes
         for (;;) {
             auto _Rep   = _Repptr._Lock_and_load();
-            bool _Equal = _Ptr.load(memory_order_relaxed) == _Old;
+            bool _Equal = _Ptr.load(memory_order_relaxed) == _Old_ptr && _Rep == _Old_rep;
```
CaseyCarter, line 4008: No change requested; we should do this in a followup if at all to avoid resetting testing, since it's not a correctness issue. Should this be reordered as:
```suggestion
            bool _Equal = _Rep == _Old_rep && _Ptr.load(memory_order_relaxed) == _Old_ptr;
```
so we short-circuit when the core-local `_Rep == _Old_rep` is `false` before potentially loading `_Ptr`'s cache line from some other core's data cache? (I suspect I understand cache coherence protocols just well enough to be dangerous.) Would any difference just be noise compared to the expense of `_Repptr._Lock_and_load()`?

CaseyCarter (approved):

stl/inc/memory
CaseyCarter: (Pulling in @AlexGuteniev to render an opinion and/or tell me how wrong I am 😄.)

StephanTLavavej: My vague understanding is that a relaxed load has no barriers, so it has no special costs. It would be fine to reorder, though, since the `.load` is at least a debug mode function call.

Alcaro: Aren't _Repptr and _Ptr in the same cache line? Sure, whatever they point to can be anywhere, but the pointers themselves are (as far as I can see) right beside each other.

(STL's comment sounds accurate, though.)

AlexGuteniev: I agree that they are always in the same cache line (it is really always, not just most of the time, because we over align to `alignas(2 * sizeof(void*))`). So from the cache perspective it does not matter.

I agree also that eliminating debug mode call can be potentially beneficial.

Still I more like the original order, because we compare `_Rep == _Old_rep;` the very last thing, as close to wait as possible, so reduce the probability of resorting to timeout.

Alcaro: Considering _Rep is just a stack-local and the real load is on the line above, I can't quite follow your reasoning.

AlexGuteniev: > Considering _Rep is just a stack-local and the real load is on the line above, I can't quite follow your reasoning.

Hm, you are right, we done loading it into memory at the same time anyway.

Then debug mode call saving could be the main decision factor here.

StephanTLavavej:
Thanks for fixing this runtime correctness bug - what a rollercoaster of despair and elation! :chart_with_downwards_trend: :chart_with_upwards_trend: :joy_cat: :tada:

stl/inc/memory
CaseyCarter: Ah, I'd forgotten that we overalign - my cache coherence point is invalid. +1 to avoiding the call in debug compiles, though. We _really_ want to minimize the time between load and check.

PR #3658, opened 2023-04-17
Ensure _Idx is only worked on if it has to be.

AtariDreams:
I don't know if whatever compile is supposed to compile this will notice, but this is the only place where this structure appears, so it seems best that the incrementation be done within the loop.

### DIFF
--- a/stl/inc/valarray
+++ b/stl/inc/valarray
@@ -1537,7 +1537,8 @@ public:
             _Ans += _Indexarr[_Idx] * _Stride[_Idx]; // compute offset
         }
 
-        while (0 < _Idx--) {
+        while (0 < _Idx) {
+            --_Idx;
             if (++_Indexarr[_Idx] < _Len[_Idx]) {
                 break; // increment done, quit
             } else {


stl/inc/valarray
```diff
@@ -1537,7 +1537,8 @@ public:
             _Ans += _Indexarr[_Idx] * _Stride[_Idx]; // compute offset
         }
 
-        while (0 < _Idx--) {
+        while (0 < _Idx) {
+            --_Idx;
```
AlexGuteniev, line 1541: I'm not sure what's the motivation.
If the performance, then I don't think there's noticeable codegen difference.
If the clarity  then with `0 < _Idx` comparison and other stuff here it is still not much better

AlexGuteniev: On another thought, `0 < _Idx` -> `_Idx > 0` can be done in another change

strega-nil-ms (approved):
Small change, but appreciated.

StephanTLavavej (approved):

StephanTLavavej:
This is correct (no effective behavioral change) because `_Idx` isn't observed after the loop, so it doesn't matter that we're changing from wrapping around the `size_t` to leaving it at `0`.

StephanTLavavej:

PR #3660, opened 2023-04-18
`<map>`, `<unordered_map>`: Allow `MapLike<K, V&>` again

StephanTLavavej:
One of the changes that I pushed to #2436 was damaging - while map-like types can't tolerate references as key types, using references as mapped types appears to be permitted (by the Standard, libstdc++, and libc++), and there are 4 projects across vcpkg and Real World Code that are doing this.

Fortunately, we can fix this before changes stop flowing into VS 2022 17.7 Preview 2, so users will never be affected. I'm removing my bad `static_assert`s, and adding test coverage to minimally verify that this scenario compiles.

(I suppose we could instead assert that the mapped types aren't function types or void types, but I'd rather just drop it entirely. Such usage is extremely unlikely and will still be rejected, just with lengthy diagnostics.)

### DIFF
--- a/stl/inc/map
+++ b/stl/inc/map
@@ -77,8 +77,6 @@ public:
         _MISMATCHED_ALLOCATOR_MESSAGE("map<Key, Value, Compare, Allocator>", "pair<const Key, Value>"));
     static_assert(is_object_v<_Kty>, "The C++ Standard forbids containers of non-object types "
                                      "because of [container.requirements].");
-    static_assert(is_object_v<_Ty>, "The C++ Standard forbids containers of non-object types "
-                                    "because of [container.requirements].");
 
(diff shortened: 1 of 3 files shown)


barcharcraz (approved):

StephanTLavavej:

cjdb:
> using references as mapped types appears to be permitted (by the Standard

[[associative.reqmts.general]/8](https://eel.is/c++draft/associative.reqmts.general#8) says that the rules applying to `value_type` instead apply to `key_type` and `mapped_type` for our map types. If references are allowed for `mapped_type`, doesn't that imply they're also allowed for `key_type` and other containers' `value_type`?

CaseyCarter:
I think the cross-references in [associative.reqmts.general] are swapped; it doesn't make sense to apply the `value_type` requirements from the allocator-aware requirements to `key_type` and `mapped_type`. https://github.com/cplusplus/draft/pull/7249 tries to fix them.

PR #3662, opened 2023-04-19
comparison of identical zoned_time objects fails

Bernhard-L:
the reason for the error in comparison is, that .get_time_zone() returns a _TimeZonePtr so in this case, the adresses of the pointers are compared

   using timePoint_T = std::chrono::zoned_time<std::chrono::system_clock::duration>;

   auto now = std::chrono::system_clock::now();
   timePoint_T x{std::chrono::current_zone(), now};
   timePoint_T y{std::chrono::current_zone(), now};

   assert (x == y);

### DIFF
--- a/stl/inc/chrono
+++ b/stl/inc/chrono
@@ -2560,7 +2560,7 @@ namespace chrono {
     _EXPORT_STD template <class _Duration1, class _Duration2, class _TimeZonePtr>
     _NODISCARD bool operator==(
         const zoned_time<_Duration1, _TimeZonePtr>& _Left, const zoned_time<_Duration2, _TimeZonePtr>& _Right) {
-        return _Left.get_time_zone() == _Right.get_time_zone() && _Left.get_sys_time() == _Right.get_sys_time();
+        return *(_Left.get_time_zone()) == *(_Right.get_time_zone()) && _Left.get_sys_time() == _Right.get_sys_time();
     }
 
     // [time.clock.utc]


cpplearner:
> so in this case, the adresses of the pointers are compared

No? The *values* of these pointers are compared, and `current_zone()` should always return the same pointer value.

And indeed, this test case prints `true` for me.
```c++
#include <cassert>
#include <chrono>
#include <iostream>

using namespace std;
using namespace std::chrono;

int main() {
    using timePoint = zoned_time<system_clock::duration>;

    auto now = system_clock::now();
    timePoint x{current_zone(), now};
    timePoint y{current_zone(), now};

    cout.setf(cout.boolalpha);
    cout << (x == y) << '\n';
}
```

StephanTLavavej:
Thanks for looking into this - we talked about this at the weekly maintainer meeting and we agree with @cpplearner's assessment that the implementation is correct as written and exactly conforms to the C++ Standard's specification (see https://eel.is/c++draft/time.zone.zonedtime.nonmembers#1 ). If you still believe that there's a bug here, please open an issue with a self-contained test case that demonstrates incorrect behavior.

Bernhard-L:
@StephanTLavavej and @cpplearner but this only works as long as the timezone gets the same adress.
We encapsulated this (in our project) in a static lib and this static lib gets included in different dlls.
I noticed, that the adress of the same timezone got different adresses and therefore the comparison failed.
Our current solution is, to derive the zoned time and overload the operator=
`
	class timePoint_T: public std::chrono::zoned_time<std::chrono::system_clock::duration>
	...
bool operator==(const timePoint_T& that) const { return *get_time_zone() == *(that.get_time_zone()) && get_sys_time() == that.get_sys_time(); }
`

and thats working with the static libs and the dlls, too

PR #3673, opened 2023-04-26
`<mdspan>`: `layout_stride` improvements

JMazurkiewicz:
* Implement WG21-P2763R1,
* Implement preconditions (currently unconditional),
* Implement (incomplete) regular tests and death tests.


tests/std/tests/P0009R18_mdspan_layout_stride/test.cpp
```diff
@@ -0,0 +1,150 @@
...
+        static_assert(Mapping::is_always_strided());
+    }
+
+    { // Check 'is_[unique/strided]' functions
+        static_assert(Mapping::is_unique());
+        static_assert(Mapping::is_strided());
+        // Tests of 'is_exhaustive' are defined in 'check_is_exhaustive' function [FIXME]
+    }
+
+    { // Check 'stride' function (intentionally not if constexpr)
```
StephanTLavavej, line 119: I'm confused by "(intentionally not if constexpr)" here - I don't see any `if` at all?

tests/std/tests/P0009R18_mdspan_layout_stride/test.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include <span>
+#include <type_traits>
+
+#include "test_mdspan_support.hpp"
+
+using namespace std;
+
+template <class IndexType, size_t... Extents, size_t... Indices>
+constexpr void do_check_members(const extents<IndexType, Extents...>& ext,
+    const array<IndexType, sizeof...(Extents)> strs, index_sequence<Indices...>) {
```
StephanTLavavej, line 18: Suggestion: I would usually expect `array` to be passed by const reference, not by const value, although this doesn't affect test correctness or lead to test instability. Occurs below.

StephanTLavavej (approved):

StephanTLavavej:
Just a small question and suggestion, but I went ahead and merged this - thank you! :heart_eyes_cat: :rocket: :white_check_mark:

tests/std/tests/P0009R18_mdspan_layout_stride/test.cpp
JMazurkiewicz: `if` statement used to be here, I forgot to fix the comment

PR #3677, opened 2023-04-28
`<typeinfo>`: Error with `import std;` and `typeid(T).name()`

StephanTLavavej:
@cdacamar explained that in this case, the MSVC compiler looks for the True Name of `type_info` in the global namespace, so we need to export it as a special exception.

I'm adding test coverage for this scenario. The `using namespace std;` in this test doesn't actually affect the repro, so I didn't bother to remove it.

Fixes DevCom-10349749 and internal VSO-1810086 / AB#1810086 .

### DIFF
--- a/stl/inc/typeinfo
+++ b/stl/inc/typeinfo
@@ -22,9 +22,7 @@ _STL_DISABLE_CLANG_WARNINGS
 
 #include <vcruntime_typeinfo.h>
 
-extern "C++" { // attach declarations to the global module, see N4944 [module.unit]/7
-_EXPORT_STD class type_info; // for typeid, MSVC looks for type_info in the global namespace
-} // extern "C++"
+_EXPORT_STD extern "C++" class type_info; // for typeid, MSVC looks for type_info in the global namespace
 
 _STD_BEGIN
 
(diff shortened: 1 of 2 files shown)


strega-nil-ms:
Is this conforming? I'm okay with "no, but it needs to not be conforming for our implementation", but I'm wondering if it'd be possible to make the compiler search in `std::` here, rather than continuing the non-conformance.

StephanTLavavej:
@cdacamar said on Discord:

> I suppose I could change that, but I don't want to introduce a compatibility problem 😦

The Standardese that I wrote, N4944 \[std.modules\]/6, says:

> *Recommended practice:* Implementations should avoid exporting any other declarations from the C++ library modules.

IMO it's slightly nonconforming to emit `::type_info` because it could clash with user code, but the same is also true for classic `<type_info>` (it's not a reserved name, nor is it one of the C names that is in a superposition between global and `std`).

In practice I think it's easiest to just make this special exception and not worry about it further.

CaseyCarter (approved):

CaseyCarter:
I suppose I can live with this as extending our pre-existing non-conformance into the Modules Universe.

StephanTLavavej:

Issue #1520, opened 2020-12-09
<chrono>: system_clock could be off after a future leap second


statementreply:
**Describe the bug**

Since Windows 10 1809 and Windows Server 2019, `FILETIME` includes leap seconds after 2018-06 (configurable system-wide, enabled by default), and will no longer be a constant offset from `system_clock`. `system_clock::now` needs to be updated to account for future leap seconds.

**Command-line test case**

1. Run the following command *once* as administrator. **Note that this inserts a fake leap second (2020-11-30T23:59:60Z). The system clock and file modification times after 2020-11-30 will be off by 1 second until Windows reboot.**

   ```
   C:\WINDOWS\system32>w32tm /leapseconds /add:+2020-11-30T23:59:59 /force /testmode
   成功地执行了命令。
   ```

   The following command can be used to check the status of leap seconds.

   ```
   C:\WINDOWS\system32>w32tm /leapseconds /getstatus /verbose
   [闰秒]
   Enabled: 1 (Local)
   Number of Leap Seconds (after June 2018): 1 (Local)
   Leap Seconds List (Local):
   +2020-11-30T23:59:59
   ```

2. Run the test case.

   ```
   D:\Temp>type leap_second_2.cpp
   ```
   ```cpp
   #include <Windows.h>
   #include <chrono>
   #include <cstdio>
   
   using namespace std;
   using namespace std::chrono;
   
   int main() {
       SYSTEMTIME win_sys_time;
       GetSystemTime(&win_sys_time);
   
       const auto stl_sys_time = system_clock::now();
       const auto stl_total = stl_sys_time.time_since_epoch();
       const auto stl_days = floor<duration<int, ratio<86400>>>(stl_total);
       const auto stl_hours = floor<duration<int, ratio<3600>>>(stl_total - stl_days);
       const auto stl_minutes = floor<duration<int, ratio<60>>>(stl_total - stl_days - stl_hours);
       const auto stl_seconds = floor<duration<int>>(stl_total - stl_days - stl_hours - stl_minutes);
       const auto stl_100ns =
           floor<duration<int, ratio<1, 10'000'000>>>(stl_total - stl_days - stl_hours - stl_minutes - stl_seconds);
   
       printf("Windows: %.4d-%.2d-%.2d %.2d:%.2d:%.2d.%.3d\n", win_sys_time.wYear, win_sys_time.wMonth, win_sys_time.wDay,
[...]

PR #3681, opened 2023-05-02
`<__msvc_chrono.hpp>`: Use `_Xtime_get_ticks` in `_To_timespec64_sys_10_day_clamped` for consistency

statementreply:
Part of #1588.

`_Xtime_get_ticks` and `system_clock::now` would no longer be equal after a future leap second when #1520 is eventually fixed. This PR changes the clock used in `_To_timespec64_sys_10_day_clamped` to be consistent with code that consumes its result, e.g. `target` in https://github.com/microsoft/STL/blob/091cad2eaaa5bc25873eb7261cae57ab123592f3/stl/src/cond.cpp#L72-L80

@BillyONeal pointed out (on Discord) that this PR could break (already somewhat broken) timed wait ABI:
> `_To_xtime_10_day_clamped`, used by `condition_variable::wait_for`, `this_thread::sleep_for`, etc., calls `system_clock::now()` internally to compute a value interpreted as `GetSystemTimeAsFileTime + offset` by another component. My understanding is that if the linker picks the old version of `_To_xtime_10_day_clamped` and the new version of `system_clock::now()` from different object files, all those wait functions could wait for a few seconds too short (assuming positive leap seconds) after future leap seconds.

### DIFF
--- a/stl/inc/__msvc_chrono.hpp
+++ b/stl/inc/__msvc_chrono.hpp
@@ -703,7 +703,7 @@ _NODISCARD bool _To_timespec64_sys_10_day_clamped(
     // Every function calling this one is TRANSITION, ABI
     constexpr _CHRONO nanoseconds _Ten_days{_CHRONO hours{24} * 10};
     constexpr _CHRONO duration<double> _Ten_days_d{_Ten_days};
-    _CHRONO nanoseconds _Tx0 = _CHRONO system_clock::now().time_since_epoch();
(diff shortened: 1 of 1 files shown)


StephanTLavavej (approved):

StephanTLavavej:
Looks good to me - no behavioral change now, and the consistency argument for the future is convincing. Thanks!

strega-nil-ms (approved):

PR #3682, opened 2023-05-04
`<mdspan>`: Diversify integer types used in tests

JMazurkiewicz:
* Improve tests of `extents` and layout mappings:
  * Check all 10 standard signed and unsigned integer types,
  * Check various combinations of static and dynamic extents.
* Fix integral conversions in `layout_stride::mapping`.

### DIFF
--- a/stl/inc/mdspan
+++ b/stl/inc/mdspan
@@ -601,7 +601,7 @@ public:
             _Strides.back() = 1;
             for (rank_type _Idx = extents_type::_Rank - 1; _Idx-- > 0;) {
                 // TRANSITION USE `_Multiply_with_overflow_check` IN DEBUG MODE
-                _Strides[_Idx] = _Strides[_Idx + 1] * _Exts.extent(_Idx + 1);
+                _Strides[_Idx] = static_cast<index_type>(_Strides[_Idx + 1] * _Exts.extent(_Idx + 1));
             }
         }
     }
@@ -752,7 +752,7 @@ public:
             }
 
             for (rank_type _Idx = 0; _Idx < extents_type::_Rank; ++_Idx) {
-                if (_Left.stride(_Idx) != _Right.stride(_Idx)) {
+                if (_STD cmp_not_equal(_Left.stride(_Idx), _Right.stride(_Idx))) {
                     return false;
                 }
             }
@@ -786,7 +786,7 @@ private:
     template <class... _IndexTypes, size_t... _Seq>
     _NODISCARD constexpr index_type _Index_impl(index_sequence<_Seq...>, _IndexTypes... _Indices) const noexcept {
         _STL_INTERNAL_STATIC_ASSERT((same_as<_IndexTypes, index_type> && ...));
-        return ((_Indices * _Strides[_Seq]) + ... + 0);
+        return static_cast<index_type>(((_Indices * _Strides[_Seq]) + ... + 0));
     }
 };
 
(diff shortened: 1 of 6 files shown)


stl/inc/mdspan
```diff
@@ -786,7 +786,7 @@ private:
     template <class... _IndexTypes, size_t... _Seq>
     _NODISCARD constexpr index_type _Index_impl(index_sequence<_Seq...>, _IndexTypes... _Indices) const noexcept {
         _STL_INTERNAL_STATIC_ASSERT((same_as<_IndexTypes, index_type> && ...));
-        return ((_Indices * _Strides[_Seq]) + ... + 0);
+        return static_cast<index_type>(((_Indices * _Strides[_Seq]) + ... + 0));
```
MattStephanson, line 789: Is the issue here just the (implicitly `int`) zero in the fold init? `_Strides` is an `array<index_type>` and `_Indices` is casted to `index_type` at the call site. If that's the case, then maybe
```suggestion
        return ((_Indices * _Strides[_Seq]) + ... + index_type{0});
```
would avoid any integer promotions.

MattStephanson: Never mind, I understand now the integer promotions happen anyway, even when the types match.

tests/std/tests/P0009R18_mdspan_layout_stride/test.cpp
```diff
@@ -131,15 +139,46 @@ constexpr void do_check_members(const extents<IndexType, Extents...>& ext,
...
+    check_members_with_different_strides_index_type<signed char>(ext, strides);
+    check_members_with_different_strides_index_type<short>(ext, strides);
+    check_members_with_different_strides_index_type<int>(ext, strides);
+    check_members_with_different_strides_index_type<long>(ext, strides);
+    check_members_with_different_strides_index_type<long long>(ext, strides);
+
+    // Check unsigned strides
+    check_members_with_different_strides_index_type<unsigned char>(ext, strides);
+    check_members_with_different_strides_index_type<unsigned short>(ext, strides);
+    check_members_with_different_strides_index_type<unsigned>(ext, strides);
```
StephanTLavavej, line 162: `unsigned int` again.

tests/std/tests/P0009R18_mdspan_layout_stride/test.cpp
```diff
@@ -131,15 +139,46 @@ constexpr void do_check_members(const extents<IndexType, Extents...>& ext,
...
+    check_members(extents<signed char, 5>{5}, array{1});
+    check_members(extents<short, 6, 7>{}, array{1, 6});
+    check_members(extents<int, 3, dynamic_extent>{3}, array{1, 3});
+    check_members(extents<long, 4>{}, array{1});
+    check_members(extents<long long, 3, 2, dynamic_extent>{3}, array{1, 3, 6});
+
+    // Check unsigned integers
+    check_members(extents<unsigned char, 5>{5}, array{1});
+    check_members(extents<unsigned short, 6, 7>{}, array{1, 6});
+    check_members(extents<unsigned, 3, dynamic_extent>{3}, array{1, 3});
```
StephanTLavavej, line 178: Ditto.

StephanTLavavej (approved):

StephanTLavavej:
I pushed minor style nitpicks and will merge after the tests pass. Thanks!

PR #3683, opened 2023-05-07
Status Chart: Remove Average Wait and Combined Wait

StephanTLavavej:
In addition to the usual weekly/dependency updates, this removes the Average Wait and Combined Wait lines. They were never very useful, and became even less so after we dramatically reduced the PR backlog. This simplifies the chart machinery by a significant amount, avoids the need to track maintainer/contributor usernames, and also accelerates chart generation because our GraphQL query can request significantly less data.

As this involves coordinated changes between TypeScript and HTML, this PR contains a "REGEN BEFORE MERGING" commit which will accumulate merge conflicts until I update and force-push it before merging.

:chart_with_downwards_trend: Live preview: [stephantlavavej.github.io/STL/](https://stephantlavavej.github.io/STL/)
===


CaseyCarter (approved):
Awesome that we no longer need to maintain the lists of contributors and maintainers!

StephanTLavavej:
Thanks for the review! Since it's Friday, I added one more line of weekly updates before merging.

PR #3688, opened 2023-05-10
`<mdspan>`: Add modules test

JMazurkiewicz:
* Add test in `test_header_units_and_modules.hpp` (addresses https://github.com/microsoft/STL/pull/3563#discussion_r1133984594)
  * Currently checks `layout_right::mapping` (not `mdspan` itself) due to ~~yet unreported MSVC bug~~ DevCom-10359857.
*  Add missing `<mdspan>` header in `include_each_header_alone_matrix.lst`.


tests/std/include/test_header_units_and_modules.hpp
```diff
@@ -389,6 +389,17 @@ void test_map() {
     assert(m[30] == 33);
 }
 
+void test_mdspan() {
```
StephanTLavavej, line 392: Need to actually call this from `all_cpp_header_tests()` below.

tests/std/include/test_header_units_and_modules.hpp
```diff
@@ -389,6 +389,17 @@ void test_map() {
...
 }
 
+void test_mdspan() {
+    using namespace std;
+    puts("Testing <mdspan>.");
+    int arr[] = {1, 0, 0, 0, 1, 0, 0, 0, 1};
+    layout_right::mapping<extents<int, 3, 3>> mp;
+    assert(arr[mp(0, 0)] == 1);
+    assert(arr[mp(1, 1)] == 1);
+    assert(arr[mp(2, 2)] == 1);
```
StephanTLavavej, line 399: This is good, but following the "unique numbers for unique purposes" principle I always recommend, I think we should use distinct numbers instead of `1`. This would catch anything being reversed, for example.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! I pushed test coverage improvements and will merge after it passes.

PR #3689, opened 2023-05-10
`<mdspan>`: Implement multidimensional subscript operator

JMazurkiewicz:
* Implement multidimensional subscript operator (Clang only),
* Remove `operator()` in all modes,
* Use `span` as an argument to `mdspan::_Index_impl`,
  * Clang required workaround: instead of writing `(*this)[indices...]` I had to use `this->operator[](indices...)` syntax (https://godbolt.org/z/shfq7TPEx).
* Tests now use `operator[indices...]` when possible, `operator[array{indices...}]` otherwise.
* Drive-by: address comments from my previous PRs.


stl/inc/mdspan
```diff
@@ -935,14 +935,15 @@ public:
     constexpr mdspan& operator=(const mdspan&) = default;
     constexpr mdspan& operator=(mdspan&&)      = default;
 
-    // TRANSITION, P2128R6 (Multidimensional subscript operator)
+#ifdef __clang__ // TRANSITION, P2128R6
```
StephanTLavavej, line 938: Should we consider testing `__cpp_multidimensional_subscript` instead? Writing "light-up" code has advantages and disadvantages:

* We properly react to newer compilers instead of having to change the STL again
* However, because the compiler can activate new STL codepaths without our explicit blessing, this can result in essentially untested configurations being immediately experienced by users
  + However x2, since MSVC compiler development and EDG updates run the STL's test suite, this would prevent broken compiler support from shipping.

I think this argues in favor of using the feature-test macro.

stl/inc/mdspan
```diff
@@ -1026,9 +1027,13 @@ public:
     }
 
 private:
-    template <class _IndexContainer, size_t... _Idx>
-    _NODISCARD constexpr reference _Index_impl(_IndexContainer&& _Indices, index_sequence<_Idx...>) const {
-        return _Acc.access(_Ptr, _Map(_STD as_const(_STD forward<_IndexContainer>(_Indices)[_Idx])...));
+    template <class _OtherIndexType, size_t... _Seq>
+    _NODISCARD constexpr reference _Index_impl(span<_OtherIndexType, rank()> _Indices, index_sequence<_Seq...>) const {
+#ifdef __clang__ // TRANSITION, P2128R6
+        return this->operator[](_STD as_const(_Indices[_Seq])...);
```
StephanTLavavej, line 1033: Do we need `this->` here? It appears to be a direct member, so I think this is unnecessary.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! Merged with a couple of suggestions for later.

PR #3690, opened 2023-05-10
`<mdspan>`: Remove some C28020 warning suppressions

JMazurkiewicz:
* Remove all suppressions from `<mdspan>` header,
* Remove suppressions from tests where possible,
* Reduce the size of code with suppressed warning and add `TRANSITION, DevCom-923103` comments.


tests/std/tests/P0009R18_mdspan_layout_left/test.cpp
```diff
@@ -75,14 +75,16 @@ constexpr void do_check_members(const extents<IndexType, Extents...>& ext, index
...
     { // Check construction from layout_stride::mapping
         array<IndexType, Ext::rank()> strides{};
         if constexpr (Ext::rank() > 0) {
             strides.front() = 1;
             for (size_t i = 1; i < Ext::rank(); ++i) {
-                strides[i] = static_cast<IndexType>(strides[i - 1] * ext.extent(i - 1));
+#pragma warning(push)
+#pragma warning(disable : 28020) // TRANSITION, DevCom-923103
+                strides[i] =
+#pragma warning(pop)
```
StephanTLavavej, line 86: It seems really weird to scope this to a subexpression instead of a full expression, even if that's possible. Although this is just a test and it's compiling fine right now, I think I would prefer to wrap the whole statement (possibly this comes from paranoia about "weird code" triggering compiler bugs).

StephanTLavavej (approved):

StephanTLavavej:
Thanks! Out of paranoia, I have adjusted one suppression to encompass the entire statement as we haven't done subexpression suppression (outside of perhaps some `_Pragma`s) elsewhere. Will merge after the tests cycle.

StephanTLavavej:
I've pushed a merge to resolve a conflict - the warning suppression is still necessary despite the change from `==` to `cmp_equal`.

StephanTLavavej (approved):

StephanTLavavej:
Thanks for all of these `<mdspan>` PRs, this is great progress! :tada:

PR #3697, opened 2023-05-13
`<format>`: Use `_STL_INTERNAL_CHECK` for calls of `to_chars`

cpplearner:
These `to_chars` should never fail unless there's a bug in the implementation, so it's more appropriate to use `_STL_INTERNAL_CHECK`.


StephanTLavavej (approved):

StephanTLavavej:

strega-nil-ms (approved):

CaseyCarter:
Thanks for internally changing this check to `_STL_INTERNAL_CHECK`! (Yeah, this is a miserable pun even by my standards. Sorry, I got nothing.)

PR #3702, opened 2023-05-14
`<mdspan>`: Add new debug checks to layout mappings

JMazurkiewicz:
* Currently not guarded (https://github.com/microsoft/STL/pull/3586#discussion_r1148075895),
* Fix bug in `P0009R18_mdspan_layout_left/test.cpp`,
* Drive-by: in #3690 I've missed one warning, reduce it here.

### DIFF
--- a/tests/std/tests/P0009R18_mdspan_layout_right/test.cpp
+++ b/tests/std/tests/P0009R18_mdspan_layout_right/test.cpp
@@ -51,8 +51,7 @@ constexpr void check_members(const extents<IndexType, Extents...>& ext, index_se
     using Ext2           = extents<OtherIndexType, Extents...>;
     using Mapping2       = layout_right::mapping<Ext2>;
 
-#pragma warning(push) // TRANSITION, "/analyze:only" BUG?
-#pragma warning(disable : 28020) // The expression '0<=_Param_(1)&&_Param_(1)<=1-1' is not true at this call
+#ifndef __clang__ // FIXME, Clang suddenly cannot digest this
     { // Check construction from other layout_right::mapping
         Mapping m1{ext};
         Mapping2 m2{m1};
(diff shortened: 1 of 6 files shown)


tests/std/tests/P0009R18_mdspan_layout_right/test.cpp
```diff
@@ -51,15 +51,15 @@ constexpr void check_members(const extents<IndexType, Extents...>& ext, index_se
...
-#pragma warning(disable : 28020) // The expression '0<=_Param_(1)&&_Param_(1)<=1-1' is not true at this call
+#ifndef __clang__ // FIXME, Clang suddenly cannot digest this
     { // Check construction from other layout_right::mapping
         Mapping m1{ext};
         Mapping2 m2{m1};
         assert(m1 == m2);
         static_assert(is_nothrow_constructible_v<Mapping2, Mapping>);
         // Other tests are defined in 'check_construction_from_other_right_mapping' function
     }
+#endif
```
StephanTLavavej, line 62: Suggestion for followup: This `#endif` should be commented. (The `#ifndef` is FIXME so it'll be resolved before the final merge, but if we need to convert it into a workaround with a cited LLVM bug, commenting the `#endif` will avoid forgetting `// __clang__` later.)

tests/std/tests/P0009R18_mdspan_layout_stride_death/test.cpp
```diff
@@ -27,13 +27,20 @@ void test_construction_from_strided_layout_mapping() {
     [[maybe_unused]] layout_stride::mapping<dextents<unsigned char, 1>> m2{m1};
 }
 
+void test_call_operator() {
+    layout_stride::mapping<extents<unsigned, 5, 3, 4>> m;
```
StephanTLavavej, line 31: Nitpick for followup: We conventionally say `unsigned int` instead of `unsigned`.

StephanTLavavej (approved):

StephanTLavavej:
Thanks, looks great! :heart_eyes_cat: I had a couple of nitpicks for followup but they weren't worth resetting testing here, they can be addressed later.

PR #3709, opened 2023-05-17
CMakeLists.txt Use targets for toolset interface

duncanspumpkin:
In general with CMake you are encouraged to use targets for applying all compile settings and directories. This splits off the toolset interface into a separate target which is then linked to each of the libraries. I'm unsure if this is something STL would like to go towards but thought it would be an initial attempt at changing a few CMake things.


duncanspumpkin:

Osyotr:
IMHO, setting "global" compile flags and definitions on an interface target and then linking to that target everywhere is actually an antipattern.

duncanspumpkin:
> IMHO, setting "global" compile flags and definitions on an interface target and then linking to that target everywhere is actually an antipattern.

There is certainly many ways to approach it. I was sort of thinking that this would be useful if the other toolsets `(/clr, /clr:pure, OneCore, Spectre)` were going to be making their way into CMake.

One thing i noticed doing this is that the headers include directory is part of the toolset section of the cmake file but really its not related. So could just scrap these changes and move that one little line to outside of that particular sections of CMake.

barcharcraz (changes requested):
I think I agree with @Osyotr here that this is worse than just using the global commands. Since the toolset actually is global. One could argue that we should be setting the toolset stuff up in a toolchain file, or at least using the "pre-everything" variables, but moving things into an interface target strikes me as somewhat pointless.

The only global command that really has problems is `include_directories` because it has somewhat surprising behavior w.r.t. previously defined targets, but even then I don't think it's worth using an interface library in this way.

duncanspumpkin:
> I think I agree with @Osyotr here that this is worse than just using the global commands. Since the toolset actually is global. One could argue that we should be setting the toolset stuff up in a toolchain file, or at least using the "pre-everything" variables, but moving things into an interface target strikes me as somewhat pointless.
> 
> The only global command that really has problems is `include_directories` because it has somewhat surprising behavior w.r.t. previously defined targets, but even then I don't think it's worth using an interface library in this way.

Okay thats fine what about moving
```
include_directories(BEFORE
    "${CMAKE_CURRENT_LIST_DIR}/inc"
```
to outside of the `# Toolset options. #` section would you like that moved since its not a toolset option.

barcharcraz:
> > I think I agree with @Osyotr here that this is worse than just using the global commands. Since the toolset actually is global. One could argue that we should be setting the toolset stuff up in a toolchain file, or at least using the "pre-everything" variables, but moving things into an interface target strikes me as somewhat pointless.
> > The only global command that really has problems is `include_directories` because it has somewhat surprising behavior w.r.t. previously defined targets, but even then I don't think it's worth using an interface library in this way.
> 
> Okay thats fine what about moving
> 
> ```
> include_directories(BEFORE
>     "${CMAKE_CURRENT_LIST_DIR}/inc"
> ```
> 
> to outside of the `# Toolset options. #` section would you like that moved since its not a toolset option.

Yeah I'm OK with that, but honestly it is nice to have all the global include directories added in one place (so the order is very, very obvious. I don't have a super strong preference either way.

duncanspumpkin:
> Yeah I'm OK with that, but honestly it is nice to have all the global include directories added in one place (so the order is very, very obvious. I don't have a super strong preference either way.

Okay i've moved it now. If not wanted can just close the PR. Have you considered moving to CMakePresets instead of CMakeSettings perhaps I could tackle that?

StephanTLavavej:
We talked about this at the weekly maintainer meeting, after @barcharcraz and @strega-nil-ms took a look at the changes. We're going to close this PR without merging because there's no benefit to changing these global properties that all of our targets have, so keeping them global makes the intention clearer. It doesn't help any potential build system composition scenarios because it's doubtful we'll ever do such composition, and they're all directory-scoped anyways. This PR isn't moving us towards idiomatic CMake conventions, like using toolchain files or platform definition files, but there's no real point in exploring those methods since we're so low in the stack of platform technology and aren't building on top of other layers.

While we're unable to merge this, thanks again for exploring this area and submitting this PR!

Issue #3568, opened 2023-03-13
Investigate why Clang/LLVM UBSan doesn't link internally


StephanTLavavej:
While merging #3452 (internal MSVC-PR-457627), I had to disable the Clang/LLVM UBSan configurations for the internal test harness, as they fail to link (with linker errors about support machinery). We use Clang/LLVM via an internally captured "NuGet package" with a subset of binaries (but most of the support libs, AFAIK) to avoid excessive space consumption. At the time, I am unsure why UBSan  doesn't work in the internal environment. It could be that we need to add more libs to the NuGet package, or it could be due to PATH ordering issues where we're mixing up MSVC and Clang/LLVM's support libs.

Thanks to @zacklj89 for reminding me to file this followup issue instead of being a bad kitty :smirk_cat: and silently working around it forever.

StephanTLavavej:
I'm now seeing this with VS 2022 17.7 Preview 1 and Clang 16:

```
code: FAIL
name: std :: tests/GH_002334_branchless_clamp:27
output: Build setup steps:
Build steps:
Build step failed unexpectedly.
[...]

PR #3711, opened 2023-05-17
Toolset update: VS 2022 17.7 Preview 1

StephanTLavavej:
# New pool
* Use PowerShell 7.3.4.
* New pool.
  + Also contains May 2023's Patch Tuesday.
* Mention VS 2022 17.7 Preview 1 in the README.
  + CMake and Ninja are unchanged.
* Require Clang 16.
  + We can't require MSVC 19.37 yet because the internal toolset hasn't been sufficiently updated.

# Remove workarounds
* ~~Remove workaround for VSO-1761088 "ICE when using a pack expansion to repeat a type N times".~~
* Remove workaround for VSO-1753916 'C++23 tuple causes x86chk assertion: !"If lookup found RDSymbol before it should find something this time, too", ParseTree.cpp 10376'.
* Remove workarounds for VSO-1799670 "EDG rejects constexpr vector in debug mode with 'attempt to access expired storage' errors, part 3".
* Remove workarounds for LLVM-44833 / llvm/llvm-project#44178.

# LLVM 16 changes from @CaseyCarter
* Clang has implemented WG21-P0960; unskip tests that now pass, and remove workaround in `P2474R2_views_repeat`.
* Clang now warns about overriding `/fp:strict` with `-ffp-contract`. Update `floating_point_model_matrix.lst` to silence these warnings.
* Fix redeclaration of `adjacent_transform_view` (atomic constraint expressions effectively cannot be reproduced since each utterance is distinct).
* Disambiguate `tuple` comparison operators.
* Workaround llvm/llvm-project#62290, which has been fixed for Clang 17, in `<ranges>`.
* Remove workarounds for a couple of `TRANSITION, Clang 16` issues in `<ranges>` and `<type_traits>`.
* Perma-workaround unfiled Clang bug in `P0896R4_views_take`.

# LLVM 16 followups from @StephanTLavavej
* Improve arrow comments.
* We `_EXPORT_STD` both declarations and definitions.

# Add/update workarounds
* Disable UBSan to work around #3568.
  + @strega-nil-ms mentioned that this is a known issue.
* Guard `_addcarry_u64`/`_subborrow_u64` for Clang.
  + I'm not sure how this worked before, but they're definitely missing now.
* `LDBL_DECIMAL_DIG` is now missing for Clang too.
  + I suspect that Clang changed how it overlays headers, but I didn't investigate the root cause.
* Skip the `tgmath.h` test because it fails in the MSVC-internal repo but passes in the GitHub repo (likely related to header overlaying weirdness).
* Cite llvm/llvm-project#62096 in `P2278R4_views_as_const`.
  + Thanks @JMazurkiewicz for finding this!
* Add workaround for llvm/llvm-project#62762 in `P0067R5_charconv`.
  + Thanks @CaseyCarter for investigating this!

# clang-format 16
* clang-format 16, no manual changes.
  + The changes are almost entirely positive, with no horrible mangling, and limited weirdness around `&`, `&&`, and `*`. No need for manual fixups.
* `.clang-format`: Update link to Clang 16, no behavioral changes.
* `.clang-format`: Sort defaults, no behavioral changes.
* `.clang-format`: Drop commented-out defaults, no behavioral changes.
  + `BreakBeforeInheritanceComma` and `BreakConstructorInitializersBeforeComma` were never documented and aren't printed out by `-dump-config` anymore.
  + The weird empty `BasedOnStyle` isn't printed out by `-dump-config` anymore.
  + `ConstructorInitializerAllOnOneLineOrOnePerLine` and `AllowAllConstructorInitializersOnNextLine` are now deprecated and aren't printed out by `-dump-config` anymore.
* `.clang-format`: `DeriveLineEnding` and `UseCRLF` were deprecated and fused into `LineEnding`, no behavioral changes.
* `.clang-format`: `AlignTrailingComments` is no longer a bool. Use the new options, no behavioral changes.
* `.clang-format`: `SortUsingDeclarations` (which we didn't customize) is no longer a bool, no behavioral changes.
* `.clang-format`: Add new defaults, no behavioral changes.
* `.clang-format`: `InsertNewlineAtEOF` is new, set it to `true`. Doesn't change any files right now.
* `.clang-format`: `RemoveSemicolon` is new, set it to `true`. Behavioral change!
* clang-format 16, no manual changes - remove semicolons.

## clang-format update process
As always, the idea is to run `clang-format --style=LLVM -dump-config` when a new version is available, and then update our `.clang-format` such that all of the defaults are commented out, followed by any customizations.


StephanTLavavej:

stl/inc/__msvc_int128.hpp
```diff
@@ -43,14 +43,17 @@ _STD_BEGIN
 #if defined(_M_X64) && !defined(_M_ARM64EC) && !defined(_M_CEE_PURE) && !defined(__CUDACC__) \
     && !defined(__INTEL_COMPILER)
 #define _STL_128_INTRINSICS 1
-#ifdef __clang__ // clang doesn't have _udiv128 / _div128
-#define _STL_128_DIV_INTRINSICS 0
-#else // ^^^ Clang / other vvv
-#define _STL_128_DIV_INTRINSICS 1
-#endif // ^^^ detect _udiv128 / _div128 ^^^
+#ifdef __clang__
+#define _STL_128_ADD_SUB_INTRINSICS 0 // clang doesn't have _addcarry_u64 / _subborrow_u64
```
CaseyCarter, line 47: Naming nit: these are 64-bit intrinsics, not 128-bit intrinsics.
More serious: This is a lie. I can reproduce the issue with the clang in 17.7p1, but not when I self-install 16.0.1. Something has gone awry in the VS clang tree, and we need to run it down.

I don't really care about the naming given that I suspect this will be a temporary workaround. I've filed #3713 to investigate this.

CaseyCarter (approved):

StephanTLavavej:
I had to restore the workaround for VSO-1761088 "ICE when using a pack expansion to repeat a type N times".

(Sorry, not enough time today to address the "nice to have" nitpicks even though I'm resetting testing.)

strega-nil-ms (approved):

CaseyCarter (approved):

StephanTLavavej:
I had to push an additional commit to skip the `tgmath.h` test.

CaseyCarter (approved):

CaseyCarter:
Thanks for enabling us to use current tools! (No joke, these toolset updates are a _ton_ of work.)

PR #3720, opened 2023-05-19
Cleanup elements when `resize` fails in `deque(n)`

### LINKED ISSUE #3717: `deque(size_t,const Allocator&)` doesn't clear up constructed elements on exception
`deque<T>(size_t,const Allocator&)` lacks a tidy guard now. When the default constructor of a `T` throws, no destructor of constructed `T`s will be called.

For example, in the code below, `test([] {std::deque<resource>f(10); });` will print `"created:5 destroyed:0"`. The expected behavior should be `"created:5 destroyed:5"` as the other code do.

```
#include<deque>
#include<iostream>

struct resource {
	inline static int created = 0, destroyed = 0;
	resource() {
		if (created == 5) { throw 0; }
		++created;
	}
	resource(const resource&) {
		if (created == 5) { throw 0; }
		++created;
	}
	~resource() { ++destroyed; }
};

void test(auto f) {
	resource::created = resource::destroyed = 0;
	try { f(); }
	catch (...) {}
	std::cout << "created:" << resource::created << " destroyed:" << resource::destroyed << std::endl;
}

int main() {
	//created:5 destroyed:0
	test([] {std::deque<resource>f(10); });
	//created:5 destroyed:5
	test([] {std::deque<resource>f; f.resize(10); });
	test([] {std::deque<resource>f; f.resize(10, resource{}); });
	test([] {std::deque<resource>f(10, resource{}); });
	test([] {std::deque<resource>f(10, resource{}, {}); });
}
```

Causes: In the implementation, the use of `resize` is wrong as it doesn't do cleanup on exception, and the construction is not completed so `~deque()` won't be called either.
https://github.com/microsoft/STL/blob/a62109595b6d89e08172fdf4beb75a2670fe0cc9/stl/inc/deque#L622-L628

CaseyCarter:
Fixes #3717

### DIFF
--- a/stl/inc/deque
+++ b/stl/inc/deque
@@ -623,7 +623,9 @@ public:
         : _Mypair(_One_then_variadic_args_t{}, _Al) {
         _Alproxy_ty _Alproxy(_Getal());
         _Container_proxy_ptr12<_Alproxy_ty> _Proxy(_Alproxy, _Get_data());
+        _Tidy_guard<deque> _Guard{this};
         resize(_Count);
+        _Guard._Target = nullptr;
         _Proxy._Release();
     }
 
--- a/tests/std/tests/GH_002769_handle_deque_block_pointers/test.cpp
+++ b/tests/std/tests/GH_002769_handle_deque_block_pointers/test.cpp
@@ -10,7 +10,7 @@
 
 using namespace std;
 
-size_t fancy_counter = 0;
+size_t counter = 0;
 
 template <class T>
 class counting_ptr {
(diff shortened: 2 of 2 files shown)


strega-nil-ms (approved):

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for fixing this major bug in the STL's most important container! :joy_cat: :lady_beetle: :tada:

PR #3725, opened 2023-05-21
remove `operator>>` for `istream >> setfill(c)`

### LINKED ISSUE #3724: `<iomanip>`: consider removing `operator>>` for `istream >> setfill(c)`
According to the standard([this version](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/n4944.pdf)), `std::setfill` is only for ostreams. Therefore, `istream >> setfill(c)` has no well-defined meaning and can lead to [confusion](https://stackoverflow.com/questions/42530114/stdsetfill-and-stdsetw-for-input-streams).
image
The following piece of code should get removed.
https://github.com/microsoft/STL/blob/a62109595b6d89e08172fdf4beb75a2670fe0cc9/stl/inc/iomanip#L28-L35

fsb4000:
Fixes #3724

### DIFF
--- a/tests/std/tests/VSO_0000000_instantiate_iterators_misc/test.compile.pass.cpp
+++ b/tests/std/tests/VSO_0000000_instantiate_iterators_misc/test.compile.pass.cpp
@@ -572,7 +572,7 @@ void future_test() {
 #endif // _M_CEE_PURE
 
 template <typename IoManipIn, typename IoManipOut>
-void iomanip_test_impl(IoManipIn in, IoManipOut out, bool test_mainp_out = true) {
+void iomanip_test_impl(IoManipIn in, IoManipOut out, bool test_manip_out = true) {
     stringstream ss{};
     ss << in;
     if (test_input_stream) {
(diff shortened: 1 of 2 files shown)


achabense:
whoa, it seems some testcases are assuming the availability of `istream >> setfill(c)` too

StephanTLavavej (approved):

tests/std/tests/VSO_0000000_instantiate_iterators_misc/test.compile.pass.cpp
```diff
@@ -571,18 +571,24 @@ void future_test() {
...
+    ss >> in;
 }
 
 template <typename IoManip>
 void iomanip_test_impl(IoManip iom) {
     iomanip_test_impl(iom, iom);
 }
 
+template <typename IoManip>
+void iomanip_test_impl_for_setfill(IoManip out) {
```
CaseyCarter, line 587: It's a little weird that this generic function is used only with `setfill` and includes `setfill` in its name. I'd call this `_output_only` or something, describing what it does instead of how it's currently used. This isn't worth resetting testing, though, so don't bother if we don't need to touch something else.

CaseyCarter (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for fixing this long-standing nonconformance! :tada: :smile_cat: :rocket:

PR #3732, opened 2023-05-25
Support x64-only symbols on ARM64EC

### LINKED ISSUE #3382: Errors when linking x64 OBJ files into ARM64EC programs
When linking x64 OBJ files into ARM64EC programs, users are encountering errors like:

```
error LNK2001: unresolved external symbol __std_reverse_trivially_swappable_8 (EC Symbol)
```

Related to #2635 and #2740.

It appears that for this scenario, we need to provide separately compiled functions in the import lib (the ARM64 import lib?) that x64 object files could be expecting. These functions can perform plain vanilla swaps/reverses/etc., they don't need to be fancy initially. I am not exactly sure what their mangled names should look like, or what the entire fix will look like as a whole. We should be sure to actually validate the fix end-to-end manually (we now have the ability to create ARM64 VMs for internal testing).

Originally reported as DevCom-10219146 and Microsoft-internal VSO-1707080 / AB#1707080 .

CaseyCarter:
* Adds support for building the STL for the ARM64EC platform. (No testing yet, just build.)
* Enable the `__std_init_once_meow` `ALTERNATENAME`s in `src/xonce2.cpp` on ARM64EC. Note that we continue not to use the `<mutex>` `ALTERNATENAME`s when compiling for ARM64EC - the linkage model is complicated - we only include the `ALTERNATENAME`s to support linking with object files built targeting x64. For example, linking x64 static libraries into an ARM64EC binary needs to work.
* Isolate x86/x64-specific code in `src/vector_algorithms.cpp` so we can enable the non-vectorized fallback implementations for ARM64EC. Again, these aren't used when compiling for ARM64EC, but we need the symbols in the import lib to support linking with x64 object files.

Fixes #3382

Note that there's no test coverage here. We're close-but-not-ready to support testing of the STL on ARM64, but a ways away from being able to compile an object file for one platform and link it into a binary built for another. I did validate the fix on an ARM64 VM using files `test.cpp`:
```c++
#include <cassert>
#include <cstddef>

extern int f(const char* ptr, std::size_t length);
extern int g();

int main() {
    static const char data[] = "this is a test, this is only a test";
    assert(f(data, sizeof(data)) == 14);
    assert(g() == 42);
}
```
and `repro.cpp`:
```c++
#include <algorithm>
#include <cstddef>
#include <mutex>

int f(const char* ptr, std::size_t length) {
    return std::find(ptr, ptr + length, ',') - ptr;
}

int g() {
    static std::once_flag nonce;
    static int x = 0;
    std::call_once(nonce, [] { x = 42; });
    return x;
}
```
compiling `repro.cpp` for x64, and linking the resulting object with `test.cpp` (EDIT: built with `/arm64EC`, of course). I performed this process once for each of `/MT`, `/MTd`, `/MD`, and `/MDd` and verified the resulting program linked and ran successfully. (Debug should make no difference, but paranoia never hurts.)


CMakeLists.txt
```diff
@@ -71,6 +73,13 @@ elseif(VCLIBS_TARGET_ARCHITECTURE MATCHES "^arm64$")
     set(VCLIBS_I386_OR_AMD64 "arm64")
     set(VCLIBS_X86_OR_X64 "arm64")
     add_compile_definitions(_ARM64_ _VCRT_WIN32_WINNT=0x0A00 _STL_WIN32_WINNT=0x0A00)
+elseif(VCLIBS_TARGET_ARCHITECTURE MATCHES "^arm64ec$")
+    set(VCLIBS_TARGET_ARCHITECTURE "arm64ec")
+    set(VCLIBS_I386_OR_AMD64 "arm64ec")
+    set(VCLIBS_X86_OR_X64 "arm64")
```
StephanTLavavej, line 79: Double-checking, is this intentional?

StephanTLavavej (approved):

CMakeLists.txt
CaseyCarter: Yes it is, and it probably deserves at least a "yes, really" comment to make that clear. `VCLIBS_X86_OR_X64` is used only as the suffix of `TOOLSET_LIB`, which is roughly `%VCToolsInstallDir%lib\${VCLIBS_X86_OR_X64}`. This directory is where the linker finds the default `lib`s (libcpmt, vcruntime, etc.) and corresponding `pdb`s, and the so-called [Link Options](https://learn.microsoft.com/en-us/cpp/c-runtime-library/link-options?view=msvc-170) (`chkstk.obj`, `legacy_stdio_float_rounding.obj`). `%VCToolsInstallDir%lib\arm64ec` only contains the Link Options (which our build doesn't need). IIUC the "vanilla" ARM64 libs are used.

I'll add a comment.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for figuring out how to fix this major bug that was affecting several customers! :hammer_and_wrench: :bulb: :heart_eyes_cat:

kobykahane:
Will this be backported to 17.6/17.7?

StephanTLavavej:
At this time, we have no plans to backport the fix. There's an escape hatch of disabling the vectorized algorithms when compiling the x64 object files.

kobykahane:
> At this time, we have no plans to backport the fix. There's an escape hatch of disabling the vectorized algorithms when compiling the x64 object files.

Thanks @StephanTLavavej. Unfortunately, in our case, we're not in a position to recompile the x64 static libraries we're attempting to consume in our ARM64EC build. If we were, we'd probably just rebuild them as ARM64 and wouldn't have opted for EC. Additionally, we're hitting the `__std_init_once_*` cases, to which disabling the vectorized algorithms doesn't seem to apply (although it seems that manually defining the alternate names could take care of that one).

PR #3733, opened 2023-05-25
`<ranges>`: Fix `cartesian_product_view::size` in debug mode

JMazurkiewicz:
Per [[range.cartesian.view]/9-10](http://eel.is/c++draft/range.cartesian.view):
> Let *p* be the product of the sizes of all the ranges in bases_.
> *Preconditions*: *p* can be represented by the return type.

Currently `cartesian_product_view::size` may fail when *p* can be represented by return type:
```c++
#include <ranges>

using namespace std;

int main() {
  const auto r1 = views::repeat(0, (numeric_limits<ptrdiff_t>::max)());
  const auto r2 = views::repeat(1, 0);
  (void)views::cartesian_product(r1, r1, r1, r2).size(); // FAIL!
}
```

Obviously `0` can be represented by any unsigned integral type. This PR fixes this bug.

### DIFF
--- a/tests/std/tests/P2374R4_views_cartesian_product/test.cpp
+++ b/tests/std/tests/P2374R4_views_cartesian_product/test.cpp
@@ -973,6 +973,8 @@ namespace check_recommended_practice_implementation { // MSVC STL specific behav
     STATIC_ASSERT(sizeof(range_difference_t<cartesian_product_view<all_t<Vec>, all_t<Vec>>>) > sizeof(ptrdiff_t));
 } // namespace check_recommended_practice_implementation
 
+// GH-3733: cartesian_product_view would incorrectly reject a call to size() claiming that big*big*big*0 is not
+// representable as range_size_t because big*big*big is not.
 constexpr void test_gh_3733() {
     const auto r1   = views::repeat(0, (numeric_limits<ptrdiff_t>::max)());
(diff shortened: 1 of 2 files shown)


tests/std/tests/P2374R4_views_cartesian_product/test.cpp
```diff
@@ -973,6 +973,14 @@ namespace check_recommended_practice_implementation { // MSVC STL specific behav
     STATIC_ASSERT(sizeof(range_difference_t<cartesian_product_view<all_t<Vec>, all_t<Vec>>>) > sizeof(ptrdiff_t));
 } // namespace check_recommended_practice_implementation
 
+constexpr void test_gh_3733() {
```
CaseyCarter, line 976: Code tends to outlive metacode databases; it's likely this test will exist long after "GH-3733" loses meaning. We conventionally try to include a one-line summary description of the issue in regression tests, something like "`cartesian_product_view` would incorrectly reject a call to `size` claiming that `big*big*big*0` is not representable because `big*big` is not."

CaseyCarter (changes requested):

CaseyCarter (approved):

StephanTLavavej (approved):

StephanTLavavej:

PR #3734, opened 2023-05-26
Fix static analysis warnings

StephanTLavavej:
The MSVC-internal repo is making another push to enable and clean up static analysis warnings in the build. The STL was previously unaffected as we've built our tests with plain `/analyze` for 3000 years, but now additional rulesets are being enabled and the STL is emitting warnings.

* Add empty braces to fix warning C26495 "Variable 'MEOW' is uninitialized. Always initialize a member variable (type.6)."
  + This is safe because Zero Is A Strict Subset Of Garbage.
* Permanently silence warning C26495 (uninitialized member variable) for `_Big_integer_flt::_Mydata`.
  + As the big comment explains, leaving this uninitialized is important for performance.
* Fix warning C28252 "Inconsistent annotation for `'RtlCaptureStackBackTrace'`: return/function has `'SAL_success(return!=0)'` on the prior instance. See `[...]\winnt.h(20721)`."
  + The return annotation was indeed missing.
* Fix warning C26439 "This kind of function should not throw. Declare it `'noexcept'` (f.6)."
  + Add `noexcept(false)` because these move ctors call `basic_ios::init()` => `ios_base::_Init()` => `new locale`.
  + `list` dynamically allocates a sentinel node.
* Fix warning C26450 "Arithmetic overflow: `'-'` operation causes overflow at compile time. Use a wider type to store the operands (io.1)." and warning C26454 "Arithmetic overflow: `'-'` operation produces a negative unsigned result at compile time (io.5)."
  + We can avoid these warnings by directly forming the values, which is also easier to read.
    ```
    C:\Temp>type meow.cpp
    ```
    ```cpp
    #include <climits>
    #include <iostream>
    using namespace std;
    
    static_assert((1ul << 31) == 0 - static_cast<unsigned long>(LONG_MIN));
    static_assert((1ull << 63) == 0 - static_cast<unsigned long long>(LLONG_MIN));
    
    int main() {
        cout << 0 - static_cast<unsigned long>(LONG_MIN) << endl;
        cout << (1ul << 31) << endl;
    
        cout << 0 - static_cast<unsigned long long>(LLONG_MIN) << endl;
        cout << (1ull << 63) << endl;
    }
    ```
    ```
    C:\Temp>cl /EHsc /nologo /W4 meow.cpp && meow
    meow.cpp
    2147483648
    2147483648
    9223372036854775808
    9223372036854775808
    ```
* Silence warning C26110 "Caller failing to hold lock 'MEOW' before calling function `'ReleaseSRWLockExclusive'`."
  + We're implementing lock machinery, so we have to assume that the lock is held here.
* In `<charconv>`, work around VSO-1826196 "False positive C26454 in charconv". This has already been fixed in the latest compiler sources, we just need to wait for the MSVC-internal toolset to be updated. Thanks to @dmitrykobets-msft for investigating this.

### DIFF
--- a/stl/inc/future
+++ b/stl/inc/future
@@ -205,8 +205,8 @@ public:
     // TRANSITION, incorrectly default constructs _Result when _Ty is default constructible
     _Associated_state(_Mydel* _Dp = nullptr)
         : _Refs(1), // non-atomic initialization
-          _Exception(), _Retrieved(false), _Ready(false), _Ready_at_thread_exit(false), _Has_stored_result(false),
-          _Running(false), _Deleter(_Dp) {}
+          _Result(), _Exception(), _Retrieved(false), _Ready(false), _Ready_at_thread_exit(false),
+          _Has_stored_result(false), _Running(false), _Deleter(_Dp) {}
 
     virtual ~_Associated_state() noexcept {
         if (_Already_has_stored_result() && !_Ready) { // registered for release at thread exit
@@ -409,7 +409,7 @@ protected:
     }
 
 public:
-    conditional_t<is_default_constructible_v<_Ty>, _Ty, _Result_holder<_Ty>> _Result{};
+    conditional_t<is_default_constructible_v<_Ty>, _Ty, _Result_holder<_Ty>> _Result;
     exception_ptr _Exception;
     mutex _Mtx;
     condition_variable _Cond;
--- a/stl/inc/future
+++ b/stl/inc/future
@@ -409,7 +409,7 @@ protected:
     }
 
 public:
-    conditional_t<is_default_constructible_v<_Ty>, _Ty, _Result_holder<_Ty>> _Result;
+    conditional_t<is_default_constructible_v<_Ty>, _Ty, _Result_holder<_Ty>> _Result{};
     exception_ptr _Exception;
     mutex _Mtx;
     condition_variable _Cond;
--- a/stl/inc/ios
+++ b/stl/inc/ios
@@ -161,9 +161,9 @@ protected:
     __CLR_OR_THIS_CALL basic_ios() {}
 
 private:
-    _Mysb* _Mystrbuf; // pointer to stream buffer
-    _Myos* _Tiestr; // pointer to tied output stream
-    _Elem _Fillch; // the fill character
+    _Mysb* _Mystrbuf{}; // pointer to stream buffer
+    _Myos* _Tiestr{}; // pointer to tied output stream
+    _Elem _Fillch{}; // the fill character
 
 public:
     __CLR_OR_THIS_CALL basic_ios(const basic_ios&)            = delete;
--- a/stl/inc/istream
+++ b/stl/inc/istream
@@ -686,7 +686,7 @@ public:
     }
 
 private:
-    streamsize _Chcount; // the character count
+    streamsize _Chcount{}; // the character count
 };
 
 #pragma vtordisp(pop) // compiler bug workaround
--- a/stl/inc/istream
+++ b/stl/inc/istream
@@ -48,7 +48,7 @@ public:
     }
 
 protected:
-    __CLR_OR_THIS_CALL basic_istream(basic_istream&& _Right) : _Chcount(_Right._Chcount) {
+    __CLR_OR_THIS_CALL basic_istream(basic_istream&& _Right) noexcept(false) : _Chcount(_Right._Chcount) {
         _Myios::init();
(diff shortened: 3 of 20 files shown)


StephanTLavavej:

StephanTLavavej:
Added one more change before mirroring.

stl/inc/future
```diff
@@ -409,7 +409,7 @@ protected:
     }
 
 public:
-    conditional_t<is_default_constructible_v<_Ty>, _Ty, _Result_holder<_Ty>> _Result;
+    conditional_t<is_default_constructible_v<_Ty>, _Ty, _Result_holder<_Ty>> _Result{};
```
CaseyCarter, line 412: `_Ty` is a user-provided type, we should avoid brace-init. And this needs to compile in C++14, so we shouldn't `= `conditional_t<...>()` (even if we did want to repeat that enormous typename). We should just add a member-initializer to the single constructor.

stl/inc/ios
```diff
@@ -161,9 +161,9 @@ protected:
     __CLR_OR_THIS_CALL basic_ios() {}
 
 private:
-    _Mysb* _Mystrbuf; // pointer to stream buffer
-    _Myos* _Tiestr; // pointer to tied output stream
-    _Elem _Fillch; // the fill character
+    _Mysb* _Mystrbuf{}; // pointer to stream buffer
+    _Myos* _Tiestr{}; // pointer to tied output stream
+    _Elem _Fillch{}; // the fill character
```
CaseyCarter, line 166: `_Elem` is potentially a user-provided type, but _incredibly_ unlikely to have issues with brace-init since the Standard makes the set of allowed types approximately implementation defined. I wouldn't mind `= _Elem()` here, but I won't complain either way.

stl/inc/list
```diff
@@ -745,8 +745,8 @@ struct _List_node_insert_op2 {
 private:
     _Alnode& _Al;
     size_type _Added; // if 0, the values of _Head and _Tail are indeterminate
-    pointer _Tail; // points to the most recently appended element; it doesn't have _Next constructed
-    pointer _Head; // points to the first appended element; it doesn't have _Prev constructed
+    pointer _Tail{}; // points to the most recently appended element; it doesn't have _Next constructed
+    pointer _Head{}; // points to the first appended element; it doesn't have _Prev constructed
```
CaseyCarter, line 749: User-provided types again; I think `= nullptr` (or even `= pointer()`) would be an improvement here.

stl/inc/istream
```diff
@@ -48,7 +48,7 @@ public:
     }
 
 protected:
-    __CLR_OR_THIS_CALL basic_istream(basic_istream&& _Right) : _Chcount(_Right._Chcount) {
+    __CLR_OR_THIS_CALL basic_istream(basic_istream&& _Right) noexcept(false) : _Chcount(_Right._Chcount) {
```
CaseyCarter, line 51: We have a warning that suggests marking potentially-throwing functions `noexcept`??!? Wat? (No change requested.)

CaseyCarter (changes requested):

stl/inc/list
CaseyCarter: Per Discord discussion, this is fairly conventional for fancy pointers. We should fix all such occurrences (in a different PR) or none.

StephanTLavavej: \[nullablepointer.requirements\] says Cpp17NullablePointer requires Cpp17DefaultConstructible, and \[tab:cpp17.defaultconstructible\] says that `T{}` must be either value-initialized or aggregate-initialized. So this is valid, and we routinely do this in `forward_list` and `list` for potentially fancy pointers.

CaseyCarter: ...and Stephan points out that _`Cpp17NullablePointer`_ requires _`Cpp17DefaultConstructible`_ which requires `T{}` to be valid, so I'm simply wrong.

stl/inc/istream
StephanTLavavej: Yeah, it is... not the greatest.

stl/inc/ios
StephanTLavavej: We have a bunch of occurrences of `_Elem{}` in `<xlocale>` and related headers, so I'd like to leave this as-is.

stl/inc/future
StephanTLavavej: I'll make this change.

CaseyCarter (approved):

PR #3736, opened 2023-05-26
Introduce `_Is_transparent` helper trait

Saalvage:
And properly standardize some SFINAE using it.

I would like to utilize this helper in implementing `flat_set`.

### DIFF
--- a/stl/inc/xutility
+++ b/stl/inc/xutility
@@ -537,6 +537,20 @@ struct less_equal<void> {
     using is_transparent = int;
 };
 
+template <class _Ty, class = void>
+_INLINE_VAR constexpr bool _Is_transparent_v = false;
+
+template <class _Ty>
+_INLINE_VAR constexpr bool _Is_transparent_v<_Ty, void_t<typename _Ty::is_transparent>> = true;
+
+template <class _Ty>
+struct _Is_transparent : bool_constant<_Is_transparent_v<_Ty>> {};
+
+#ifdef __cpp_lib_concepts // TRANSITION, GH-395
+template <class _Ty>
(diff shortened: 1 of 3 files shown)


stl/inc/xutility
```diff
@@ -537,6 +537,20 @@ struct less_equal<void> {
...
+_INLINE_VAR constexpr bool _Is_transparent_v<_Ty, void_t<typename _Ty::is_transparent>> = true;
+
+template <class _Ty>
+struct _Is_transparent : bool_constant<_Is_transparent_v<_Ty>> {};
+
+#ifdef __cpp_lib_concepts // TRANSITION, GH-395
+template <class _Ty>
+concept _Transparent = _Is_transparent_v<_Ty>;
+#endif // __cpp_lib_concepts
+
```
StephanTLavavej, line 553: This is great, but `<xutility>` is unnecessarily high up/widely included in the (poorly structured and undocumented) dependency graph. It would be better to add this to `<xmemory>`, which is the internal header that all of the containers drag in (and which is lower/leafier than `<xutility>`). I checked your in-progress `<flat_set>` implementation which drags in `<vector>`, so it's also getting `<xmemory>`.

StephanTLavavej (approved):

StephanTLavavej:
This is an *excellent* refactoring, thank you! The SFINAE was more complicated than I remembered, and the way you're simplifying it is great. (It was weird/unusual to be doing SFINAE with separate template parameters.) I've pushed a small change to move code around for increased throughput.

CaseyCarter (approved):

strega-nil-ms (approved):

StephanTLavavej:

PR #3743, opened 2023-05-31
Silence 3 static analysis warnings

StephanTLavavej:
Followup to #3734.

I've filed a tracking bug VSO-1828677 "False positive warning C26450 (arithmetic overflow) in charconv and memory_resource" for the static analysis team to investigate, and @dmitrykobets-msft has added fully reduced repros there. Our code is correct, so we need to silence this warning until the compiler is fixed.

For `locale::_Ptr`, attempting to initialize it will break the STL catastrophically. I'm permanently suppressing the warning, with an explanation of the squirrelly :chipmunk: behavior involved. In theory, this is not ABI-sensitive since `classic_locale` is isolated to the locale0.cpp TU, but iostreams is extremely complicated and fragile and I am not eager to mess with this logic at this time unless absolutely forced to.

### DIFF
--- a/stl/inc/xlocale
+++ b/stl/inc/xlocale
@@ -257,7 +257,17 @@ public:
         }
     }
 
+#pragma warning(push)
+#pragma warning(disable : 26495) // Variable 'std::locale::_Ptr' is uninitialized.
+                                 // Always initialize a member variable (type.6).
+
+    // _Ptr must remain uninitialized here.
+    // In locale0.cpp, locale::_Init() uses True Placement New at classic_locale's address,
+    // and classic_locale is constructed from the _Noinit enumerator of type _Uninitialized.
+    // The sequencing is highly unusual; the True Placement New happens before the _Uninitialized construction,
+    // so attempting to initialize _Ptr here will overwrite the result of the True Placement New.
     locale(_Uninitialized) {} // defer construction
+#pragma warning(pop)
 
     locale(const locale& _Right) noexcept : _Ptr(_Right._Ptr) {
         _Ptr->_Incref();
--- a/stl/inc/xlocale
+++ b/stl/inc/xlocale
@@ -261,11 +261,12 @@ public:
 #pragma warning(disable : 26495) // Variable 'std::locale::_Ptr' is uninitialized.
                                  // Always initialize a member variable (type.6).
 
-    // _Ptr must remain uninitialized here.
+    // We must not explicitly initialize _Ptr here; we rely on it maintaining the value
+    // previously created in its storage. To be precise:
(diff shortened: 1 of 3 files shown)


stl/inc/xlocale
```diff
@@ -257,7 +257,17 @@ public:
...
 
+#pragma warning(push)
+#pragma warning(disable : 26495) // Variable 'std::locale::_Ptr' is uninitialized.
+                                 // Always initialize a member variable (type.6).
+
+    // _Ptr must remain uninitialized here.
+    // In locale0.cpp, locale::_Init() uses True Placement New at classic_locale's address,
+    // and classic_locale is constructed from the _Noinit enumerator of type _Uninitialized.
+    // The sequencing is highly unusual; the True Placement New happens before the _Uninitialized construction,
+    // so attempting to initialize _Ptr here will overwrite the result of the True Placement New.
```
CaseyCarter, line 268: IMO this comment doesn't illustrate clearly enough the eldritch horrors in the code. 

Formally:
The phrase "attempting to initialize" is meaningless. This constructor unconditionally initializes the member subobject `_Ptr`; there is no way for it to "remain uninitialized". The form of initialization here is *default-initialization*, which (remember, "formally") results in an indeterminate value rather than "the previous contents of this storage interpreted is a value of type `_Locimp*`".

I think we should make it painfully clear that what we're relying on the compiler to leave the storage for `_Ptr` untouched and to interpret the former contents of that storage as the object representation of `_Ptr`. It's not unusual for us to rely on toolchain-specific behavior, but this particular case is right on the verge of breaking. I believe the last few versions of GCC would miscompile this, for example.

How about:
```suggestion
    // We must not explicitly initialize _Ptr here, we rely on it maintaining the value 
    // previously created in its storage. To be precise:
    // In locale0.cpp, locale::_Init() uses True Placement New at classic_locale's address,
    // and classic_locale is constructed from the _Noinit enumerator of type _Uninitialized.
    // The sequencing is highly unusual; the True Placement New happens before the _Uninitialized construction,
    // so while _Ptr here formally has indeterminate value, we expect it to actually keep the previous value.
```

CaseyCarter (changes requested):

stl/inc/xlocale
StephanTLavavej: Thanks! I'll apply this (and fix the comma splice).

CaseyCarter: > (and fix the comma splice).

Oh, wow, that's terrible. This is what happens when you edit the text three or four times and don't give it a final read-through.

CaseyCarter (approved):

Issue #3391, opened 2023-02-05
`<ranges>`: Add missing test coverage for `ranges::view_interface::c(begin|end)`


JMazurkiewicz:
#3187 introduced new members to `ranges::view_interface`: `cbegin` and `cend`. While there is small coverage for those functions in `P0896R4_ranges_subrange/test.compile.pass.cpp`, I believe that we should add extra assertions to each `views` tests (just like #3350 did).

<details>
<summary>TODO</summary>

  * [ ] `as_const`,
  * [ ] `as_rvalue`,
  * [ ] `adjacent` (not yet implemented),
  * [ ] `adjacent_transform` (not yet implemented),
  * [ ] `cartesian_product` (not yet implemented),
  * [ ] `chunk`,
  * [ ] `chunk_by`,
  * [ ] `common_view`,
  * [ ] `drop` (#3390),
  * [ ] `drop_while` (#3390),
  * [x] `elements` (#3350),
  * [ ] `empty`,
  * [ ] `enumerate` (soon?),
  * [ ] `filter`,
  * [ ] `iota`,
  * [ ] `istream`,
  * [ ] `join`,
  * [ ] `join_with`,
  * [ ] `lazy_split`,
  * [ ] `owning_view`,
  * [ ] `ref_view`
  * [ ] `repeat`,
  * [ ] `reverse`,
  * [ ] `single`,
  * [ ] `slide`,
  * [ ] `split`,
  * [ ] `stride`,
  * [ ] `subrange`,
  * [ ] `take` (#3390),
  * [ ] `take_while` (#3390),
  * [ ] `transform`,
  * [ ] `zip`,
  * [x] `zip_transform` (#3322),
</details>

PR #3753, opened 2023-06-07
Make `P2415R2_owning_view/test.cpp` test something

### LINKED ISSUE #3613: Testing: `P2415R2_owning_view\test.cpp` does not test anything
Currently `P2415R2_owning_view\test.cpp` looks like this:

```c++
// [banner, headers]

struct instantiator {
    template <ranges::input_range R>
    static constexpr void call() {
        if constexpr (move_constructible<R>) {
            // [ALL TESTS]
        }
    }
};

int main() {
    STATIC_ASSERT((test_in<instantiator, int>(), true));
    test_in<instantiator, int>();
}
```

When we add extra constraint to `instantiator::call`, which is negation of the condition in `if constexpr` statement:

```c++
struct instantiator {
    template <ranges::input_range R>
        requires (!move_constructible<R>)
    static constexpr void call() {
        if constexpr (move_constructible<R>) {
            // [ALL TESTS]
        }
    }
};
```

The test still passes, which means that the `// [ALL TESTS]` part is never executed.

JMazurkiewicz:
* Closes #3613.
* Drive-by:
  * Test `c(begin|end)` too (towards #3391),
  * Fix `owning_view::size` test.

### DIFF
--- a/tests/std/include/range_algorithm_support.hpp
+++ b/tests/std/include/range_algorithm_support.hpp
@@ -915,6 +915,8 @@ namespace test {
     public:
         using I = iterator<Category, Element, Diff, Eq, Proxy, WrappedState::wrapped>;
         using S = conditional_t<to_bool(IsCommon), I, sentinel<Element, WrappedState::wrapped>>;
+        using RebindAsMoveOnly = range<Category, Element, IsSized, Diff, IsCommon, Eq, Proxy, IsView,
+            Copyability::move_only>;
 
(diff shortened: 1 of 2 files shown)


StephanTLavavej (approved):

StephanTLavavej:

strega-nil-ms (approved):

StephanTLavavej:
Thanks for noticing and fixing this major omission in test coverage! :hole: :adhesive_bandage: :sparkles:

PR #3759, opened 2023-06-10
Cleanups for `<mutex>`  etc.

achabense:
- Removes `condition_variable::_Wait_until1(...const _timespec64*...)`, which is not called anywhere.
- Fuses some ugly methods like `timed_mutex::_Try_lock_until`.

### DIFF
--- a/stl/inc/mutex
+++ b/stl/inc/mutex
@@ -762,18 +762,6 @@ private:
         }
     }
 
-    template <class _Predicate>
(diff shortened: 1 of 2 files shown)


StephanTLavavej (approved):
Perfect, thank you!

strega-nil-ms (approved):

StephanTLavavej:
Thanks for clearing out even more dead code! :skull: :skull_and_crossbones: :magic_wand:

PR #3768, opened 2023-06-13
Remove some unnecessary clang-format hacks

CaseyCarter:
clang-format 16 doesn't need these anymore, and in fact they often make things _worse_. Note that "I don't like how clang formats this" isn't enough reason to mess with the formatting; the threshold for using annotations is "the code without annotations is so bad that having noise from annotations is an improvement."

### DIFF
--- a/stl/inc/__msvc_iter_core.hpp
+++ b/stl/inc/__msvc_iter_core.hpp
@@ -165,8 +165,8 @@ template <class>
 struct _Iterator_traits_base {};
 
 template <class _It>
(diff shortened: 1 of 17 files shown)


StephanTLavavej (approved):

StephanTLavavej:
Thanks! Those parentheses used to be necessary but they were really messing things up now!

StephanTLavavej:

StephanTLavavej:
Thanks for noticing and cleaning up all of these formatting relics! :broom: :broom: :broom:

PR #3769, opened 2023-06-14
Toolset update: VS 2022 17.7 Preview 2

StephanTLavavej:
* Updated build compiler to VS 2022 17.7 Preview 2.
* Updated Python to 3.11.4.
* Switched our VM SKU to Standard_D32ads_v5, which 1ES has already been doing for us behind the scenes.
* Updated our VM pool, also containing June's Patch Tuesday.
* Removed compiler bug workarounds:
  + VSO-1292293 "EDG incorrectly rejects defaulted `noexcept` dtor when a data member has a throwing dtor".
  + VSO-1222776 "EDG doesn't know that `noexcept`-specifiers of hidden friend functions are complete-class contexts".
* Reduce timeouts. It looks like the new pool doesn't avoid the timeouts we've been suffering recently, but this will waste less time.
  + `azure-pipelines.yml`:
    - Reduce the timeout for Code Format Validation from 20 min to 5 min. This is still a ~5x safety margin as it takes ~56 seconds total.
  + `azure-devops/cmake-configure-build.yml`:
    - Reduce the timeout for building the STL from 10 min to 5 min. This is still a ~3x safety margin as it takes ~1m 43s to build.
    - Reduce the timeout for building the benchmarks from 10 min to 2 min. This is a massive safety margin as it takes ~5s to build.
  + `azure-devops/run-tests.yml`:
    - Reduce the timeout for actually running the tests from 120 min to 20 min. This is still a ~2x safety margin as the tests currently take ~8.5 min to run for the slowest slice.
    - Reduce the timeout for publishing the tests for 10 min to 5 min. This is a massive safety margin as it takes ~11s to publish, although it may be network-dependent, hence the caution.
  + `azure-devops/cross-build.yml`:
  + `azure-devops/native-build-test.yml`:
    - Reduce the overall timeouts from 360 minutes to 25 minutes. This gives us a ~2x safety margin, as the tests typically finish in ~12 min, and are virtually guaranteed to be doomed if they've run for over 15 minutes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ Just try to follow these rules, so we can spend more time fixing bugs and implem
 
 # How To Build With The Visual Studio IDE
 
-1. Install Visual Studio 2022 17.7 Preview 1 or later.
+1. Install Visual Studio 2022 17.7 Preview 2 or later.
     * Select "Windows 11 SDK (10.0.22000.0)" in the VS Installer.
     * We recommend selecting "C++ CMake tools for Windows" in the VS Installer.
     This will ensure that you're using supported versions of CMake and Ninja.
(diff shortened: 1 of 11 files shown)


CaseyCarter (approved):

StephanTLavavej:
I had to restore the workaround for VSO-1761088 "ICE when using a pack expansion to repeat a type N times" because the MSVC-internal build is using a 17.6-based compiler, whereas this bug was fixed in 17.7p1. I'll try again in a later PR.

CaseyCarter (approved):

PR #3773, opened 2023-06-15
Fix `fstream.seekp(0, ios::cur)`

### LINKED ISSUE #3572: `<fstream>`: seeking 0 with std::ios::cur ignored with bidirectional streams
**Describe the bug**
C++ Standard, [filebuf], paragraph 2 states that:

> The restrictions on reading and writing a sequence controlled by an object of class `basic_filebuf<charT, traits>` are the same as for reading and writing with the C standard library `FILE`s

So when referring to C standard §7.19.5.3, paragraph 6 states that:

> However, output shall not be directly followed by input without an intervening call to the `fflush` function or to a file positioning function (`fseek`, `fsetpos`, or `rewind`), and input shall not be directly followed by output without an intervening call to a file positioning function, unless the input operation encounters end-of-file

One could therefore expect that for a bidirectional `std::fstream` you could perform `stream.seekp(0, std::ios::cur)` to satisfy this. In libstdc++ and libc++, this works but this does not work with MSVC STL.

Note that `stream.seekp(stream.tellp())` works. So does a +1 and -1 on `std::ios:cur`.

**Command-line test case**
```cpp
#include <fstream>
#include <iostream>

int main() {
	std::ofstream init("test.txt");
	init << "abcd";
	init.close();

	char buffer[5] = {};

	std::fstream inout("test.txt");
	inout.read(buffer, 2);
	inout.seekp(0, std::ios::cur); // Should allow switch to write.
	inout << "ba";
	inout.close();

	std::ifstream out("test.txt");
	out.read(buffer, 4);
	std::cout << buffer << std::endl;
	out.close();
}
```

**Expected behavior**
The output is "abba".

MSVC STL outputs "abcd" - the write operation is completely ignored.

**STL version**
```
Microsoft Visual Studio Enterprise 2022
Version 17.5.2
```

**Additional context**

C++ standard, [filebuf.virtuals], paragraph 13 states:

> Let `width` denote `a_codecvt.encoding()`. If `is_open() == false`, or `off != 0 && width <= 0`, then the positioning operation fails. Otherwise, if `way != basic_ios::cur` or `off != 0`, and if the last operation was output, then update the output sequence and write any unshift sequence. Next, seek to the new position: if `width > 0`, call `fseek(file, width * off, whence)`, otherwise call `fseek(file, 0, whence)`

If I've not misread this: although `stream.seekp(0, std::ios::cur)` should not touch the output/unshift sequence, it still should call `fseek` to my understanding.

achabense:
Fixes #3572. As discovered in the issue, we need `fseek` for side effects even when `(off,way)==(0,ios::cur)`.


strega-nil-ms (approved):
This seems reasonable to me.

CaseyCarter (changes requested):
The _fix_ looks fine, but I'd like to see a regression test (a test that fails without the fix but succeeds with it) to validate it.

achabense:
I have no idea how to handle this formatting problem. I feel exhausted now.

StephanTLavavej:
Please note that we strongly recommend configuring your editor to clang-format on save. (VSCode can do this with Microsoft's C/C++ Extension or xaver's clang-format extension, among others.) The CI system is a shared resource, so it's best to avoid putting unnecessary load on it.

tests/std/tests/GH_003572_fstream_seekp_0_cur/env.lst
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+RUNALL_INCLUDE ..\usual_latest_matrix.lst
```
StephanTLavavej, line 4: It appears that this test could be run down to C++17 (with `string_view`). We use the "latest" matrix only when C++23 features are necessary.

tests/std/tests/GH_003572_fstream_seekp_0_cur/test.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include <fstream>
+
+int main() {
+    using namespace std;
+
+    fstream f("test.txt", ios::in | ios::out | ios::trunc);
+
+    f << "123";
+
+    auto check = [&](string_view content) {
```
StephanTLavavej, line 14: Need to include `<string_view>`.

tests/std/tests/GH_003572_fstream_seekp_0_cur/test.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+    f.seekg(0);
+    (void) f.get();
+    f.seekp(f.tellp());
+    f << "*";
+    check("1*3");
+
+    f.seekg(0);
+    (void) f.get();
+    f.seekp(0, std::ios::cur);
```
StephanTLavavej, line 29: This test is already `using namespace std;`. Occurs below.

tests/std/tests/GH_003572_fstream_seekp_0_cur/test.cpp
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <cassert>
+#include <fstream>
+
+int main() {
+    using namespace std;
+
+    fstream f("test.txt", ios::in | ios::out | ios::trunc);
```
StephanTLavavej, line 10: Technically need to include `<ios>` for `ios::in` etc.

tests/std/tests/GH_003572_fstream_seekp_0_cur/test.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+    fstream f("test.txt", ios::in | ios::out | ios::trunc);
+
+    f << "123";
+
+    auto check = [&](string_view content) {
+        char buffer[4]{};
+        f.seekg(0);
+        f.read(buffer, 3);
+        assert(f && buffer == content);
```
StephanTLavavej, line 18: Suggestion: It's better to separately `assert(X);` and `assert(Y);` because if one of them fails, the line number will say which one it was. We try to follow this convention in our tests (not always perfectly).

StephanTLavavej (approved):

StephanTLavavej:
Thanks for fixing this bug and testing the fix! :heart_eyes_cat: I validated and pushed minor cleanups to the test.

CaseyCarter:
> Please note that we strongly recommend configuring your editor to clang-format on save.

There's also a cmake `format` target that clang-formats everything, but format-on-save is both impossible to forget and (generally) faster since the cmake target literally clang-formats everything.

tests/std/tests/GH_003572_fstream_seekp_0_cur/env.lst
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+RUNALL_INCLUDE ..\usual_17_matrix.lst
```
CaseyCarter, line 4: Since the fix applies to C++14, it would be nice for the test to do so as well.

CaseyCarter (approved):
I'll push a change to run the test in C++14 mode.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks again for this correctness fix! :lady_beetle: :white_check_mark: :100:

PR #3776, opened 2023-06-16
Avoid evil macros

StephanTLavavej:
ICU previously macroized these reserved names. This was fixed by https://github.com/unicode-org/icu/pull/1555 on 2021-02-01, but old versions of ICU are still being used by various projects. Let's avoid these names for now. I haven't added test coverage because the UCRT uses some of these reserved names (which it's getting away with, due to the specific order in which headers are being included).

Fixes VSO-1837381 "\[RWC\]\[prod/fe\]\[Regression\] PDFium failed with error C2209: `'std::_Aligned_storage_t<64,8>'`: aliases cannot be used in constructor declarations".

* `_Cs` => `_Cs_storage`
  + This introduces a slight but acceptable asymmetry with `mutex.cpp`.
  + This is the actual fix for the regression introduced by #3763. We were getting away with the others.
* `_No` => `_Nope`
  + This makes `<deque>` `_Is_bidi` consistent with `<format>` `_Add_newline`, `<xlocale>` `_Case_sensitive`, etc.
* `_Mn` => `_Min_`, also `_Mx` => `_Max_` for consistency
  + This is our conventional pattern for avoiding parameters shadowing data members.
* `_Cc` => `_Cx`, also `_Yy` => `_Yx` and `_Nn` => `_Nx` for consistency
  + This is our common pattern for "single letter" variable names.
* `_Pi` => `_Pi_val`

### DIFF
--- a/stl/inc/__msvc_bit_utils.hpp
+++ b/stl/inc/__msvc_bit_utils.hpp
@@ -49,19 +49,19 @@ _INLINE_VAR constexpr int _Unsigned_integer_digits = sizeof(_UInt) * CHAR_BIT;
 // see "Hacker's Delight" section 5-3
 template <class _Ty>
 _NODISCARD constexpr int _Countl_zero_fallback(_Ty _Val) noexcept {
-    _Ty _Yy = 0;
+    _Ty _Yx = 0;
 
-    unsigned int _Nn = _Unsigned_integer_digits<_Ty>;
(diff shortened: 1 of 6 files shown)


fsb4000 (approved):

strega-nil-ms (approved):

StephanTLavavej:

PR #3786, opened 2023-06-19
`<mdspan>`: `extents` improvements

JMazurkiewicz:
* Guard `_STL_VERIFY` with `#if _CONTAINER_DEBUG_LEVEL > 0` (towards https://github.com/microsoft/STL/pull/3586#discussion_r1148075895). I believe that this is the best solution, because:
  * The checks are cheap,
  * The checks don't require ABI-changing bookkeeping.
* Improve `extents`' constructors:
  * `extents(other extents)` constructor avoids using `for` loop and uses `index_sequence` instead,
  * `extents(exts...)` constructor avoids creating temporary arrays and using `for` loop. Instead, it puts `exts...` in a tuple and then constructs `_Dynamic_extents` array directly,
  * `extents(exts...)` now always checks elements in `exts` when it is possible. Previously it checked element when all elements were "checkable" (were `integral`),
  * `extents(span/array)` is more efficient.
* Add extra internal checks in `_(Fwd/Rev)_prod_of_extents`,
* Improve test code:
  * Add parentheses around `std::min` in `test_mdspan_support.hpp`,
  * Add extra death test to verify new `extents(exts...)`'s behaviour.

### DIFF
--- a/stl/inc/mdspan
+++ b/stl/inc/mdspan
@@ -118,28 +118,34 @@ public:
 
     constexpr extents() noexcept = default;
 
+    template <class _OtherIndexType, size_t... _OtherExtents, size_t... _Indices>
+        requires (sizeof...(_OtherExtents) == rank())
+              && ((_OtherExtents == dynamic_extent || _Extents == dynamic_extent || _OtherExtents == _Extents) && ...)
+    constexpr explicit extents(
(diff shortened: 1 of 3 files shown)


StephanTLavavej (approved):

StephanTLavavej:
Looks great, thank you! :sparkles:

PR #3816, opened 2023-06-20
Standard Library Modules: Add workaround for `<format>` `/utf-8`

StephanTLavavej:
Thanks to C++ MVP Marc Gregoire for reporting that `import std;` didn't allow `<format>` machinery to work with `/utf-8`.

First, this adds test coverage to the Standard Library Modules test. We were already exercising the affected function, we just need to add a `/utf-8` configuration to light up the corresponding `if constexpr` paths.

Second, this adds a product code workaround. As @cdacamar noted, this is internal VSO-1538698 "Better handling for non-exported friend function declarations" again, where granting friendship to templated, non-module-`export`ed (i.e. `_Ugly`) functions doesn't work.

I've audited the STL and I believe we're consistently applying the workaround now. We just need to remember that `friend` will trigger this compiler bug.


MarcGregoire:
Thanks for the quick reaction!

CaseyCarter (approved):

StephanTLavavej:

PR #3819, opened 2023-06-21
Deallocate at least

CaseyCarter:
#3712 broke BitCoin in Microsoft's internal Real World Code (RWC) test suite. They publicly derive an allocator from `std::allocate`, implementing `allocate` and `deallocate` but not `allocate_at_least`
(https://github.com/bitcoin/bitcoin/blob/f1b4975461364d5d40d2bfafc6b165dd5d7eec30/src/support/allocators/secure.h#L19-L56).
When `vector` allocates memory with `std::allocator::allocate_at_least` and tries to free it with `secure_allocator::deallocate`, terrible things happen.

We suspect this pattern is widespread, so we're reverting the change for now.

There's a second commit here to reapply the `basic_string` drive-by fix which we _don't_ want to revert. (We want to merge these commits separately to make it easier to revert the revert in the future.)

Fixes VSO-1839141 / AB#1839141.


StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
I've pushed a merge with `main` to resolve a trivial adjacent add/remove conflict in `tests/std/test.lst`.

PR #3821, opened 2023-06-22
`<mdspan>`: Improve debug checks

JMazurkiewicz:
* Guard all existing debug checks with `#if _CONTAINER_DEBUG_LEVEL > 0` (addresses https://github.com/microsoft/STL/pull/3586#discussion_r1148075895),
* `extents`:
  * Make `_Static_extents` array public and use it when possible,
  * Add new `_Multidim_index_space_size_is_always_zero` static member variable,
  * Split `_Is_index_space_size_representable` into two functions: `_Is_static_multidim_index_space_size_representable` and `_Is_dynamic_multidim_index_space_size_representable`,
  * Report DevCom-10398426 and silence it,
  * Improve debug check in `extents(_OtherIndexTypes... _Exts)` constructor,
* `layout_(left|right|stride)`
  * Implement missing debug checks,
  * Replace scary looking immediately invoked lambda expressions with simple `for` loops (`layout_(left|right)`'s constructors from `layout_stride`),
  * Move precondition checks from `operator()` to corresponding `_Index_impl` functions (to avoid casting `_Indices` to `index_type` twice),
  * Drive-by: Remove unnecessary `[[maybe_unused]]` attributes in tests,
  * Drive-by: Simplify calls to `check_members_with_various_extents` in tests,
  * Drive-by: use `ranges::adjacent_find` instead of `views::pairwise_transform` in `get_mapping_properties` function,
  * Drive-by: fix exhaustiveness check in `get_mapping_properties` - previously, non-unique exhaustive mappings were recognized as non-exhaustive,
* `mdspan`:
  * Implement missing debug checks,
  * Manually inline call to `extents_type::static_extent` in `mdspan::static_extent` and add precondition check,
  * Drive-by: improve performance of `mdspan::empty`,
  * Implement new death tests.

### DIFF
--- a/tests/std/tests/P0009R18_mdspan_extents_death/test.cpp
+++ b/tests/std/tests/P0009R18_mdspan_extents_death/test.cpp
@@ -54,6 +54,13 @@ void test_construction_from_pack_with_unrepresentable_as_index_type_values_2() {
     [[maybe_unused]] extents<unsigned char, 1, dynamic_extent> e{ConvertibleToInt<unsigned char>{.val = 1}, 256};
 }
 
+void test_construction_from_pack_with_unrepresentable_as_index_type_values_3() {
+    static_assert(signed_integral<char>, "char is not signed integral");
+    // Either sizeof...(exts) must be equal to 0 or each element of exts must be nonnegative and must be representable
+    // as value of type index_type
+    [[maybe_unused]] extents<signed char, 1, dynamic_extent> e{static_cast<char>(-1)};
+}
+
 void test_construction_from_span_with_invalid_values() {
     int vals[] = {1, 2};
     span s{vals};
@@ -93,6 +100,7 @@ int main(int argc, char* argv[]) {
         test_construction_from_pack_with_invalid_values,
         test_construction_from_pack_with_unrepresentable_as_index_type_values_1,
         test_construction_from_pack_with_unrepresentable_as_index_type_values_2,
+        test_construction_from_pack_with_unrepresentable_as_index_type_values_3,
         test_construction_from_span_with_invalid_values,
         test_construction_from_span_with_unrepresentable_as_index_type_values,
         test_construction_from_array_with_invalid_values,
--- a/tests/std/tests/P0009R18_mdspan_layout_stride_death/test.cpp
+++ b/tests/std/tests/P0009R18_mdspan_layout_stride_death/test.cpp
@@ -12,6 +12,13 @@
 
 using namespace std;
 
+void test_default_construction() {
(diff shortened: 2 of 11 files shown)


tests/std/tests/P0009R18_mdspan_extents_death/test.cpp
```diff
@@ -54,6 +54,13 @@ void test_construction_from_pack_with_unrepresentable_as_index_type_values_2() {
     [[maybe_unused]] extents<unsigned char, 1, dynamic_extent> e{ConvertibleToInt<unsigned char>{.val = 1}, 256};
 }
 
+void test_construction_from_pack_with_unrepresentable_as_index_type_values_3() {
+    static_assert(signed_integral<char>, "char is not signed integral");
```
StephanTLavavej, line 58: I find this `static_assert` message to be confusing, because it's somewhat ambiguous whether it's mentioning something that should be the case, or something that actually is the case. Here's it's "actually is the case", but there's still some residual confusion because it's not talking about any actual Standardese guarantee, just an implementation/test assumption.

I think it would be better to replace this message with `"This test assumes that it isn't being compiled with /J"` or something along those lines.

tests/std/tests/P0009R18_mdspan_layout_stride_death/test.cpp
```diff
@@ -1,24 +1,49 @@
...
 
 #include <test_death.hpp>
+#include <test_mdspan_support.hpp>
 
 using namespace std;
 
-void test_construction_from_extents_and_array() {
+void test_default_construction() {
+    using Ext = extents<signed char, dynamic_extent, 4, 5, 7>;
+    // Value of layout_right::mapping<extents_type>().required_span_size() must be
```
StephanTLavavej, line 18: Copy-paste? Should this say `layout_stride`?

tests/std/tests/P0009R18_mdspan_mdspan_death/test.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include <mdspan>
+
+#include <test_death.hpp>
+
+using namespace std;
+
+constexpr array<int, 256> some_ints{};
+
+void test_construction_from_other_mdspan() {
+    auto mds1 = mdspan{some_ints.data(), 8, 2, 8};
```
StephanTLavavej, line 16: This is an occurrence of "almost always auto" which we conventionally avoid. `mdspan mds1{some_ints.data(), 8, 2, 8};` is the preferred form.

Occurs below.

StephanTLavavej (approved):

StephanTLavavej:
Thank you! I have minor comments for followup but nothing that prevented a merge.

PR #3825, opened 2023-06-23
`<mdspan>`: Make `extents` truly empty when `rank_dynamic() == 0`

JMazurkiewicz:
* Introduce `_Calculate_rank_dynamic` template variable for easier `rank_dynamic()` computations,
* Introduce `_Maybe_empty_array` - a base class for `std::extents` and (in the future) for `std::layout_stride::mapping`,
* Remove `extents::_Static_extents_only` and `extents::_Dynamic_extents`,
* Add tests.

### DIFF
--- a/stl/inc/mdspan
+++ b/stl/inc/mdspan
@@ -25,19 +25,19 @@ _STL_DISABLE_CLANG_WARNINGS
 #undef new
 
 _STD_BEGIN
-template <class _IndexType, size_t _RankDynamic>
-struct _Dynamic_extents_array {
-    array<_IndexType, _RankDynamic> _Dynamic_extents{};
+template <class _IndexType, size_t _Size>
+struct _Maybe_empty_array {
+    array<_IndexType, _Size> _Array{};
 };
 
 template <class _IndexType>
-struct _Dynamic_extents_array<_IndexType, 0> {};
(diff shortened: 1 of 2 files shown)


stl/inc/mdspan
```diff
@@ -25,9 +25,19 @@ _STL_DISABLE_CLANG_WARNINGS
...
+template <class _IndexType, size_t _Size>
+struct _Maybe_empty_array {
+    array<_IndexType, _Size> _Array{};
+};
+
+template <class _IndexType>
+struct _Maybe_empty_array<_IndexType, 0> {};
+
+template <size_t... _Extents>
+inline constexpr size_t _Calculate_rank_dynamic = ((_Extents == dynamic_extent) + ... + 0);
```
StephanTLavavej, line 37: Suggestion: This is adding `bool`s which is kind of weird. On left 43 / right 53 there used to be a `static_cast` making this clearer. I think a `static_cast<size_t>` could help make things clearer here (a rare case where adding unnecessary code might be a good idea).

StephanTLavavej (approved):

StephanTLavavej:
Awesome, thank you! I had one suggestion for followup. :heart_eyes_cat: :tada:

PR #3846, opened 2023-07-01
Status Chart: Add C++26 line

StephanTLavavej:
* Weekly updates.
* Dependency updates.
* As Node.js 20 is the latest major version, mention it in `README.md` and require it in `compile-typescript.yml`. (The current version is 20.4.0 but we don't need to be that specific.)
* In `.vscode/settings.json`, we don't need to set `C_Cpp.autoAddFileAssociations`. This branch doesn't contain any C++ files that could cause the Microsoft C/C++ Extension to want to modify `settings.json`. (This was a relic of how `gh-pages` was initially derived from `main`.)
* Add a cxx26 line. While we aren't yet accepting C++26 PRs in general, the work is still accumulating. This also properly excludes C++26 features from the overall number of non-feature/non-LWG issues.
  + We have a limited number of nice colors from Primer CSS, so I've chosen to simply "rotate" back to the C++17 color. I didn't bother to make lines dashed, since there's plenty of separation between C++17 and C++26 to avoid any visual ambiguity.
  + Note that for the labels (mentioned once, near "avoid double-counting"), we are using different colors than the chart lines; the repo is using the same color for C++20 and C++26, so I have mirrored that here.

Because this involves coordinated changes between TS and HTML, there's a "REGEN BEFORE MERGING" commit that will accumulate merge conflicts.

:chart_with_upwards_trend: Live preview: [stephantlavavej.github.io/STL/](https://stephantlavavej.github.io/STL/)
===


StephanTLavavej:
Rebased and updated the weekly data and dependencies again. (As usual, the Status Chart is a special exception to our policy of discouraging force-pushes in active PRs.)

barcharcraz (approved):

PR #3847, opened 2023-07-05
`<mdspan>`: More EBO

JMazurkiewicz:
* Add new base classes required for EBO implementation:
  * `_Maybe_fully_static_extents` - EBO for `layout_left::mapping`,`layout_right::mapping` and `layout_stride::mapping`,
  * `_Mdspan_mapping_base` - EBO for `mdspan` (`_Map` member only),
  * `_Mdspan_accessor_base` - EBO for `mdspan` (`_Acc` member only),
* Add `/* [[no_unique_address]] */` comments when EBO cannot be applied.
  * Ideally we would get rid of those base classes and just use this attribute (see #1364).
* Drive-by: address comment from #3825 (`static_cast<size_t>` in `_Calculate_rank_dynamic`).
* Add tests.

### DIFF
--- a/stl/inc/mdspan
+++ b/stl/inc/mdspan
@@ -456,7 +456,7 @@ struct _Maybe_fully_static_extents<_Extents> {
     template <class _OtherExtents>
     constexpr explicit _Maybe_fully_static_extents([[maybe_unused]] const _OtherExtents& _Exts_) {
 #if _CONTAINER_DEBUG_LEVEL > 0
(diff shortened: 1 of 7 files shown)


StephanTLavavej (approved):

StephanTLavavej:
Thanks! Noticed a couple of issues in test comments that can be fixed in a followup PR, but they weren't worth resetting testing. :heart_eyes_cat:

PR #3856, opened 2023-07-10
Status Chart: Fix error TS2590

StephanTLavavej:
The Status Chart's automated daily update started failing with "error TS2590: Expression produces a union type that is too complex to represent." This is because we've accumulated so many days of history that TypeScript's advanced type inference is reaching internal limits.

The fix is very simple. When generating our `DailyRow[]` table, add `as DailyRow` type assertions to each element. This avoids making TypeScript deduce what each element could be, because we already know what type we ultimately want.

This PR also contains unrelated housekeeping - minor dependency updates, and increasing the mentioned Node.js version to 20.4.0 (I've gone back and forth on this; I think we should be a bit more diligent about keeping this updated).

Finally, this contains a "DROP BEFORE MERGING" commit to make the live preview work, which will accumulate merge conflicts.

:chart_with_upwards_trend: Live preview: [stephantlavavej.github.io/STL/](https://stephantlavavej.github.io/STL/)
===

### DIFF
--- a/.github/workflows/compile-typescript.yml
+++ b/.github/workflows/compile-typescript.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: ">=20.0.0"
+          node-version: ">=20.4.0"
       - name: Install Packages
         run: |
           npm ci
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ This branch generates the STL's [Status Chart][].
 
 ## Repo
 
-1. Install [Node.js][] 20.0.0 or newer.
+1. Install [Node.js][] 20.4.0 or newer.
     + You can accept all of the installer's default options.
 2. Open a new Command Prompt.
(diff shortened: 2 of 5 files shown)


AraHaan:
Cant dependabot be configured to do the dependency updates on it automatically as automatic PRs? That would help I think.

strega-nil (approved):
It's unfortunate this is necessary, but thanks!

strega-nil-ms (approved):

AraHaan (approved):
This looks great, although I feel like the package.json and the package-lock.json files could be automated like how I commented above. Overall this change looks good though.

StephanTLavavej:
Dependabot might be able to automate some of the dependency updates but we really don't need more microsoft/STL PRs - batching them up with monthly-ish updates to the manually gathered weekly tables is less work for us overall.

AraHaan:
Yeah, when they implement batch pr's that will bundle all updates in a single PR would be perfect for the microsoft/STL repository in this case as well.

PR #3863, opened 2023-07-11
Initialize array<T, 0>'s single element

CaseyCarter:
... which shouldn't even be there, but can't be removed because ABI.

Fixes DevCom-10299275 / VSO-1759562 / AB#1759562.

Notably this will fail with `T`'s that cannot be `{}`-initialized, e.g., types with explicit default constructors. We could work around that in C++17-and-later by initializing the array with `{T()}`, but I didn't want to obfuscate the code and I feel these two corner cases are incredibly unlikely to occur together.

### DIFF
--- a/stl/inc/array
+++ b/stl/inc/array
@@ -761,7 +761,7 @@ public:
 
     conditional_t<disjunction_v<is_default_constructible<_Ty>, _Is_implicitly_default_constructible<_Ty>>, _Ty,
         _Empty_array_element>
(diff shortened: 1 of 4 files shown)


StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for fixing this notoriously annoying bug! :lady_beetle: :hammer_and_wrench: :sparkles:

PR #3864, opened 2023-07-11
Reallocate at least

CaseyCarter:
This reverts the reversion of #3712 from #3819. It also detects allocators publicly derived from `std::allocator`, and refuses to use `allocate_at_least` in such a case to avoid potential mismatch between a user's `deallocate` and the default `allocate_at_least`.

I expect we'll squash merge this, but I've kept it as two separate commits for ease of review: one that reverts the revert, and one that implements the fix.

### DIFF
--- a/stl/inc/xmemory
+++ b/stl/inc/xmemory
@@ -2174,16 +2174,33 @@ _NODISCARD constexpr bool _Allocators_equal(const _Alloc& _Lhs, const _Alloc& _R
     }
 }
 
+#if _HAS_CXX23
+template <class _Ty, class = void>
+inline constexpr bool _Has_member_from_primary = false;
+template <class _Ty>
+inline constexpr bool _Has_member_from_primary<_Ty, void_t<typename _Ty::_From_primary>> = true;
+
+// Avoid using allocate_at_least when the allocator publicly derives from std::allocator:
+// "old" allocators might override allocate and deallocate but fail to override allocate_at_least.
+template <class _Alloc>
+inline constexpr bool _Allocate_at_least_is_safe =
+    !_Has_member_from_primary<_Alloc>
+    && _Has_member_allocate_at_least<_Alloc, typename allocator_traits<_Alloc>::size_type>;
+#endif // _HAS_CXX23
+
 template <class _Alloc>
 _NODISCARD_RAW_PTR_ALLOC _CONSTEXPR20 typename allocator_traits<_Alloc>::pointer _Allocate_at_least_helper(
     _Alloc& _Al, _CRT_GUARDOVERFLOW typename allocator_traits<_Alloc>::size_type& _Count) {
 #if _HAS_CXX23
-    auto [_Ptr, _Allocated] = allocator_traits<_Alloc>::allocate_at_least(_Al, _Count);
-    _Count                  = _Allocated;
-    return _Ptr;
-#else // _HAS_CXX23
-    return _Al.allocate(_Count);
+    if constexpr (_Allocate_at_least_is_safe<_Alloc>) {
+        auto [_Ptr, _Allocated] = _Al.allocate_at_least(_Count);
+        _Count                  = _Allocated;
+        return _Ptr;
+    } else
 #endif // _HAS_CXX23
+    {
+        return _Al.allocate(_Count);
+    }
 }
 
 _EXPORT_STD template <class _FwdIt, class _Ty>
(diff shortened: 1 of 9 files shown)


stl/inc/xmemory
```diff
@@ -2174,6 +2174,35 @@ _NODISCARD constexpr bool _Allocators_equal(const _Alloc& _Lhs, const _Alloc& _R
...
+#if _HAS_CXX23
+template <class _Ty, class = void>
+inline constexpr bool _Has_member_from_primary = false;
+template <class _Ty>
+inline constexpr bool _Has_member_from_primary<_Ty, void_t<typename _Ty::_From_primary>> = true;
+
+// Avoid using allocate_at_least when the allocator publicly derives from std::allocator:
+// "old" allocators might override allocate and deallocate but fail to override allocate_at_least.
+template <class _Alloc>
+inline constexpr bool _Allocate_at_least_is_safe =
```
AlexGuteniev, line 2186: Looks like not true for `std::allocator`, as it has `_From_primary` too, so apparently `_Allocate_at_least_is_safe` is false for `std::allocator`. Why?

CaseyCarter: Because `allocator::allocate_at_least` currently does nothing useful, it simply returns the result of calling `allocate`. That said, we should have a comment in `allocator::allocate_at_least` pointing out that `_Allocate_at_least_is_safe` will need adjustment if and when we improve `allocate_at_least`.

CaseyCarter: I suppose there's a case to be made that `std::allocator::allocate_at_least` is _safe_ despite that we don't want to use it. Should we rename `_Allocate_at_least_is_safe` to `_Should_allocate_at_least` or something?

AlexGuteniev: Yes. Or better yet, start using `std::allocator::allocate_at_least` and avoid the need to rename and a comment.

StephanTLavavej: I like `_Should_allocate_at_least`, I'll validate and push changes. I don't like the idea of calling `std::allocator::allocate_at_least` at this time - that's adding an unnecessary function call in debug mode, and asking the inliner to do extra work in release mode, for no user benefit at this time (or for the foreseeable future).

stl/inc/xmemory
```diff
@@ -2174,6 +2174,35 @@ _NODISCARD constexpr bool _Allocators_equal(const _Alloc& _Lhs, const _Alloc& _R
...
 }
 
+#if _HAS_CXX23
+template <class _Ty, class = void>
+inline constexpr bool _Has_member_from_primary = false;
+template <class _Ty>
+inline constexpr bool _Has_member_from_primary<_Ty, void_t<typename _Ty::_From_primary>> = true;
+
+// Avoid using allocate_at_least when the allocator publicly derives from std::allocator:
+// "old" allocators might override allocate and deallocate but fail to override allocate_at_least.
```
StephanTLavavej, line 2184: Terminology nitpick: only `virtual` functions can be overridden. These are being hidden.

tests/std/tests/GH_003570_allocate_at_least/test.cpp
```diff
@@ -0,0 +1,131 @@
...
+struct icky_allocator : allocator<T> {
+    template <class U>
+    struct rebind {
+        using other = icky_allocator<U>;
+    };
+
+    allocation_result<T*, size_t> allocate_at_least(size_t) {
+        // The initial implementation of this feature had a problem with (icky) allocators that
+        // publicly derive from std::allocator and implement allocate/deallocate. The STL would
+        // call allocate_at_least in the std::allocator base and the overridden deallocate in
```
StephanTLavavej, line 112: Similarly, "overridden" is improper here.

tests/std/tests/GH_003570_allocate_at_least/test.cpp
```diff
@@ -0,0 +1,131 @@
...
+        // the derived class. We now detect public derivation from std::allocator and avoid
+        // using allocate_at_least in that case.
+        static_assert(always_false<T>);
+    }
+};
+
+void test_inheriting_allocator() {
+    std::vector<int, icky_allocator<int>> vec{2, 1, 4, 7, 5, 6, 3, 8};
+    assert(std::accumulate(vec.begin(), vec.end(), 0, plus<>{}) == 36);
```
StephanTLavavej, line 121: This test is already `using namespace std;`.

StephanTLavavej, line 121: Technically should include `<functional>` for `plus`.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for restoring this optimization! May the RWC be ever in your favor. :joy_cat: :rocket: :tada:

PR #3866, opened 2023-07-12
Toolset update: VS 2022 17.7 Preview 3, Clang 16.0.5

### LINKED ISSUE #3713: Clang in VS 17.7p1 doesn't grok `_addcarry_u64`/`_subborrow_u64`
The compiler is emitting calls as if these were separately-compiled functions instead of intrinsics that it should expand inline. This is not the case when installing LLVM 16.0.1 directly, so I suspect some kind of corruption of the VS in-box Clang. We need to figure out what's going on here, and communicate with the MS-internal folks who handle the in-box Clang to get it straightened out before 17.7 ships.

EDIT:
This bug has the same [root cause](https://github.com/microsoft/STL/issues/3713#issuecomment-1564942172) as DevCom-10369658/VSO-1822920 "Clang 16.0 offsetof behavior difference". The issue is in how VS packages LLVM.

StephanTLavavej:
* Remove workarounds for VSO-1664463 "`/clr` C++20 `alignas` emits error C3821 'managed type or function cannot be used in an unmanaged function' instead of falling back to native codegen".
* Remove workarounds for VSO-1152705 "`_malloca_crt` should suppress PREfast warning 6386".
* Remove workaround for DevCom-10334808 "Pointer operands are not correctly converted to the composite pointer type in three-way comparison".
* Followup changes after #3497.
  + Enable the recently-added, off-by-default warning C5267 (for deprecated copy constructors and copy assignment operators) in the `std` and `tr1` test suites to ensure that we remain clean. (`libcxx` contains too many occurrences in its own code to make this feasible.)
  + Fix `aligned_type` in `P0220R1_any` to avoid the warning. We need to work around VSO-1846195 "`/clr` `alignas` emits fatal error C1193: an error expected in yyaction.cpp(2971) not reached".
* Require Node.js 20.4.0 when building the Status Chart.
* Go back to Standard_D32ds_v5. This is an attempt to avoid stalled checks, although it probably won't help.
* Use PowerShell 7.3.5 to build the pool.
* New pool.
* VS 2022 17.7 Preview 3 contains Clang 16.0.5.
* Remove `<xtree>` workaround for VSO-761321 "Failure with local type alias in `if constexpr` condition", which was fixed at some point.
  + #3862 is removing the `<xstring>` workaround, so I'm not changing that here.
* 17.7 Preview 3 fixes #3713, so we can revert the workarounds that I had to introduce in #3711:
  + `_addcarry_u64`/`_subborrow_u64`
  + `LDBL_DECIMAL_DIG`
  + `tgmath.h`
* Remove workaround for LLVM-62762 "32-bit-only misoptimization of `printf` calls on Windows", fixed by Clang 16.0.3.
* Style followup from #3711: Attach `// TRANSITION, VSO-1690654` comment.
* Style followup from #3711: Attempt to improve formatting of `_Can_construct_from_tuple_like`.
* Remove workaround for VSO-1285779 "EDG rejects `source_location` intrinsics within `consteval` functions".
  + This was fixed on 2021-09-07 in VS 2022 17.0.
* Remove workaround for VSO-1268984 "New compiler intrinsics for Layout-compatibility and Pointer-interconvertibility Traits need to be understood by EDG".
  + This was fixed on 2022-03-07 in VS 2022 17.2.
* Remove workaround for VSO-657455 "EDG: dependent name in nested generic lambda not seen as dependent".
  + This was fixed on 2023-01-04 in VS 2022 17.5.
* Rerun clang-format 16.0.5, no manual changes.
* Add workarounds for VSO-1849453 "EDG mishandles `is_pointer_interconvertible_base_of_v<EmptyBase, StatefulDerived>`".
* Add workaround for VSO-1849458 "EDG mishandles `is_layout_compatible_v<const int[], int[]>`".
* Perma-workaround EDG behavioral differences in `source_location::column()`.
  + I consider `column()` to be relatively unimportant and I don't think that these behavioral differences are worth reporting.
* Add workaround for VSO-1849463 "EDG silently rejects `source_location::current()` as a default argument".

x86chk stlimport passed internally.


CaseyCarter (approved):

StephanTLavavej:

PR #3873, opened 2023-07-13
fixing std::shared_ptr<void> initialization from void*

MiguelBarro:
class `shared_ptr` is optimized to avoid keeping a deleter object for non-primitive types which are detected using `_Can_scalar_delete`. This class relies on SFINAE to detect if the operator delete is valid for a given pointer type. It turns out that delete a void* is valid in msvc, though, according to the standard is undefined behaviour (7.6.2.9).  By specializing for void is possible to prevent direct initialization of `shared_ptr<void>` from a void* and prevent bugs. Note that `default_deleter<void>` is not valid too.


achabense:
Is this a compiler bug?
```cpp
#include <memory>

constexpr int foo() {
  void* p = new int;
  delete p;
  return 0;
}

int main() { constexpr auto f = foo(); }
```
[/std:c++14](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:41,endLineNumber:9,positionColumn:41,positionLineNumber:9,selectionStartColumn:41,selectionStartLineNumber:9,startColumn:41,startLineNumber:9),source:'%23include+%3Cmemory%3E%0A%0Aconstexpr+int+foo()+%7B%0A++void*+p+%3D+new+int%3B%0A++delete+p%3B%0A++return+0%3B%0A%7D%0A%0Aint+main()+%7B+constexpr+auto+f+%3D+foo()%3B+%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:vcpp_v19_latest_x64,deviceViewOpen:'1',filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'/std:c%2B%2B14',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x64+msvc+v19.latest+(Editor+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)
[/std:c++20](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:41,endLineNumber:9,positionColumn:41,positionLineNumber:9,selectionStartColumn:41,selectionStartLineNumber:9,startColumn:41,startLineNumber:9),source:'%23include+%3Cmemory%3E%0A%0Aconstexpr+int+foo()+%7B%0A++void*+p+%3D+new+int%3B%0A++delete+p%3B%0A++return+0%3B%0A%7D%0A%0Aint+main()+%7B+constexpr+auto+f+%3D+foo()%3B+%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:vcpp_v19_latest_x64,deviceViewOpen:'1',filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'/std:c%2B%2B20',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x64+msvc+v19.latest+(Editor+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

fsb4000:
>Is this a compiler bug?

@achabense  , no. We have `constexpr new` only since C++20.

stl/inc/memory
```diff
@@ -1486,6 +1486,8 @@ template <class _Yty, class = void>
 struct _Can_scalar_delete : false_type {};
 template <class _Yty>
 struct _Can_scalar_delete<_Yty, void_t<decltype(delete _STD declval<_Yty*>())>> : true_type {};
+template <>
+struct _Can_scalar_delete<void> : false_type {};
```
CaseyCarter, line 1490: We need specializations for `const`-, `volatile`-, and `const volatile`-`void` here as well. Or we could drop this change and have the above partial specialization instead derive from negation<is_void<_Yty>>`.

CaseyCarter (changes requested):

CaseyCarter:
FYI, I'm calling this "bug" rather than "enhancement" since [[util.smartptr.shared.const]/4](https://eel.is/c++draft/util.smartptr.shared.const#4) says we should be diagnosing this as ill-formed.

stl/inc/memory
```diff
@@ -1485,9 +1485,8 @@ private:
 template <class _Yty, class = void>
 struct _Can_scalar_delete : false_type {};
 template <class _Yty>
-struct _Can_scalar_delete<_Yty, void_t<decltype(delete _STD declval<_Yty*>())>> : true_type {};
-template <>
-struct _Can_scalar_delete<void> : false_type {};
+struct _Can_scalar_delete<_Yty, void_t<decltype(delete _STD declval<_Yty*>())>>
+    : std::integral_constant<bool, !bool(is_void<_Yty>::value)> {};
```
CaseyCarter: We conventionally prefer the `bool_constant` alias to `integral_constant<bool, `, and `is_void_v` to `is_void<...>::value` (both of which are already `bool`s). I'll verify and push a style change.

CaseyCarter (approved):

StephanTLavavej (approved):

fsb4000:
`##[error]Path does not exist: D:\build\test-results.xml`

I guess it's because the toolset was updated and the old pool was deleted...

I think after merging commits from main branch the CI will be green.

StephanTLavavej:

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for fixing this bug and congratulations on your first microsoft/STL commit! :rocket: :tada: :partying_face:

PR #3881, opened 2023-07-17
Follow-up enhancements for `<filesystem>`

achabense:
- Fix another problematic `noexcept` signature for `_Read_symlink`
https://github.com/microsoft/STL/blob/f51733ca5352c712165e46fdbc5d0d395971f0e7/stl/inc/filesystem#L3233
- Merge `_Absolute` into `absolute`, as it's simple enough
- Nitpicks for my previous prs: add a missed `// for exception safety` comment; make benchmark cases look better; make the vector reserve slightly more space in `lexically_normal`

### DIFF
--- a/stl/inc/filesystem
+++ b/stl/inc/filesystem
@@ -621,7 +621,7 @@ namespace filesystem {
         friend inline path current_path(error_code& _Ec);
         friend inline void current_path(const path& _To);
         friend inline void current_path(const path& _To, error_code& _Ec) noexcept;
-        friend inline __std_win_error _Read_symlink(const path& _Symlink_path, path& _Result) noexcept;
+        friend inline __std_win_error _Read_symlink(const path& _Symlink_path, path& _Result);
 
     public:
         using value_type  = wchar_t;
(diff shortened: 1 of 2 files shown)


stl/inc/filesystem
```diff
@@ -2969,45 +2969,35 @@ namespace filesystem {
...
-            if (_Full_path_result._Size < _Requested_size) {
-                break;
-            }
-        }
-
-        return _Err;
-    }
-
-    _EXPORT_STD _NODISCARD inline path absolute(const path& _Input) {
+    _EXPORT_STD _NODISCARD inline path absolute(const path& _Input, error_code& _Ec) {
```
achabense, line 2972: I make `absolute(const path&, error_code&)` defined before `absolute(const path&)`. I notice that `...(...,error_code&)` is always defined after `...(...)` in the standard; however, as there are already a lot of examples in this order(`exists`, `read_symlink`, `create_directories` etc.), I think this change is ok.

strega-nil-ms (approved):

stl/inc/filesystem
```diff
@@ -3230,8 +3220,7 @@ namespace filesystem {
             return _Err;
         }
 
-        _Result._Text.resize(_Length);
-        _CSTD memcpy(_Result._Text.data(), _Offset, _Length * sizeof(wchar_t));
+        _Result._Text.assign(_Offset, _Length);
```
StephanTLavavej, line 3223: No change requested: This is great, thank you for noticing it!

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks again for noticing and fixing another one of these bugs! :eagle: :eye: :lady_beetle:

PR #3884, opened 2023-07-17
Fix `string::replace` under ASan

### LINKED ISSUE #3883: `<string>`: `basic_string::replace` fails under ASan when `pos + count > size`, and `count2 < count`
```cxx
#include <string>

int main()
{
  std::string t = "0123456789ABCDEF"; // large string
  t.replace(0, 30, 7, 'A'); // Address Sanitizer Error: Container overflow.
}
```

Looks like we (I) screwed up the logic here.

Reported in DevCom-10369462

strega-nil-ms:
Fixes #3883


tests/std/tests/GH_003883_replace_asan_failure/env.lst
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) Microsoft Corporation.
```
CaseyCarter, line 1: Could we add the test case to the existing string ASan test so we don't need to duplicate this giant `env.lst`? If not, then can we avoid duplication by promoting this up a directory, naming it `usual_asan_something`, and having the three asan tests share it?

StephanTLavavej: The string and vector `env.lst`s differ somewhat. I think we should fuse this into the existing string test that already tests a bugfix. I can validate and push changes.

tests/std/tests/GH_003883_replace_asan_failure/test.cpp
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <string>
+
+int main() {
+    std::string t = "0123456789ABCDEF"; // large string
+    t.replace(0, 30, 7, 'A');
```
StephanTLavavej, line 8: In general, after calling a modifying function, we should verify the final state. This helps catch all sorts of bugs, including those we might not have expected.

tests/std/tests/GH_003883_replace_asan_failure/env.lst
StephanTLavavej: Also, there was a missing `test.lst` change here.

stl/inc/xstring
```diff
@@ -3807,12 +3807,13 @@ public:
             // either we are shrinking, or the growth fits
             // may temporarily overflow; OK because size_type must be unsigned
             const auto _New_size = _Old_size + _Count - _Nx;
-            _ASAN_STRING_MODIFY(*this, _Old_size, _New_size);
+            _ASAN_STRING_REMOVE(*this);
             _Mypair._Myval2._Mysize = _New_size;
```
StephanTLavavej, line 3811: No change requested: We could fuse `_New_size` now, or move its definition below the `_ASAN_STRING_REMOVE`. This is fine as-is though.

StephanTLavavej (approved):

zacklj89 (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for this important ASAN fix! :lady_beetle: :hammer_and_wrench: :tada:

PR #3893, opened 2023-07-20
Remove unnecessary `typename`s for all standard headers introduced in C++20

achabense:
The headers newly introduced in C++20 are:
```
<bit>
<compare>
<concepts>
<coroutine>
<format>
<numbers>
<ranges>
<source_location>
<span>
<syncstream>
<version>
<barrier>
<latch>
<semaphore>
<stop_token>
```

The headers that contain `typename` are:
```
<compare>: Removes two `typename` in `using` clauses; one left as necessary.
<concepts>: All necessary.
<coroutine>: X(Removes one in `using` clause; one left as necessary) update: kept unchanged
<format>: Tackled in last pr.
<numbers>: Removes the only one.
<ranges>: Removes those in `using` clause; the rest(a lot) are necessary.
<span>: All necessary.
<syncstream>: Removes those in `using` clause; one left as necessary.
```

Also removes all `typename` in `<stacktrace>`

Works towards #3718.

### DIFF
--- a/stl/inc/coroutine
+++ b/stl/inc/coroutine
@@ -40,7 +40,7 @@ struct _Coroutine_traits {};
 
 template <class _Ret>
 struct _Coroutine_traits<_Ret, void_t<typename _Ret::promise_type>> {
-    using promise_type = typename _Ret::promise_type;
+    using promise_type = _Ret::promise_type;
 };
 
 _EXPORT_STD template <class _Ret, class...>
--- a/stl/inc/coroutine
+++ b/stl/inc/coroutine
@@ -40,7 +40,7 @@ struct _Coroutine_traits {};
 
 template <class _Ret>
 struct _Coroutine_traits<_Ret, void_t<typename _Ret::promise_type>> {
-    using promise_type = _Ret::promise_type;
+    using promise_type = typename _Ret::promise_type;
 };
 
 _EXPORT_STD template <class _Ret, class...>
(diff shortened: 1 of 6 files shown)


stl/inc/coroutine
```diff
@@ -40,7 +40,7 @@ struct _Coroutine_traits {};
 
 template <class _Ret>
 struct _Coroutine_traits<_Ret, void_t<typename _Ret::promise_type>> {
-    using promise_type = typename _Ret::promise_type;
+    using promise_type = _Ret::promise_type;
```
CaseyCarter, line 43: Despite being added in C++20, we support this header in sub-c++20 modes with `/await:strict` to enable folks to migrate to the new/better/supported coroutine ABI independently of migrating their code to C++20. Let's revert this one:
```suggestion
    using promise_type = typename _Ret::promise_type;
```

CaseyCarter (changes requested):

StephanTLavavej:
Thanks! BTW I have changes to clean up all other headers, I will create a separate PR.

StephanTLavavej (approved):

CaseyCarter (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for these additional `typename` cleanups! :chart_with_upwards_trend: :broom: :heart_eyes_cat:

PR #3892, opened 2023-07-20
Remove unnecessary `typename`s for `<format>`

achabense:
- Remove all unnecessary `typename`s for `<format>` header (#3718).
- (Also, move `_Is_slash_oper & _Is_slash` downstream from `<xfilesystem_abi.h>` to `<filesystem>` as it is only needed there.)

PR #3894, opened 2023-07-20
Remove unnecessary `typename`s in all remaining headers

StephanTLavavej:
Followup to #3892 and #3893. Works towards #3718 - I believe this finishes the product code, but we're open to cleaning up the test code here.

Reported VSO-1854035 "C++20 Down With Typename is confused by fully qualified names".

### DIFF
--- a/stl/inc/vector
+++ b/stl/inc/vector
@@ -226,7 +226,7 @@ public:
 template <class _Myvec>
 struct pointer_traits<_Vector_const_iterator<_Myvec>> {
     using pointer         = _Vector_const_iterator<_Myvec>;
-    using element_type    = const typename pointer::value_type;
+    using element_type    = const typename pointer::value_type; // TRANSITION, VSO-1854115, omit typename
     using difference_type = pointer::difference_type;
 
     _NODISCARD static constexpr element_type* to_address(const pointer _Iter) noexcept {
--- a/stl/inc/vector
+++ b/stl/inc/vector
@@ -227,7 +227,7 @@ template <class _Myvec>
 struct pointer_traits<_Vector_const_iterator<_Myvec>> {
     using pointer         = _Vector_const_iterator<_Myvec>;
     using element_type    = const typename pointer::value_type;
-    using difference_type = typename pointer::difference_type;
+    using difference_type = pointer::difference_type;
 
     _NODISCARD static constexpr element_type* to_address(const pointer _Iter) noexcept {
 #if _ITERATOR_DEBUG_LEVEL != 0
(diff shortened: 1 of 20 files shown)


stl/inc/vector
```diff
@@ -227,7 +227,7 @@ template <class _Myvec>
 struct pointer_traits<_Vector_const_iterator<_Myvec>> {
     using pointer         = _Vector_const_iterator<_Myvec>;
     using element_type    = const typename pointer::value_type;
```
CaseyCarter, line 229: This `typename` still seems to be very "up".

stl/inc/xstring
```diff
@@ -1153,7 +1153,7 @@ template <class _Traits>
 struct pointer_traits<_String_view_iterator<_Traits>> {
     using pointer         = _String_view_iterator<_Traits>;
     using element_type    = const typename pointer::value_type;
```
CaseyCarter, line 1155: Ditto "still up".

CaseyCarter (changes requested):

stl/inc/vector
StephanTLavavej: This is by design - the presence of the `const` means that it's embedded within a type, not appearing at the top level. (At least, that's my understanding of the rule, and MSVC certainly rejects it.)

StephanTLavavej:
@CaseyCarter Please double-check the Standardese for `const typename cats::meow`; I believe that this is not a Down With Typename context.

stl/inc/vector
CaseyCarter: I believe `value_type` is "the terminal name of a _type-specifier_ of a _defining-type-id_", in which case it is in a type-only context per [temp.res]/4.2.2. That said, removing `typename` is such a low-value change that I'm not sure it's worth our time to file a compiler bug.

CaseyCarter: GCC and Clang seem to agree: https://godbolt.org/z/YzYvYvq3P.

CaseyCarter: Fine, I'll file a bug. Guilt me into it, why don't you.

CaseyCarter: This is VSO-1854115 "C++20 Down With Typename is confused by cv-qualifiers".

CaseyCarter (approved):

CaseyCarter:
FYI, I pushed a small change to annotate instances where we can't remove `typename` due to VSO-1854115.

StephanTLavavej:
LGTM, thanks!

StephanTLavavej:

PR #3895, opened 2023-07-21
Drop `#pragma once` in STL headers

StephanTLavavej:
This drops `#pragma once` in `stl/inc` (and the generator `grapheme_break_property_data_gen.py`), relying on our classic idempotency guards (combined with modern compiler abilities to recognize and optimize them; C1XX gained that ability a while ago, so there should be no throughput regressions). This is an attempt to fix VSO-1830439 "\[Feedback\]\[std:c++latest\] C++: 'is not a member' error when importing modules", a scenario which I don't fully understand, but which @cdacamar says will be fixed if the STL stops using `#pragma once`.

Elsewhere, we generally rely on `#pragma once` alone, which is fine. This contains a couple of tiny related cleanups for that (as they are extremely tiny and isolated into separate commits, they shouldn't interfere with reviewing the large mechanical change):

* Move `#pragma once` to the top of `tests/std/include/new_counter.hpp`.
  + This may not be strictly necessary, but it is absolutely conventional elsewhere.
* Rely on `#pragma once` in `tests/tr1/include/cvt_xtest.h`, as it isn't product code anymore.
  + There are all sorts of other cleanups that could be performed here, but we generally avoid messing with `tr1`. In this case, I thought it was worth a change to follow the principle "product code avoids `#pragma once`, test code uses it".

:warning: We should also create PRs for the feature branches in order to avoid this re-appearing.

### DIFF
--- a/stl/inc/__msvc_bit_utils.hpp
+++ b/stl/inc/__msvc_bit_utils.hpp
@@ -3,7 +3,6 @@
 // Copyright (c) Microsoft Corporation.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-#pragma once
 #ifndef __MSVC_BIT_UTILS_HPP
 #define __MSVC_BIT_UTILS_HPP
 #include <yvals_core.h>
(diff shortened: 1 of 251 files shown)


CaseyCarter (approved):

StephanTLavavej:

cdacamar (approved):

PR #3898, opened 2023-07-22
Fix `is_assignable` test coverage

StephanTLavavej:
* Use lvalue references for the LHS of `is_assignable_v` and `is_nothrow_assignable_v` for `variant` and `atomic<shared_ptr>`.
  + We expect lvalues even though it technically works with rvalues.
* Check const rvalues and const lvalues in `P2321R2_proxy_reference`.
  + This is a "nice to have" expansion of test coverage, as both work, and this makes it clear that we didn't test them unintentionally.
* Cite LWG-3661 with the usual syntax.


strega-nil-ms (approved):

StephanTLavavej:

PR #3900, opened 2023-07-22
Update preprocessor `#else` comment in product code to clarify which mode is where

AlexGuteniev:
Towards #351. Product headers.
* Product files, all found places except a few I'm not sure about (plan to do a separate subsequent change).
* No `/tests` yet (but there are fewer places)
* Changed comments like `#else // _HAS_EXCEPTIONS` to  `#else // ^^^ _HAS_EXCEPTIONS / !_HAS_EXCEPTIONS vvv`
* In some of one line in each branch cases removed comment altogether
* Fixed some of pre-existing `^^^ / vvv` comments, mostly like `!_M_CEE` -> `!defined(_M_CEE)`
* Fixed some `#endif` coments to `#endif // defined(...)` / `#endif // !defined(...)`


StephanTLavavej:
I pushed a bunch of commits since it was just easier to fix stuff than do my usual process of adding a GitHub comment followed by a paired commit. Tried to keep them organized:

* Pre-existing: `not _USE_STD_VECTOR_ALGORITHMS` => `!_USE_STD_VECTOR_ALGORITHMS`
* Pre-existing: Simplify `_HAS_CXX20` comments in `<coroutine>`.
* Pre-existing: `^^^ / vvv` => `/`
* `^^^ / vvv` => `/`
* Fix `!!defined(__AVX2__)` typos.
* Fix `_CPPUNWIND` arrow comment.
* Extra space: `//  defined(_CPPUNWIND)` => `// defined(_CPPUNWIND)`
* Nested preprocessor logic should always be commented.
* Update `#ifdef __cpp_lib_concepts` comments (mostly pre-existing).
* Fix `_CMPXCHG_MASK_OUT_PADDING_BITS` arrow comment.
* Consistent arrow comments.
* Add missing spaces.
* After `#ifndef`, comment `!defined`.
* Typo: `___FORCE_INSTANCE` had 3 underscores.
* Drop an unnecessary empty line.
* Fix arrow comment.
* Within `_HAS_CXX20`, note that we have a workaround for lacking concepts.
* Fix comments that should have said `defined`; use the shorter `warning disabled / warning enabled` form.
* Non-comment style changes to align with comments: Always use parentheses with `defined`.

StephanTLavavej (approved):

AlexGuteniev:
I see da43a31 restored a couple of `#else // _M_CEE`.
Will not push a commit to address this, as there's already one approve, and I was planning a subsequent change for few remaining occurrences anyway.

StephanTLavavej:
The principle is that in nested preprocessor code, we shouldn't omit comments. (The only exception are in the feature-test macro test, and the SCL/HID defaults, where they would lead to unproductive clutter.) Arrow comments would be fine, but omitting the inner layer of comments is contrary to our current convention, even though these are pretty simple cases.

strega-nil-ms (approved):

StephanTLavavej:
I've pushed a merge with `main` to resolve a trivial merge conflict in `yvals_core.h`; this PR edited an `#endif` comment next to the addition of `__cpp_lib_formatters`.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
`#endif // ^^^ THANKS! ^^^`
===

Issue #3858, opened 2023-07-11
`<bitset>`: Investigate further performance improvements


StephanTLavavej:
Followup to #3838 which is adding a benchmark - thanks @achabense!

* @AlexGuteniev suggested using `basic_string::resize_and_overwrite`, creating an `_Ugly` version for unconditional use internally. (We generally avoid adding totally novel secret machinery to Standard classes as it can become a maintenance burden, but simply having `_Ugly` names to access Future Technology is easy and common.)
* @AlexGuteniev also suggested investigating whether branchless codegen for assigning `_Elem0` vs. `_Elem1` would be superior. (Note: we cannot assume that their values are consecutive, they could be `'M'` and `'E'`.)
* I suggest investigating SIMD, as this would seem to be highly amenable to vectorization.

Regardless of what's investigated, I recommend profiling each change in isolation.

achabense:
Somewhat off-topic, but I think we should also pay some attention to the maintainance issue of `_Getword`:
https://github.com/microsoft/STL/blob/2261f7edb760eb3fe0726187c818b796dc7ea798/stl/inc/bitset#L447-L449
https://github.com/microsoft/STL/blob/2261f7edb760eb3fe0726187c818b796dc7ea798/tests/std/tests/Dev10_860410_bitset_ctors/test.cpp#L255-L257

AlexGuteniev:
Vectorization approach suggested by @Alcaro : 
```c++
void bits_to_str16_walrus(char* dest, uint16_t src)
{
	__m128i bits = _mm_cvtsi32_si128(src);
	bits = _mm_unpacklo_epi8(bits, bits);
	bits = _mm_unpacklo_epi8(bits, bits);
	bits = _mm_unpacklo_epi8(bits, bits);
	bits = _mm_and_si128(bits, _mm_set1_epi64x(0x8040201008040201));
	bits = _mm_cmpeq_epi8(bits, _mm_setzero_si128());
	bits = _mm_and_si128(bits, _mm_set1_epi8(0x01));
	bits = _mm_xor_si128(bits, _mm_set1_epi8(0x31));
	_mm_storeu_si128((__m128i*)dest, bits);
}
```
https://godbolt.org/z/hK1YaY8fK
(saving it here for now, so that it is not lost while reviewing the current PR)

PR #3904, opened 2023-07-25
Use `string::resize_and_overwrite` in `bitset` to avoid buffer initialization

AlexGuteniev:
Towards #3858.

> * @AlexGuteniev suggested using `basic_string::resize_and_overwrite`, creating an `_Ugly` version for unconditional use internally. (We generally avoid adding totally novel secret machinery to Standard classes as it can become a maintenance burden, but simply having `_Ugly` names to access Future Technology is easy and common.)

This is what was addressed

> * @AlexGuteniev also suggested investigating whether branchless codegen for assigning `_Elem0` vs. `_Elem1` would be superior. (Note: we cannot assume that their values are consecutive, they could be `'M'` and `'E'`.)

This is confirmed to be non applicable. Before the change it was `cmovne` to set `'1'`. After it became `setne al` to set to 0 or 1, then `add al, '0'`. Both are branchless. I think, the compiler will be able to make efficient branchless code for non-consecutive values by going back to `cmovcc`

Benchmark is already there - thanks @achabense!

Benchmark results before:
```
BM_bitset_to_string<15, char>                    922 ns          924 ns       896000
BM_bitset_to_string<64, char>                 125791 ns       125558 ns         5600
BM_bitset_to_string_large_single<char>          2449 ns         2431 ns       263529
BM_bitset_to_string<7, wchar_t>                  539 ns          547 ns      1000000
BM_bitset_to_string<64, wchar_t>                4399 ns         4325 ns       144516
BM_bitset_to_string_large_single<wchar_t>       4792 ns         4708 ns       149333
```
after
```
BM_bitset_to_string<15, char>                    235 ns          234 ns      3200000
BM_bitset_to_string<64, char>                 119085 ns       119978 ns         5600
BM_bitset_to_string_large_single<char>          2111 ns         2100 ns       320000
BM_bitset_to_string<7, wchar_t>                  153 ns          153 ns      4480000
BM_bitset_to_string<64, wchar_t>                4106 ns         4081 ns       172308
BM_bitset_to_string_large_single<wchar_t>       2137 ns         2148 ns       320000
```

The anomality of `BM_bitset_to_string<64, char>` is caused by too angry loop unrolling, and, IMO, should be fixed in compiler, if needs fixing.

### DIFF
--- a/stl/inc/bitset
+++ b/stl/inc/bitset
@@ -347,13 +347,13 @@ public:
     _NODISCARD _CONSTEXPR23 basic_string<_Elem, _Tr, _Alloc> to_string(
         const _Elem _Elem0 = static_cast<_Elem>('0'), const _Elem _Elem1 = static_cast<_Elem>('1')) const {
         // convert bitset to string
-        basic_string<_Elem, _Tr, _Alloc> _Str(_Bits, _Elem0);
-        for (size_t _Pos = 0; _Pos < _Bits; ++_Pos) {
-            if (_Subscript(_Bits - 1 - _Pos)) {
-                _Str[_Pos] = _Elem1;
+        basic_string<_Elem, _Tr, _Alloc> _Str;
+        _Str._Resize_and_overwrite(_Bits, [this, _Elem0, _Elem1](_Elem* _Buf, size_t _Len) {
+            for (size_t _Pos = 0; _Pos < _Len; ++_Pos) {
+                _Buf[_Pos] = _Subscript(_Len - 1 - _Pos) ? _Elem1 : _Elem0;
             }
-        }
-
+            return _Len;
+        });
         return _Str;
     }
 
--- a/stl/inc/xstring
+++ b/stl/inc/xstring
@@ -4161,7 +4161,7 @@ public:
 #pragma warning(push)
 #pragma warning(disable : 4018) // '<=': signed/unsigned mismatch (we compare to 0 before _New_size below, so it's safe)
     template <class _Operation>
-    constexpr void 
+    constexpr void
 #if _HAS_CXX23
         resize_and_overwrite
 #else // ^^^ _HAS_CXX23 / !_HAS_CXX23 vvv
@@ -4191,7 +4191,7 @@ public:
     constexpr void _Resize_and_overwrite(_CRT_GUARDOVERFLOW const size_type _New_size, _Operation _Op) {
         return resize_and_overwrite(_New_size, _Op);
     }
-#endif // !_HAS_CXX23
+#endif // _HAS_CXX23
 
     _NODISCARD _CONSTEXPR20 size_type capacity() const noexcept {
         return _Mypair._Myval2._Myres;
--- a/stl/inc/xstring
+++ b/stl/inc/xstring
@@ -4189,7 +4189,7 @@ public:
 #if _HAS_CXX23
     template <class _Operation>
     constexpr void _Resize_and_overwrite(_CRT_GUARDOVERFLOW const size_type _New_size, _Operation _Op) {
-        return resize_and_overwrite(_New_size, _Op);
+        resize_and_overwrite(_New_size, _Op);
     }
 #endif // _HAS_CXX23
 
--- a/stl/inc/xstring
+++ b/stl/inc/xstring
@@ -4158,11 +4158,16 @@ public:
         }
     }
 
-#if _HAS_CXX23
 #pragma warning(push)
 #pragma warning(disable : 4018) // '<=': signed/unsigned mismatch (we compare to 0 before _New_size below, so it's safe)
     template <class _Operation>
-    constexpr void resize_and_overwrite(_CRT_GUARDOVERFLOW const size_type _New_size, _Operation _Op) {
+    constexpr void 
+#if _HAS_CXX23
+        resize_and_overwrite
+#else // ^^^ _HAS_CXX23 / !_HAS_CXX23 vvv
+        _Resize_and_overwrite
+#endif // ^^^ !_HAS_CXX23 ^^^
+        (_CRT_GUARDOVERFLOW const size_type _New_size, _Operation _Op) {
         if (_Mypair._Myval2._Myres < _New_size) {
             _Reallocate_grow_by(_New_size - _Mypair._Myval2._Mysize,
                 [](_Elem* const _New_ptr, const _Elem* const _Old_ptr, const size_type _Old_size) {
@@ -4180,7 +4185,13 @@ public:
         _Eos(static_cast<size_type>(_Result_size));
     }
 #pragma warning(pop)
-#endif // _HAS_CXX23
+
+#if _HAS_CXX23
+    template <class _Operation>
+    constexpr void _Resize_and_overwrite(_CRT_GUARDOVERFLOW const size_type _New_size, _Operation _Op) {
+        return resize_and_overwrite(_New_size, _Op);
+    }
+#endif // !_HAS_CXX23
 
     _NODISCARD _CONSTEXPR20 size_type capacity() const noexcept {
         return _Mypair._Myval2._Myres;


stl/inc/bitset
```diff
@@ -347,13 +347,13 @@ public:
     _NODISCARD _CONSTEXPR23 basic_string<_Elem, _Tr, _Alloc> to_string(
         const _Elem _Elem0 = static_cast<_Elem>('0'), const _Elem _Elem1 = static_cast<_Elem>('1')) const {
         // convert bitset to string
-        basic_string<_Elem, _Tr, _Alloc> _Str(_Bits, _Elem0);
-        for (size_t _Pos = 0; _Pos < _Bits; ++_Pos) {
-            if (_Subscript(_Bits - 1 - _Pos)) {
-                _Str[_Pos] = _Elem1;
+        basic_string<_Elem, _Tr, _Alloc> _Str;
+        _Str._Resize_and_overwrite(_Bits, [this, _Elem0, _Elem1](_Elem* _Buf, size_t _Len) {
```
StephanTLavavej, line 351: No change requested: I observed that we could ignore `_Len` and take advantage of our knowledge that it will always be the compile-time constant `_Bits`, which is available without capturing. I checked the codegen, and this indeed allows the compiler to bake in the constant. However, while benchmarking showed some improvement for `BM_bitset_to_string<64, char>`, it showed significant pessimization for `BM_bitset_to_string<15, char>` .

```
BEFORE USING _Bits
BM_bitset_to_string<15, char>                    140 ns          144 ns      4977778
BM_bitset_to_string<64, char>                   3336 ns         3348 ns       224000
BM_bitset_to_string_large_single<char>          2008 ns         2040 ns       344615
BM_bitset_to_string<7, wchar_t>                 88.3 ns         87.9 ns      7466667
BM_bitset_to_string<64, wchar_t>                2211 ns         2197 ns       298667
BM_bitset_to_string_large_single<wchar_t>       1451 ns         1430 ns       448000

AFTER USING _Bits
BM_bitset_to_string<15, char>                    640 ns          642 ns      1120000
BM_bitset_to_string<64, char>                   2646 ns         2609 ns       263529
BM_bitset_to_string_large_single<char>          2010 ns         2040 ns       344615
BM_bitset_to_string<7, wchar_t>                 82.3 ns         82.0 ns      8960000
BM_bitset_to_string<64, wchar_t>                2407 ns         2441 ns       320000
BM_bitset_to_string_large_single<wchar_t>       1372 ns         1360 ns       448000
```

StephanTLavavej (approved):

stl/inc/bitset
AlexGuteniev: I think despite some pessimization of the certain case, it is still the right thing to do

stl/inc/xstring
```diff
@@ -4180,6 +4185,12 @@ public:
         _Eos(static_cast<size_type>(_Result_size));
     }
 #pragma warning(pop)
+
+#if _HAS_CXX23
+    template <class _Operation>
+    constexpr void _Resize_and_overwrite(_CRT_GUARDOVERFLOW const size_type _New_size, _Operation _Op) {
+        return resize_and_overwrite(_New_size, _Op);
```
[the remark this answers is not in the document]
AlexGuteniev: Might be it worth to make `_Resize_and_overwrite` calling `resize_and_overwrite` instead of the opposite, and use `_Pass_fn` in `resize_and_overwrite` ?

stl/inc/bitset
AlexGuteniev: Oh, actually the pessimization is too much. I think, angry loop unrolling strikes here too.

AlexGuteniev: Let's keep it as a parameter for now then

stl/inc/xstring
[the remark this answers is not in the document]
StephanTLavavej: > Might be it worth to make `_Resize_and_overwrite` calling `resize_and_overwrite` instead of the opposite

That's what's currently happening - did you intend to say the other order?

I think it's desirable for the Standard name, when available, to not be a wrapper. I don't think we do that consistently elsewhere, but in this new case it seems reasonable. Thus I think the code is fine as-is.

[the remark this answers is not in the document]
AlexGuteniev: Yes I meant the opposite.

> I think it's desirable for the Standard name, when available, to not be a wrapper

Ok. I did it for another reason, specifically to avoid extra copying of a potentially heavy object, but let's keep it for this reason then.

stl/inc/bitset
achabense: As `_Bits` is the natural decision but we have to use `_Len` for performance reasons, I think we need to add a comment for it.

achabense: Hmm, how to compare benchmark results before and after change in the same run in a convenient way?

<details>
<summary>Benchmark</summary>

```cpp
#include <array>
#include <bit>
#include <bitset>

#include <climits>
#include <cstddef>
#include <cstdint>
#include <random>
#include <iostream>

using namespace std;

struct impl_old {
    template <class _Elem = char, size_t _Bits, class _Tr = char_traits<_Elem>, class _Alloc = allocator<_Elem>>
    static basic_string<_Elem, _Tr, _Alloc> to_string(const std::bitset<_Bits>& bs,
        const _Elem _Elem0 = static_cast<_Elem>('0'), const _Elem _Elem1 = static_cast<_Elem>('1')) {
        basic_string<_Elem, _Tr, _Alloc> _Str(_Bits, _Elem0);
        for (size_t _Pos = 0; _Pos < _Bits; ++_Pos) {
            if (bs[_Bits - 1 - _Pos]) {
                _Str[_Pos] = _Elem1;
            }
        }
        return _Str;
    }
};

struct impl_new_Bits {
    template <class _Elem = char, size_t _Bits, class _Tr = char_traits<_Elem>, class _Alloc = allocator<_Elem>>
    static basic_string<_Elem, _Tr, _Alloc> to_string(const std::bitset<_Bits>& bs,
        const _Elem _Elem0 = static_cast<_Elem>('0'), const _Elem _Elem1 = static_cast<_Elem>('1')) {
        basic_string<_Elem, _Tr, _Alloc> _Str;
        _Str.resize_and_overwrite(_Bits, [&bs, _Elem0, _Elem1](_Elem* _Buf, size_t) {
            for (size_t _Pos = 0; _Pos < _Bits; ++_Pos) {
                _Buf[_Pos] = bs[_Bits - 1 - _Pos] ? _Elem1 : _Elem0;
            }
            return _Bits;
            });
        return _Str;
    }
};

struct impl_new_Len {
    template <class _Elem = char, size_t _Bits, class _Tr = char_traits<_Elem>, class _Alloc = allocator<_Elem>>
    static basic_string<_Elem, _Tr, _Alloc> to_string(const std::bitset<_Bits>& bs,
        const _Elem _Elem0 = static_cast<_Elem>('0'), const _Elem _Elem1 = static_cast<_Elem>('1')) {
        basic_string<_Elem, _Tr, _Alloc> _Str;
        _Str.resize_and_overwrite(_Bits, [&bs, _Elem0, _Elem1](_Elem* _Buf, size_t _Len) {
            for (size_t _Pos = 0; _Pos < _Len; ++_Pos) {
                _Buf[_Pos] = bs[_Len - 1 - _Pos] ? _Elem1 : _Elem0;
            }
            return _Len;
            });
        return _Str;
    }
};

namespace {
    const auto random_bits = [] {
        mt19937_64 rnd{};
        array<uint64_t, 32> arr;
        for (auto& d : arr) {
            d = rnd();
        }
        return arr;
        }();

        template <class op, size_t N, class charT>
        void BM_bitset_to_string(benchmark::State& state) {
            for (auto _ : state) {
                for (const auto& bits : random_bits) {
                    bitset<N> bs{ bits };
                    benchmark::DoNotOptimize(op::template to_string<charT>(bs));
                }
            }
        }

        template <class op, class charT>
        void BM_bitset_to_string_large_single(benchmark::State& state) {
            const auto large_bitset = bit_cast<bitset<CHAR_BIT * sizeof(random_bits)>>(random_bits);
            for (auto _ : state) {
                benchmark::DoNotOptimize(op::template to_string<charT>(large_bitset));
            }
        }
} // namespace

#define MAKE_BENCHMARKS(impl)\
BENCHMARK(BM_bitset_to_string<impl, 15, char>)->Name("\nBM_bitset_to_string<" #impl "_old, 15, char>");\
BENCHMARK(BM_bitset_to_string<impl, 64, char>);\
BENCHMARK(BM_bitset_to_string_large_single<impl, char>);\
BENCHMARK(BM_bitset_to_string<impl, 7, wchar_t>);\
BENCHMARK(BM_bitset_to_string<impl, 64, wchar_t>);\
BENCHMARK(BM_bitset_to_string_large_single<impl, wchar_t>);

MAKE_BENCHMARKS(impl_old);
MAKE_BENCHMARKS(impl_new_Bits);
MAKE_BENCHMARKS(impl_new_Len);

BENCHMARK_MAIN();
```
</details>

(And I'm feeling nervious that it seems the effect is very complex and can even vary between different runs)
image
```
---------------------------------------------------------------------------------------------------
Benchmark                                                         Time             CPU   Iterations
---------------------------------------------------------------------------------------------------

BM_bitset_to_string<impl_old_old, 15, char>                   1531 ns         1549 ns       373333
BM_bitset_to_string<impl_old, 64, char>                        8230 ns         8231 ns       112000
BM_bitset_to_string_large_single<impl_old, char>               6477 ns         6487 ns       149333
BM_bitset_to_string<impl_old, 7, wchar_t>                       642 ns          638 ns       857710
BM_bitset_to_string<impl_old, 64, wchar_t>                     7436 ns         7533 ns       112000
BM_bitset_to_string_large_single<impl_old, wchar_t>            4339 ns         4325 ns       144516

BM_bitset_to_string<impl_new_Bits_old, 15, char>              1156 ns         1172 ns       560000
BM_bitset_to_string<impl_new_Bits, 64, char>                   6346 ns         6452 ns        89600
BM_bitset_to_string_large_single<impl_new_Bits, char>          2344 ns         2354 ns       278756
BM_bitset_to_string<impl_new_Bits, 7, wchar_t>                  210 ns          207 ns      4824615
BM_bitset_to_string<impl_new_Bits, 64, wchar_t>                6676 ns         6696 ns        74667
BM_bitset_to_string_large_single<impl_new_Bits, wchar_t>       4516 ns         4450 ns       154483

BM_bitset_to_string<impl_new_Len_old, 15, char>                251 ns          254 ns      3200000
BM_bitset_to_string<impl_new_Len, 64, char>                    5659 ns         5755 ns        89600
BM_bitset_to_string_large_single<impl_new_Len, char>           3870 ns         3850 ns       280000
BM_bitset_to_string<impl_new_Len, 7, wchar_t>                   255 ns          255 ns      3733333
BM_bitset_to_string<impl_new_Len, 64, wchar_t>                15309 ns        15415 ns        72983
BM_bitset_to_string_large_single<impl_new_Len, wchar_t>        7019 ns         7150 ns        89600
```

achabense: Are there problems with my comparing benchmark? I'm finding `resize_and_overwrite`&`_Len` approach stably slow for large wchar_t cases in my testing.

AlexGuteniev: It may be CPU-dependent. 
For `_Bits` the compiler is more likely to advantage of knowing `_Bits` and unroll loops or vectorize something.
For `_Len` the compiler may be restricted from doing that.

Whether the optimizations like loop unrolling are useful or harmful -- dependent on CPU type.

Also non-unrolled loop (`_Len`) is subject to JCC errata pessimization on some CPUs, `/QIntel-jcc-erratum` might help in this case https://learn.microsoft.com/ru-ru/cpp/build/reference/qintel-jcc-erratum?view=msvc-170

I'm neutral with regards to `_Bits` vs `_Len` if none of them are absolutely advantage.

achabense: I tend to believe that by using `_Bits` we can get general improvements over previous impl, though not significant in some cases like `small_bitset.to_string<char>`. By using `_Len` we are introducing some trade-offs.

StephanTLavavej (approved):

stl/inc/bitset
AlexGuteniev: I suggest don't think too much on `_Bits` vs `_Len`. The next step would be vectorization, it would be obviously faster than this, and obviously will use `_Bits`.

strega-nil-ms (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for further optimizing this function! :rabbit2: :rabbit2: :rabbit2:

PR #3912, opened 2023-07-30
Cleanups for mutex utilities

achabense:
- Removes declaration of `_Mtx_timedlock` and `_Mtx_getconcrtcs` from `<xthreads.h>`, as they are not used in the library.
- Slightly simplifies `mtx_do_lock` in `"mutex.cpp"`


AlexGuteniev (approved):

stl/src/mutex.cpp
```diff
@@ -77,26 +77,28 @@ void _Mtx_destroy(_Mtx_t mtx) { // destroy mutex
 }
 
 static _Thrd_result mtx_do_lock(_Mtx_t mtx, const _timespec64* target) { // lock mutex
+    // TRANSITION, ABI: the use of `const _timespec64*` is preserved for `_Mtx_timedlock`
+    const long current_thread_id = static_cast<long>(GetCurrentThreadId());
```
jovibor, line 81: Nitpick:
No need to repeat the type twice, `auto` is enough.

stl/src/mutex.cpp
```diff
@@ -123,31 +126,25 @@ static _Thrd_result mtx_do_lock(_Mtx_t mtx, const _timespec64* target) { // lock
...
         }
 
-        switch (res) {
-        case WAIT_OBJECT_0:
-        case WAIT_ABANDONED:
+        if (res == WAIT_OBJECT_0) {
             return _Thrd_result::_Success;
-
-        case WAIT_TIMEOUT:
+        } else { // WAIT_TIMEOUT
```
jovibor, line 142: Nitpick:
No need to `else` after the `return`.

stl/src/mutex.cpp
```diff
@@ -107,11 +109,12 @@ static _Thrd_result mtx_do_lock(_Mtx_t mtx, const _timespec64* target) { // lock
             }
 
         } else { // check timeout
+            // TRANSITION, ABI: this branch is preserved for `_Mtx_timedlock`
```
achabense, line 112: This is the only branch that has a condition when `current_thread_id = static_cast<long>(GetCurrentThreadId());` becomes an unnecessary overhead. That is, `while(now.tv_sec < target->tv_sec || now.tv_sec == target->tv_sec && now.tv_nsec < target->tv_nsec)` break directly on the first try.
This happens when `"target>0"` (so the second branch is passed), but `"target<=now"` on the first try.

However, I also notice that, as commented in the second branch, we attempt to do a try-lock when the call is `"timed wait for time that has passed"`, but there is no such attempt on this branch. Also, it seems the second branch recognize `"time that has passed"` as `"target<=0"`, but this branch recognize as `"target<=now"`.

I think we could simply ignore this problem and makes no changes here, as this branch is already outdated, and the real fix is to drop this branch entirely when we can safely remove `on the first try`.

stl/src/mutex.cpp
```diff
@@ -77,26 +77,28 @@ void _Mtx_destroy(_Mtx_t mtx) { // destroy mutex
 }
 
 static _Thrd_result mtx_do_lock(_Mtx_t mtx, const _timespec64* target) { // lock mutex
+    // TRANSITION, ABI: the use of `const _timespec64*` is preserved for `_Mtx_timedlock`
+    const long current_thread_id = static_cast<long>(GetCurrentThreadId());
     if ((mtx->_Type & ~_Mtx_recursive) == _Mtx_plain) { // set the lock
-        if (mtx->_Thread_id != static_cast<long>(GetCurrentThreadId())) { // not current thread, do lock
+        if (mtx->_Thread_id != current_thread_id) { // not current thread, do lock
```
achabense, line 83: It seems that this branch `if ((mtx->_Type & ~_Mtx_recursive) == _Mtx_plain)` is only used in the outdated `_Thrd_create`?

achabense: The legacy burden looks very heavy...

stl/inc/xthreads.h
```diff
@@ -111,10 +111,8 @@ _CRTIMP2_PURE void __cdecl _Mtx_destroy_in_situ(_Mtx_t);
 _CRTIMP2_PURE int __cdecl _Mtx_current_owns(_Mtx_t);
 _CRTIMP2_PURE _Thrd_result __cdecl _Mtx_lock(_Mtx_t);
 _CRTIMP2_PURE _Thrd_result __cdecl _Mtx_trylock(_Mtx_t);
-_CRTIMP2_PURE _Thrd_result __cdecl _Mtx_timedlock(_Mtx_t, const _timespec64*);
 _CRTIMP2_PURE _Thrd_result __cdecl _Mtx_unlock(_Mtx_t); // TRANSITION, ABI: Always succeeds
```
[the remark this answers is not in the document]
achabense: The same goes for `_Cnd_init` and `_Cnd_destroy`.
Can we define `_Mtx_init`, `_Mtx_destroy`, and `_Cnd_init`, `_Cnd_destroy` into `"cthread.cpp"` (they are used by `_Thrd_create` there) and hide them from `<xthreads.h>`?

stl/src/mutex.cpp
achabense: Can we define new versions of  `_Mtx_lock` and `_Mtx_trylock` for their real active users (`std::mutex`; `std::recursive_mutex`)?

achabense:
As these functionalities are highly related, can we collect all outdated functions into a single file? (Including `_Mtx_lock`, if [this comment](https://github.com/microsoft/STL/pull/3912#discussion_r1278572188) holds valid)

stl/inc/xthreads.h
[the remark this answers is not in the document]
achabense: Further, as these functionalities are highly related, can we collect all outdated functions into a single file? (Including `_Mtx_lock`, if [this comment](https://github.com/microsoft/STL/pull/3912#discussion_r1278572188) holds valid)

stl/src/mutex.cpp
achabense: I believe this should be a do-while loop.

stl/src/mutex.cpp
```diff
@@ -123,31 +126,25 @@ static _Thrd_result mtx_do_lock(_Mtx_t mtx, const _timespec64* target) { // lock
             }
```
achabense, line 126:
```suggestion
            do {
                if (mtx->_Thread_id == current_thread_id
                    || TryAcquireSRWLockExclusive(get_srw_lock(mtx)) != 0) { // stop waiting
                    res = WAIT_OBJECT_0;
                    break;
                }
                _Timespec64_get_sys(&now);
            } while (now.tv_sec < target->tv_sec || now.tv_sec == target->tv_sec && now.tv_nsec < target->tv_nsec);
```

stl/src/mutex.cpp
StephanTLavavej: While that doesn't save any characters, it is indeed consistent with our usual conventions.

stl/src/mutex.cpp
StephanTLavavej: Also consistent with our usual conventions, although this one's debatable (we don't *strictly* avoid `else` when there's little logic afterwards).

stl/inc/xthreads.h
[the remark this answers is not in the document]
StephanTLavavej: We can guard them with `#ifdef _CRTBLD` for a start. (This macro is defined only when building the STL's separately compiled sources; it is not defined when users use the STL's headers.)

I can add such comments.

Moving functions around, or adding files, is possible, but it's a lot of work (including internal build system work for new files) for generally little reward. I'm not too enthusiastic about such changes.

stl/src/mutex.cpp
StephanTLavavej: I think that it's safe to add a comment.

Adding variants of `_Mtx_lock` and `_Mtx_trylock` would need to be done in the import lib. I cannot emphasize strongly enough how difficult it is to mess with the STL's separately compiled code (specifically the interface between the headers and the separately compiled code; changes that are entirely on one side are less risky) - we need to do it to add new features, and it can be worth it to fix important bugs or improve performance measurably. However, for pure cleanups, it is almost never worth the effort and risk. (The XP/Vista cleanups were an important counterexample, where we were able to eradicate so much code it was worth it, and this eventually enabled `constexpr mutex`.)

This is not exactly a "no" but it is a plea for caution and thought before attempting to make such changes. We would be persuaded by measurable benchmark improvements.

Because `mtx_do_lock` is `static`, you could create a variant of it for use by `_Mtx_lock`, but I am not sure if that would help improve performance or clarity.

stl/src/mutex.cpp
StephanTLavavej: Changing control flow is risky. I'm not especially eager to review changes with risk of introducing correctness problems, when they affect separately compiled code that's been working for a long time.

(The background here is that the maintainer team is *severely* overloaded and we don't have a lot of review capacity for the foreseeable future, so I would like to encourage PRs to focus on features and bugfixing, like our libcxx failure backlog, instead of changes to existing code that don't significantly improve observable characteristics.)

StephanTLavavej: Resolving conversation without code changes as it is unclear whether anything needs to be added to this PR (which is already getting quite big).

StephanTLavavej (approved):

StephanTLavavej:
I've pushed changes to address accumulated comments:

* Style: Use `auto` after `static_cast`.
* Style: After `if`-`return`, avoid `else`.
* Comment that `_Mtx_plain` is for `_Thrd_create`.

I've pushed one enhancement (related to a comment) that hides more declarations from users. We've seen this pattern elsewhere in the STL - there are some separately compiled functions that are used only between separately compiled STL TUs (in the DLL or static LIB), but we marked them as dllexported and declared them for users. We can at least hide them from users, but we have to be careful to preserve the dllexporting:

* Enhancement: Guard 6 declarations with `#ifdef _CRTBLD`, comment why.

Finally, there is a major correctness fix. (Check the output of `dumpbin /exports` if you don't believe me :smile_cat:) This is related to pre-existing weirdness in the STL that it's time to fix up:

* Fix bincompat: Copy `_CRTIMP2_PURE` and `__cdecl` from `xthreads.h`.
  + Also mark `_Thrd_abort`, `_Thrd_create`, `_Thrd_current`, `_Thrd_equal`, `_Thrd_exit`, `_Thrd_start` as `__cdecl`. This is the default when we build the STL, and this is also how they were originally declared (see internal MSVC-PR-135576 when the declarations were removed).
  + Wrap `cond.cpp` and `mutex.cpp` in `_EXTERN_C`, copied from `xthreads.h`. (Other functions declared in `xthreads.h` were already being defined in `_EXTERN_C` blocks.) This also affects the following:
    - `struct _Cnd_internal_imp_t`
      * Good, it was already declared within `_EXTERN_C` in `xthreads.h`.
    - `enum class __stl_sync_api_modes_enum`
      * Good, it's used only as a parameter type for `__set_stl_sync_api_mode()` which is `extern "C"`.
    - `get_srw_lock()` and `mtx_do_lock()`
      * Good, they're `static`.

I've verified that `dumpbin /exports` is now unaffected.

stl/src/mutex.cpp
achabense: Thanks! I learned a lot from these comments.

StephanTLavavej: You're welcome! I apologize for sounding grumpy in some of them - I was running on little sleep for entirely unrelated reasons. You've been doing careful work (don't worry about the `_CRTIMP2_PURE` thing, nobody knows about that one until they run into it) and your PRs have identified good things to clean up (like the `WAIT_ABANDONED` dead code here). It is true that some cleanups aren't worth the risk, but we do still try to continuously improve the codebase which is what keeps the STL a nice place to implement features in, instead of an ossifying mass of legacy code. This PR is definitely a good cleanup!

achabense:
I've made another push to remove two redundant assignments (`res = WAIT_TIMEOUT`)

StephanTLavavej (approved):

StephanTLavavej:
After carefully looking at the control flow, I agree that the assignments were redundant. (Had to consider the impact of loops.)

CaseyCarter (approved):

stl/inc/xthreads.h
```diff
@@ -104,19 +104,21 @@ enum { // mutex types
     _Mtx_recursive = 0x100
 };
 
-_CRTIMP2_PURE _Thrd_result __cdecl _Mtx_init(_Mtx_t*, int);
-_CRTIMP2_PURE void __cdecl _Mtx_destroy(_Mtx_t);
+#ifdef _CRTBLD
+_CRTIMP2_PURE _Thrd_result __cdecl _Mtx_init(_Mtx_t*, int); // used only by _Thrd_create
```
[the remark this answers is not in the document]
StephanTLavavej: I think that `#ifdef _CRTBLD` *means* "used only internally", so such a comment is redundant.

I see your point about bit-rotting, so because I added `#ifdef _CRTBLD`, I think that supersedes the need for comments.

We have traditionally reserved "// TRANSITION, ABI: preserved for binary compatibility" comments for functions that are *completely* unused, except for potential historical callers. The functions in question are currently being used, just STL-internally. Thus I think that TRANSITION comments aren't warranted. We have #3941 tracking the potential for a `_CRTBLD` cleanup.

StephanTLavavej:
@CaseyCarter I've pushed a commit to drop the "used only by" comments as they are superseded by `_CRTBLD`.

StephanTLavavej (approved):

StephanTLavavej:

CaseyCarter (approved):

StephanTLavavej (approved):

StephanTLavavej:
I pushed a merge with `main` to resolve a trivial conflict in `stl/src/cond.cpp` where #3935 removed `static_assert`s next to where this PR is adding `_CRTIMP2_PURE` and `__cdecl` to `_Cnd_init_in_situ`.

CaseyCarter (approved):

StephanTLavavej:
Thanks for eliminating these unnecessary declarations, and prompting us to fix long-standing codebase inconsistency! :tada: :smile_cat: :rocket:

Issue #2919, opened 2022-07-25
P2286R8 Formatting Ranges


StephanTLavavej:
WG21-P2286R8 Formatting Ranges
WG21-P2585R1 Improve Default Container Formatting
WG21-P2713R1 Escaping Improvements In `std::format`
LWG-3631 `basic_format_arg(T&&)` should use `remove_cvref_t<T>` throughout (affects `formattable` concept)
LWG-3750 Too many papers bump `__cpp_lib_format`
LWG-3839 `range_formatter`'s `set_separator`, `set_brackets`, and `underlying` functions should be `noexcept`
LWG-3881 Incorrect formatting of container adaptors backed by `std::string`
LWG-3892 Incorrect formatting of nested ranges and tuples
LWG-3925 Concept `formattable`'s definition is incorrect

Feature-test macro:
```cpp
#define __cpp_lib_format_ranges 202207L
```

PR #3913, opened 2023-07-30
Implement debug-enabled `formatter` specializations

JMazurkiewicz:
* This PR implements debug-enabled standard `formatter` specializations,
* I intentionally didn't mention this in `yvals_core.h` - I think those changes should have been a part of #3656 and current comment is enough.
* Drive-by: rename tests added in #3656 - use `text_formatting_` prefix instead of `formatting_ranges_` (for consistency) and longer names,
* Towards #2919.


stl/inc/format
```diff
@@ -3554,8 +3562,18 @@ struct formatter {
 _FMT_P2286_BEGIN
 template <class _Ty, class _CharT, _Basic_format_arg_type _ArgType>
 struct _Formatter_base {
+private:
     using _Pc = basic_format_parse_context<_CharT>;
 
+public:
+#if _HAS_CXX23
+    constexpr void set_debug_format() noexcept
+        requires (_Is_debug_enabled_fmt_type(_ArgType))
```
cpplearner, line 3571: I'm slightly concerned that this affects program-defined types that inherit from `std::formatter<int>`. Although this function is constrained, its name can be found by name lookup.

```c++
void set_debug_format(auto&); // #1

struct my_formatter : std::formatter<int> {
    auto parse(auto& x) {
        set_debug_format(*this); // before: OK, calls #1
                                 // after: invalid. Name lookup finds the constrained member
        return x.begin();
    }
};
```

stl/inc/format
```diff
@@ -125,10 +125,18 @@ static_assert(static_cast<int>(_Basic_format_arg_type::_Custom_type) < 16, "must
...
 }
+
 _NODISCARD constexpr bool _Is_arithmetic_fmt_type(_Basic_format_arg_type _Ty) {
     return _Ty > _Basic_format_arg_type::_None && _Ty <= _Basic_format_arg_type::_Long_double_type;
 }
 
+#if _HAS_CXX23
+_NODISCARD consteval bool _Is_debug_enabled_fmt_type(_Basic_format_arg_type _Ty) {
+    return (_Ty == _Basic_format_arg_type::_Char_type) || (_Ty == _Basic_format_arg_type::_CString_type)
+        || (_Ty == _Basic_format_arg_type::_String_type);
```
StephanTLavavej, line 136: Style: Equality has well-known higher precedence than logical operators, so we don't wrap `==` in parens here.

tests/std/tests/P2286R8_text_formatting_debug_enabled_specializations/test.cpp
```diff
@@ -0,0 +1,214 @@
...
+    static_assert(!DebugEnabledSpecialization<formatter<unsigned long long, CharT>>);
+
+    static_assert(!DebugEnabledSpecialization<formatter<bool, CharT>>);
+    static_assert(!DebugEnabledSpecialization<formatter<nullptr_t, CharT>>);
+    static_assert(!DebugEnabledSpecialization<formatter<void*, CharT>>);
+    static_assert(!DebugEnabledSpecialization<formatter<const void*, CharT>>);
+
+    // NB: wchar_t might be defined as a typedef for unsigned short (with '/Zc:wchar_t-')
+    static_assert(!DebugEnabledSpecialization<formatter<unsigned short, CharT>>
+                  == (same_as<CharT, char> || !same_as<CharT, unsigned short>) );
```
StephanTLavavej, line 56: This can be simplified. `same_as<CharT, char>` implies `!same_as<CharT, unsigned short>` so we can just test the latter. Then this is testing `!X == !Y` so we can simplify again.

tests/std/tests/P2286R8_text_formatting_debug_enabled_specializations/test.cpp
```diff
@@ -0,0 +1,214 @@
...
+// holder-format-specs:
+//   member debug-format(opt)
+// member:
+//   0 1 2 ... N (index of member object, single digit)
+// debug-format:
+//   $ (use debug format)
+template <class CharT>
+struct std::formatter<Holder<CharT>, CharT> {
+public:
+    constexpr formatter() : member_index{-1} {}
```
StephanTLavavej, line 81: Suggestion: We could use a data member initializer.

StephanTLavavej (approved):

StephanTLavavej:
I've pushed a merge with `main` to resolve a simple merge conflict in `test_format_support.hpp`, where #3861 added `choose()` overloads for characters, and this PR added overloads for characters and `string_view`s, so we want to keep both.

StephanTLavavej (approved):

StephanTLavavej (approved):

StephanTLavavej:
I've pushed a merge with `main` to resolve a trivial conflict where this PR added `_Is_debug_enabled_fmt_type()` above `_Auto_id_tag` which I enhanced with an explicit default constructor.

barcharcraz (approved):
Looks good, I'm not sure we should be compiling out all the internal stuff on pre-23 mode, but it's consistent with the other parts of this feature and I can't really think of a great reason to _not_ do it this way.

StephanTLavavej:

StephanTLavavej:
Thanks for implementing part of this C++23 `<format>` feature! :tada: :heart_eyes_cat: :rocket:

PR #3916, opened 2023-08-01
`stacktrace.cpp`: Fix `snprintf()` usage

StephanTLavavej:
[C23 WP N3096](https://open-std.org/JTC1/SC22/WG14/www/docs/n3096.pdf) 7.23.6.5 "The `snprintf` function"/2:

> The `snprintf` function is equivalent to `fprintf`, except that the output is written into an array (specified by argument `s`) rather than to a stream. If `n` is zero, nothing is written, and `s` may be a null pointer. Otherwise, output characters beyond the `n`-1<sup>st</sup> are discarded rather than being written to the array, and a null character is written at the end of the characters actually written into the array.

In English, `n` should be the buffer size, **INCLUDING** room for a null terminator.

`stacktrace.cpp` consistently said `sizeof("meow") - 1`, which excludes the null terminator. This was a mistake.

So far I haven't seen actual misbehavior caused by this.

### DIFF
--- a/stl/src/stacktrace.cpp
+++ b/stl/src/stacktrace.cpp
@@ -148,7 +148,7 @@ namespace {
             }
 
             if (displacement != 0) {
-                constexpr size_t max_disp_num = sizeof("+0x1122334455667788") - 1; // maximum possible offset
+                constexpr size_t max_disp_num = sizeof("+0x1122334455667788"); // maximum possible offset
 
                 off = string_fill(fill, off + max_disp_num, str, [displacement, off](char* s, size_t) {
                     const int ret = std::snprintf(s + off, max_disp_num, "+0x%llX", displacement);
@@ -214,7 +214,7 @@ namespace {
             off = source_file(address, str, off, &line, fill);
 
             if (line != 0) {
-                constexpr size_t max_line_num = sizeof("(4294967295): ") - 1; // maximum possible line number
+                constexpr size_t max_line_num = sizeof("(4294967295): "); // maximum possible line number
 
                 off = string_fill(fill, off + max_line_num, str, [line, off](char* s, size_t) {
                     const int ret = std::snprintf(s + off, max_line_num, "(%u): ", line);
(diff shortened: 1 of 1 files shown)


AlexGuteniev (changes requested):
These constanst are shared as `snprintf` parameter,, and `resize_and_overwrite` parameter. The later apears not to include `\0`

AlexGuteniev (changes requested):
Actually there is nothing wrong with the current `snprintf` usage.
When null terminator does not fit, `snprintf` still fills other characters, and return the same result.
The null terminator will be set anyway by `string::resize_and_overwrite`

CaseyCarter (approved):

stl/src/stacktrace.cpp
```diff
@@ -148,7 +148,7 @@ namespace {
             }
 
             if (displacement != 0) {
-                constexpr size_t max_disp_num = sizeof("+0x1122334455667788") - 1; // maximum possible offset
+                constexpr size_t max_disp_num = sizeof("+0x1122334455667788"); // maximum possible offset
 
                 off = string_fill(fill, off + max_disp_num, str, [displacement, off](char* s, size_t) {
                     const int ret = std::snprintf(s + off, max_disp_num, "+0x%llX", displacement);
```
CaseyCarter, line 154: I think we actually want:
```suggestion
                constexpr size_t max_disp_num = sizeof("+0x1122334455667788") - 1; // maximum possible offset

                off = string_fill(fill, off + max_disp_num, str, [displacement, off](char* s, size_t) {
                    const int ret = std::snprintf(s + off, max_disp_num + 1, "+0x%llX", displacement);
```
here. See https://github.com/microsoft/STL/pull/3916#issuecomment-1660870358.

StephanTLavavej:
After thinking about this, Casey's suggested pattern appears to be best. I thought about avoiding the `- 1` followed by the `+ 1` but that's really just embedding the implicit count of the null terminator so it doesn't really simplify things.

AlexGuteniev (approved):

StephanTLavavej:

PR #2671, opened 2022-04-26
add format target to cmake

strega-nil-ms:
This replaces the `parallelize.cpp` hack, as well as allows users to run `format` with CMake.

Additionally, this publishes the diff, so that people without the specific version of clang-format on their machine can download the diff and apply it themselves.

PR #3919, opened 2023-08-01
Modernize `validate.cpp`: Use `<print>` instead of `<cstdio>`

JMazurkiewicz:
This PR replaces uses of `fwprintf` with calls to `print(ln)`.

### DIFF
--- a/tools/validate/validate.cpp
+++ b/tools/validate/validate.cpp
@@ -5,12 +5,14 @@
 #include <algorithm>
 #include <array>
 #include <cassert>
-#include <cstdarg>
 #include <cstdio>
 #include <cstdlib>
 #include <filesystem>
+#include <format>
+#include <print>
 #include <string>
 #include <string_view>
+#include <type_traits>
 #include <utility>
 #include <vector>
 using namespace std;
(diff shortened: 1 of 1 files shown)


tools/validate/validate.cpp
```diff
@@ -192,7 +187,7 @@ void scan_file(
         static_assert(ranges::is_sorted(checked_extensions));
 
         if (ranges::binary_search(checked_extensions, filepath.extension().wstring())) {
-            validation_failure(any_errors, filepath, L"file contains %zu lines with more than %zu columns.\n",
+            validation_failure(any_errors, filepath, "file contains {} lines with more than {} columns.",
```
StephanTLavavej, line 190: No change requested: This fixes a mistake that was introduced by #2671. `validation_failure()` prints a newline, so we don't want a second one here.

StephanTLavavej (approved):

CaseyCarter (approved):

StephanTLavavej:
Thanks for simplifying our beloved validator! :white_check_mark: :tada: :rocket:

PR #3925, opened 2023-08-05
Vectorize `ranges::find_last`

### LINKED ISSUE #3274: `<algorithm>`: `find_last()` could probably be vectorized
It has occurred to us that it should be possible to add a vectorized implementation of `find_last`.

AlexGuteniev:
Resolves #3274

The implementation in `vector_algorithm.cpp` is similar to the forward find, except:
 * Advancing the `_Last` to negative direction
    - For negative advance, `_Rewind_bytes` was introduced. This helps avoiding casting `size_t` to `ptrdiff_t` and running into potential UB due to overflow when the range is more than half addressable space
    - The existing  `_Advance_bytes` was made template to fix pre-existing potentially large `size_t` to `ptrdiff_t` conversion
 * Advance before the indirection, so that start with past-the-last pointer and stop on first
 * Still returning last pointer on failure, save it for such case
 * `_BitScanForward` -> _`BitScanReverse`. `_lzcnt_u32` -> `_tzcnt_u32`
   - both `_BitScanForward` and_`BitScanReverse` index from the least significant bit, so no changes here
   - `_tzcnt_u32` index from most significant bit, so reverse using `31 - x`
   - AVX2 imply `_tzcnt_u32` is present, same as  `_lzcnt_u32`. We have precedent in `<__msvc_bit_utils.hpp>`

The integration in `<algorithm>` is similar to `ranges::find`, except that we don't support unsized ranges (in this regard, similar to `ranges::count`).

The test uses the same random data as for forward find. I made sure it covers all branches of the algorithm.


StephanTLavavej (changes requested):

AlexGuteniev:
I observe that #4004 does a bulk change, which applies here too, will do when that PR lands

stl/inc/algorithm
```diff
@@ -2834,6 +2860,21 @@ namespace ranges {
...
+                          && sized_sentinel_for<_Se, _It>) {
+                const auto _First_ptr = _To_address(_First);
+                const auto _Last_ptr  = _First_ptr + (_Last - _First);
+
+                const auto _Result = __std_find_last_trivial(_First_ptr, _Last_ptr, _Value);
+                if constexpr (is_pointer_v<_It>) {
+                    return {_First + (_Result - _First_ptr), _Last};
+                } else {
+                    return {_STD move(_Result), _STD move(_Last)};
+                }
```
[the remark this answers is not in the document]
AlexGuteniev: I tried to address this, not sure if it is right.

stl/inc/algorithm
```diff
@@ -2834,6 +2860,23 @@ namespace ranges {
...
+                if (!_STD is_constant_evaluated()) {
+                    const auto _First_ptr = _To_address(_First);
+                    const auto _Last_ptr  = _First_ptr + (_Last - _First);
+
+                    const auto _Result = __std_find_last_trivial(_First_ptr, _Last_ptr, _Value);
+                    if constexpr (is_pointer_v<_It>) {
+                        return {_Result, _STD move(_Last)};
+                    } else {
+                        return {_STD move(_First) + (_Result - _First_ptr), _STD move(_Last)};
+                    }
```
StephanTLavavej, line 2875: This is better, but `_STD move(_Last)` is attempting to return `_Se _Last` as the end of a `subrange<_It>`. We know that `sized_sentinel_for<_Se, _It>` holds here, but in general we can't convert `_Se` to `_It`.

@CaseyCarter self-assigned this PR and presumably knows how to fix this (again I could try and probably get something to work).

stl/src/vector_algorithms.cpp
```diff
@@ -1396,6 +1420,58 @@ namespace {
...
+            const void* _Stop_at     = _Last;
+            _Rewind_bytes(_Stop_at, _Avx_size);
+            do {
+                _Rewind_bytes(_Last, 32);
+                const __m256i _Data = _mm256_loadu_si256(static_cast<const __m256i*>(_Last));
+                const int _Bingo    = _mm256_movemask_epi8(_Traits::_Cmp_avx(_Data, _Comparand));
+
+                if (_Bingo != 0) {
+                    const unsigned long _Offset = _lzcnt_u32(_Bingo);
+                    _Advance_bytes(_Last, _Offset ^ 0x1F - (sizeof(_Ty) - 1));
```
StephanTLavavej, line 1443: `31 - _Offset` was changed to `_Offset ^ 0x1F` via a commit saying "xor for better codegen". I'm willing to believe that, but I find the code much less clear, so I think this deserves a comment.

StephanTLavavej, line 1443: This is mixing bitwise `^` with arithmetic `-`, so our stylistic conventions demand parentheses here. In fact, because arithmetic `-` has much higher precedence than bitwise `^`, this looks like a bug.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -399,7 +450,47 @@ void test_various_containers() {
     test_one_container<list<int>>(); // bidi, not vectorizable
 }
 
+#if _HAS_CXX20
+constexpr bool test_constexpr() {
```
StephanTLavavej, line 454: This is adding `constexpr` test coverage for most vectorized algorithms (good!) but I observe that `reverse`, `ranges::reverse`, `reverse_copy`, `ranges::reverse_copy`, `swap_ranges`, `ranges::swap_ranges` are missing here.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -399,7 +450,47 @@ void test_various_containers() {
...
+    assert(max_element(begin(data), end(data)) == end(data) - 2);
+    assert(get<0>(minmax_element(begin(data), end(data))) == begin(data) + 1);
+    assert(get<1>(minmax_element(begin(data), end(data))) == end(data) - 2);
+
+#if defined(__cpp_lib_concepts)
+    assert(ranges::min_element(data) == begin(data) + 1);
+    assert(ranges::max_element(data) == end(data) - 2);
+    assert(ranges::minmax_element(data).min == begin(data) + 1);
+    assert(ranges::minmax_element(data).max == end(data) - 2);
+#endif
```
StephanTLavavej, line 481: Nitpick: This is "far away" so the `#endif` deserves a comment.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -399,7 +450,47 @@ void test_various_containers() {
     test_one_container<list<int>>(); // bidi, not vectorizable
 }
 
+#if _HAS_CXX20
+constexpr bool test_constexpr() {
+    int data[] = {2, 1, 3, 3, 3, 3, 4, 6, 5};
```
StephanTLavavej, line 455: This could be `const`.

StephanTLavavej, line 455: Nitpick: This shadows `std::data`. I've probably lost this war against shadowing (so many tests are saying `expected`, etc.) but I'll mention it anyways.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -399,7 +450,47 @@ void test_various_containers() {
     test_one_container<list<int>>(); // bidi, not vectorizable
 }
 
+#if _HAS_CXX20
+constexpr bool test_constexpr() {
+    int data[] = {2, 1, 3, 3, 3, 3, 4, 6, 5};
+
+    assert(count(begin(data), end(data), 3) == 4);
```
StephanTLavavej, line 457: In test code, I recommend following the principle "unique numbers for unique purposes". This makes it a lot easier to understand what various values mean. Here, because the test data uses small integers, there's visual ambiguity between element values like `3`, counts of elements like `4`, and distances like `2`.

If the test data were simply multiples of 10, then `30` would obviously be an element, making the counts like `4` and distances like `2` visually distinctive.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -399,7 +450,47 @@ void test_various_containers() {
...
+    assert(ranges::count(data, 3) == 4);
+#endif
+
+    assert(find(begin(data), end(data), 3) == begin(data) + 2);
+#if defined(__cpp_lib_concepts)
+    assert(ranges::find(data, 3) == begin(data) + 2);
+#endif
+
+#if defined(__cpp_lib_concepts) && _HAS_CXX23
+    assert(begin(ranges::find_last(data, 3)) == begin(data) + 5);
```
StephanTLavavej, line 468: Can we also check the end of the subrange?

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -123,6 +139,29 @@ void test_find(mt19937_64& gen) {
     }
 }
 
+#if _HAS_CXX23 && defined(__cpp_lib_concepts)
+template <class T>
+void test_case_find_last(const vector<T>& input, T v) {
+    auto expected = last_known_good_find_last(input.begin(), input.end(), v);
+    auto range    = ranges::find_last(input.begin(), input.end(), v);
+    auto actual   = range.begin();
+    assert(expected == actual);
```
StephanTLavavej, line 148: Can we also check the end of the subrange?

StephanTLavavej (changes requested):

stl/src/vector_algorithms.cpp
StephanTLavavej: Even if it's not a bug, we need parentheses to make it clear that this is intentional.

AlexGuteniev: `^` reverted, DevCom-10487304 reported instead.

AlexGuteniev: `^` reverted, DevCom-10487304 reported instead.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -399,7 +449,76 @@ void test_various_containers() {
...
+#ifdef __cpp_lib_concepts
+    assert(ranges::min_element(a) == begin(a) + 1);
+    assert(ranges::max_element(a) == end(a) - 2);
+    assert(ranges::minmax_element(a).min == begin(a) + 1);
+    assert(ranges::minmax_element(a).max == end(a) - 2);
+#endif // defined(__cpp_lib_concepts)
+
+    int b[size(a)];
+    reverse_copy(begin(a), end(a), begin(b));
+    assert(equal(rbegin(a), rend(a), begin(b)));
```
StephanTLavavej, line 485: We should always use 4-arg `equal()` since 3-arg is hazardous. Occurs below.

StephanTLavavej (approved):

stl/src/vector_algorithms.cpp
```diff
@@ -1396,6 +1420,58 @@ namespace {
...
+                _Rewind_bytes(_Last, 32);
+                const __m256i _Data = _mm256_loadu_si256(static_cast<const __m256i*>(_Last));
+                const int _Bingo    = _mm256_movemask_epi8(_Traits::_Cmp_avx(_Data, _Comparand));
+
+                if (_Bingo != 0) {
+                    const unsigned long _Offset = _lzcnt_u32(_Bingo);
+                    _Advance_bytes(_Last, (31 - _Offset) - (sizeof(_Ty) - 1));
+                    return _Last;
+                }
+
```
CaseyCarter, line 1446: Not worth resetting testing, but we conventionally avoid (Casey's pet peeve) empty lines at the start/end of blocks:
```suggestion
```

stl/src/vector_algorithms.cpp
```diff
@@ -1396,6 +1420,58 @@ namespace {
...
+                const __m128i _Data = _mm_loadu_si128(static_cast<const __m128i*>(_Last));
+                const int _Bingo    = _mm_movemask_epi8(_Traits::_Cmp_sse(_Data, _Comparand));
+
+                if (_Bingo != 0) {
+                    unsigned long _Offset;
+                    _BitScanReverse(&_Offset, _Bingo); // lgtm [cpp/conditionallyuninitializedvariable]
+                    _Advance_bytes(_Last, _Offset - (sizeof(_Ty) - 1));
+                    return _Last;
+                }
+
```
CaseyCarter, line 1467: Ditto:
```suggestion
```

CaseyCarter (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for optimizing this new algorithm! :rocket: :rocket: :rocket:

stl/src/vector_algorithms.cpp
```diff
@@ -1143,6 +1153,20 @@ namespace {
         return _Ptr;
     }
 
+    template <class _Ty>
+    const void* _Find_trivial_last_tail(const void* _First, const void* _Last, const void* _Real_last, _Ty _Val) {
```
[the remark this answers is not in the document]
AlexGuteniev: Addressed in #4113

stl/src/vector_algorithms.cpp
AlexGuteniev: Addressed in #4113

stl/src/vector_algorithms.cpp
AlexGuteniev: Addressed in #4113

PR #3927, opened 2023-08-06
Repeat export specifiers on definitions, so that they stay if the declarations are gone

AlexGuteniev:
We should have export specifiers on both function declaration **and** definition.
This is because after some future changes the declaration may be gone - either entirely, or header be no longer included.
By repeating export specifiers we make sure the function type and its impact on export surface stays the same.

See https://github.com/microsoft/STL/pull/3912#issuecomment-1661423190

Additionally, this simplifies the declarations in `stl/src/xmtx.hpp` to match the definitions (the declarations are functionally unchanged).

### DIFF
--- a/stl/src/xmtx.cpp
+++ b/stl/src/xmtx.cpp
@@ -11,16 +11,16 @@ _EXTERN_C_UNLESS_PURE
 
 // Win32 critical sections are recursive
 
-void __CLRCALL_PURE_OR_CDECL _Mtxinit(_Rmtx* _Mtx) noexcept { // initialize mutex
+_MRTIMP2_PURE void __CLRCALL_PURE_OR_CDECL _Mtxinit(_Rmtx* _Mtx) noexcept { // initialize mutex
     InitializeCriticalSectionEx(_Mtx, 4000, 0);
 }
 
-void __CLRCALL_PURE_OR_CDECL _Mtxdst(_Rmtx* _Mtx) noexcept { // delete mutex
+_MRTIMP2_PURE void __CLRCALL_PURE_OR_CDECL _Mtxdst(_Rmtx* _Mtx) noexcept { // delete mutex
     DeleteCriticalSection(_Mtx);
 }
 
 _RELIABILITY_CONTRACT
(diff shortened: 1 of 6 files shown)


StephanTLavavej:
I understand what you're doing here, but can you please provide a PR description so that others can understand (both now, and people in the future doing PR archaeology)? For a PR fixing typos in comments, a title of "Fix typos" with no description would indeed be sufficient, but this PR is doing something much less obvious.

AlexGuteniev:
Updated.

stl/src/xmtx.cpp
```diff
@@ -11,24 +11,24 @@ _EXTERN_C_UNLESS_PURE
 
 // Win32 critical sections are recursive
 
-void __CLRCALL_PURE_OR_CDECL _Mtxinit(_Rmtx* _Mtx) noexcept { // initialize mutex
+_MRTIMP2_PURE void __CLRCALL_PURE_OR_CDECL _Mtxinit(_Rmtx* _Mtx) noexcept { // initialize mutex
```
StephanTLavavej, line 14: This is good, but we should simplify `xmtx.hpp` accordingly.

StephanTLavavej (approved):

stl/src/xmtx.cpp
AlexGuteniev: That would be merge conflict with #3900

StephanTLavavej: Trivial to resolve, thanks for the heads up.

StephanTLavavej:

CaseyCarter (approved):

StephanTLavavej (approved):

StephanTLavavej:
I've pushed a merge with `main` to resolve a conflict in `stl/src/xmtx.hpp` where #3900 updated preprocessor comments. Resolved by accepting this PR's elimination of the preprocessor logic.

StephanTLavavej:
Thanks for fixing more of these inconsistencies and defending against future mistakes! :hammer_and_wrench: :shield: :smile_cat:

PR #3932, opened 2023-08-07
update stl/inc/xtrlcommon: _Is_any_of_v

GoodenoughPhysicsLab:
This is a very simple modification. The original _Is_any_of_v produces a lot of bloated unrolling. Using fold-expression is a better alternative. Because the change was so simple, I didn't issue it, and the tests were fine.

### DIFF
--- a/stl/inc/xtr1common
+++ b/stl/inc/xtr1common
@@ -168,7 +168,7 @@ _INLINE_VAR constexpr bool disjunction_v = disjunction<_Traits...>::value;
 
 template <class _Ty, class... _Types>
 _INLINE_VAR constexpr bool _Is_any_of_v = // true if and only if _Ty is in _Types
-    disjunction_v<is_same<_Ty, _Types>...>;
+    (is_same_v<_Ty, _Types> || ...);
 
 _NODISCARD constexpr bool _Is_constant_evaluated() noexcept { // Internal function for any standard mode
     return __builtin_is_constant_evaluated();
--- a/stl/inc/xtr1common
+++ b/stl/inc/xtr1common
@@ -168,7 +168,15 @@ _INLINE_VAR constexpr bool disjunction_v = disjunction<_Traits...>::value;
 
 template <class _Ty, class... _Types>
 _INLINE_VAR constexpr bool _Is_any_of_v = // true if and only if _Ty is in _Types
+#if (__cplusplus >= 201703L) // >= c++17
+
     (is_same_v<_Ty, _Types> || ...);
+
+#else
+
+    disjunction_v<is_same<_Ty, _Types>...>;
+
+#endif
 
 _NODISCARD constexpr bool _Is_constant_evaluated() noexcept { // Internal function for any standard mode
     return __builtin_is_constant_evaluated();


stl/inc/xtr1common
```diff
@@ -168,7 +168,7 @@ _INLINE_VAR constexpr bool disjunction_v = disjunction<_Traits...>::value;
 
 template <class _Ty, class... _Types>
 _INLINE_VAR constexpr bool _Is_any_of_v = // true if and only if _Ty is in _Types
-    disjunction_v<is_same<_Ty, _Types>...>;
+    (is_same_v<_Ty, _Types> || ...);
```
fsb4000, line 171: Fold expressions are a C++17 feature, but the STL must work in C++14 mode. 
The function '_is_any_of_v' starts with an underscore and may be used for internal C++14 machinery (although I am not sure if it is actually used).

fsb4000: Tests failed.

The machinary is actually used in C++14 mode.

From the CI:

```
D:\build\out\inc\xtr1common(171): error C7518: fold expressions require at least '/std:c++17'
D:\build\out\inc\xlocnum(1216): note: see reference to variable template 'const bool _Is_any_of_v<char,char,wchar_t>' being compiled
D:\build\out\inc\ostream(293): note: see reference to class template instantiation 'std::num_put<_Elem,std::ostreambuf_iterator<_Elem,_Traits>>' being compiled
        with
        [
            _Elem=char,
            _Traits=std::char_traits<char>
        ]
```

Sorry, I think your current PR can't be merged...

If you want to create a PR you could take the issue: https://github.com/microsoft/STL/issues/3929

stl/inc/xtr1common
```diff
@@ -168,7 +168,15 @@ _INLINE_VAR constexpr bool disjunction_v = disjunction<_Traits...>::value;
 
 template <class _Ty, class... _Types>
 _INLINE_VAR constexpr bool _Is_any_of_v = // true if and only if _Ty is in _Types
+#if (__cplusplus >= 201703L) // >= c++17
+
+    (is_same_v<_Ty, _Types> || ...);
+
+#else
```
fsb4000, line 175:
```suggestion
#else // ^^^ _HAS_CXX17 / !_HAS_CXX17 vvv
```

stl/inc/xtr1common
```diff
@@ -168,7 +168,15 @@ _INLINE_VAR constexpr bool disjunction_v = disjunction<_Traits...>::value;
...
 _INLINE_VAR constexpr bool _Is_any_of_v = // true if and only if _Ty is in _Types
+#if (__cplusplus >= 201703L) // >= c++17
+
+    (is_same_v<_Ty, _Types> || ...);
+
+#else
+
     disjunction_v<is_same<_Ty, _Types>...>;
+
+#endif
```
fsb4000, line 179:
```suggestion
#endif // !_HAS_CXX17
```

GoodenoughPhysicsLab:
Thanks, I'll try it again.

StephanTLavavej:
`disjunction_v` might look like it does a lot of work, but the MSVC compiler front-end actually special-cases it to be highly efficient. This might make a difference for Clang, but they could always implement the same kind of compiler optimization.

PR #3933, opened 2023-08-07
update stl/inc/xtr1common: _Is_any_of_v

GoodenoughPhysicsLab:
This is a very simple modification. The original _Is_any_of_v produces a lot of bloated unrolling. Using fold-expression is a better alternative. Because the change was so simple, I didn't issue it, and the tests were fine.

### DIFF
--- a/stl/inc/xtr1common
+++ b/stl/inc/xtr1common
@@ -168,7 +168,11 @@ _INLINE_VAR constexpr bool disjunction_v = disjunction<_Traits...>::value;
 
 template <class _Ty, class... _Types>
 _INLINE_VAR constexpr bool _Is_any_of_v = // true if and only if _Ty is in _Types
+#if _HAS_CXX17
+    (is_same_v<_Ty, _Types> || ...);
+#else
     disjunction_v<is_same<_Ty, _Types>...>;
+#endif /* _HAS_CXX17 */
 
 _NODISCARD constexpr bool _Is_constant_evaluated() noexcept { // Internal function for any standard mode
     return __builtin_is_constant_evaluated();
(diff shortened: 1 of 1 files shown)


GoodenoughPhysicsLab:
Sorry, I was so reckless.

GoodenoughPhysicsLab:
This was my first attempt to submit a pr to a project like this, and there were a lot of things I didn't do well

fsb4000:
The CI will be restarted later.

I have no idea what's wrong: "Received request to deprovision: The request was cancelled by the remote provider."

But your code is fine. 

No changes needed.

CaseyCarter:
> I have no idea what's wrong: "Received request to deprovision: The request was cancelled by the remote provider."

We use so-called "spot" VMs in Azure to minimize CI costs. They cost less, but are subject to being pulled out from underneath us without warning and reallocated to full paying customers.

StephanTLavavej:
The MSVC compiler actually internally optimizes both `disjunction_v` and `is_same_of_v`, so I expect that this will make no different for MSVC, but might help Clang and shouldn't hurt.

StephanTLavavej (approved):

StephanTLavavej:
We use `_Is_any_of_v` commonly enough that this seems like a reasonable, targeted change to make. I'd be skeptical of further changes to conditionally replace usage of `conjunction_v`/`disjunction_v` with fold expressions.

StephanTLavavej:

StephanTLavavej:
Thanks for improving compiler throughput here, and congratulations on your first microsoft/STL commit! :tada: :heart_eyes_cat: :partying_face:

PR #3934, opened 2023-08-08
Consistently use `override` in tests

StephanTLavavej:
As a general rule, we avoid significant cleanups of test code. This is partly because some tests are deliberately doing strange things to exercise weird corners of the product, and "cleaning up" the strangeness could affect test coverage. It's also because test code is much less important for long-term maintainability - if the tests keep passing, we generally don't care about them after they've been written.

However, we do try to enforce some basic conventions across the entire codebase including the tests. In the product code, we now follow a convention of saying only `override`. This avoids redundancy and allows `virtual` and `override` to serve different purposes (now when we see `virtual` we know we're introducing a new virtual member function). And using `override` is *much* safer than not using it.

This is so important for safety, and so non-disruptive, that it's worth making the tests consistent. (Including within the legacy `tr1` test suite that we even more strongly avoid changing.) This PR consists of two commits performing related systematic changes:

* Test style: Drop `virtual` when we `override`.
* Test style: Use `override` instead of `virtual`.

### DIFF
--- a/tests/std/tests/Dev09_199123_tr1_mem_fun_abstract_classes/test.cpp
+++ b/tests/std/tests/Dev09_199123_tr1_mem_fun_abstract_classes/test.cpp
@@ -25,7 +25,7 @@ private:
 
 class Derived : public Base {
 private:
-    virtual void purr() {}
+    void purr() override {}
 };
 
 int main() {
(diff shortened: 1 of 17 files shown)


AlexGuteniev:
Have you ran tests against Core Guidelines with only virtual rules ? `
See example from #2094 : https://godbolt.org/z/jfx9dzb1T

StephanTLavavej:
No - we are not yet attempting to be Core Guidelines clean (there's a subset of important rules that the internal build will complain about).

AlexGuteniev:
Sure I was asking just about few Core Guidelines which catch missing `override` and superfluous `virtual` / `override`

CaseyCarter (approved):

StephanTLavavej:

PR #3935, opened 2023-08-08
Various cleanups

StephanTLavavej:
This PR is admittedly somewhat hypocritical as I've mentioned that maintainer capacity is sharply limited and cleanups take time away from features and bugfixes (including libcxx skip reduction which I'd like to see). That said, while reviewing PRs, I notice a lot of things that aren't worth resetting testing of those PRs, but collectively add up.

We *strongly* discourage "grab bag" PRs, so I've separated out #3934 as a fairly large semi-mechanical test change. The remaining changes here are numerous, but individually small and safe, so it's easier to review (and merge) one PR with well-structured commits instead of 20 ultra-tiny PRs. (The problem with "grab bag" PRs is when large changes are mixed with other changes, or when major semantic changes are mixed with cleanups. The impact of each change here is very small.)

Nevertheless, I can decompose this into smaller PRs if desired (that's the great thing about having fine-grained commits).

The commits:

* Guard `_Xtime_diff_to_millis2` with `#ifdef _CRTBLD`.
  + This isn't needed by `stl/inc` headers, so we don't need to emit this declaration for users. (This doesn't affect the export surface since we continue to declare it for `stl/src` *and* the declaration matches the definition. We shouldn't have exported it at all, which we can fix during vNext - we don't really need an explicit comment saying so, since the presence of `#ifdef _CRTBLD` is sufficient.)
* Guard `_Execute_once`, `_Execute_once_fp_t` with `#ifdef _CRTBLD`.
  + Ditto.
* Style: Group `__msvc_sanitizer_annotate_container.hpp` with other unconditionally included headers.
  + This was added at the bottom during maintenance, but there's (fortunately) no ordering requirement. We should let clang-format sort this normally.
* Formatting: Drop a spurious newline in a `basic_istream` constructor.
* Drop `_CONSTEVAL`, `_Cx_exp2()` can be plain `constexpr`.
  + The only usage of `_Cx_exp2()` is initializing `static constexpr _Cy_t _Cy` and `static constexpr _Ty _Scale1`, so it's always being evaluated at compile-time.
* Style: Change unnecessarily `const _Auto_id_tag` unnamed parameters.
  + Most were already non-`const`.
* Add `explicit _Auto_id_tag() = default;` to defend against mistakes.
  + The Standard now does this conventionally (it defends against `{}`), so we should do this for our own tags.
* Fix citation to nonexistent N5687.
* Update links: `docs.microsoft.com` => `learn.microsoft.com`
  + This skips a domain redirection. This also updates a `support.microsoft.com` URL in the README that's being redirected. Finally, this adds specific links to `StlCompareStringA.cpp`, `StlCompareStringW.cpp`, `StlLCMapStringA.cpp`, `StlLCMapStringW.cpp`. Intentionally not changing `SECURITY.md`.
* Add space to comment.
* Style: `typename` => `class` in product code.
* Fuse adjacent Standard modes.
* Simplify away `_THREAD_CHECK`, `_THREAD_CHECKX`.
  + We **never** defined `_THREAD_CHECK` in the internal or GitHub builds, so this code simply collapses to `#ifdef _DEBUG`.
* Drop polluting using-declarations within `stdext`.
  + We don't need to qualify `size_t` at all. Intentionally not attempting to make `<hash_map>` and `<hash_set>` less polluting.
  + This is the only potentially source-breaking change, but it seems unlikely that users would rely on this pollution, and fixing affected code is trivial.
* Mark `_Big_uint128::operator<` as `_NODISCARD`.
* Mark `uncaught_exception()` and `uncaught_exceptions()` as `_NODISCARD`.
  + I've updated the definitions to repeat the attributes, and to drop useless comments (we know from the Standard what these functions are supposed to do). Note that `_EXPORT_STD extern "C++"` is intentionally not repeated as the separately compiled STL is always built clasically.
* Simplify CMakeLists.txt: `MATCHES "\^(\w+)\$"` => `STREQUAL "$1"`
* Simple test cleanup: After `STATIC_ASSERT(CanViewElements<Rng&>)`, assume it's true.
  + We caught most occurrences of this during code review, but missed this one. It wasn't being deliberately weird.
* `<random>`: Decompose complicated conditional expression.
  + This is more lines, but easier to follow. (In modern code, we prefer to avoid mixing side-effects with other expressions when they aren't highly idiomatic, and we usually prefer to avoid conditional expressions when there isn't a strong motivation for them.) Note that the result of the compound assignment is the left operand. The `if`-`return`-`else`-`return` is deliberate.
* Simplify `_Cnd_internal_imp_t` size/align constants.
  + The `stl_condition_variable_max_MEOW` constants in `primitives.hpp` were exactly identical to the `_Cnd_internal_imp_MEOW` constants in `xthreads.h`, because we use `_Aligned_storage` with no other data members. This unifies them, keeping the `xthreads.h` names but the simpler `primitives.hpp` definition technique and clearer comments. Finally, this makes the `static_assert`s in `cond.cpp` tautological, so this removes them.


stl/src/mutex.cpp
```diff
@@ -19,19 +19,13 @@ extern "C" [[noreturn]] _CRTIMP2_PURE void _Thrd_abort(const char* msg) { // abo
...
-#define _THREAD_CHECKX 0
-#endif // defined(_THREAD_CHECK) || defined(_DEBUG)
-
-#if _THREAD_CHECKX
+#ifdef _DEBUG
 #define _THREAD_QUOTX(x)          #x
 #define _THREAD_QUOT(x)           _THREAD_QUOTX(x)
 #define _THREAD_ASSERT(expr, msg) ((expr) ? (void) 0 : _Thrd_abort(__FILE__ "(" _THREAD_QUOT(__LINE__) "): " msg))
-#else // _THREAD_CHECKX
+#else // defined(_DEBUG)
```
AlexGuteniev, line 26: This conflicts with #3900 , maybe easier to postpone this part to avoid conflicts?

stl/src/StlCompareStringA.cpp
```diff
@@ -18,7 +18,8 @@
 //
 // Entry:
 //        LPCWSTR LocaleName  - locale context for the comparison.
-//        DWORD   dwCmpFlags  - see docs.microsoft.com
+//        DWORD   dwCmpFlags  - see:
+//                    https://learn.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-comparestringex
```
AlexGuteniev, line 22: 😕 
Can this be formatted better?

stl/src/mutex.cpp
StephanTLavavej: That's a really simple merge conflict to resolve, but thanks for the heads up. I'll remember to upgrade this one to arrow comments.

stl/src/StlCompareStringA.cpp
StephanTLavavej: This is the best I could do with a 120 column limit, unless you have a better idea. (One occurrence could technically have been unwrapped but I felt it was best to make them all consistently wrapped.)

AlexGuteniev: * "see `CompareStringEx` in Platform SDK documentation"
* an URL shortener, if there's one that can be trusted, like from microsoft.com domain

StephanTLavavej: I do actually have the ability to create aka ms links, I can do that here.

jovibor: `https://learn.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-comparestringex`
is equivalent to:
`learn.microsoft.com/windows/win32/api/stringapiset/nf-stringapiset-comparestringex`

14 chars less.

stl/inc/xcall_once.h
```diff
@@ -28,12 +28,14 @@ _EXPORT_STD struct once_flag { // opaque data structure for call_once()
     void* _Opaque;
 };
 
+#ifdef _CRTBLD
```
CaseyCarter, line 31: Should we file a TODO to move anything `#ifdef _CRTBLD` from public to private headers?

stl/inc/exception
```diff
@@ -21,10 +21,10 @@ _STL_DISABLE_CLANG_WARNINGS
 _STD_BEGIN
 
 #if _HAS_DEPRECATED_UNCAUGHT_EXCEPTION
-_EXPORT_STD extern "C++" _CXX17_DEPRECATE_UNCAUGHT_EXCEPTION _CRTIMP2_PURE bool __CLRCALL_PURE_OR_CDECL
+_EXPORT_STD extern "C++" _CXX17_DEPRECATE_UNCAUGHT_EXCEPTION _NODISCARD _CRTIMP2_PURE bool __CLRCALL_PURE_OR_CDECL
```
CaseyCarter, line 24: Before I noticed your comment, it occurred to me that we should annotate the definitions consistently. I'll let you decide if that adds enough weight to do so in this PR.

stl/inc/xthreads.h
```diff
@@ -59,23 +59,15 @@ struct _Mtx_internal_imp_t {
...
-_INLINE_VAR constexpr size_t _Cnd_internal_imp_size      = 40;
-_INLINE_VAR constexpr size_t _Cnd_internal_imp_alignment = 4;
-#endif // _WIN64
-#endif // _CRT_WINDOWS
+#if defined(_CRT_WINDOWS) // for Windows-internal code
+_INLINE_VAR constexpr size_t _Cnd_internal_imp_size = 2 * sizeof(void*);
+#elif defined(_WIN64) // ordinary 64-bit code
+_INLINE_VAR constexpr size_t _Cnd_internal_imp_size = 72;
+#else // vvv ordinary 32-bit code vvv
+_INLINE_VAR constexpr size_t _Cnd_internal_imp_size = 40;
```
CaseyCarter, line 67: No change requested, but I note these could be unified as `8 * sizeof(void*) + 8`.

CaseyCarter (approved):

stl/inc/xcall_once.h
StephanTLavavej: Filed #3941.

stl/inc/xthreads.h
StephanTLavavej: That's true, but I think I would be weakly opposed to such a refactoring, even though it's shorter. It implies a structure to the storage space that doesn't necessarily exist.

stl/inc/exception
StephanTLavavej: Now I can't find the precedent that I thought I found earlier. I'll apply the attributes (including the deprecations) to the definitions.

StephanTLavavej:
@CaseyCarter I've pushed a commit to address one of your comments.

StephanTLavavej:

StephanTLavavej:
I've pushed a merge with `main` to resolve merge conflicts where #3900 updated preprocessor comments.

+ Resolved in `stl/inc/xthreads.h` by accepting this PR's new definitions of `_Cnd_internal_imp_MEOW` with well-commented preprocessor logic.
+ Resolved in `stl/src/mutex.cpp` by upgrading this PR's preprocessor comments to `defined(_DEBUG)` arrow comments.

CaseyCarter (approved):

PR #3939, opened 2023-08-08
Toolset update: VS 2022 17.8 Preview 1

StephanTLavavej:
This updates the pool for VS 2022 17.8 Preview 1, including August's Patch Tuesday.

We're not yet able to require 17.8 because the internal build is still starting with 17.6 (!). No changes to CMake/Ninja/Clang versions.

Update to PowerShell 7.3.6 when building the pool.

We're able to remove a couple of compiler bug workarounds:

* Remove workaround for VSO-1659408 "`/clr` `alignas` emits error C2711 "this function cannot be compiled as managed" instead of falling back to native codegen"
  + Validated internally.
  + One workaround had to be retained due to an apparently different bug that has now been revealed. I filed VSO-1867037 "`/clr` C++20 System.AccessViolationException with ranges uninitialized algorithms".
  + Two more workarounds had to be retained due to VSO-1659695 "`/clr` x86 runtime assertions/crashes with parallel algorithms".
* Remove workaround for VSO-1849458 "REPORTED: EDG mishandles `is_layout_compatible_v<const int[], int[]>`"

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ Just try to follow these rules, so we can spend more time fixing bugs and implem
 
 # How To Build With The Visual Studio IDE
 
-1. Install Visual Studio 2022 17.7 Preview 3 or later.
+1. Install Visual Studio 2022 17.8 Preview 1 or later.
     * Select "Windows 11 SDK (10.0.22000.0)" in the VS Installer.
     * We recommend selecting "C++ CMake tools for Windows" in the VS Installer.
     This will ensure that you're using supported versions of CMake and Ninja.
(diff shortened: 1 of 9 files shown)


CaseyCarter (approved):

StephanTLavavej:

StephanTLavavej:
@CaseyCarter I had to push an additional commit after you approved, to restore a couple of `/clr` workarounds and update their citations. (I had validated x64 but these are specific to x86.)

CaseyCarter (approved):

PR #3951, opened 2023-08-11
Update `#else` transition comment

AlexGuteniev:
Towards #351.

I'm not sure which is really TRANSITION,
so I want this reviewed separately.

### DIFF
--- a/stl/inc/system_error
+++ b/stl/inc/system_error
@@ -654,7 +654,7 @@ _NODISCARD const _Ty& _Immortalize_memcpy_image() noexcept {
     [[_Clang::__require_constant_initialization__]] static _Ty _Static;
     return _Static;
 }
-#elif !defined(_M_CEE) // TRANSITION, VSO-1153256
+#elif !defined(_M_CEE)
 template <class _Ty>
 struct _Constexpr_immortalize_impl {
     union {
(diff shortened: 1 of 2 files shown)


stl/src/excptptr.cpp
```diff
@@ -67,7 +67,7 @@ namespace {
     [[nodiscard]] _Ty& _Immortalize() noexcept {
         return _Immortalize_impl<_Ty>._Storage;
     }
-#else // choose immortalize strategy
+#else // ^^^ !defined(_M_CEE) / defined(_M_CEE), TRANSITION, VSO-1153256 vvv
```
CaseyCarter, line 70: With the TRANSITION here and on line 89, I'm inclined to remove the comment on 46 so the comments bracket the section of code that will actually go away when VSO-1153256 is fixed.

AlexGuteniev: Good idea!

CaseyCarter (approved):

stl/inc/system_error
```diff
@@ -676,7 +676,7 @@ _NODISCARD const _Ty& _Immortalize_memcpy_image() noexcept {
     static _Constexpr_immortalize_impl<_Ty> _Static;
     return _Static._Storage;
 }
-#else // ^^^ no workaround / workaround vvv
+#else // ^^^ !defined(_M_CEE) / defined(_M_CEE), TRANSITION, VSO-1153256  vvv
```
StephanTLavavej, line 679: No change requested: There are two spaces here but it's not worth resetting testing, I'll fix it up later.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:broom: :scroll: :tada:
===

PR #3960, opened 2023-08-13
SSE2 vectorization for `bitset::to_string`

### LINKED ISSUE #3858: `<bitset>`: Investigate further performance improvements

AlexGuteniev:
Resolves #3858

@Alcaro suggested the original approach with forming mask with `_mm_and_si128(_Vec4, _mm_set1_epi64x(0x0102040810204080))` and populating 2 bytes to low and high 8 bytes of SSE vector via repeated `_mm_unpacklo_epi8`

## Results

Without vectorization:
```
BM_bitset_to_string<15, char>                    281 ns          283 ns      2488889
BM_bitset_to_string<64, char>                   9675 ns         9835 ns        74667
BM_bitset_to_string_large_single<char>          9366 ns         9417 ns        74667
BM_bitset_to_string<7, wchar_t>                  210 ns          209 ns      3446154
BM_bitset_to_string<64, wchar_t>                5583 ns         5625 ns       100000
BM_bitset_to_string_large_single<wchar_t>       2689 ns         2679 ns       280000
```

With vectorization:
```
BM_bitset_to_string<15, char>                    378 ns          374 ns      2133333
BM_bitset_to_string<64, char>                   3593 ns         3641 ns       154483
BM_bitset_to_string_large_single<char>           339 ns          345 ns      2036364
BM_bitset_to_string<7, wchar_t>                  192 ns          192 ns      4072727
BM_bitset_to_string<64, wchar_t>                3845 ns         3767 ns       165926
BM_bitset_to_string_large_single<wchar_t>        613 ns          614 ns      1120000
```
(`15, char` and `7, wchar_t` results are variation between runs, but there's still a strong indication that others are better with vectorization)


stl/src/vector_algorithms.cpp
```diff
@@ -1496,5 +1497,121 @@ __declspec(noalias) size_t
...
+        const uint16_t _Valsw = _byteswap_ushort(_Val);
+        const __m128i _Vec1   = _mm_cvtsi32_si128(_Valsw);
+        const __m128i _Vec2   = _mm_unpacklo_epi8(_Vec1, _Vec1);
+        const __m128i _Vec3   = _mm_unpacklo_epi8(_Vec2, _Vec2);
+        const __m128i _Vec4   = _mm_unpacklo_epi8(_Vec3, _Vec3);
+        const __m128i _Msk    = _mm_and_si128(_Vec4, _mm_set1_epi64x(0x0102040810204080));
+        const __m128i _Zeros  = _mm_cmpeq_epi8(_Msk, _mm_setzero_si128());
+        const __m128i _Elem0s = _mm_and_si128(_Zeros, _Elem0_v);
+        const __m128i _Elem1s = _mm_andnot_si128(_Zeros, _Elem1_v);
+        const __m128i _Elems  = _mm_or_si128(_Elem0s, _Elem1s);
```
AlexGuteniev, line 1517: This can be improved with:
 * `_mm_shuffle_epi8` input value populating (replaces byteswap and upacks)
 * `_mm_blendv_epi8` output blending (replaces and/andnot/or)
We don't have discrete SSSE3 / SSE4.1 detection, both can be detected as SSE4.2

I think this can be done in a subsequent PR, so that we step gradually, as usually, but can also be done in this.

stl/src/vector_algorithms.cpp
```diff
@@ -1496,5 +1497,121 @@ __declspec(noalias) size_t
...
+    __m128i
+#ifdef _M_X64
+        __vectorcall
+#endif
+        _Bitset_to_string_1_step(const uint16_t _Val, const __m128i _Elem0_v, const __m128i _Elem1_v) {
+        const uint16_t _Valsw = _byteswap_ushort(_Val);
+        const __m128i _Vec1   = _mm_cvtsi32_si128(_Valsw);
+        const __m128i _Vec2   = _mm_unpacklo_epi8(_Vec1, _Vec1);
+        const __m128i _Vec3   = _mm_unpacklo_epi8(_Vec2, _Vec2);
+        const __m128i _Vec4   = _mm_unpacklo_epi8(_Vec3, _Vec3);
```
Alcaro, line 1512: If you swap the last unpacklo for shuffle_epi32, you can then save one instruction by merging the byteswap into that shuffle.

stl/src/vector_algorithms.cpp
Alcaro, line 1517: If you pass different arguments, you can swap these three for and(val1^val0) + xor(val1). (Or andnot(val1^val0) + xor(val0).)

Would make it harder to upgrade to blendv, though. Your call whether it's worth it.

stl/src/vector_algorithms.cpp
```diff
@@ -1496,5 +1497,114 @@ __declspec(noalias) size_t
...
+    if (_Use_sse2()) {
+        const __m128i _Xor_v = _mm_set1_epi8(_Elem0 ^ _Elem1);
+        const __m128i _And_v = _mm_set1_epi8(_Elem1);
+        if (_Size_bits > 16) {
+            char* _Pos = _Dest + _Size_bits;
+            _Size_bits &= 0xF;
+            char* const _Stop_at = _Dest + _Size_bits;
+            do {
+                uint16_t _Val;
+                memcpy(&_Val, _Src, 2);
```
AlexGuteniev, line 1539: this `memcpy` is to avoid aliasing problems. But other places of this file do `reinterpret_cast` instead with a comment `NOTE: UNALIGNED MEMORY ACCESSES` and not even mentioning incorrect aliasing. Should this be done here too?

stl/src/vector_algorithms.cpp
```diff
@@ -1496,5 +1497,114 @@ __declspec(noalias) size_t
...
+
+extern "C" {
+
+__declspec(noalias) void __stdcall __std_bitset_to_string_1(
+    char* const _Dest, const void* _Src, size_t _Size_bits, const char _Elem0, const char _Elem1) noexcept {
+#ifndef _M_ARM64EC
+    if (_Use_sse2()) {
+        const __m128i _Xor_v = _mm_set1_epi8(_Elem0 ^ _Elem1);
+        const __m128i _And_v = _mm_set1_epi8(_Elem1);
+        if (_Size_bits > 16) {
```
Alcaro, line 1533: why is that > and not >= ? You'll get UB from the __assume below if size_bits is exactly 16.

AlexGuteniev: a typo, the other version had `>=`, thanks

stl/src/vector_algorithms.cpp
```diff
@@ -1496,5 +1497,114 @@ __declspec(noalias) size_t
...
+            do {
+                uint16_t _Val;
+                memcpy(&_Val, _Src, 2);
+                const __m128i _Elems = _Bitset_to_string_1_step(_Val, _Px0, _Px1);
+                _Pos -= 16;
+                _mm_storeu_si128(reinterpret_cast<__m128i*>(_Pos), _Elems);
+                _Advance_bytes(_Src, 2);
+            } while (_Pos != _Stop_at);
+        }
+        if (_Size_bits > 0) {
```
StephanTLavavej, line 1546: Style: We conventionally put a newline between non-chained `if` statements. Occurs below.

stl/src/vector_algorithms.cpp
```diff
@@ -1496,5 +1497,114 @@ __declspec(noalias) size_t
...
+            }
+            const __m128i _Elems = _Bitset_to_string_1_step(_Val, _Px0, _Px1);
+            char _Tmp[16];
+            _mm_storeu_si128(reinterpret_cast<__m128i*>(_Tmp), _Elems);
+            const char* const _Tmpd = _Tmp + (16 - _Size_bits);
+            for (size_t _Ix = 0; _Ix < _Size_bits; ++_Ix) {
+                _Dest[_Ix] = _Tmpd[_Ix];
+            }
+        }
+        return;
```
StephanTLavavej, line 1562: Style: `return;` at the end of a `void` function is unnecessary, so we always omit this.

stl/src/vector_algorithms.cpp
StephanTLavavej: A followup PR would be fine. (It looks like you altered the code after making this comment, but aren't yet using instructions beyond SSE2, so potential improvements are still possible.)

stl/src/vector_algorithms.cpp
StephanTLavavej: I'm ok with the strictly correct `memcpy` although I suppose there's a consistency argument.

StephanTLavavej:
Thanks! :heart_eyes_cat: (And sorry for taking so long to review this. :snail:)

I pushed a merge with `main`, fixed a stealth merge conflict, and pushed some minor improvements, nothing affecting your core vectorized algorithms. I enhanced the test coverage by testing non-default characters `'o', 'x'` and adding randomized testing of a `bitset<2048>`.

StephanTLavavej:

StephanTLavavej:
Thanks for optimizing this function! :rocket: :rocket: :tada:

PR #3964, opened 2023-08-14
Remove some `#ifdef _CRTBLD` blocks

achabense:
Works towards #3941.

- Remove all `#ifdef _CRTBLD` blocks from `<xthreads.h>`.
- Remove some `#ifdef _CRTBLD` blocks from `<yvals.h>`.

### DIFF
--- a/stl/src/_tolower.cpp
+++ b/stl/src/_tolower.cpp
@@ -17,6 +17,8 @@
 #undef _tolower
 #undef tolower
 
+#define _cpp_isleadbyte(c) (__pctype_func()[static_cast<unsigned char>(c)] & _LEADBYTE)
+
 _EXTERN_C_UNLESS_PURE
 
 // int _Tolower(c) - convert character to lower case
--- a/stl/src/_tolower.cpp
+++ b/stl/src/_tolower.cpp
@@ -17,8 +17,6 @@
 #undef _tolower
 #undef tolower
 
-#define _cpp_isleadbyte(c) (__pctype_func()[static_cast<unsigned char>(c)] & _LEADBYTE)
-
 _EXTERN_C_UNLESS_PURE
 
 // int _Tolower(c) - convert character to lower case
--- a/stl/src/cthread.cpp
+++ b/stl/src/cthread.cpp
@@ -109,6 +109,12 @@ _CRTIMP2_PURE unsigned int __cdecl _Thrd_hardware_concurrency() { // return numb
     return info.dwNumberOfProcessors;
 }
 
+// TRANSITION, ABI: these functions are used only by _Thrd_create()
+_CRTIMP2_PURE _Thrd_result __cdecl _Mtx_init(_Mtx_t*, int);
+_CRTIMP2_PURE void __cdecl _Mtx_destroy(_Mtx_t);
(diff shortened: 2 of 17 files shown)


stl/src/cthread.cpp
```diff
@@ -114,10 +114,14 @@ _CRTIMP2_PURE _Thrd_result __cdecl _Thrd_create(_Thrd_t* thr, _Thrd_start_t func
     _Thrd_result res;
     _Thrd_binder b;
     int started = 0;
-    _Cnd_t cond;
-    _Mtx_t mtx;
-    _Cnd_init(&cond);
-    _Mtx_init(&mtx, _Mtx_plain);
+
+    std::_Aligned_storage_t<_Cnd_internal_imp_size, _Cnd_internal_imp_alignment> cond_storage;
+    _Mtx_internal_imp_t mtx_storage{};
```
achabense, line 119: Is it ok to make cond and mtx adjacent here?

AlexGuteniev: Do you mean false sharing?
It is currently avoided by oversize of `_Cnd_internal_imp_size`, at least for 64 bits.
In general, I think, you can avoid it by `alignas(hardware_constructive_interference_size)`, but:
 * Not sure if it ever worth for `mutex` and `condition_variable` used in conjunction as they are used by the same waiting thread, and then the same signaling thread.
 * Not sure if it worth for such heavy operation as thread start.

achabense: Yes. I have only vague awareness of this type of problem.
(As to the align cost if there are, I think it will not be significant, as I've already replaced `calloc` versions with in-place versions.)

achabense:
I'm temporarily moving this pr to WIP state, as I'm hestitant whether this decouplement is worth bothering with implementation details of legacy code...

stl/src/cthread.cpp
achabense: I've force-pushed this pr to avoid such changes.

achabense:
I've reworked and force-pushed this pr to make it ready for review.

achabense:
I have another commit 06121fd35845e82ece46fd7d1b5a00e242af5724; the change is simple there; I will make a push if this looks ok.

strega-nil-ms (approved):
Thanks!

stl/src/_tolower.cpp
```diff
@@ -17,6 +17,8 @@
 #undef _tolower
 #undef tolower
 
+#define _cpp_isleadbyte(c) (__pctype_func()[static_cast<unsigned char>(c)] & _LEADBYTE)
```
CaseyCarter, line 20: I'm not too happy about duplicating this definition in two source files. Could we move it into an internal header like `awint.hpp` or something instead?

CaseyCarter (changes requested):

CaseyCarter (approved):

StephanTLavavej (approved):

StephanTLavavej:
Thanks, these are great cleanups! I'm not sure why we aren't using `isleadbyte` directly (there may be some semantic difference I'm missing), but this change is a strict, low-risk improvement.

StephanTLavavej:

StephanTLavavej:
:tada: :broom: :smile_cat:
===

PR #3965, opened 2023-08-14
Bugfixes for `<any>`

achabense:
This pr contains two fixes for `std::any`:
- Fixes `any`'s self-move-assignment (which will result in the object being destroyed).
- Works towards #140. Now `any` can hold `holder<incomplete>*` etc.


tests/std/tests/Hmm_I_dont_know_how_to_name_this_folder/test.cpp
```diff
@@ -0,0 +1,37 @@
...
+};
+
+int main() {
+    using foo = wrapper<incomplete>*;
+    struct bar {
+        foo p{};
+        nontrivial q{};
+    };
+    struct baz {
+        foo p[10]{};
```
achabense, line 26: foo, bar and baz represent trivial/small/large modes respectively.

tests/std/tests/Hmm_I_dont_know_how_to_name_this_folder/test.cpp
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
```
achabense, line 1: How to name the folder for this test 😳?

tests/std/tests/Hmm_I_dont_know_how_to_name_this_folder/test.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+int main() {
+    using foo = wrapper<incomplete>*;
+    struct bar {
+        foo p{};
+        nontrivial q{};
+    };
+    struct baz {
+        foo p[20]{};
+    };
```
achabense, line 27: This is for any's support for pointers to templates holding incomplete types; `foo, bar and baz` represent trivial/small/large modes respectively.

How to name the folder for this test 😳?

AlexGuteniev:
```
GH_003965_any_incomplete_types
```

AlexGuteniev: Merging is also fine to me.
Merging serves a good purpose of saving tests run time.

achabense: Is it desirable to group ADL-proof tests into tests like `GH_000140_ADL_proof_.../`?

tests/std/tests/GH_003965_robust_against_adl_any/env.lst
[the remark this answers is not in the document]
achabense: hmm, I'm too careless...

achabense:
* sign *

tests/std/tests/GH_003965_robust_against_adl_any/test.cpp
```diff
@@ -0,0 +1,38 @@
...
+        foo p{};
+        nontrivial q{};
+    };
+    struct baz {
+        foo p[20]{};
+    };
+
+    std::any a{foo{}}, b(bar{}), c{baz{}};
+    std::any *ap = &a, *bp = &b, *cp = &c;
+    a = *ap, b = *bp, c = *cp;
```
AlexGuteniev, line 31: if the pointers are to fool the compiler to suppress some waring, I'd consider actual suppression instead to make it more obvious what you do here.

achabense: I've suppressed the warnings in this file. As to the one in `P0220R1_any`, as there is only one such usage I think it's ok to keep it unchanged.

tests/std/tests/P0220R1_any/test.cpp
```diff
@@ -3160,6 +3160,28 @@ namespace msvc {
...
+            foo& operator=(const foo&) = default;
+            foo(foo&&)                 = default;
+            foo& operator=(foo&&)      = default;
+            ~foo() {
+                val = 0;
+            }
+        };
+        void run_test() {
+            any a(foo{});
+            any* ap = &a;
```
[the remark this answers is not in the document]
achabense: I think it's ok to keep it unchanged here as there is only one usage, actually there are a lot of things like `a = (std::any&)a;` to do the same thing in this file.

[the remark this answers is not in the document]
achabense: What about `a = (std::any&&) a; // move`?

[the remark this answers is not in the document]
achabense: https://github.com/microsoft/STL/blob/8674b3d732a2bae43105fb7654d3b8c1c57adb56/tests/std/tests/P0220R1_any/test.cpp#L2564-L2573
Oh, there are actually usage of the suppression, and actually this check fails to catch the error. I think I should discard my check and instead fix this check directly.

tests/std/tests/GH_003965_robust_against_adl_any/test.cpp
```diff
@@ -0,0 +1,39 @@
...
+        foo p[20]{};
+    };
+
+    std::any a{foo{}}, b(bar{}), c{baz{}};
+
+    a = foo{}, b = bar{}, c = baz{};
+
+    a.emplace<foo>();
+    b.emplace<bar>();
+    c.emplace<baz>();
```
achabense, line 34: Should I also add tests for `make_any` / `in_place_type` construction?

tests/std/tests/GH_003965_robust_against_adl_any/test.cpp
frederick-vs-ja: No change requested: there can be a concentrated test for self move (and/or copy) assignment in the future, and we may close #1485 at that time.

stl/inc/any
```diff
@@ -168,8 +168,10 @@ public:
     }
 
     any& operator=(any&& _That) noexcept {
-        reset();
-        _Move_from(_That);
+        if (this != &_That) {
+            reset();
+            _Move_from(_That);
+        }
```
achabense, line 174: I think this "fix" doesn't essentially resolve the problem. For example:
```
    any a = std::make_any<any>(1); // a has an `any(1)`
    a = any_cast<any&&>(a);// ugh...
```
Though risking higher performance impact, I think we have to move-construct a local any then `_Move_from` it to do the real fix...

stl/inc/any
```diff
@@ -163,15 +163,12 @@ public:
 
     // Assignment [any.assign]
     any& operator=(const any& _That) {
-        *this = any{_That};
+        _Reset_and_move_from(_That);
```
achabense, line 166: This change will hopefully not affect performance of `const&`

stl/inc/any
```diff
@@ -178,7 +177,7 @@ public:
                                     int> = 0>
     any& operator=(_ValueType&& _Value) {
         // replace contained value with an object of type decay_t<_ValueType> initialized from _Value
-        *this = any{_STD forward<_ValueType>(_Value)};
+        _Assign(any{_STD forward<_ValueType>(_Value)});
```
achabense, line 180: This change will hopefully not affect performance of `operator=(const&)` and `operator=(auto&&)`.
For `operator=(any&&)`, it seems for correctness we have to make a local variable as well.

tests/std/tests/Hmm_I_dont_know_how_to_name_this_folder/test.cpp
achabense: merged to `P0220R1_any`

tests/std/tests/GH_003965_robust_against_adl_any/test.cpp
achabense: simplified to construction only.

tests/std/tests/P0220R1_any/test.cpp
```diff
@@ -3160,6 +3184,27 @@ namespace msvc {
...
+
+        void run_test() {
+            using _trivial = wrapper<incomplete>*;
+            using _small   = std::pair<_trivial, small>;
+            using _large   = std::pair<_trivial, large>;
+
+            globalMemCounter.disable_allocations = true;
+            any a{_trivial{}};
+            any b{_small{}};
+            globalMemCounter.disable_allocations = false;
```
achabense, line 3204: This is a test-compile test; I'm adding `globalMemCounter.disable_allocations = true / false;` to make it able to pass the runtime check.

tests/std/tests/P0220R1_any/test.cpp
```diff
@@ -3220,4 +3241,5 @@ int main() {
     msvc::size_and_alignment::run_test();
     msvc::small_type::run_test();
     msvc::trivial::run_test();
+    msvc::gh_140_robust_against_adl::run_test();
```
achabense, line 3244: This is a test-compile test; in `run_test` I'm adding `globalMemCounter.disable_allocations = true / false;` to make it pass the runtime check. Is it ok to run this function here as with other run-time tests?

achabense:
https://github.com/microsoft/STL/blob/60f18856c56389001c38a9929ef37bd5c01c4e47/tests/std/tests/P0220R1_any/test.cpp#L2051
I find some tests in `P0220R1_any` are actually ill-formed ~ an `any&` is not allowed to `any_cast` to an rvalue.

Though I think this *should* be well-formed; `any_cast` is performing cast, but the standard is requiring constructible instead of castable, I think this requirement is too restrictive.

frederick-vs-ja:
> I find some tests in `P0220R1_any` are actually ill-formed ~ an `any&` is not allowed to `any_cast` to an rvalue reference.

Whether this case is ill-formed (currently) is unspecified - it's unspecified whether any violation of _Mandates_: is substitution failure when it can be. MSVC STL tends to make the ill-formedness SFINAE-unfriendly, which allows emitting informative message in `static_assert`.

StephanTLavavej (approved):

CaseyCarter (approved):

StephanTLavavej:

StephanTLavavej (approved):

StephanTLavavej:
I had to push an additional commit to work around VSO-1659496 "`/clr` emits bogus error C2079 "uses undefined struct 'Incomplete'" despite `std::variant`'s ADL defenses". Someday I'll remember this before preparing a merge.

CaseyCarter (approved):

StephanTLavavej:
:star: :lady_beetle: :hammer_and_wrench:
===

PR #3967, opened 2023-08-14
Fix bad tr1/algorithm test cases

CaseyCarter:
These two `is_permutation(first, last, x)` calls violate the precondition that `[x, x + (last - first))` is a valid range.

Fixes VSO-1854237 / AB#1854237


CaseyCarter:
Who knew this Address Sanitizer thing could actually be useful!

StephanTLavavej:
Dropping the `bug` label, which we reserve for bugs in product code only.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:lady_beetle: :hammer_and_wrench: :heart_eyes_cat:
===

PR #3968, opened 2023-08-15
Avoid calling `size()` in valarray loops

### LINKED ISSUE #3915: `<valarray>`: Consider replacing `size()` calls with `_Mysize`
Followup issue after discussion in #3911, to improve debug codegen for a minor editing cost.

As @AlexGuteniev mentioned in https://github.com/microsoft/STL/pull/3911#issuecomment-1656875296 , `<valarray>` is currently extremely careful to extract the size to a local variable to avoid aliasing, which is important for vectorizing its following loops. When replacing calls of `size()` with direct access to `_Mysize`, the local variables should be preserved. We may even want to add comments like `// avoid aliasing` as a small reminder about this somewhat unusual pattern.

AlexGuteniev:
Resolves #3915

I've chosen "eliminating indirection helps vectorization" wording that does not mention aliasing.

The correctly looking formula would be "avoid suspected false aliasing" and it is confusing. The wording like "inhibit aliasing" is not quite correct, because the actual aliasing is not happening anyway.

I'm sure that wording mentioning "indirection" and "vectorization" strongly suggests what happening for those who familiar with problem, and for those who are not, the margin of this assignment statement is too small to fit an explanation.

### DIFF
--- a/stl/inc/valarray
+++ b/stl/inc/valarray
@@ -166,7 +166,7 @@ public:
     }
 
     valarray& operator=(const _Ty& _Val) {
-        const size_t _Size = size();
+        const size_t _Size = _Mysize; // eliminating indirection helps vectorization
         for (size_t _Idx = 0; _Idx < _Size; ++_Idx) {
             _Myptr[_Idx] = _Val;
         }
@@ -192,7 +192,7 @@ public:
     valarray& operator=(const indirect_array<_Ty>& _Indarr); // defined below
 
     _NODISCARD valarray operator+() const {
-        const size_t _Size = size();
+        const size_t _Size = _Mysize; // eliminating indirection helps vectorization
         valarray<_Ty> _Ans(_Size);
         for (size_t _Idx = 0; _Idx < _Size; ++_Idx) {
             _Ans[_Idx] = +_Myptr[_Idx];
@@ -201,7 +201,7 @@ public:
     }
 
     _NODISCARD valarray operator-() const {
-        const size_t _Size = size();
+        const size_t _Size = _Mysize; // eliminating indirection helps vectorization
         valarray<_Ty> _Ans(_Size);
         for (size_t _Idx = 0; _Idx < _Size; ++_Idx) {
             _Ans[_Idx] = -_Myptr[_Idx];
@@ -210,7 +210,7 @@ public:
     }
 
     _NODISCARD valarray operator~() const {
-        const size_t _Size = size();
+        const size_t _Size = _Mysize; // eliminating indirection helps vectorization
         valarray<_Ty> _Ans(_Size);
         for (size_t _Idx = 0; _Idx < _Size; ++_Idx) {
             _Ans[_Idx] = ~_Myptr[_Idx];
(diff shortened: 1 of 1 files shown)


stl/inc/valarray
```diff
@@ -555,10 +555,12 @@ private:
 
     void _Tidy_deallocate() noexcept {
         if (_Myptr) { // destroy elements
-            for (size_t _Idx = 0; _Idx < _Mysize; ++_Idx) {
-                _Destroy_in_place(_Myptr[_Idx]);
+            if constexpr (!is_trivially_destructible<_Ty>::value) {
```
StephanTLavavej, line 558: We conventionally use variable templates for type traits.

StephanTLavavej (approved):

AlexGuteniev:
I've also created DevCom-10442357 to optimize even without pre-indirection after more checks, it has some votes, but I'm skeptical, as other compilers miss the vectorization without pre-indirected size too,

stl/inc/valarray
```diff
@@ -166,7 +166,7 @@ public:
     }
 
     valarray& operator=(const _Ty& _Val) {
-        const size_t _Size = size();
+        const size_t _Size = _Mysize; // eliminating indirection helps vectorization
```
CaseyCarter, line 169: I think these comments are all unnecessary, but it's not worth resetting testing to remove them. It's also not a hill I'm willing to die on if you think we're likely to break this in the future.

stl/inc/valarray
```diff
@@ -555,10 +555,12 @@ private:
...
     void _Tidy_deallocate() noexcept {
         if (_Myptr) { // destroy elements
-            for (size_t _Idx = 0; _Idx < _Mysize; ++_Idx) {
-                _Destroy_in_place(_Myptr[_Idx]);
+            if constexpr (!is_trivially_destructible_v<_Ty>) {
+                const size_t _Size = _Mysize; // eliminating indirection helps vectorization
+                for (size_t _Idx = 0; _Idx < _Size; ++_Idx) {
+                    _Destroy_in_place(_Myptr[_Idx]);
+                }
             }
```
CaseyCarter, line 563: Could we simply:
```suggestion
            _Destroy_range(_Myptr, _Myptr + _Mysize);
```
here? It already optimizes for trivially destructible types.

stl/inc/valarray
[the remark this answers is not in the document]
CaseyCarter: Ha! This was resolved before I submitted it.

stl/inc/valarray
StephanTLavavej: This code pattern is weird (in `vector` we wouldn't do this) so I would prefer to keep the comments, even if it seems like they aren't super valuable.

stl/inc/valarray
StephanTLavavej: I believe this is a strict improvement, yes. Good catch - `valarray` predates `_Destroy_range` and we never noticed we could call it here.

stl/inc/valarray
CaseyCarter: Fair enough.

CaseyCarter (approved):

CaseyCarter:

stl/inc/valarray
AlexGuteniev: I think this is tricky enough issue that definitely deserves a comment.

I've created DevCom-10442357 and SO question https://stackoverflow.com/q/76920847/2945027 while struggling to understand why indirect size completely inhibits vectorization, whereas other indirections don't.

(I'm also against `TRANSITION, DevCom-10442357` comment here too, as since each of the other compilers demonstrate the issue, it could be something fundamental, without a feasible compiler-level fix.)

CaseyCarter:
Thanks for convincing the compiler that `size()` doesn't change in these loops.

PR #3944, opened 2023-08-10
mdspan STL's review comments - easy part

AlexGuteniev:
* Remove superfluous `std::`, the tests do `using namespace std;`
 * Cast `bool` to `size_t` explicitly
 * Check the correct constructor

PR #3852, opened 2023-07-07
`<mdspan>`: Final cleanups

JMazurkiewicz:
* Privatize implementation-defined constructors,
* Strengthen various functions,
* Address comments from previous PRs:
  * https://github.com/microsoft/STL/pull/3829,
  * https://github.com/microsoft/STL/pull/3847,
* Add extra `_STL_INTERNAL_STATIC_ASSERT` in various places,
* Simplify `_Make_dynamic_indices_inv` and fix size of `_Dynamic_indices_inv` array,
* Make `check_members_with_various_extents` test less heavy.

After this PR `<mdspan>` will be ready for final review 😸

PR #3972, opened 2023-08-16
P0009R18 `<mdspan>`

### LINKED ISSUE #2909: P0009R18 `<mdspan>`
WG21-P0009R18 `<mdspan>`
WG21-P2599R2 `mdspan`: `index_type`, `size_type`
WG21-P2604R0 `mdspan`: `data_handle_type`, `data_handle()`, `exhaustive`
WG21-P2613R1 `mdspan`: `empty()`
WG21-P2763R1 Fixing `layout_stride`'s Default Constructor For Fully Static Extents
LWG-3876 Default constructor of `std::layout_`*`meow`*`::mapping` misses precondition

```cpp
#define __cpp_lib_mdspan 202207L
```

Internal VS-PR-420058 has taught the IDE about this new extensionless header - thanks @CaseyCarter!

StephanTLavavej:
Fixes #2909.

Thanks to @MattStephanson for starting this major feature branch with #3502, and to @JMazurkiewicz for finishing it with an incredible amount of work (#3534 #3535 #3560 #3564 #3580 #3586 #3593 #3603 #3615 #3616 #3632 #3672 #3673 #3674 #3682 #3688 #3689 #3690 #3693 #3694 #3698 #3701 #3702 #3715 #3749 #3767 #3786 #3821 #3825 #3829 #3847 #3852 wow!!!). Also thanks to @AlexGuteniev (#3944) and @frederick-vs-ja (#3638 #3901) for additional contributions. The feature branch also contains my own code review feedback (#3563 #3885 #3954 #3971), and merges from `main` by @CaseyCarter and myself to keep the branch active.

This will merge the feature branch `feature/mdspan2` into `main`. It's a separate branch on my own fork so we can push additional last-minute commits if necessary, without going through multiple PRs. (Also, GitHub gets confused when directly merging between branches in the same fork.) Currently, this contains 3 additional commits:

* Update the `#endif` comment next to the feature-test macro in `<yvals_core.h>` for consistency with recent cleanups.
* Merge with `main` to resolve a trivial merge conflict in `<yvals_core.h>` around `__cpp_lib_mdspan`, as the formerly-adjacent `__cpp_lib_modules` has moved to the C++20 section.
* Fix the stealth merge conflict that I foretold in #3977.

Followed by a bunch of commits from @CaseyCarter's code review.

### ⚠️ Reminder: this PR needs additional changes when merged to the internal MSVC repository ⚠️
Per https://github.com/microsoft/STL/wiki/Files-To-Edit-When-Adding-Or-Removing-Files.


stl/inc/mdspan
```diff
@@ -0,0 +1,1464 @@
...
+        _Result[_Rank] = _Counter;
+        return _Result;
+    }
+
+    static constexpr array<rank_type, _Rank + 1> _Dynamic_indices = _Make_dynamic_indices();
+
+    _NODISCARD static consteval auto _Make_dynamic_indices_inv() noexcept {
+        array<rank_type, _Rank_dynamic> _Result{};
+        rank_type _Counter = 0;
+        for (rank_type _Idx = 0; _Idx < _Rank; ++_Idx) {
```
jovibor, line 79: `_Counter` might be declared in the `for (rank_type _Idx = 0, _Counter = 0; _Idx < _Rank; ++_Idx)`, for locality.

StephanTLavavej: This is a style judgment - it's true that that would reduce `_Counter`'s scope, and it's generally preferable to reduce variable scope when possible. However, this function simply ends immediately afterwards, so there is no real clarity gain there. I believe that there's a clarity loss by defining `_Counter` within the `for`-loop itself. Right now, the `_Idx` by itself is very clearly a classic iteration, nothing interesting to see there. As soon as a `for`-loop defines multiple variables (especially more than an `_End`), it's doing something fancier that deserves extra scrutiny, which is not necessary here. `_Counter` doesn't participate in the `for`-loop's bounds, it's effectively iterating (sometimes) through the result array, so mixing that into the `for`-loop is more confusing.

Thanks for noting this, but I recommend against a change here.

jovibor: It has another property - reducing LOC. But I don't know whether this is generally the goal of this repo.

JMazurkiewicz (approved):

stl/inc/mdspan
```diff
@@ -0,0 +1,1464 @@
...
+    }
+
+    struct _Construct_from_tuple {
+        explicit _Construct_from_tuple() = default;
+    };
+
+    template <class _ExtsTuple, size_t... _Indices>
+        requires (tuple_size_v<_ExtsTuple> == _Rank_dynamic)
+    constexpr explicit extents(_Construct_from_tuple, _ExtsTuple _Tpl, index_sequence<_Indices...>) noexcept
+        : _Base{static_cast<index_type>(_STD move(_STD get<_Indices>(_Tpl)))...} {}
```
achabense, line 123: I find this is specified by the standard; just curious, why does the standard say the elements should get `move`d before casting to an integer type?

stl/inc/mdspan
StephanTLavavej: All else being equal, fewer lines of code are better. However, making lines more complicated horizontally in order to save space vertically is often counterproductive. For example, we strongly try to avoid expressions with too many side effects since they're complicated to reason about. Having an extra variable definition in a `for`-loop isn't the most complicated thing ever, but overall I think it's just moving complexity around instead of eliminating it, and moving it to a slightly more surprising place which is undesirable.

stl/inc/mdspan
StephanTLavavej: My understanding is that they're allowed to be class types that are convertible to integer types; the test suite has a few "weird indices" with such conversions. (And indeed they caught an over-simplification that I had briefly committed while working on my feedback changes, where I thought I could take a shortcut that was actually disallowed.)

crtrott: Yeah, generally constructors and other access operators in mdspan and friends are meant to allow to take things convertible to index_type (which itself however must be an integral). The primary intent is to allow for strongly typed indices. For example a bunch of our codes distinguish between things like edge and vertice indices. For most of those the move wouldn't really matter, but at that point LWG said: ok lets do this the right away even if the "indices" become even weirder.

crtrott:
Hi, would it help to move this along if I organized a code review with some experienced reviewers from our side (i.e. co-authors of mdspan who are implementers of the original reference implementation and the one we are pushing into llvm)? I would be really excited to see this hitting the main branch :-)

CaseyCarter:
> Hi, would it help to move this along if I organized a code review with some experienced reviewers from our side?

I don't know if it would speed things up - we are critically short of review resources for a couple of weeks and we need at least one set of experienced STL maintainer eyeballs - but I imagine it would _enormously_ benefit the quality of implementation to have domain expert reviewers.

StephanTLavavej:
I've pushed a merge with `main` to resolve a trivial merge conflict in `<yvals_core.h>` around `__cpp_lib_mdspan`, as the formerly-adjacent `__cpp_lib_modules` has moved to the C++20 section.

StephanTLavavej:
And I've pushed an additional commit to fix the stealth merge conflict that I foretold in #3977 and then forgot about over the last two weeks.

CaseyCarter (changes requested):
I've prepared fixes for these comments except for "`_Fwd_prod_of_extents::_Calculate` can potentially overflow when `mdspan::size`'s precondition is satisfied", which I think deserves a bit of deep thinking and a followup issue. I'm going to PR a series of nicely fine-grained commits into STL's fork for ease of review.

stl/inc/mdspan
```diff
@@ -0,0 +1,1464 @@
...
+    { _Mp::is_always_exhaustive() } -> same_as<bool>;
+    { _Mp::is_always_unique() } -> same_as<bool>;
+    bool_constant<_Mp::is_always_strided()>::value;
+    bool_constant<_Mp::is_always_exhaustive()>::value;
+    bool_constant<_Mp::is_always_unique()>::value;
+};
+
+template <class _Extents>
+class layout_stride::mapping : private _Maybe_fully_static_extents<_Extents>,
+                               private _Maybe_empty_array<typename _Extents::index_type, _Extents::rank()> {
```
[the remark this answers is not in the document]
StephanTLavavej: This was reverted.

stl/inc/mdspan
```diff
@@ -0,0 +1,1464 @@
...
+        _STL_INTERNAL_STATIC_ASSERT((same_as<_OtherIndexTypes, index_type> && ...));
+#if _CONTAINER_DEBUG_LEVEL > 0
+        _STL_VERIFY(this->_Map.extents()._Contains_multidimensional_index(make_index_sequence<rank()>{}, _Indices...),
+            "I must be a multidimensional index in extents() (N4950 [mdspan.mdspan.members]/3).");
+#endif // _CONTAINER_DEBUG_LEVEL > 0
+
+        return this->_Acc.access(_Ptr, static_cast<size_t>(this->_Map(_Indices...)));
+    }
+
+    /* [[no_unique_address]] */ data_handle_type _Ptr{};
```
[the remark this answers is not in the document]
StephanTLavavej: I didn't see a commit for this comment.

stl/inc/mdspan
```diff
@@ -0,0 +1,1464 @@
...
+class extents : private _Maybe_empty_array<_IndexType, _Calculate_rank_dynamic<_Extents...>> {
+public:
+    using index_type = _IndexType;
+    using size_type  = make_unsigned_t<index_type>;
+    using rank_type  = size_t;
+
+    static_assert(_Is_standard_integer<index_type>,
+        "IndexType must be a signed or unsigned integer type (N4950 [mdspan.extents.overview]/1.1).");
+    static_assert(((_Extents == dynamic_extent || _STD in_range<index_type>(_Extents)) && ...),
+        "Each element of Extents must be either equal to dynamic_extent, or must be representable as a value of type "
```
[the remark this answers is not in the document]
StephanTLavavej: The commit here introduced a typo, I'll fix it.

stl/inc/mdspan
```diff
@@ -0,0 +1,1464 @@
...
+                "Value of exts_arr[r] must be equal to extent(r) for each r for which extent(r) is a static extent "
+                "(N4950 [mdspan.extents.cons]/7.1)");
+        }(make_index_sequence<rank()>{});
+#endif // _CONTAINER_DEBUG_LEVEL > 0
+    }
+
+    template <class _OtherIndexType, size_t _Size, size_t... _Indices>
+        requires is_convertible_v<const _OtherIndexType&, index_type>
+              && is_nothrow_constructible_v<index_type, const _OtherIndexType&> && (_Size == _Rank_dynamic)
+    constexpr explicit extents(span<_OtherIndexType, _Size> _Dynamic_exts, index_sequence<_Indices...>) noexcept
```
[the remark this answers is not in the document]
StephanTLavavej: Yep, this is good. You had a followup commit that I'll squash, "Don't cast `size_t{_Rank_dynamic}`, let the conversion be implicit" - note that there is no conversion at all, it's `rank_type _Rank_dynamic` with `using rank_type  = size_t;`.

stl/inc/mdspan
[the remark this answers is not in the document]
StephanTLavavej: I've made this change.

stl/inc/mdspan
```diff
@@ -0,0 +1,1464 @@
...
+    _NODISCARD constexpr size_type size() const noexcept {
+#if _CONTAINER_DEBUG_LEVEL > 0
+        if constexpr (rank_dynamic() != 0) {
+            _STL_VERIFY(this->_Map.extents().template _Is_dynamic_multidim_index_space_size_representable<size_type>(),
+                "The size of the multidimensional index space extents() must be representable as a value of type "
+                "size_type (N4950 [mdspan.mdspan.members]/7).");
+        }
+#endif // _CONTAINER_DEBUG_LEVEL > 0
+        return static_cast<size_type>(
+            _Fwd_prod_of_extents<extents_type>::_Calculate(this->_Map.extents(), extents_type::_Rank));
```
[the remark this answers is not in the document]
StephanTLavavej: Filed #4023.

CaseyCarter (approved):

CaseyCarter:
# 🚀

StephanTLavavej:

StephanTLavavej:
Thanks again @JMazurkiewicz, @MattStephanson, @AlexGuteniev, @frederick-vs-ja, and @CaseyCarter!

:tada: :tada: :tada:
===

PR #3974, opened 2023-08-17
Cleanups for `condition_variable`'s waiting functions

achabense:
Simplifies ~~, and fixes some minor problems in~~ `condition_variable`'s waiting functions. The correctness can be confirmed by looking through commits in order.

Works towards #718; after this change it will be much easier to fix `condition_variable`'s clock-awareness problem.

`condition_variable_any` are not effected by this pr.

### DIFF
--- a/stl/inc/mutex
+++ b/stl/inc/mutex
@@ -574,7 +574,7 @@ public:
     cv_status wait_for(unique_lock<mutex>& _Lck, const chrono::duration<_Rep, _Period>& _Rel_time) {
         // wait for duration
         if (_Rel_time <= chrono::duration<_Rep, _Period>::zero()) {
-            // we don't lock-and-relock _Lck for this case because it's not observable
+            // we don't unlock-and-relock _Lck for this case because it's not observable
             return cv_status::timeout;
         }
         return wait_until(_Lck, _To_absolute_time(_Rel_time));
@@ -601,7 +601,7 @@ public:
         for (;;) {
             const auto _Now = _Clock::now();
             if (_Abs_time <= _Now) {
-                // we don't lock-and-relock _Lck for this case because it's not observable
+                // we don't unlock-and-relock _Lck for this case because it's not observable
                 return cv_status::timeout;
             }
 
--- a/stl/inc/mutex
+++ b/stl/inc/mutex
@@ -588,16 +588,21 @@ public:
 #if _HAS_CXX20
         static_assert(chrono::is_clock_v<_Clock>, "Clock type required");
 #endif // _HAS_CXX20
+        if (!_Mtx_current_owns(_Lck.mutex()->_Mymtx())) {
+            _Throw_Cpp_error(_OPERATION_NOT_PERMITTED);
+        }
         for (;;) {
             const auto _Now = _Clock::now();
             if (_Abs_time <= _Now) {
                 return cv_status::timeout;
             }
 
+            // TRANSITION, ABI: incorrectly relying on system clock
             _timespec64 _Tgt;
             (void) _To_timespec64_sys_10_day_clamped(_Tgt, _Abs_time - _Now);
-            const cv_status _Result = _Wait_until_sys_time(_Lck, &_Tgt);
-            if (_Result == cv_status::no_timeout) {
+            // Nothing to do to comply with LWG-2135 because std::mutex lock/unlock are nothrow
+            const _Thrd_result _Res = _Cnd_timedwait(_Mycnd(), _Lck.mutex()->_Mymtx(), &_Tgt);
+            if (_Res == _Thrd_result::_Success) {
                 return cv_status::no_timeout;
             }
         }
@@ -634,22 +639,6 @@ private:
 
     _Cnd_t _Mycnd() noexcept { // get pointer to _Cnd_internal_imp_t inside _Cnd_storage
         return reinterpret_cast<_Cnd_t>(&_Cnd_storage);
-    }
-
-    cv_status _Wait_until_sys_time(unique_lock<mutex>& _Lck, const _timespec64* _Abs_time) {
-        // wait for signal with timeout
-        if (!_Mtx_current_owns(_Lck.mutex()->_Mymtx())) {
-            _Throw_Cpp_error(_OPERATION_NOT_PERMITTED);
-        }
-
-        // Nothing to do to comply with LWG-2135 because std::mutex lock/unlock are nothrow
-        const _Thrd_result _Res = _Cnd_timedwait(_Mycnd(), _Lck.mutex()->_Mymtx(), _Abs_time);
-
-        if (_Res == _Thrd_result::_Success) {
-            return cv_status::no_timeout;
-        } else {
-            return cv_status::timeout;
-        }
     }
 };
 
--- a/stl/inc/mutex
+++ b/stl/inc/mutex
@@ -573,20 +573,7 @@ public:
     template <class _Rep, class _Period>
     cv_status wait_for(unique_lock<mutex>& _Lck, const chrono::duration<_Rep, _Period>& _Rel_time) {
         // wait for duration
-        if (_Rel_time <= chrono::duration<_Rep, _Period>::zero()) {
-            return cv_status::timeout;
-        }
-
-        // TRANSITION, ABI: The standard says that we should use a steady clock,
-        // but unfortunately our ABI relies on the system clock.
-        _timespec64 _Tgt;
-        const bool _Clamped     = _To_timespec64_sys_10_day_clamped(_Tgt, _Rel_time);
-        const cv_status _Result = _Wait_until_sys_time(_Lck, &_Tgt);
-        if (_Clamped) {
-            return cv_status::no_timeout;
-        }
-
-        return _Result;
+        return wait_until(_Lck, _To_absolute_time(_Rel_time));
     }
 
     template <class _Rep, class _Period, class _Predicate>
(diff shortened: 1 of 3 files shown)


stl/inc/mutex
```diff
@@ -592,7 +592,7 @@ public:
     template <class _Rep, class _Period, class _Predicate>
     bool wait_for(unique_lock<mutex>& _Lck, const chrono::duration<_Rep, _Period>& _Rel_time, _Predicate _Pred) {
         // wait for signal with timeout and check predicate
-        return _Wait_until1(_Lck, _To_absolute_time(_Rel_time), _Pred);
+        return wait_until(_Lck, _To_absolute_time(_Rel_time), _Pass_fn(_Pred));
```
achabense, line 582: The standard says `std::move(_Pred)`, but I think passing by `_Pass_fn` will be ok - ~~at least I'm not making behavorial changes here~~.

stl/inc/mutex
```diff
@@ -624,14 +624,7 @@ public:
...
         while (!_Pred()) {
-            const auto _Now = _Clock::now();
-            if (_Abs_time <= _Now) {
-                return false;
-            }
-
-            _timespec64 _Tgt;
-            const bool _Clamped = _To_timespec64_sys_10_day_clamped(_Tgt, _Abs_time - _Now);
-            if (_Wait_until_sys_time(_Lck, &_Tgt) == cv_status::timeout && !_Clamped) {
+            if (wait_until(_Lck, _Abs_time) == cv_status::timeout) {
```
achabense, line 619: About this change:
1. This is specified by the standard and is in the most maintainable form. The correctness is now obviously only dependent on `wait_until(lock,abs_time,pred)`.
2. Actually this change fixes a bug - `if (_Abs_time <= _Now)` should return `_Pred()` too.
3. The efficiency loss is not significant IMO, and we will gain some efficiency back after fixing the clock-awareness bug.
4. `condition_variable_any` already does like this.

stl/inc/mutex
```diff
@@ -624,14 +624,7 @@ public:
         static_assert(chrono::is_clock_v<_Clock>, "Clock type required");
 #endif // _HAS_CXX20
         while (!_Pred()) {
-            const auto _Now = _Clock::now();
-            if (_Abs_time <= _Now) {
-                return false;
-            }
-
```
achabense: About this change:
1. This is specified by the standard and is in the most maintainable form. The correctness is now obviously only dependent on `wait_until(lock,abs_time,pred)`.
2. Actually this change fixes a bug - `if (_Abs_time <= _Now)` should return `_Pred()` too.
3. The efficiency loss is not significant IMO, and we will gain some efficiency back after fixing the clock-awareness bug.
4. `condition_variable_any` already does like this.

stl/inc/mutex
```diff
@@ -588,16 +588,21 @@ public:
 #if _HAS_CXX20
         static_assert(chrono::is_clock_v<_Clock>, "Clock type required");
 #endif // _HAS_CXX20
+        if (!_Mtx_current_owns(_Lck.mutex()->_Mymtx())) {
+            _Throw_Cpp_error(_OPERATION_NOT_PERMITTED);
+        }
```
achabense, line 593: Now the clock-awareness of all `wait_` functions are uniquely depentent on this usage of `_Cnd_timedwait` (which is not correct so far).
I'm not sure, should we add this in the beginning?
```cpp
if (_Abs_time <= _Clock::now()) {
    _Lck.unlock();
    _Lck.lock();
    return cv_status::timeout;
}
```
 The standard says the effects are
`(18.1) — Atomically calls lock.unlock() and blocks on *this.`
`(18.2) — When unblocked, calls lock.lock() (possibly blocking on the lock), then returns.`
`condition_variable_any`'s `wait_for` does things like this.

StephanTLavavej (approved):

stl/inc/mutex
```diff
@@ -601,16 +588,21 @@ public:
...
```
CaseyCarter, line 593: (Pre-existing) We conventionally avoid throwing exceptions in response to precondition violations. This should be a debug check:
```suggestion
#if _CONTAINER_DEBUG_LEVEL > 0
        _STL_VERIFY(_Lck.owns_lock(), "wait_until's caller must own the lock argument (N4958 [thread.condition.condvar]/17)");
        _STL_VERIFY(_Mtx_current_owns(_Lck.mutex()->_Mymtx()),
            "wait_until's calling thread must hold the lock argument's mutex (N4958 [thread.condition.condvar]/17)");
#endif // _CONTAINER_DEBUG_LEVEL > 0
```

stl/inc/mutex
```diff
@@ -573,20 +573,7 @@ public:
...
-        _timespec64 _Tgt;
-        const bool _Clamped     = _To_timespec64_sys_10_day_clamped(_Tgt, _Rel_time);
-        const cv_status _Result = _Wait_until_sys_time(_Lck, &_Tgt);
-        if (_Clamped) {
-            return cv_status::no_timeout;
-        }
-
-        return _Result;
+        return wait_until(_Lck, _To_absolute_time(_Rel_time));
     }
```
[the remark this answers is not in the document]
CaseyCarter: (1) The standard doesn't specify how things are implemented, it specifies how implementations behave. Keeping things close to the standard's wording is nice when 

(2) is conforming: `wait_until` is permitted to unblock spuriously by [thread.condition.condvar]/18.3. It's not perfect QoI, but waits longer than ten days are, as you say, exceedingly rare.

(3) If you're referring to the lock-and-relock degenerate case, it's unobservable (as I mention below).

I'm not averse to this change if it eases #718, but it seems like the microoptimization on lines 576-578 may be worth preserving even if we do delegate to `wait_until`.

stl/inc/mutex
```diff
@@ -601,16 +588,21 @@ public:
 #if _HAS_CXX20
         static_assert(chrono::is_clock_v<_Clock>, "Clock type required");
 #endif // _HAS_CXX20
+        if (!_Mtx_current_owns(_Lck.mutex()->_Mymtx())) {
+            _Throw_Cpp_error(_OPERATION_NOT_PERMITTED);
+        }
         for (;;) {
             const auto _Now = _Clock::now();
             if (_Abs_time <= _Now) {
                 return cv_status::timeout;
```
CaseyCarter, line 597:
```suggestion
                // we don't lock-and-relock _Lck for this case because it's not observable
                return cv_status::timeout;
```

stl/inc/mutex
```diff
@@ -588,16 +588,21 @@ public:
...
             }
 
+            // TRANSITION, ABI: incorrectly relying on system_clock
             _timespec64 _Tgt;
             (void) _To_timespec64_sys_10_day_clamped(_Tgt, _Abs_time - _Now);
-            const cv_status _Result = _Wait_until_sys_time(_Lck, &_Tgt);
-            if (_Result == cv_status::no_timeout) {
+            // Nothing to do to comply with LWG-2135 because std::mutex lock/unlock are nothrow
+            const _Thrd_result _Res = _Cnd_timedwait(_Mycnd(), _Lck.mutex()->_Mymtx(), _Abs_time);
+            if (_Res == _Thrd_result::_Success) {
```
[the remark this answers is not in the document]
CaseyCarter: `condition_variable_any` works with program-defined types. Here, `_Lck` is a `unique_lock<mutex>` which we control. The unlock-and-relock cycle isn't observable. (I've suggested adding a comment to record this fact above.)

stl/inc/mutex
CaseyCarter: This is fine. There's an implicit understanding in LWG that it's ok for implementations to move or copy a program-supplied object fewer times than the wording implies, or to move from a copy instead of copying again - nobody's figured out how to write blanket wording for it.

stl/inc/mutex
```diff
@@ -601,16 +588,21 @@ public:
...
+        if (!_Mtx_current_owns(_Lck.mutex()->_Mymtx())) {
+            _Throw_Cpp_error(_OPERATION_NOT_PERMITTED);
+        }
         for (;;) {
             const auto _Now = _Clock::now();
             if (_Abs_time <= _Now) {
                 return cv_status::timeout;
             }
 
+            // TRANSITION, ABI: incorrectly relying on system clock
```
CaseyCarter, line 600:
```suggestion
            // TRANSITION, ABI: should use a steady clock
```

CaseyCarter (changes requested):

stl/inc/mutex
CaseyCarter: (2) is not a bug. `return false` _is_ returning the result of the most-recent evaluation of `_Pred()`. (I do agree with the rest of this comment ;))

stl/inc/mutex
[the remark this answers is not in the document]
achabense: I've restored that `_Rel_time<=0` branch, though I think it is an unlikely case and may miss some opportunities to catch ownership bugs.

stl/inc/mutex
achabense: Hmm, should it be `unlock-and-relock`?

CaseyCarter: Derp, yes it should.

StephanTLavavej (approved):

stl/inc/mutex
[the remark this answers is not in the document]
achabense: I'll leave it as is in this pr. I'd prefer not to keep this branch, but it looks not worth resetting testing.

StephanTLavavej:

StephanTLavavej:
:broom: :sparkles: :tada:
===

PR #3977, opened 2023-08-18
Support the Standard Library Modules in C++20 mode

### LINKED ISSUE #3945: Supporting `import std;` in C++20
Currently, we block the C++23 Standard Library Modules from being used in C++20 mode. There's no technical reason for this, it's purely a policy choice. Supporting `import std;` in C++20 mode would be easy (remove the block, add test coverage, it'll take me just a few minutes - at least in the toolset, then we'll need to talk to the build systems), because we don't need anything from the C++23 Core Language here. However, after granting this ability to users, *we can never take it back*.

Therefore, we're looking for cross-vendor agreement. Next Wednesday (2023-08-16), the MSVC STL maintainers will talk about this at our weekly scheduled meeting and I expect we'll agree that this is a reasonable thing to do. If the libc++ and libstdc++ maintainer teams could discuss this in their own channels and then add comments here saying "The lib\<meow\>c++ maintainers want to support `import std;` in C++20 mode", then we can move forward together.

(This equally applies to `import std.compat;`, of course. It does not apply to any future modules that might be added to C++26 and beyond. Please note that we're not running a user survey here.)

StephanTLavavej:
Fixes #3945.

* `stl/inc/ctime`
  + Activate the workaround for internal OS-33790456 "`static inline` functions cannot be called when using modules" whenever `_BUILD_STD_MODULE` is defined, instead of additionally guarding it with `_HAS_CXX23`. This follows what `<new>` is already doing to detect whether we're building `std.ixx`: https://github.com/microsoft/STL/blob/60f18856c56389001c38a9929ef37bd5c01c4e47/stl/inc/new#L20
* `stl/inc/yvals_core.h`
  + Move the paper comment from the C++23 section to the C++20 section.
  + Originally, I silently ignored attempts to define `_BUILD_STD_MODULE` in too-old Standard modes. If they were then attempting to build `std.ixx`, they'd be confused by the results with no `export` appearing, and if they weren't building `std.ixx` then I was letting them get away with something very squirrely :chipmunk:. Now, `_EXPORT_STD` is simply directly controlled by `_BUILD_STD_MODULE`, but we have enforcement that immediately rejects attempts to do this in C++17 and older.
  + Move the feature-test macro from C++23 to C++20.
* `stl/modules/std.ixx`
  + Guard the 5 C++23-only headers (so far).
  + :warning: Stealth merge conflict with the remaining C++23 headers: `<mdspan>`, `<flat_map>`, `<flat_set>`, `<generator>`
* `tests/std/tests/VSO_0157762_feature_test_macros/test.compile.pass.cpp`
  + Update the feature-test macro test accordingly.
* `tests/std/tests/P1502R1_standard_library_header_units/env.lst`
  + Originally, I was testing the Standard Library Header Units (an official C++20 feature) only in `/std:c++latest` mode because it took a while to build, and we didn't have a lot of C++23 content yet. Now we should be more disciplined and actually test it in all possible Standard modes like everything else. Although Header Units emit macros, this test machinery is shared with the Named Modules test, so I'm having the test detect the Standard mode with a dedicated `TEST_STANDARD` macro (following a different convention than the product code, to avoid confusion).
* `tests/std/tests/P2465R3_standard_library_modules/env.lst`
  + Add C++20 coverage in an identical manner.
* `tests/std/include/test_header_units_and_modules.hpp`
  + Update the shared test code to guard the C++23-only headers.
* `tests/std/tests/P1502R1_standard_library_header_units/test.cpp`
  + Also guard the `import`s of C++23-only headers.


CaseyCarter (approved):

CaseyCarter:

CaseyCarter:
Really late thought: will we have ODR problems in a project that includes both C++20 TUs and C++23 TUs that `import std`?

StephanTLavavej:
Not any more than we already have with classic headers.

PR #3981, opened 2023-08-20
Fix incorrect parsing of floating numbers

### LINKED ISSUE #3980: <iostream>: Wrong reading of float values
# Describe the bug
"1e-07" is read from std::istream as `1e-06`

# Command-line test case
```
#include <iostream>
#include <sstream>
using namespace std;

int main()
{
    istringstream istr("1e-07");
    double xx = 0;
    istr >> xx;
    cout << "xx = " << xx << endl;
    return 0;
}
```

# STL version
Visual Studio 2022 17.7
Please use Runtime Library not from DLL

# Additional context
First found here: https://stackoverflow.com/q/76937234/7325599
This is also DevCom-10450662 / VSO-1876474 / AB#1876474

achabense:
Fixes #3980.

### DIFF
--- a/stl/inc/xlocnum
+++ b/stl/inc/xlocnum
@@ -1019,10 +1019,6 @@ private:
                 _Seendigit = true;
             }
 
-            if (_Seendigit) {
-                *_Ptr++ = '0'; // put one back
-            }
-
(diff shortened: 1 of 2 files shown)


achabense:
Hmm, I find it very terrible to name a branch as `GH-XXXX`...

PR #3982, opened 2023-08-20
Fix `num_get`'s float-parsing problem

### LINKED ISSUE #3980: <iostream>: Wrong reading of float values

achabense:
Fixes #3980 / DevCom-10450662 / VSO-1876474 / AB#1876474

Also fixes DevCom-10445082 / VSO-1873604 / AB#1873604.

Likely fixes VSO-1874745 / AB#1874745.

Likely fixes VSO-1876601 / AB#1876601.

### DIFF
--- a/stl/inc/xlocnum
+++ b/stl/inc/xlocnum
@@ -1019,10 +1019,6 @@ private:
                 _Seendigit = true;
             }
 
-            if (_Seendigit) {
-                *_Ptr++ = '0'; // put one back
-            }
-
             for (size_t _Idx; _First != _Last && (_Idx = _STD _Find_elem(_Atoms, *_First)) < _Offset_dec_digit_end;
                  _Seendigit = true, (void) ++_First) {
                 if (_Exponent_part < PTRDIFF_MAX / 10
--- a/tests/std/tests/LWG2381_num_get_floating_point/test.cpp
+++ b/tests/std/tests/LWG2381_num_get_floating_point/test.cpp
@@ -532,6 +532,23 @@ void test_gh_3378() {
     }
 }
 
+// Also test GH-3980 <iostream>: Wrong reading of float values
+template <class Flt>
+void test_gh_3980() {
+    auto test_case = [](const char* str, double expected) {
+        Flt val = 0;
+        istringstream{str} >> val;
+        assert((ostringstream{} << val).str() == (ostringstream{} << expected).str());
+    };
+
+    test_case("0x1p+07", 0x1p+07);
+    test_case("1e+07", 1e+07);
+    test_case("1e+0", 1);
+    test_case("1e-0", 1);
+    test_case("1e-07", 1e-07);
+    test_case("0x1p-07", 0x1p-07);
+}
+
 #if _HAS_CXX17
 void test_float_from_char_cases() {
     for (const auto& test_case : float_from_chars_test_cases) {
@@ -588,6 +605,10 @@ int main() {
 
     test_gh_3378<double>();
     test_gh_3378<long double>();
+
+    test_gh_3980<float>();
+    test_gh_3980<double>();
+    test_gh_3980<long double>();
 
 #if _HAS_CXX17
     test_float_from_char_cases();


tests/std/tests/LWG2381_num_get_floating_point/test.cpp
```diff
@@ -532,6 +532,30 @@ void test_gh_3378() {
...
+
+    istringstream{"1e-0"} >> val;
+    assert((ostringstream{} << val).str() == "1");
+
+    istringstream{"1e-01"} >> val;
+    assert((ostringstream{} << val).str() == (ostringstream{} << 1e-01).str());
+
+    istringstream{"1e-07"} >> val;
+    assert((ostringstream{} << val).str() == (ostringstream{} << 1e-07).str());
+}
```
achabense, line 557: These can also pass the test, but I think there is no need to add them.
```
    istringstream{"1e07"} >> val;
    assert((ostringstream{} << val).str() == (ostringstream{} << 1e07).str());

    istringstream{"1e01"} >> val;
    assert((ostringstream{} << val).str() == "10");

    istringstream{"1e0"} >> val;
    assert((ostringstream{} << val).str() == "1");
```

stl/inc/xlocnum
```diff
@@ -1019,10 +1019,6 @@ private:
                 _Seendigit = true;
             }
 
-            if (_Seendigit) {
-                *_Ptr++ = '0'; // put one back
-            }
-
```
achabense, line 1025: This branch is useless, as the integral and fraction part has been finished, and the exponent part will be written here:
https://github.com/microsoft/STL/blob/60f18856c56389001c38a9929ef37bd5c01c4e47/stl/inc/xlocnum#L1092-L1096

achabense:
I haven't checked the standard carefully, but please check whether this is just another problem.

If the function sees exponent and finds no following digits, it will consider the whole section as invalid:
```
#include <iostream>
#include <sstream>
using namespace std;

int main()
{
    istringstream istr("1.3e");
    double x = 0;
    istr >> x;
    cout << x; // "0"

    double y = stod("1.3e");
    cout << y; // "1.3"
}
```

Relevant logic:
https://github.com/microsoft/STL/blob/6c69a73911b33892919ec628c0ea5bbf0caf8a6a/stl/inc/xlocnum#L1003-L1006
https://github.com/microsoft/STL/blob/6c69a73911b33892919ec628c0ea5bbf0caf8a6a/stl/inc/xlocnum#L1042-L1044

StephanTLavavej (approved):

StephanTLavavej:

statementreply:
> I haven't checked the standard carefully, but please check whether this is just another problem: If the function sees exponent character and finds no following digits, it will consider the whole section as invalid: [(maybe not a problem, as gcc and clang have the same output)](https://godbolt.org/z/YE789x3K9)

I think this is the correct behavior according to [[facet.num.get.virtuals]/3.3.6](https://eel.is/c++draft/facet.num.get.virtuals#3.3.6):
> The numeric value to be stored can be one of:
> - zero, if the conversion function does not convert the entire field[.](https://eel.is/c++draft/facet.num.get.virtuals#3.3.6.sentence-1)
> - ...

achabense:
For the tests introduced in this pr, [clang accepts it and gcc is rejecting those with `0x` prefixes](https://godbolt.org/z/Mh16KbEEq).
```
#include<sstream>
#include<iostream>
using namespace std;

int main() {
    using Flt = double;
    auto test_case = [](const char* str, double expected) {
        Flt val = 0;
        istringstream{ str } >> val;
        if ((ostringstream{} << val).str() != (ostringstream{} << expected).str()) {
            cout << val << "vs" << expected << endl;
        }
        };

    test_case("0x1p+07", 0x1p+07); // gcc rejects this case
    test_case("1e+07", 1e+07);
    test_case("1e+0", 1);
    test_case("1e-0", 1);
    test_case("1e-07", 1e-07);
    test_case("0x1p-07", 0x1p-07); // gcc rejects this case
}
```

tests/std/tests/LWG2381_num_get_floating_point/test.cpp
```diff
@@ -532,6 +532,23 @@ void test_gh_3378() {
     }
 }
 
+// Also test GH-3980 <iostream>: Wrong reading of float values
+template <class Flt>
```
CaseyCarter, line 536: Not worth resetting testing: vwls pls.
`Flt` isn't _that_ much shorter than `Float`. If you want a descriptive name, be descriptive. If you want a short name, `T` would work fine here and honestly `std::floating_point T` (or `static_assert(std::is_floating_point_v<T>)` in the function body) would be even better.

CaseyCarter (approved):
Just style nitpicks; ignore me.

tests/std/tests/LWG2381_num_get_floating_point/test.cpp
StephanTLavavej: `Float` specifically is dangerous due to the potential confusion with `float`. `Floating` or any other name avoids that hazard.

StephanTLavavej:
Thanks for fixing this major regression! :hammer_and_wrench: :sparkles: :tada:

Issue #2912, opened 2022-07-25
P1222R4 `<flat_set>`


StephanTLavavej:
WG21-P1222R4 `<flat_set>`
WG21-P3567R2 `flat_meow` Fixes
LWG-3751 Missing feature macro for `flat_set`
LWG-3774 `<flat_set>` should include `<compare>`
LWG-3786 Flat maps' deduction guides need to default `Allocator` to be useful
LWG-3803 `flat_`*`meow`* constructors taking `KeyContainer` lack `KeyCompare` parameter
LWG-3879 `erase_if` for `flat_{,multi}set` is incorrectly specified
LWG-3884 `flat_`*`meow`* is missing allocator-extended copy/move constructors
LWG-3987 Including `<flat_meow>` doesn't provide `std::begin/end`
LWG-4384 `flat_set::erase(iterator)` is underconstrained

Feature-test macro:
```cpp
#define __cpp_lib_flat_set 202511L
```

Internal VS-PR-420058 has taught the IDE about this new extensionless header - thanks @CaseyCarter!

WG21-P3567R2 "`flat_meow` Fixes" was voted into C++26 at the November 2025 meeting; while it wasn't officially a C++23 DR, we should obviously implement it as part of the initial release.

Saalvage:
I have just started on a branch implementing this feature. :)
It will need a lot more tests and a few improvements but I'd love to contribute it!

StephanTLavavej:
We're starting a multi-collaborator feature branch, `feature/flat_set`, with @Saalvage's implementation, so that it can be refined and test coverage can be added.

vmichal:
I have checked our support of papers tracked by this issue (also by #2910 ).
I would recommend removing the implemented papers from the list in the description of this issue and #2910 (if not to close these issues completely) to keep track of what is still missing.

Fully supported:
- LWG-3751 "Missing feature macro for `flat_set`"
- LWG-3774 "`<flat_set>` should include `<compare>`"
- LWG-3786 "Flat map's deduction guides need to default `Allocator` to be useful."
- LWG-3803 "`flat_meow` constructors taking `KeyContainer` lack `KeyCompare` parameter."
- LWG-3987 "Including `<flat_meow>` doesn't provide `std::begin/end`"
- LWG-3879 "`erase_if` for `flat_{,multi}set` is incorrectly specified"

Comments needed:
[...]

PR #3985, opened 2023-08-22
Various bugfixes / enhancements for `<flat_set>`

achabense:
`This pr is becoming large and hard to review, I'm going to replace this pr with a new one with all commits re-grouped when it is finished.`

This pr contains a lot of bugfixes/enhancements for `<flat_set>`:

About each commit:
- `1st`: Fix usage of `_Compressed_pair`. `_Compressed_pair` tries to compress its first member, but the implementation was erroneously trying to compress the container instead of `key_compare`.
- `2nd`: Add `#pragma pack(push, _CRT_PACKING)` etc.
- `3rd`: (This turns out to be merely a nitpick. see [this comment](https://github.com/microsoft/STL/pull/3985#discussion_r1302665643).)
- `4th`: `_Emplace_hint` should return `iterator` instead of `void`, and `const_iterator _Where` may be not be convertible to `iterator`. (As to `const const_iterator _Begin = cbegin(); const const_iterator _End   = cend();`, they are enhancements to match with `_Where`.)
- `5th`: Nitpicks to enhance constness.
- `6th`: Regroup/reorder non-constructor methods to match with the standard. `insert(initializer_list)` is found missing.
- `7th`: Implement `insert(initializer_list)`.
- `8th`: Improve implementation of `contains` by using `binary_search`.
- `9th`: Fix `insert` methods. Especially, `insert(const _Kty&)` was broken before this commit.
- `10th`: `operator=(initializer_list)` was not observable from `flat_set/multiset`. `flat_set/multiset` should `using _Mybase::operator=;`. (this turns out to be problematic; fixed by 14th commit)
- `11th`: Various cleanups.
- `12th`: `insert(const iter&, const iter&)` should be `insert(iter, iter)`
- `13th` : Implement LWG-3884 (`*flat_foo is missing allocator-extended copy/move constructors*`), as applied in N4958.
- `14th`: Fix 10th commit. Add tests for insertion methods, the new constructor and ebco optimization; `test_insert` is a general test covering commit `4, 7, 9 and 12`.
- `15th`: `emplace` forgot to return the result of `_Emplace`; use `const key_compare&` for comparision consistently; some other nitpicks.
- `16th`: `emplace` and `emplace_hint` should not construct temporary for `key_type` (so that a `key_type&&` will not be `move`d away if insertion fails); try to add debug check for `flat_set::insert([hint,]auto&&)`.
- `17th`: For conformance and efficiency, `flat_multiset::emplace` should use `upper_bound` instead of `lower_bound`.
- `18th`: Fix incorrect bound-finding behavior of `emplace_hint` (`flat_set` should always look for `lower_bound`); update test.

Works towards https://github.com/microsoft/STL/issues/2912.

### DIFF
--- a/stl/inc/flat_set
+++ b/stl/inc/flat_set
@@ -208,11 +208,21 @@ public:
     // modifiers
     template <class... _Args>
     auto emplace(_Args&&... _Vals) {
-        return _Emplace(_Kty{_STD forward<_Args>(_Vals)...});
+        constexpr bool _Is_key_type = _In_place_key_extract_set<_Kty, remove_cvref_t<_Args>...>::_Extractable;
+        if constexpr (_Is_key_type) {
+            return _Emplace(_STD forward<_Args>(_Vals)...);
+        } else {
+            return _Emplace(_Kty{_STD forward<_Args>(_Vals)...});
+        }
     }
     template <class... _Args>
     iterator emplace_hint(const_iterator _Hint, _Args&&... _Vals) {
-        return _Emplace_hint(_Hint, _Kty{_STD forward<_Args>(_Vals)...});
+        constexpr bool _Is_key_type = _In_place_key_extract_set<_Kty, remove_cvref_t<_Args>...>::_Extractable;
+        if constexpr (_Is_key_type) {
+            return _Emplace_hint(_Hint, _STD forward<_Args>(_Vals)...);
+        } else {
+            return _Emplace_hint(_Hint, _Kty{_STD forward<_Args>(_Vals)...});
+        }
     }
 
     auto insert(const value_type& _Val) {
@@ -435,12 +445,22 @@ private:
         const iterator _End   = end();
         const iterator _Where = lower_bound(_Val);
         if constexpr (_Multi) {
+            _STL_INTERNAL_STATIC_ASSERT(is_same_v<remove_cvref_t<_Ty>, _Kty>);
             return _Cont.emplace(_Where, _STD forward<_Ty>(_Val));
         } else {
             if (_Where != _End && _Keys_equal(*_Where, _Val)) {
                 return pair{_Where, false};
             }
-            return pair{_Cont.emplace(_Where, _STD forward<_Ty>(_Val)), true};
+            if constexpr (is_same_v<remove_cvref_t<_Ty>, _Kty>) {
+                return pair{_Cont.emplace(_Where, _STD forward<_Ty>(_Val)), true};
+            } else {
+                // flat_set::insert(auto&&)
+                _STL_INTERNAL_STATIC_ASSERT(_Keylt_transparent && is_constructible_v<_Kty, _Ty>);
+                _Kty _Keyval{_STD forward<_Ty>(_Val)};
+                _STL_ASSERT(lower_bound(_Keyval) == _Where && !_Keys_equal(_Keyval, *_Where),
+                    "find(val) was not equal to find(key_type{forward(val)})");
+                return pair{_Cont.emplace(_Where, _STD move(_Keyval)), true};
+            }
         }
     }
 
@@ -468,12 +488,22 @@ private:
         }
 
         if constexpr (_Multi) {
+            _STL_INTERNAL_STATIC_ASSERT(is_same_v<remove_cvref_t<_Ty>, _Kty>);
             return _Cont.emplace(_Where, _STD forward<_Ty>(_Val));
         } else {
             if (_Where != _End && _Keys_equal(_Val, *_Where)) {
                 return _Cont.begin() + (_Where - _Begin);
             }
-            return _Cont.emplace(_Where, _STD forward<_Ty>(_Val));
+            if constexpr (is_same_v<remove_cvref_t<_Ty>, _Kty>) {
+                return _Cont.emplace(_Where, _STD forward<_Ty>(_Val));
+            } else {
+                // flat_set::insert(hint,auto&&)
+                _STL_INTERNAL_STATIC_ASSERT(_Keylt_transparent && is_constructible_v<_Kty, _Ty>);
+                _Kty _Keyval{_STD forward<_Ty>(_Val)};
+                _STL_ASSERT(lower_bound(_Keyval) == _Where && !_Keys_equal(_Keyval, *_Where),
+                    "find(val) was not equal to find(key_type{forward(val)})");
+                return _Cont.emplace(_Where, _STD move(_Keyval));
+            }
         }
     }
 
--- a/stl/inc/flat_set
+++ b/stl/inc/flat_set
@@ -612,7 +612,7 @@ private:
     using _Mybase = _Base_flat_set<_Kty, _Keylt, _Container, false, flat_set, sorted_unique_t>;
 
 public:
-    using _Mybase::_Base_flat_set;
+    using _Mybase::_Mybase;
     using _Mybase::operator=;
 };
 
@@ -623,7 +623,7 @@ private:
     using _Mybase = _Base_flat_set<_Kty, _Keylt, _Container, true, flat_multiset, sorted_equivalent_t>;
 
 public:
-    using _Mybase::_Base_flat_set;
+    using _Mybase::_Mybase;
     using _Mybase::operator=;
 };
 
--- a/stl/inc/flat_set
+++ b/stl/inc/flat_set
@@ -247,9 +247,12 @@ public:
         _Restore_invariants_after_insert<false>(_Old_size);
     }
 
-    // TODO, missing
-    void insert(initializer_list<_Kty> _Ilist);
-    void insert(_Tsorted, initializer_list<_Kty> _Ilist);
+    void insert(initializer_list<_Kty> _Ilist) {
+        _Insert_range<false>(_Ilist.begin(), _Ilist.end());
+    }
+    void insert(_Tsorted, initializer_list<_Kty> _Ilist) {
+        _Insert_range<true>(_Ilist.begin(), _Ilist.end());
+    }
 
     _NODISCARD container_type extract() && {
         container_type& _Cont = _Get_cont();
--- a/stl/inc/flat_set
+++ b/stl/inc/flat_set
@@ -111,11 +111,14 @@ public:
     _Base_flat_set(_Iter _First, _Iter _Last, const _Alloc& _Al) : _Base_flat_set(container_type(_First, _Last, _Al)) {}
 
     template <_Container_compatible_range<_Kty> _Rng>
-    _Base_flat_set(from_range_t, _Rng&& _Range, const key_compare& _Comp = key_compare())
-        : _Base_flat_set(container_type(from_range, _STD forward<_Rng>(_Range)), _Comp) {}
+    _Base_flat_set(from_range_t, _Rng&& _Range)
+        : _Base_flat_set(container_type(from_range, _STD forward<_Rng>(_Range))) {}
     template <_Container_compatible_range<_Kty> _Rng, _Allocator_for<container_type> _Alloc>
     _Base_flat_set(from_range_t, _Rng&& _Range, const _Alloc& _Al)
         : _Base_flat_set(container_type(from_range, _STD forward<_Rng>(_Range), _Al)) {}
+    template <_Container_compatible_range<_Kty> _Rng>
+    _Base_flat_set(from_range_t, _Rng&& _Range, const key_compare& _Comp)
+        : _Base_flat_set(container_type(from_range, _STD forward<_Rng>(_Range)), _Comp) {}
     template <_Container_compatible_range<_Kty> _Rng, _Allocator_for<container_type> _Alloc>
     _Base_flat_set(from_range_t, _Rng&& _Range, const key_compare& _Comp, const _Alloc& _Al)
         : _Base_flat_set(container_type(from_range, _STD forward<_Rng>(_Range), _Al), _Comp) {}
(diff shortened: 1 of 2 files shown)


achabense:
I'm going to do more reviews tomorrow; I will make more pushes to this pr if I find other problems.

achabense:
It seems the standard doesn't say what `flat_set::emplace` do? (only `flat_multiset::emplace` is specified)

frederick-vs-ja:
> It seems the standard doesn't say what `flat_set::emplace` should do? (only `flat_multiset::emplace` is specified)

May be specified in [[associative.reqmts.general]/49](https://eel.is/c++draft/associative.reqmts.general#49).

But it seems to me that `flat_set::emplace` possibly uses `Alloc::construct`, while `flat_multiset::emplace` doesn't. An LWG issue may be needed.

achabense:
https://github.com/microsoft/STL/blob/feature/flat_set/stl/inc/flat_set#L155-L163
Is it preferable to use `ranges::method_name(_Get_cont())`? I think it's too restrictive to define these function as `_Get_cont().method__name()`.

achabense:
I've just made a push to fix `_Emplace_hint`'s signature problem. However, I'm afraid that the current implementation of `flat_(multiset)::emplace/insert` methods are somewhat broken, and the situation is worsened by some unclarity in the standard.

Hope I'm mistaken.

stl/inc/flat_set
```diff
@@ -432,11 +432,11 @@ private:
 
     template <class _Ty>
         requires (_Keylt_transparent && is_constructible_v<_Kty, _Ty>) || is_same_v<_Ty, _Kty>
-    void _Emplace_hint(const_iterator _Where, _Ty&& _Val) {
-        _Container& _Cont     = _Get_cont();
-        _Keylt& _Compare      = _Get_comp();
-        const iterator _Begin = begin();
-        const iterator _End   = end();
+    iterator _Emplace_hint(const_iterator _Where, _Ty&& _Val) {
```
achabense, line 435: Also, according to how it is used, this function should be called `_Insert_hint`.

stl/inc/flat_set
```diff
@@ -513,8 +513,8 @@ private:
     template <class _Other>
         requires _Keylt_transparent || is_same_v<_Other, _Kty>
     _NODISCARD const_iterator _Find(const _Other& _Val) const {
-        const iterator _End   = end();
-        const iterator _Where = lower_bound(_Val);
+        const const_iterator _End   = end();
+        const const_iterator _Where = lower_bound(_Val);
```
achabense, line 517: Rationale: `iterator` is not convertible to `const_iterator` (return type) in many cases.

stl/inc/flat_set
```diff
@@ -453,7 +454,7 @@ private:
             if (_Where == _End || !_Keys_equal(_Val, *_Where)) {
                 return _Cont.insert(_Where, _STD forward<_Ty>(_Val));
             }
-            return _Where;
+            return _Cont.begin() + (_Where - _Begin);
```
achabense, line 457: I'm just making this line work (`const_iterator` to `iterator`); may need further improvements in the future.

achabense:
https://github.com/microsoft/STL/blob/d6e6b5ac5d7e43b2210d98deb6cea20263abe2af/stl/inc/flat_set#L137-L141
I'm afraid this is not observable from `flat_set`, as `flat_set/multiset` is currently not using 's base class's `operator=`

stl/inc/flat_set
```diff
@@ -201,23 +201,23 @@ public:
...
     }
     template <class... _Args>
     iterator emplace_hint(const_iterator _Hint, _Args&&... _Vals) {
         return _Emplace_hint(_Hint, _Kty{_STD forward<_Args>(_Vals)...});
     }
 
     auto insert(const value_type& _Val) {
-        return _Insert<_Kty>(_Val);
+        return _Emplace(_Val);
     }
```
achabense, line 213: Before this change, even the following code will fail:
```cpp
flat_set<int> s;
int i = 1;
s.insert(i);
```

stl/inc/flat_set
```diff
@@ -467,8 +466,7 @@ private:
     }
 
     template <class _Ty>
-        requires (!_Multi && _Keylt_transparent && is_constructible_v<_Kty, _Ty>) || is_same_v<_Ty, _Kty>
```
achabense, line 469: about this removal:
- it's enough to constrain in the caller site (`insert` / `emplace`).
- shouldn't be `is_same_v<_Ty, _Kty>`, as _Ty can be `const _Kty&`

stl/inc/flat_set
duanqn: Can you explain a bit? I thought it was required to have `iterator` convertible to `const_iterator` https://eel.is/c++draft/container.reqmts#6

achabense: This will fail; is this a bug in the current implementation?
```
    vector<int> i;
    auto nonc = i.begin();
    nonc = i.cbegin();
```

achabense: Ugh, my mistake; `const_iterator` is not required to convert to `iterator`. (The fix is correct.)

achabense: Ugh, my mistake; correction: const_iterator is not required to convert to iterator.
Update: the 3rd commit doesn't "fix" anything (but can be seen as nitpick/enhancement); there are real iterator fixes, see 4th commit.

achabense:
For this function, I can imagine it causing a lot of surprises in practice. I think we should try to add debug check in the future.
```cpp
    template <class _Other>
        requires (!_Multi && _Keylt_transparent && is_constructible_v<_Kty, _Other>)
    auto insert(_Other&& _Val) {
        return _Emplace(_STD forward<_Other>(_Val));
    }
```
image

achabense:
https://github.com/microsoft/STL/blob/d6e6b5ac5d7e43b2210d98deb6cea20263abe2af/stl/inc/flat_set#L548-L550
I think `key_compare` should be `regular_invocable`? (so that when referred-to explicitly, it's always preferred to use `const key_compare& _Compare` instead)

achabense:
I've force-pushed this pr to add serial numbers for each commit description. Nothing else is changed.

tests/std/tests/P1222R4_flat_set/test.cpp
```diff
@@ -132,9 +119,74 @@ void test_constructors() {
     flat_set<int> a{};
     a = {1, 7, 7, 7, 2, 100, -1};
```
achabense, line 120: Previously, this was passing as the `initializer_list` was implicitly converted to `flat_multiset` and `operator=(flat_set&&)` was called here. Now I can make sure this is calling `operator=(initializer_list)`

achabense:
Hmm, is this UB or another implementation bug?
```   
    std::flat_set<int, std::less<>> s{ 1,2,3,5,7 };
    s.insert(2.5);
    for (auto i : s) {
        std::cout << i << ' '; // 1 2 2 3 5 7
    }
```

achabense:
There is no definition of `emplace_hint` in `flat_meow` part.
`emplace_hint` defined in [`associative.reqmts.general`](https://eel.is/c++draft/associative.reqmts.general#58):
image
`flat_multiset::emplace` defined [here](https://eel.is/c++draft/flat.multiset#modifiers-2):
image
Then, [I find](https://eel.is/c++draft/flat.multiset#modifiers):
image

What is this @frederick-vs-ja?

stl/inc/flat_set
```diff
@@ -208,11 +208,21 @@ public:
     // modifiers
     template <class... _Args>
     auto emplace(_Args&&... _Vals) {
-        return _Emplace(_Kty{_STD forward<_Args>(_Vals)...});
+        constexpr bool _Is_key_type = _In_place_key_extract_set<_Kty, remove_cvref_t<_Args>...>::_Extractable;
+        if constexpr (_Is_key_type) {
+            return _Emplace(_STD forward<_Args>(_Vals)...);
```
achabense, line 213: As to `emplace` and `emplace_hint`:
When input is `(const)_Kty(&/&&)`, I believe there shouldn't be a temporary before insertion, even for `flat_multiset`; especially, for `flat_set` a `_Kty&&` input should not be moved to a temporary if insertion fails.
For example, the following code holds true for `set` and `unordered_set`
```cpp
    Set<shared_ptr<int>> s;
    auto a = shared_ptr<int>(new int);
    auto b = a, c = a;
    assert(a == b && a == c);
    assert(!(a < b) && !(b < a));
    s.insert(a);
    s.emplace(move(b));
    assert(b != nullptr);
    s.insert(move(c));
    assert(c != nullptr);
```

stl/inc/flat_set
```diff
@@ -435,12 +445,22 @@ private:
...
+            if constexpr (is_same_v<remove_cvref_t<_Ty>, _Kty>) {
+                return pair{_Cont.emplace(_Where, _STD forward<_Ty>(_Val)), true};
+            } else {
+                // flat_set::insert(auto&&)
+                _STL_INTERNAL_STATIC_ASSERT(_Keylt_transparent && is_constructible_v<_Kty, _Ty>);
+                _Kty _Keyval{_STD forward<_Ty>(_Val)};
+                _STL_ASSERT(lower_bound(_Keyval) == _Where && !_Keys_equal(_Keyval, *_Where),
+                    "find(val) was not equal to find(key_type{forward(val)})");
+                return pair{_Cont.emplace(_Where, _STD move(_Keyval)), true};
+            }
```
achabense, line 463: This is for debug of `flat_set::insert([hint,]auto&&)`; there are fairly dangerous use patterns like:
```cpp
    std::flat_set<int, std::less<>> s{ 1,2,3,5,7 };
    s.insert(2.5);
```
Which can break invariant unexpectedly. See details in [this comment](https://github.com/microsoft/STL/pull/3985#issuecomment-1691614165).

Otherwise, this is actually equivalent to the `if constexpr (is_same_v<remove_cvref_t<_Ty>, _Kty>)` branch (as `_Cont.emplace(_Where, _STD forward<_Ty>(_Val))` can does the same thing).

**Due to the trouble it introduces, I believe `flat_set::insert([hint,]auto&&)` should be renamed in the standard to avoid accidental misuse.**

duanqn: I don't get why this is dangerous. Wouldn't `2.5` be converted to `2` first and cause the insertion to fail? I tested the 2.5 insertion with `flat_map` and the flat_map was unchanged.

achabense: find(2.5) whould get `end()`; find(int(2.5))->find(2) would get 2; this is precondition violation and is UB.

achabense: For `flat_set::emplace`, I think a conversion to the `key_type` is required before insertion, as this is explicitly required in `flat_multiset::emplace`:
image
Conversion happens before insertion even if `flat_multiset` can have transparent comparators...

image
This means I cannot call `emplace(_hint)` directly, as in `emplace`, for non-key types I need to firstly convert `auto&&` to a `key_type`, which may change the value of `auto&&`.

duanqn: > find(2.5) whould get `end()`; find(int(2.5))->find(2) would get 2; this is precondition violation and is UB.
> 
> See [this link](https://github.com/microsoft/STL/pull/3985#issuecomment-1691024595)

So the implementation is not wrong then ... ? It's just counter-intuitive?

achabense: I'm still not confident whether the implementation is correct for `insert([hint,]auto&&)`; but yes, `s.insert(2.5);` is just UB behavior, and I think this method is too tricky and can cause a lot of surprises in practice.

stl/inc/flat_set
```diff
@@ -424,63 +440,101 @@ private:
...
+                    // _Val >= *(_Where-1)
+                    // upper_bound is _Where
+                } else {
+                    // _Val < *(_Where-1)
+                    // upper_bound is in [_Begin,_Where-1]
+                    _Where = _STD upper_bound(_Begin, _Where - 1, _Val);
+                }
             } else {
-                // _Val < (*_Where - 1)
-                _Where = _STD upper_bound(_Begin, _Where, _Val, _Compare);
```
achabense, line 442: This line was making the following test failing:
```cpp
        flat_set<int, lt, C> a{0, 5};
        assert_all_requirements_and_equals(a, {0, 5});
        a.emplace_hint(a.end());
        assert_all_requirements_and_equals(a, {0, 5}); // < got {0, 0, 5}!
```
flat_set should always use `lower_bound` for correctness; `flat_multiset` should use upper_bound for efficiency (and this is also specified by he standard).

stl/inc/flat_set
```diff
@@ -441,13 +441,13 @@ private:
...
     template <class _Ty>
     auto _Emplace(_Ty&& _Val) {
-        _Container& _Cont     = _Get_cont();
-        const iterator _End   = end();
-        const iterator _Where = lower_bound(_Val);
+        _Container& _Cont = _Get_cont();
         if constexpr (_Multi) {
             _STL_INTERNAL_STATIC_ASSERT(is_same_v<remove_cvref_t<_Ty>, _Kty>);
-            return _Cont.emplace(_Where, _STD forward<_Ty>(_Val));
+            return _Cont.emplace(upper_bound(_Val), _STD forward<_Ty>(_Val));
```
achabense, line 447: image

achabense:
This pr is getting large, and some commits have contents that become outdated after newer pushes. I think the pr is becoming hard to review.
I'm going to replace this pr with a new one with all changes re-grouped when all chages are done. This pr will be kept in WIP state until removed.

achabense:
For ease of review, this pr now replaced with #3993.

PR #3989, opened 2023-08-23
Don't `memcmp(x, y, -1)` from `ranges::lexicographical_compare`

CaseyCarter:
... even with two potentially-infinite ranges. 

Drive by: Pull non-dependent test cases out of the `instantiator::call` template magic in `tests/P0896R4_ranges_alg_lexicographical_compare`.

Fixes VSO-1854238 / AB#1854238

### DIFF
--- a/stl/inc/algorithm
+++ b/stl/inc/algorithm
@@ -10654,7 +10654,7 @@ namespace ranges {
             constexpr bool _Is_sized2         = sized_sentinel_for<_Se2, _It2>;
             if constexpr (!is_void_v<_Memcmp_classification_pred> && _Sized_or_unreachable_sentinel_for<_Se1, _It1>
                           && _Sized_or_unreachable_sentinel_for<_Se2, _It2> && same_as<_Pj1, identity>
-                          && same_as<_Pj2, identity>) {
+                          && same_as<_Pj2, identity> && (_Is_sized1 || _Is_sized2)) {
                 if (!_STD is_constant_evaluated()) {
                     size_t _Num1;
                     if constexpr (_Is_sized1) {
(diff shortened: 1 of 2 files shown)


strega-nil-ms:
I think we should maybe reconsider calling `memcmp(x, y, -1)` in this case entirely, at least under `__SANITIZE_ADDRESS__`, since, while `memcmp` is UB when you don't call it with fully valid buffers[^1], calling `lexicographical_compare(i1, unreachable_sentinel, i2, unreachable_sentinel)` is valid (at least, that's my understanding; I may fully be wrong about that and this is a test of an additional guarantee we provide).

[^1]: https://trust-in-soft.com/blog/2015/12/21/memcmp-requires-pointers-to-fully-valid-buffers/

CaseyCarter:
> I think we should maybe reconsider calling `memcmp(x, y, -1)` in this case entirely, at least under `__SANITIZE_ADDRESS__`, since, while `memcmp` is UB when you don't call it with fully valid buffers[1](#user-content-fn-1-2b8d102f01b61a4a0d055974745ca21d), calling `lexicographical_compare(i1, unreachable_sentinel, i2, unreachable_sentinel)` is valid (at least, that's my understanding; I may fully be wrong about that and this is a test of an additional guarantee we provide).
> 
> ## Footnotes
> 1. https://trust-in-soft.com/blog/2015/12/21/memcmp-requires-pointers-to-fully-valid-buffers/ [↩](#user-content-fnref-1-2b8d102f01b61a4a0d055974745ca21d)

Yeah, it's hard to argue in support of `memcmp(x, y, -1)`. The initial motivation was "it works, and doing so means we don't need a special-case check for the case where both ranges are 'infinite'". But given how questionable this is in the first place and how quickly we've found a case where it doesn't "work", I should just add the special-case check and be done with it.

CaseyCarter:
Okay, fixed. Thanks for reminding me how icky this was in the first place.

strega-nil-ms (approved):

CaseyCarter:

PR #3990, opened 2023-08-23
Don't violate preconditions in tr1/regex3

CaseyCarter:
This test constructs a `regex_iterator` and a `regex_token_iterator` with invalid character ranges. The first occurrence is benign - the test never directs the library to examine the bogus part of the range - but the second case ventures into the land of pure undefined behavior. Caught by ASan.

Fixes VSO-1854241 / AB#1854241


strega-nil-ms (approved):

StephanTLavavej (approved):

StephanTLavavej:
I have no idea what this code was thinking - great catch.

PR #3991, opened 2023-08-23
Tell `P0881R7_stacktrace` that we `HAS_DEBUG_INFO`

CaseyCarter:
... when building with ASan enabled.

Fixes VSO-1854234 / AB#1854234

### DIFF
--- a/tests/std/tests/P0881R7_stacktrace/test.cpp
+++ b/tests/std/tests/P0881R7_stacktrace/test.cpp
@@ -156,6 +156,11 @@ string to_string_using_to_string(const stacktrace& st) {
     return to_string(st) + "\n";
 }
 
+#if !defined(HAS_DEBUG_INFO) && defined(__SANITIZE_ADDRESS__)
+// We always use /Zi with /fsanitize=address
+#define HAS_DEBUG_INFO 1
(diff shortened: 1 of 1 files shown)


AlexGuteniev:
Why is this needed?

strega-nil-ms:
I also do not understand the relationship between the bug and this code.

CaseyCarter:
I've been trying to split out small changes so I don't have to merge a single huge monolothic change, but this one doesn't stand alone. If you imagine that `usual_latest_matrix.lst` adds some new lines with `/fsanitize=address /Zi`, which then hits the local `env.lst` for this test:
```
# Copyright (c) Microsoft Corporation.

RUNALL_INCLUDE ..\usual_latest_matrix.lst
RUNALL_CROSSLIST
PM_CL="/Zi /DHAS_DEBUG_INFO" PM_LINK="/debug"
PM_CL="/DHAS_EXPORT"
PM_CL=""
```
you'll see that some of the final configurations in the resulting cross product have `/fsanitize=address /Zi` but not `/DHAS_DEBUG_INFO`. Those configurations will fail. It's easier to teach the test that `defined(__SANITIZE_ADDRESS__)` implies `HAS_DEBUG_INFO` than it is to expand the cross product and pass `/DHAS_DEBUG_INFO` everywhere explicitly.

I'll just close this for now and bring it as part of the larger change.

PR #3993, opened 2023-08-26
Various bugfixes / enhancements for `<flat_set>`

achabense:
Works towards #2912.

Replace #3985; the commits are reorganized for ease of review.
Major changes are:
1. Fix usage of `_Compressed_pair`; the implementation was erroneously trying to compress the container instead of `key_compare`.
2. Implement LWG-3884 (`flat_meow(flat_meow&&, al)`).
3. Implement missing `insert(initializer_list)`, and make `operator=(initializer_list)` available for users of `flat_meow`.
4. Fix `emplace[_hint]` and `insert([hint,]val)` functions.
4.1. Some functions couldn't compile, including `insert(const key&)` and hint functions.
4.2. Some function signatures were non-conformant.
4.3. The bound searching algorithm in `_Emplace_hint` was wrong. `flat_set` must find `lower_bound` for correctness, and `flat_multiset` should find `upper_bound` for efficiency and conformance to the standard.
5. Various other cleanups and enhancements.

Thanks to @duanqn and @frederick-vs-ja for previous review on #3985 😽

### DIFF
--- a/stl/inc/flat_set
+++ b/stl/inc/flat_set
@@ -211,31 +211,33 @@ public:
     // modifiers
     template <class... _Args>
     auto emplace(_Args&&... _Vals) {
-        insert<_Kty>(_Kty{_STD forward<_Args>(_Vals)...});
+        constexpr bool _Is_key_type = _In_place_key_extract_set<_Kty, remove_cvref_t<_Args>...>::_Extractable;
+        if constexpr (_Is_key_type) {
+            return _Emplace(_STD forward<_Args>(_Vals)...);
+        } else {
+            return _Emplace(_Kty{_STD forward<_Args>(_Vals)...});
+        }
     }
-
     template <class... _Args>
     iterator emplace_hint(const_iterator _Hint, _Args&&... _Vals) {
-        return _Emplace_hint(_Hint, _Kty{_STD forward<_Args>(_Vals)...});
+        constexpr bool _Is_key_type = _In_place_key_extract_set<_Kty, remove_cvref_t<_Args>...>::_Extractable;
+        if constexpr (_Is_key_type) {
+            return _Emplace_hint(_Hint, _STD forward<_Args>(_Vals)...);
+        } else {
+            return _Emplace_hint(_Hint, _Kty{_STD forward<_Args>(_Vals)...});
+        }
     }
 
     auto insert(const value_type& _Val) {
-        return _Insert<_Kty>(_Val);
+        return _Emplace(_Val);
     }
     auto insert(value_type&& _Val) {
-        return _Insert(_STD move(_Val));
+        return _Emplace(_STD move(_Val));
     }
-
     template <class _Other>
         requires (!_Multi && _Keylt_transparent && is_constructible_v<_Kty, _Other>)
     auto insert(_Other&& _Val) {
-        return _Insert(_STD forward<_Other>(_Val));
-    }
-
-    template <class _Other>
-        requires (!_Multi && _Keylt_transparent && is_constructible_v<_Kty, _Other>)
-    iterator insert(const_iterator _Hint, _Other&& _Val) {
-        return _Emplace_hint(_Hint, _STD forward<_Other>(_Val));
+        return _Emplace(_STD forward<_Other>(_Val));
     }
 
     iterator insert(const_iterator _Hint, const value_type& _Val) {
@@ -244,22 +246,32 @@ public:
     iterator insert(const_iterator _Hint, value_type&& _Val) {
         return _Emplace_hint(_Hint, _STD move(_Val));
     }
-
-    template <input_iterator _Iter>
-    void insert(const _Iter& _First, const _Iter& _Last) {
-        _Insert_range<false>(_First, _Last);
+    template <class _Other>
+        requires (!_Multi && _Keylt_transparent && is_constructible_v<_Kty, _Other>)
+    iterator insert(const_iterator _Hint, _Other&& _Val) {
+        return _Emplace_hint(_Hint, _STD forward<_Other>(_Val));
     }
 
+    template <input_iterator _Iter>
+    void insert(_Iter _First, _Iter _Last) {
+        _Insert_range<false>(_First, _Last);
+    }
     template <input_iterator _Iter>
     void insert(_Tsorted, _Iter _First, _Iter _Last) {
         _Insert_range<true>(_First, _Last);
     }
-
     template <_Container_compatible_range<_Kty> _Rng>
     void insert_range(_Rng&& _Range) {
         const size_type _Old_size = size();
         _Get_cont().append_range(_STD forward<_Rng>(_Range));
         _Restore_invariants_after_insert<false>(_Old_size);
+    }
+
+    void insert(initializer_list<_Kty> _Ilist) {
+        _Insert_range<false>(_Ilist.begin(), _Ilist.end());
+    }
+    void insert(_Tsorted, initializer_list<_Kty> _Ilist) {
+        _Insert_range<true>(_Ilist.begin(), _Ilist.end());
     }
 
     _NODISCARD container_type extract() && {
@@ -435,36 +447,84 @@ private:
     }
 
     template <class _Ty>
-        requires (_Keylt_transparent && is_constructible_v<_Kty, _Ty>) || is_same_v<_Ty, _Kty>
-    void _Emplace_hint(const_iterator _Where, _Ty&& _Val) {
-        _Container& _Cont     = _Get_cont();
-        _Keylt& _Compare      = _Get_comp();
-        const iterator _Begin = begin();
-        const iterator _End   = end();
-
-        if (_Where == _End || !_Compare(*_Where, _Val)) {
-            // _Val <= *_Where
-            // Left of _Where
-            if (_Where == _Begin || !_Compare(_Val, *(_Where - 1))) {
-                // _Val >= (*_Where - 1)
-                // Insert before _Where
+    auto _Emplace(_Ty&& _Val) {
+        _Container& _Cont = _Get_cont();
+        if constexpr (_Multi) {
+            _STL_INTERNAL_STATIC_ASSERT(is_same_v<remove_cvref_t<_Ty>, _Kty>);
+            return _Cont.emplace(upper_bound(_Val), _STD forward<_Ty>(_Val));
+        } else {
+            const iterator _End   = end();
+            const iterator _Where = lower_bound(_Val);
+            if (_Where != _End && _Keys_equal(*_Where, _Val)) {
+                return pair{_Where, false};
+            }
+            if constexpr (is_same_v<remove_cvref_t<_Ty>, _Kty>) {
+                return pair{_Cont.emplace(_Where, _STD forward<_Ty>(_Val)), true};
             } else {
-                // _Val < (*_Where - 1)
-                _Where = _STD upper_bound(_Begin, _Where, _Val, _Compare);
+                // flat_set::insert(auto&&)
+                _STL_INTERNAL_STATIC_ASSERT(_Keylt_transparent && is_constructible_v<_Kty, _Ty>);
+                _Kty _Keyval{_STD forward<_Ty>(_Val)};
+                _STL_ASSERT(lower_bound(_Keyval) == _Where && !_Keys_equal(_Keyval, *_Where),
+                    "find(val) was not equal to find(key_type{forward(val)})");
+                return pair{_Cont.emplace(_Where, _STD move(_Keyval)), true};
+            }
+        }
+    }
+
+    template <class _Ty>
+    iterator _Emplace_hint(const_iterator _Where, _Ty&& _Val) {
+        _Container& _Cont           = _Get_cont();
+        const key_compare& _Compare = _Get_comp();
+        const const_iterator _Begin = cbegin();
+        const const_iterator _End   = cend();
+
+        if constexpr (_Multi) {
+            // Find upper_bound for flat_multiset
+            if (_Where == _End || _Compare(_Val, *_Where)) {
+                // _Val < *_Where
+                if (_Where == _Begin || !_Compare(_Val, *(_Where - 1))) {
+                    // _Val >= *(_Where-1) ~ upper_bound is _Where
+                } else {
+                    // _Val < *(_Where-1) ~ upper_bound is in [_Begin,_Where-1]
+                    _Where = _STD upper_bound(_Begin, _Where - 1, _Val);
+                }
+            } else {
+                // _Val >= *_Where ~ upper_bound is in [_Where+1,_End]
+                _Where = _STD upper_bound(_Where + 1, _End, _Val);
             }
         } else {
-            // _Val > *_Where
-            // Right of _Where
-            _Where = _STD lower_bound(_Where + 1, _End, _Val, _Compare);
+            // Find lower_bound for flat_set
+            if (_Where == _End || !_Compare(*_Where, _Val)) {
+                // _Val <= *_Where
+                if (_Where == _Begin || _Compare(*(_Where - 1), _Val)) {
+                    // _Val > *(_Where-1) ~ lower_bound is _Where
+                } else {
+                    // _Val <= *(_Where-1) ~ lower_bound is in [_Begin,_Where-1]
+                    _Where = _STD lower_bound(_Begin, _Where - 1, _Val);
+                }
+            } else {
+                // _Val > *_Where ~ lower_bound is in [_Where+1,_End]
+                _Where = _STD lower_bound(_Where + 1, _End, _Val);
+            }
         }
 
         if constexpr (_Multi) {
-            return _Cont.insert(_Where, _STD forward<_Ty>(_Val));
+            _STL_INTERNAL_STATIC_ASSERT(is_same_v<remove_cvref_t<_Ty>, _Kty>);
+            return _Cont.emplace(_Where, _STD forward<_Ty>(_Val));
         } else {
-            if (_Where == _End || !_Keys_equal(_Val, *_Where)) {
-                return _Cont.insert(_Where, _STD forward<_Ty>(_Val));
+            if (_Where != _End && _Keys_equal(_Val, *_Where)) {
+                return _Cont.begin() + (_Where - _Begin);
             }
-            return _Where;
+            if constexpr (is_same_v<remove_cvref_t<_Ty>, _Kty>) {
+                return _Cont.emplace(_Where, _STD forward<_Ty>(_Val));
+            } else {
+                // flat_set::insert(hint,auto&&)
+                _STL_INTERNAL_STATIC_ASSERT(_Keylt_transparent && is_constructible_v<_Kty, _Ty>);
+                _Kty _Keyval{_STD forward<_Ty>(_Val)};
+                _STL_ASSERT(lower_bound(_Keyval) == _Where && !_Keys_equal(_Keyval, *_Where),
+                    "find(val) was not equal to find(key_type{forward(val)})");
+                return _Cont.emplace(_Where, _STD move(_Keyval));
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


tests/std/tests/P1222R4_flat_set/test.cpp
```diff
@@ -132,9 +119,97 @@ void test_constructors() {
     flat_set<int> a{};
     a = {1, 7, 7, 7, 2, 100, -1};
     assert_all_requirements_and_equals(a, {-1, 1, 2, 7, 100});
+    assert_all_requirements_and_equals(flat_set<int>(a, allocator<int>{}), {-1, 1, 2, 7, 100});
+    assert_all_requirements_and_equals(flat_set<int>(std::move(a), allocator<int>{}), {-1, 1, 2, 7, 100});
     flat_multiset<int> b{};
     b = {1, 7, 7, 7, 2, 100, -1};
```
achabense, line 125: This was passing because the a `flat_set` was constructed from `initializer-list` and then `operator=(flat_set&&)` was called. Now it's `operator=(initializer_list)`

stl/inc/flat_set
```diff
@@ -187,33 +208,36 @@ public:
...
+        if constexpr (_Is_key_type) {
+            return _Emplace_hint(_Hint, _STD forward<_Args>(_Vals)...);
+        } else {
+            return _Emplace_hint(_Hint, _Kty{_STD forward<_Args>(_Vals)...});
+        }
     }
 
     auto insert(const value_type& _Val) {
-        return _Insert<_Kty>(_Val);
+        return _Emplace(_Val);
```
achabense, line 232: This would fail to compile before the change, as `_Insert<_Kty>(_Val)` was actually `_Insert(_Kty&&)`.
```
flat_set<int> st;
const int a = 1;
st.insert(a);
```

tests/std/tests/P1222R4_flat_set/test.cpp
```diff
@@ -132,9 +119,97 @@ void test_constructors() {
...
+    assert_all_requirements_and_equals(flat_multiset<int>(std::move(b), allocator<int>{}), {-1, 1, 2, 7, 7, 7, 100});
+}
+
+template <class C>
+void test_insert_1() {
+    using lt = std::less<int>;
+
+    const int _dat[]{0, 1, 2};
+    const int* volatile _beg = _dat;
+    const int* volatile _end = std::end(_dat);
```
achabense, line 137: `volatile` is for `insert(iter, iter)` which was `insert(cons iter&, const iter&)` before the change.

stl/inc/flat_set
```diff
@@ -424,63 +447,95 @@ private:
...
+                // _Val < *_Where
+                if (_Where == _Begin || !_Compare(_Val, *(_Where - 1))) {
+                    // _Val >= *(_Where-1) ~ upper_bound is _Where
+                } else {
+                    // _Val < *(_Where-1) ~ upper_bound is in [_Begin,_Where-1]
+                    _Where = _STD upper_bound(_Begin, _Where - 1, _Val);
+                }
             } else {
-                // _Val < (*_Where - 1)
-                _Where = _STD upper_bound(_Begin, _Where, _Val, _Compare);
```
achabense, line 442: After making `_Emplace_hint` compilable with `emplace_hint` functions, I realized that this line was making the following test fail:
```
flat_set<int, lt, C> a{0, 5};
assert_all_requirements_and_equals(a, {0, 5});
a.emplace_hint(a.end());
assert_all_requirements_and_equals(a, {0, 5}); // < got {0, 0, 5}!
```

stl/inc/flat_set
```diff
@@ -424,63 +447,95 @@ private:
...
+                return pair{_Where, false};
+            }
+            if constexpr (is_same_v<remove_cvref_t<_Ty>, _Kty>) {
+                return pair{_Cont.emplace(_Where, _STD forward<_Ty>(_Val)), true};
+            } else {
+                // flat_set::insert(auto&&)
+                _STL_INTERNAL_STATIC_ASSERT(_Keylt_transparent && is_constructible_v<_Kty, _Ty>);
+                _Kty _Keyval{_STD forward<_Ty>(_Val)};
+                _STL_ASSERT(lower_bound(_Keyval) == _Where && !_Keys_equal(_Keyval, *_Where),
+                    "find(val) was not equal to find(key_type{forward(val)})");
```
achabense, line 468: This assertion is to avoid invariant breakage. It can capture errorneous usages like:
```cpp
std::flat_set<int, std::less<>> s{ 1,2,3,5,7 };
s.insert(2.5);
```
`Which is UB as find(2.5)->end, but find(int(2.5))->2`; However, I'm highly uncertain what this precondition implies to the implementation:
image
Is `flat_set::insert([hint,]auto&&)` really well-defined by the standard?

stl/inc/flat_set
```diff
@@ -403,18 +426,18 @@ public:
 
 private:
     void _Assert_after_sorted_input() const {
-        _STL_ASSERT(_STD is_sorted(begin(), end(), _Get_comp()), "Input was not sorted!");
+        _STL_ASSERT(_STD is_sorted(cbegin(), cend(), _Get_comp()), "Input was not sorted!");
         if constexpr (!_Multi) {
-            _STL_ASSERT(_Is_unique(), "Input was not unique!");
+            _STL_ASSERT(_Is_unique(), "Input was sorted but not unique!");
```
achabense, line 431: As a side note, if just for debug efficiency, `is_sorted && _Is_unique` can be replaced by checking `[0] lt [1] lt [2] ...` in a single pass.

stl/inc/flat_set
```diff
@@ -522,33 +577,22 @@ private:
...
         }
     }
 
     template <bool _Presorted>
-    void _Restore_invariants_after_insert(const size_type& _Old_size) {
-        key_compare& _Compare   = _Get_comp();
-        const iterator _Old_end = begin() + static_cast<difference_type>(_Old_size);
-        const iterator _New_end = end();
+    void _Restore_invariants_after_insert(const size_type _Old_size) {
+        const key_compare& _Compare = _Get_comp();
```
achabense, line 593: I made all `key_compare&` to `const key_compare&`, as I believe there shouldn't be any difference between `operator(...)&` and `operator(...)const&` (but I cannot find a formal citation for this constraint :(

tests/std/tests/P1222R4_flat_set/test.cpp
```diff
@@ -132,9 +119,97 @@ void test_constructors() {
...
+        flat_multiset<int, lt, C> a{0, 5};
+        assert_all_requirements_and_equals(a, {0, 5});
+        a.emplace_hint(a.end());
+        assert_all_requirements_and_equals(a, {0, 0, 5});
+        a.emplace_hint(a.end(), 0);
+        assert_all_requirements_and_equals(a, {0, 0, 0, 5});
+        a.insert(a.begin(), 6);
+        assert_all_requirements_and_equals(a, {0, 0, 0, 5, 6});
+        a.insert(a.begin(), _dat[1]);
+        assert_all_requirements_and_equals(a, {0, 0, 0, 1, 5, 6});
```
achabense, line 211: As a side note, `assert_all_requirements_and_equals` is making the test hard to debug, as it holds `assert`s which will always report fixed lines in itself. I think we should replace these function calls with something in the form of `assert(something is true)` to catch errors more easily.
https://github.com/microsoft/STL/blob/d6e6b5ac5d7e43b2210d98deb6cea20263abe2af/tests/std/tests/P1222R4_flat_set/test.cpp#L69-L76

stl/inc/flat_set
```diff
@@ -506,8 +561,8 @@ private:
     template <class _Other>
         requires _Keylt_transparent || is_same_v<_Other, _Kty>
     _NODISCARD const_iterator _Find(const _Other& _Val) const {
-        const iterator _End   = end();
-        const iterator _Where = lower_bound(_Val);
+        const const_iterator _End   = cend();
+        const const_iterator _Where = lower_bound(_Val);
         if (_Where != _End && _Keys_equal(*_Where, _Val)) {
```
achabense, line 566: As a side note, most of usages of `_Keys_equal` can be simplified to a single comparision, as it is used after `lower_bound`. However I think it's better not to include such changes in this pr.

stl/inc/flat_set
```diff
@@ -164,6 +183,7 @@ public:
     _NODISCARD const_reverse_iterator rend() const noexcept {
```
achabense, line 183: For `rbegin()`, I think it's not necessary to call `_Get_cont().rend()`, as it is required to be `reverse_iterator(begin())`, and the container does not necessarily offer `rbegin()` method (am I mistaken?). The same goes for `rend(), crbegin() and crend()`.

achabense:
Also, I'm having a sense that a lot of functions are still not rubost against exceptions. We may need more usages  `_Clear_scope_guard` here and there. (I think we need a single issue/pr for this problem.)

stl/inc/flat_set
achabense: For example, can I assume that `lower_bound(val)==lower_bound(keyval)`? If so, how is this guaranteed? if not, then how should `hint` be used?

achabense: Also, for the first overload `insert(K&& x)`, the `bool` cannot indicate whether `x` is moved if it was rvalue, as it's unclear whether the `emplace` must success.

stl/inc/flat_set
frederick-vs-ja: > the container does not necessarily offer `rbegin()` method

No, `rbegin` must be provided since `flat_set`/`flat_multiset` requires the adapted container to be a reversible container ([[flat.set.overview]/2](https://eel.is/c++draft/flat.set.overview#2), [[flat.multiset.overview]/2](https://eel.is/c++draft/flat.multiset.overview#2)).

achabense: However, that's the requirements for `flat_meow`; the overview also says `"Any sequence container supporting Cpp17RandomAccessIterator can be used to instantiate flat_meow"` ([ref](https://eel.is/c++draft/flat.multiset.overview#7))

stl/inc/flat_set
```diff
@@ -424,63 +447,95 @@ private:
...
+            if (_Where != _End && _Keys_equal(_Val, *_Where)) {
+                return _Cont.begin() + (_Where - _Begin);
+            }
+            if constexpr (is_same_v<remove_cvref_t<_Ty>, _Kty>) {
+                return _Cont.emplace(_Where, _STD forward<_Ty>(_Val));
+            } else {
+                // flat_set::insert(hint,auto&&)
+                _STL_INTERNAL_STATIC_ASSERT(_Keylt_transparent && is_constructible_v<_Kty, _Ty>);
+                _Kty _Keyval{_STD forward<_Ty>(_Val)};
+                _STL_ASSERT(lower_bound(_Keyval) == _Where && !_Keys_equal(_Keyval, *_Where),
```
achabense, line 524: (Changes needed) I find this check is still problematic, as `_Where` can be `_End` so `*_Where` shouldn't be used..

(Also, `_Insert` and `_Insert_hint` seem to be more suitable names)

stl/inc/flat_set
achabense: My latest understanding is that, `less<>` is not supposed to be used as comparators for `flat_meow`, as it's too general. The acceptable types should be limited to a set of pre-defined types, and should be compared as if comparing the same thing (for example, there are `foo::key`, `bar::key` and `baz::key`, and the transparent comparator should compare them as if by comparing the `key`.) So `lower_bound(val)==lower_bound(key{forward(val)})` should really hold true, and `emplace` is supposed to succeed after finding no equivalent keys.

stl/inc/flat_set
frederick-vs-ja: Oh. I was just confused. The requirement on reversible containers is always additionally specified and not implied by other requirements.

tests/std/tests/P1222R4_flat_set/test.cpp
```diff
@@ -65,10 +65,52 @@ void assert_reversible_container_requirements(const T& s) {
     static_assert(is_convertible_v<typename T::reverse_iterator, typename T::const_reverse_iterator>);
 }
 
+template <class T>
+void test_ebco() {
+    using container_type = T::container_type;
+    if constexpr (is_empty_v<typename T::key_compare>) {
```
achabense, line 71: image
`_Compressed_pair`'s ebco is limited by final-ness; however I think it ok not to add `is_final_v` here, as long as we don't bother testing `final` comparators.

stl/inc/flat_set
```diff
@@ -187,33 +208,36 @@ public:
...
 
+    // modifiers
     template <class... _Args>
     auto emplace(_Args&&... _Vals) {
-        insert<_Kty>(_Kty{_STD forward<_Args>(_Vals)...});
+        constexpr bool _Is_key_type = _In_place_key_extract_set<_Kty, remove_cvref_t<_Args>...>::_Extractable;
+        if constexpr (_Is_key_type) {
+            return _Emplace(_STD forward<_Args>(_Vals)...);
+        } else {
+            return _Emplace(_Kty{_STD forward<_Args>(_Vals)...});
```
achabense, line 218: Should this be `_Kty{_STD forward<_Args>(_Vals)...}` or `_Kty(_STD forward<_Args>(_Vals)...)`? (There are currenctly 4 occurrences of `_Kty{...}` in `emplace, emplace_hint, _Emplace and _Emplace_hint`)

frederick-vs-ja: We should use `_Kty(_STD forward<_Args>(_Vals)...)` (direct-non-list-initialization).

achabense:
Here are some more commits I'm hesitant to push:
1. https://github.com/microsoft/STL/commit/312c496ab22efbf08808fc3103c23a15ed2cd240 More conservative `insert(auto&&)` that allows cases described in [this comment](https://github.com/microsoft/STL/pull/3993#issuecomment-1699066181). 

2. https://github.com/microsoft/STL/commit/68bcb91ccf4e6eab9affd71d9dd223e5e43d7559 Support `rbegin/...` even if the container doesn't support them. (or, should we use `_RANGES rbegin/...` instead?)

3. https://github.com/microsoft/STL/commit/df36cb77ba9bd81dcbdf8debcd2698e345644082 convert some `_Keys_equal` to a single comparision. (There are more candidates, but these are the most obvious ones. Only `_Erase_dupes_if_needed` actually needs `_Key_equal`.)

achabense:
Here are some more commits I'm hesitant to push:
1. https://github.com/microsoft/STL/commit/312c496ab22efbf08808fc3103c23a15ed2cd240 More conservative `insert(auto&&)` that allows cases shown above.  (**It seems that, for cases where `lower_bound(val) cannot be always be lower_bound(key(val))` (like the above case), the time complexity of `insert(hint,auto&&)` cannot be `amortized constant`, as at least one hintless `lower_bound` search is required, either to decide `?contains(val)` or to find `lower_bound(key(val))` for insertion position.**)

2. https://github.com/microsoft/STL/commit/68bcb91ccf4e6eab9affd71d9dd223e5e43d7559 Support `rbegin/...` even if the container doesn't support them. (or, should we use `_RANGES rbegin/...` instead?)

3. https://github.com/microsoft/STL/commit/df36cb77ba9bd81dcbdf8debcd2698e345644082 convert some `_Keys_equal` to a single comparision. (There are more candidates, but these are the most obvious ones. Only `_Erase_dupes_if_needed` actually needs `_Key_equal`.)

stl/inc/flat_set
```diff
@@ -187,33 +208,36 @@ public:
         return _Get_cont().max_size();
     }
 
+    // modifiers
     template <class... _Args>
     auto emplace(_Args&&... _Vals) {
```
achabense, line 213: Nitpick: I thinks `_Vals` is slightly confusing. Better be `(_Types&&... _Args)`.

stl/inc/flat_set
```diff
@@ -187,33 +208,36 @@ public:
...
 
+    // modifiers
     template <class... _Args>
     auto emplace(_Args&&... _Vals) {
-        insert<_Kty>(_Kty{_STD forward<_Args>(_Vals)...});
+        constexpr bool _Is_key_type = _In_place_key_extract_set<_Kty, remove_cvref_t<_Args>...>::_Extractable;
+        if constexpr (_Is_key_type) {
+            return _Emplace(_STD forward<_Args>(_Vals)...);
+        } else {
+            return _Emplace(_Kty(_STD forward<_Args>(_Vals)...));
```
achabense, line 218: As `_Emplace[_hint]` does single-element insertion, I'm not sure whether it's better to name as `_Insert[_hint]`.

stl/inc/flat_set
```diff
@@ -423,64 +446,112 @@ private:
...
+    }
+
     template <class _Ty>
-        requires (_Keylt_transparent && is_constructible_v<_Kty, _Ty>) || is_same_v<_Ty, _Kty>
-    void _Emplace_hint(const_iterator _Where, _Ty&& _Val) {
-        _Container& _Cont     = _Get_cont();
-        _Keylt& _Compare      = _Get_comp();
-        const iterator _Begin = begin();
-        const iterator _End   = end();
+    auto _Emplace(_Ty&& _Val) {
```
achabense, line 466: From the call sites it might be better to name as `_Insert[_hint]`, as the function does single-element insertion. (though `_Emplace[_hint]` might be ok as the function uses `container.emplace` for performance guarantee...)

stl/inc/flat_set
```diff
@@ -423,64 +446,112 @@ private:
...
         if constexpr (_Multi) {
-            return _Cont.insert(_Where, _STD forward<_Ty>(_Val));
+            _STL_INTERNAL_STATIC_ASSERT(is_same_v<remove_cvref_t<_Ty>, _Kty>);
+            _STL_INTERNAL_CHECK(_Check_where(_Where, _Val));
+            return _Cont.emplace(_Where, _STD forward<_Ty>(_Val));
         } else {
-            if (_Where == _End || !_Keys_equal(_Val, *_Where)) {
-                return _Cont.insert(_Where, _STD forward<_Ty>(_Val));
+            if (_Where != _End && _Keys_equal(_Val, *_Where)) {
+                return _Cont.begin() + (_Where - _Begin);
```
achabense, line 532: (not planned in this pr) This could be `_Make_iter` for STD containers.

tests/std/tests/P1222R4_flat_set/test.cpp
```diff
@@ -132,9 +131,95 @@ void test_constructors() {
     flat_set<int> a{};
     a = {1, 7, 7, 7, 2, 100, -1};
     assert_all_requirements_and_equals(a, {-1, 1, 2, 7, 100});
+    assert_all_requirements_and_equals(flat_set<int>(a, allocator<int>{}), {-1, 1, 2, 7, 100});
```
StephanTLavavej, line 134: Should include `<memory>` for `allocator`.

stl/inc/flat_set
```diff
@@ -571,29 +631,35 @@ private:
...
-        _Sort_potentially_sorted(_Begin, _End);
-        _STL_INTERNAL_CHECK(_STD is_sorted(begin(), end(), _Get_comp()));
+        // O(N) if already sorted.
+        const key_compare& _Compare    = _Get_comp();
+        const iterator _Begin_unsorted = _STD is_sorted_until(_Begin, _End, _Compare);
+
+        _STD sort(_Begin_unsorted, _End, _Compare);
+        _STD inplace_merge(begin(), _Begin_unsorted, _End, _Compare);
+
+        _STL_INTERNAL_CHECK(_STD is_sorted(_Begin, _End, _Compare));
```
StephanTLavavej, line 641: After the code fusion here, we can pass `_Begin` instead of `begin()` to `inplace_merge()`.

stl/inc/flat_set
```diff
@@ -423,64 +446,112 @@ private:
...
+        if constexpr (_Multi) {
+            _STL_INTERNAL_STATIC_ASSERT(is_same_v<remove_cvref_t<_Ty>, _Kty>);
+            return _Cont.emplace(upper_bound(_Val), _STD forward<_Ty>(_Val));
+        } else {
+            const iterator _End   = end();
+            const iterator _Where = lower_bound(_Val);
+            if (_Where != _End && _Keys_equal(*_Where, _Val)) {
+                return pair{_Where, false};
+            }
+            if constexpr (is_same_v<remove_cvref_t<_Ty>, _Kty>) {
```
StephanTLavavej, line 477: Style: We conventionally put newlines between "non-chained" `if` statements (including `if constexpr`) as this helps us see where control flow happpens.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! I pushed changes for minor nitpicks and will merge after the tests pass.

I noticed one mostly-pre-existing issue - there's code that's moving from a container's comparator, but this has the potential to break container invariants. Notably, moving from a `std::function` can do this (the moved-from `std::function` is left in a valid-but-unspecified state, and in our implementation we empty it out).

IIRC implementations vary on how they handle this case (which the Standard really should provide better clarity on). In the associative containers and container adaptors, we've chosen to copy from comparators to avoid this problem. It seems that `flat_meow` should do the same as `priority_queue` etc.

StephanTLavavej:
:pancakes: :smile_cat: :tada:
===

PR #3994, opened 2023-08-26
P2675R1 `<format>` generator converted from C++ to Python

### LINKED ISSUE #3908: Convert P2675R1 `<format>` generator from C++ to Python
Followup to #3903 implementing #3446.

This PR is introducing a generator `tools/unicode_properties_parse/format_width_estimate_intervals.cpp`, cited in `<format>`, that processes a Unicode file `EastAsianWidth.txt`. Our other `<format>` generators are all Python, and this one should be too for consistency.

fsb4000:
Fixes #3908


achabense:
This line should be changed too:
https://github.com/microsoft/STL/blob/60f18856c56389001c38a9929ef37bd5c01c4e47/stl/inc/format#L982

tools/unicode_properties_parse/format_width_estimate_intervals.py
```diff
@@ -0,0 +1,184 @@
...
+    return table
+
+
+def main():
+    print("Old table:")
+    old_table = get_table_cpp20()
+    old_table.print_intervals()
+
+    print("\nNew table:\nInput path for EastAsianWidth.txt: ", end="")
+    path = input()
```
barcharcraz, line 172: we can just load this like the other generators (I forget if they use parameters or just load from next to the script)

tools/unicode_properties_parse/format_width_estimate_intervals.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def fill_range(self, rng: tuple, width: bool):
+        from_, to_ = rng[0], rng[1]
+        assert from_ <= to_, "impl assertion failed"
+        assert to_ <= self.max_u, "impl assertion failed"
+        self.table[from_ : to_ + 1] = [width] * (to_ - from_ + 1)
+
+    def print_intervals(self):
+        # Print table for _Width_estimate_intervals_v2.
+        c = 0
```
strega-nil-ms, line 32: it'd be nice if this were a more descriptive name; no change requested though since it was pre-existing (i feel this way generally about this file)

tools/unicode_properties_parse/format_width_estimate_intervals.py
strega-nil-ms, line 1: This is certainly pre-existing, but I think it'd be nice if the ranges were made to be open at the end; there are a ton of `+ 1`s that seem kind of unnecessary to me if we just made `max_u` and the range ends to be past-the-end rather than at-the-end.

tools/unicode_properties_parse/format_width_estimate_intervals.py
```diff
@@ -0,0 +1,174 @@
...
+            assert match, "invalid line"
+            from_val = int(match.group(1), base=16)
+            width = get_width(match.group(3))
+            if match.group(2):
+                # range (HEX..HEX)
+                to_val = int(match.group(2)[2:], base=16)
+                table.fill_range((from_val, to_val), width)
+            else:
+                # single character (HEX)
+                table.fill_range((from_val, from_val), width)
```
strega-nil-ms, line 135: pre-existing, so no change requested, but I'd prefer if there was a specific "just assign a single value"

tools/unicode_properties_parse/format_width_estimate_intervals.py
achabense: About `max_u`: It is the max valid Unicode code point, and does not directly mean the length of table. If we need a value for past-the-end of the table, we should not change its value directly. Instead, we need to introduce a new name based on it.

About `range_u`: The data file and the standard uses closed end. For example `0000..001F` means `[0000,001F]`. We should be careful about the conversion if we choose to use open end for `range_u`.

tools/unicode_properties_parse/format_width_estimate_intervals.cpp
```diff
@@ -1,209 +0,0 @@
...
-                if (++c == 12) {
-                    c = 0;
-                    cout << endl;
-                }
-            }
-            last = table[u];
-        }
-        cout << endl;
-    }
-
```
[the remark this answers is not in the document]
achabense: Aside from comments for the functions, for `print_intervals`, we'd better:
1. assert that `width[0]` is 1.
2. use index-based loop, starts from [1], and replace `last` with direct `if [I]!=[I-1] then print I` logic.

tools/unicode_properties_parse/format_width_estimate_intervals.py
```diff
@@ -4,57 +4,73 @@
...
-        c = 0
-        last = self.table[0]
-        for u, el in enumerate(self.table):
-            if el != last:
+        """
+        Generates _Width_estimate_intervals_v2.
+        That is, starting from the second code point ([1])
+        until and including the last code point([MAX_UNICODE_POINT]),
+        for code point [U], if width[U]!=width[U-1], we print I to indicate the new range.
+        """
```
achabense, line 40: Emm, I was explaining the intent of these functions, they are not polished enough to serve as comments directly :|

fsb4000: >I'll try to think of a more accurate version, please give me some time

Sure. We can change the comments more.

achabense:
I've made a review pr for this branch https://github.com/fsb4000/STL/pull/1 please have a look :)

tools/unicode_properties_parse/format_width_estimate_intervals.py
```diff
@@ -0,0 +1,181 @@
...
+            return UnicodeWidth.IS_2
+        else:
+            assert str == "A" or str == "H" or str == "N" or str == "Na"
+            return UnicodeWidth.IS_1
+
+    for line in source:
+        line = line.strip()
+        if line and not line.startswith("#"):
+            match = LINE_REGEX.fullmatch(line)
+            assert match, line # invalid line
```
achabense, line 133: I find that the regex has been slightly outdated due to some added whitespaces in  recently-updated `EastAsianWidth-15.1.0.txt` (in https://www.unicode.org/Public/UCD/latest/ucd/EastAsianWidth.txt))

(the regex still works for 15.0.0)

image
(Was `0000..001F;N     # Cc    [32] <control-0000>..<control-001F>
`,
|now `0000..001F     ; N  # Cc    [32] <control-0000>..<control-001F>`)

achabense:
I've made another pr for this branch. Hope it helps.

StephanTLavavej (approved):

StephanTLavavej:
I've pushed a merge with `main` followed by:

* Print the filename, timestamp, and C++ array.
* Don't bother carefully wrapping; we need to clang-format anyways.
* Use `join` to avoid a trailing comma interfering with clang-format.
* Use a template string to make the C++ output clearer, like the other generators.
* Format with Prettier.
* Regenerate, capturing the filename and timestamp.
* Cite the Standardese.

Because this table is changing over time, it's important to capture the filename/timestamp (with version info) so we can easily see when the table was last regenerated. The other changes allow the output to be copy-pasted into C++ and then clang-formatted, which is how the other generators behave too.

tools/unicode_properties_parse/format_width_estimate_intervals.py
```diff
@@ -0,0 +1,199 @@
...
+        assert from_ <= to_, "invalid range"
+        assert to_ <= self.MAX_CODE_POINT, "invalid range"
+        self.table[from_ : to_ + 1] = [width] * (to_ - from_ + 1)
+
+    def width_estimate_intervals(self):
+        """
+        Divide [0..MAX_CODE_POINT] into ranges with different width estimations.
+        Represent each range with their starting values.
+        The starting value of the first range is always 0 and omitted.
+        The width estimation should be 1 for the first range, then alternate between 2 and 1.
```
achabense, line 38: These are important maintenance information I want to keep, but I think they look strange and definitely need some rewordings :( .

strega-nil-ms, line 38:
```suggestion
        Creates a string representation of the map (in `self.table`) from
        unicode code points to their width, using hexadecimal unsigned integer literals.
        Since there are long runs of code points of one width or the other,
        this representation is a list of code points where the width switches.
        Additionally, the width is assumed to start at `1` from the beginning of the list.
        For example, `[1, 1, 2, 2, 2, 1]` would be represented as `"0x2u, 0x5u"`.
```

How do you feel about this?

achabense: Looks good to me.

strega-nil-ms (approved):

strega-nil-ms:
@StephanTLavavej, I've pushed changes (to change a comment explanation) after you approved.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:book: :zap: :snake:
===

PR #4008, opened 2023-09-05
ASAN pipeline testing

CaseyCarter:
This isn't ready to merge yet - I need to squash and write up a more detailed commit message - I just want to test that the non-ASAN test pipeline continues to work with the same test coverage.


CaseyCarter:
Ah, I see:
```
pr:
  drafts: false
```
apparently overrides even manual pipeline triggering. Guess I need to publish this PR.

CaseyCarter:
> Azure Pipelines failed to run 1 pipeline(s).

I suppose that's sort-of an improvement =)

CaseyCarter:
There we go: 108,564 tests passed, exactly as was the case before.

PR #4017, opened 2023-09-08
Fix `deque`'s usage of `_Allocate_at_least_helper`

achabense:
`deque::_Mapsize()` should always be power of 2. Therefore, in `deque::_Growmap`, `_Allocate_at_least_helper` should not have chance to change `_Newsize` directly.

### DIFF
--- a/stl/inc/deque
+++ b/stl/inc/deque
@@ -1557,9 +1557,15 @@ private:
             _Newsize *= 2;
         }
 
+        size_type _Allocsize = _Newsize;
+
         size_type _Myboff = _Myoff() / _Block_size;
-        _Mapptr _Newmap   = _Allocate_at_least_helper(_Almap, _Newsize);
+        _Mapptr _Newmap   = _Allocate_at_least_helper(_Almap, _Allocsize);
         _Mapptr _Myptr    = _Newmap + _Myboff;
+        _STL_ASSERT(_Allocsize >= _Newsize, "_Allocsize >= _Newsize");
+        while (_Allocsize / 2 >= _Newsize) {
+            _Newsize *= 2;
+        }
 
         _Count = _Newsize - _Mapsize();
 
--- a/stl/inc/deque
+++ b/stl/inc/deque
@@ -1563,7 +1563,7 @@ private:
         _Mapptr _Newmap   = _Allocate_at_least_helper(_Almap, _Allocsize);
         _Mapptr _Myptr    = _Newmap + _Myboff;
         _STL_ASSERT(_Allocsize >= _Newsize, "_Allocsize >= _Newsize");
-        while (_Allocsize / 2 >= _Newsize) {
+        while (_Newsize <= _Allocsize / 2) {
             _Newsize *= 2;
         }
 
--- a/tests/std/tests/GH_003570_allocate_at_least/test.cpp
+++ b/tests/std/tests/GH_003570_allocate_at_least/test.cpp
@@ -51,7 +51,7 @@ struct signalling_allocator {
 
     allocation_result<T*> allocate_at_least(size_t count) {
         allocate_at_least_signal.set();
-        return {allocate(count * 2), count * 2};
+        return {allocate(count * 2 + 1), count * 2 + 1};
     }
 
     void deallocate(T* ptr, size_t) noexcept {
@@ -72,10 +72,15 @@ void test_container() {
 }
 
 void test_deque() {
-    deque<int, signalling_allocator<int>> d;
-    d.resize(100);
+    deque<size_t, signalling_allocator<size_t>> d;
+    for (size_t i = 0; i < 100; i++) {
+        d.push_back(i);
+    }
     assert(allocate_at_least_signal.consume());
     assert(d.size() == 100);
+    for (size_t i = 0; i < 100; i++) {
+        assert(d[i] == i);
+    }
 }
 
 void test_stream_overflow(auto& stream) {
--- a/tests/std/tests/GH_003570_allocate_at_least/test.cpp
+++ b/tests/std/tests/GH_003570_allocate_at_least/test.cpp
@@ -73,12 +73,12 @@ void test_container() {
 
 void test_deque() {
     deque<size_t, signalling_allocator<size_t>> d;
-    for (size_t i = 0; i < 100; i++) {
+    for (size_t i = 0; i < 100; ++i) {
         d.push_back(i);
     }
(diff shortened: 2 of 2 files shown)


tests/std/tests/GH_003570_allocate_at_least/test.cpp
```diff
@@ -51,7 +51,7 @@ struct signalling_allocator {
 
     allocation_result<T*> allocate_at_least(size_t count) {
         allocate_at_least_signal.set();
-        return {allocate(count * 2), count * 2};
+        return {allocate(count * 2 + 1), count * 2 + 1};
```
achabense, line 54: `count * 2` happens to work; the deque will break easily if users try to implement their own `allocate_at_least`.

strega-nil-ms (approved):

stl/inc/deque
```diff
@@ -1557,9 +1557,15 @@ private:
...
+        size_type _Allocsize = _Newsize;
+
         size_type _Myboff = _Myoff() / _Block_size;
-        _Mapptr _Newmap   = _Allocate_at_least_helper(_Almap, _Newsize);
+        _Mapptr _Newmap   = _Allocate_at_least_helper(_Almap, _Allocsize);
         _Mapptr _Myptr    = _Newmap + _Myboff;
+        _STL_ASSERT(_Allocsize >= _Newsize, "_Allocsize >= _Newsize");
+        while (_Allocsize / 2 >= _Newsize) {
+            _Newsize *= 2;
+        }
```
strega-nil-ms, line 1568: This seems inefficient to try and get the biggest power of two less than `_Allocsize`.

CaseyCarter: Do we have an `_Ugly` (i.e., usable in C++14 mode) version of `bit_floor`?

CaseyCarter:
Thanks much for separating the bugfix from the cleanup!

stl/inc/deque
achabense: I think it's ok to (temporarily) keep this while loop, as it's highly likely that `_Allocsize` is just equal to or marginally larger than `_Newsize` so the loop will skip immediatly or very short (And I'm going to do a series of cleanups for `<deque>` recently :) we can consider making its above while loop to a bit_ceil too )

achabense: I made a tiny cleanup & optimization for its above loop: https://github.com/microsoft/STL/commit/850d8aee1e8abadfff47f18726be6897c94dc5c4; I think this can partly compensate for the cost of new loop. Can I make a push to this pr?

achabense: [godbolt result](https://godbolt.org/z/E5fbM4Ysn); the codegen looks really better

achabense: hmm, I'm having difficulty deciding whether we can safely assume `_Minimum_map_size` will not cause overflow against `max_size()`.
The optimization may become "buggy" under some extreme (non practical) cases (for example, a `deque<char>` with an allocator whose `max_size()` is less than `128`.

And worse, the comparision against `_Alty::max_size` in current code looks wrong to me. I'm doubting we should use `_Almap::max_size` instead.

Due to these problems I'd suggest putting off that commit to future prs. (And in that link I've added some comments about the problems in detail.)

tests/std/tests/GH_003570_allocate_at_least/test.cpp
```diff
@@ -72,10 +72,15 @@ void test_container() {
 }
 
 void test_deque() {
-    deque<int, signalling_allocator<int>> d;
-    d.resize(100);
+    deque<size_t, signalling_allocator<size_t>> d;
+    for (size_t i = 0; i < 100; i++) {
```
StephanTLavavej, line 76: Style: We conventionally prefer preincrement, using postincrement only when it's necessary. Occurs below.

stl/inc/deque
```diff
@@ -1557,9 +1557,15 @@ private:
...
         }
 
+        size_type _Allocsize = _Newsize;
+
         size_type _Myboff = _Myoff() / _Block_size;
-        _Mapptr _Newmap   = _Allocate_at_least_helper(_Almap, _Newsize);
+        _Mapptr _Newmap   = _Allocate_at_least_helper(_Almap, _Allocsize);
         _Mapptr _Myptr    = _Newmap + _Myboff;
+        _STL_ASSERT(_Allocsize >= _Newsize, "_Allocsize >= _Newsize");
+        while (_Allocsize / 2 >= _Newsize) {
```
StephanTLavavej, line 1566: Style: In new code we try to avoid "Yoda conditions" where the varying thing is on the right. Here, `_Newsize` is varying, so it's easier to read a condition where it's on the left.

StephanTLavavej (approved):

StephanTLavavej:
@strega-nil-ms I pushed minor stylistic changes after you approved.

stl/inc/deque
```diff
@@ -1557,9 +1557,15 @@ private:
             _Newsize *= 2;
         }
 
+        size_type _Allocsize = _Newsize;
+
         size_type _Myboff = _Myoff() / _Block_size;
-        _Mapptr _Newmap   = _Allocate_at_least_helper(_Almap, _Newsize);
+        _Mapptr _Newmap   = _Allocate_at_least_helper(_Almap, _Allocsize);
         _Mapptr _Myptr    = _Newmap + _Myboff;
+        _STL_ASSERT(_Allocsize >= _Newsize, "_Allocsize >= _Newsize");
```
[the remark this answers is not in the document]
achabense: Hmm, also, is `_STL_ASSERT` a suitable choice here?

CaseyCarter: Use of `_STL_ASSERT`, but the check really belongs in `_Allocate_at_least_helper` with an error message like "your `allocate_at_least` is broken: it didn't allocate \"at least\"".

StephanTLavavej:
> I think we should put debug check in `_Allocate_at_least_helper` instead.

That would be reasonable.

> (And is `_STL_ASSERT` a suitable choice for such check?)

Yeah, for users violating preconditions.

achabense:
Thanks for the confirmation! For easier workflow I will address the change in another pr (or optionally, I think we can make a good-first-issue around this).

StephanTLavavej:

StephanTLavavej:
Thanks for noticing and fixing this regression before anyone else did! :eagle: :eye: :lady_beetle:

PR #4026, opened 2023-09-13
Silence a couple of `/Wall` warnings

StephanTLavavej:
Works towards DevCom-10368801 / VSO-1822115 "compiling `std.ixx` causes compiler warnings, even with `/external:anglebrackets` and `/external:W0`". (Additional changes are needed in the MSVC-internal repo: MSVC-PR-497556)

We don't attempt to be `/Wall` clean in general, but users are experiencing this as a regression due to the addition of the automatic build of the Standard Library Modules. Only a few warnings are being emitted so we can just clean them up. While this is motivated by the Modules, the changes are not modules-related.

I'm dealing with two warnings in this PR. First, I reported VSO-1885306 "warning C5039 shouldn't be emitted when an `extern "C"` function is marked `noexcept(false)`". `<stacktrace>` is correctly using `noexcept(false)`, so we should suppress this warning in the header. The warning is potentially useful elsewhere, so I don't want to globally suppress it for the STL.

The second change is to globally suppress warning C5220 for the STL in `<yvals_core.h>`. This is a classically obnoxious warning saying that the compiler is going to do what the Standard requires it to do, which we never need to be told. (It affects `<pplcancellation_token.h>` and `<ppltasks.h>`; my MSVC-internal changes will enhance those headers to use the `<yvals_core.h>` suppression machinery.)

### DIFF
--- a/stl/inc/stacktrace
+++ b/stl/inc/stacktrace
@@ -25,6 +25,11 @@ _STL_DISABLE_CLANG_WARNINGS
 #pragma push_macro("new")
 #undef new
 
+// TRANSITION, VSO-1885306: warning C5039 shouldn't be emitted when an extern "C" function is marked noexcept(false)
+#pragma warning(disable : 5039) // '__std_stacktrace_meow': pointer or reference to potentially throwing function
+                                // passed to 'extern "C"' function under -EHc. Undefined behavior may occur if this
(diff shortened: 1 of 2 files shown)


StephanTLavavej:

CaseyCarter (approved):

PR #4034, opened 2023-09-16
More bugfixes for `flat_set`

achabense:
This pr is separated from #4019 and contains all (and only) the bugfixes in it. Contents in this pr:
1. (Not strictly a bug) Fix `signed/unsigned` mismatch in `count` and `_Erase`.
2. Fix missed comp in `_Emplace_hint` (my mistake :(
3. Fix `erase(auto&&)`'s constraint. (Thanks to @cpplearner for sharing the link about [heterogeneous erasure](https://wg21.link/P2077R3#whyfwd)!

### DIFF
--- a/stl/inc/flat_set
+++ b/stl/inc/flat_set
@@ -294,9 +294,10 @@ public:
         return _Erase(_Val);
     }
     template <class _Other>
-        requires _Keylt_transparent
+        requires (
+            _Keylt_transparent && !is_convertible_v<_Other &&, iterator> && !is_convertible_v<_Other &&, iterator>)
     size_type erase(_Other&& _Val) {
-        return _Erase(_STD forward<_Other>(_Val));
+        return _Erase(_Val);
     }
     iterator erase(const_iterator _First, const_iterator _Last) {
         return _Get_cont().erase(_First, _Last);
@@ -556,7 +557,7 @@ private:
 
     template <class _Ty>
         requires _Keylt_transparent || is_same_v<_Ty, _Kty>
-    size_type _Erase(_Ty&& _Val) {
+    size_type _Erase(const _Ty& _Val) {
         const auto [_First, _Last] = equal_range(_Val);
 
         const auto _Removed = static_cast<size_type>(_Last - _First);
--- a/stl/inc/flat_set
+++ b/stl/inc/flat_set
@@ -502,11 +502,11 @@ private:
                     // _Val >= *(_Where-1) ~ upper_bound is _Where
                 } else {
                     // _Val < *(_Where-1) ~ upper_bound is in [_Begin,_Where-1]
-                    _Where = _STD upper_bound(_Begin, _Where - 1, _Val);
+                    _Where = _STD upper_bound(_Begin, _Where - 1, _Val, _Compare);
                 }
             } else {
                 // _Val >= *_Where ~ upper_bound is in [_Where+1,_End]
-                _Where = _STD upper_bound(_Where + 1, _End, _Val);
+                _Where = _STD upper_bound(_Where + 1, _End, _Val, _Compare);
             }
         } else {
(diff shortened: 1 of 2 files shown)


tests/std/tests/P1222R4_flat_set/test.cpp
```diff
@@ -288,6 +299,33 @@ void test_extract() {
     assert(ranges::equal(cont, elements));
 }
 
+// TRANSITION, too simple
+void test_erase_1() {
+    flat_set<int> fs{1};
+    fs.erase(1);
```
achabense, line 305: This was rejected by the compiler, as `_Erase(auto&&)` got `const key&` which was not same as `key`.

stl/inc/flat_set
```diff
@@ -294,9 +294,10 @@ public:
         return _Erase(_Val);
     }
     template <class _Other>
-        requires _Keylt_transparent
+        requires (
+            _Keylt_transparent && !is_convertible_v<_Other &&, iterator> && !is_convertible_v<_Other &&, iterator>)
     size_type erase(_Other&& _Val) {
-        return _Erase(_STD forward<_Other>(_Val));
+        return _Erase(_Val);
```
achabense, line 300: I believe it's safe to use `const auto&` in `_Erase`, as the `auto&&` in `erase` seems only to serve for constraint dispatching.

stl/inc/flat_set
```diff
@@ -556,10 +557,10 @@ private:
 
     template <class _Ty>
         requires _Keylt_transparent || is_same_v<_Ty, _Kty>
-    size_type _Erase(_Ty&& _Val) {
+    size_type _Erase(const _Ty& _Val) {
```
achabense, line 560: I believe it's safe to use `const auto&` in `_Erase`, as the `auto&&` in `erase` seems only to serve for constraint dispatching.

tests/std/tests/P1222R4_flat_set/test.cpp
```diff
@@ -288,6 +299,33 @@ void test_extract() {
...
+        return std::move(t);
+    }
+};
+
+void test_erase_2() {
+    using C = flat_set<int, std::less<>>;
+    C fs{1, 2, 3};
+    assert_all_requirements_and_equals(fs, {1, 2, 3});
+    // this should be allowed per P2077R3:
+    fs.erase(holder<C::iterator>{fs.begin()});
```
achabense, line 322: This is the case demonstrated in [`Why K&& rather than const K&`](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2077r3.html#whyfwd)

StephanTLavavej, line 322: Missing `const_iterator` coverage here.

tests/std/tests/P1222R4_flat_set/test.cpp
```diff
@@ -297,6 +335,15 @@ void test_erase_if() {
     assert(ranges::equal(fs, erased_result));
 }
 
+// TRANSITION, too simple
+void test_count() {
+    flat_set<int> fs{2};
+    assert(fs.count(1) == 0);
+
+    flat_multiset<int> fs2{1, 2, 2, 3};
+    assert(fs2.count(2) == 2);
```
StephanTLavavej, line 344: No change requested in this PR: It's not a huge deal, but in test code, I recommend "unique numbers for unique purposes". That is, if there's a `2` in the test code, ideally it should have the same meaning everywhere. Here, `2` is being used as an element value, and also as a number of elements. This test would be clearer if the elements were `10, 20, 20, 30`, for example.

StephanTLavavej (approved):

StephanTLavavej:
Thanks for finding and fixing these bugs! :lady_beetle:

PR #4037, opened 2023-09-20
`<mdspan>`: Fixes signed overflow in `_Fwd_prod_of_extents`

### LINKED ISSUE #4023: `<mdspan>`: `_Fwd_prod_of_extents::_Calculate` can potentially overflow when `mdspan::size`'s precondition is satisfied
In https://github.com/microsoft/STL/pull/3972#discussion_r1322286522, @CaseyCarter observed:

> `_Calculate` computes the product in `index_type`, which may overflow into UB land even for some values that are representable in `size_type`.

MattStephanson:
Fixes #4023.

### DIFF
--- a/stl/inc/mdspan
+++ b/stl/inc/mdspan
@@ -348,11 +348,13 @@ public:
         if constexpr (_Extents::rank() == 0) {
             return 1;
         } else {
-            typename _Extents::index_type _Result = 1;
+            // Use an unsigned type to prevent overflow when called from mdspan::size.
+            typename _Extents::size_type _Result = 1;
             for (size_t _Dim = 0; _Dim < _Idx; ++_Dim) {
-                _Result *= _Exts.extent(_Dim);
+                _Result *= static_cast<_Extents::size_type>(_Exts.extent(_Dim));
             }
-            return _Result;
+
+            return static_cast<_Extents::index_type>(_Result);
         }
     }
 };
--- a/tests/std/tests/GH_004023_mdspan_fwd_prod_overflow/test.compile.pass.cpp
+++ b/tests/std/tests/GH_004023_mdspan_fwd_prod_overflow/test.compile.pass.cpp
@@ -1,0 +1,80 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <cassert>
+#include <cstdint>
+#include <mdspan>
+#include <type_traits>
+
+using namespace std;
+
+struct layout_packed_upper { // LAPACK packed storage
+    template <class Extents>
+    class mapping {
+    public:
+        using extents_type = Extents;
+        using index_type   = typename extents_type::index_type;
+        using size_type    = typename extents_type::size_type;
+        using rank_type    = typename extents_type::rank_type;
+        using layout_type  = layout_packed_upper;
+
+        constexpr mapping() noexcept = default;
+        constexpr mapping(const extents_type& ex_) : ex{ex_} {}
+
+        constexpr const extents_type& extents() const {
+            return ex;
+        }
+
+        constexpr index_type required_span_size() const {
+            if (dim % 2 == 0) {
+                return (dim / 2) * (dim + 1);
+            } else {
+                return ((dim + 1) / 2) * dim;
+            }
+        }
+
+    private:
+        extents_type ex = extents_type();
+        index_type dim  = ex.extent(0);
+    };
+};
+
+int main() {
+    constexpr int32_t dim        = 47'000; // sqrt(2^32) > dim > sqrt(2^31)
+    constexpr auto expected_size = []() {
+        int32_t unused;
+        bool overflow = _Mul_overflow(dim, dim, unused);
+        assert(overflow);
+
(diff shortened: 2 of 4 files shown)


stl/inc/mdspan
```diff
@@ -348,11 +348,13 @@ public:
         if constexpr (_Extents::rank() == 0) {
             return 1;
         } else {
-            typename _Extents::index_type _Result = 1;
+            // Use an unsigned type to prevent overflow when called from mdspan::size.
+            typename _Extents::size_type _Result = 1;
             for (size_t _Dim = 0; _Dim < _Idx; ++_Dim) {
-                _Result *= _Exts.extent(_Dim);
+                _Result *= static_cast<_Extents::size_type>(_Exts.extent(_Dim));
```
MattStephanson, line 354: I think the `static_cast` here is redundant because it will be covered by [[expr.arith.conv]/p1.4.1](https://eel.is/c++draft/expr.arith.conv#1.4.1) or [p1.4.3.1](https://eel.is/c++draft/expr.arith.conv#1.4.3.1), either of which results in `size_type` being the common type.

stl/inc/mdspan
```diff
@@ -348,11 +348,13 @@ public:
...
-            typename _Extents::index_type _Result = 1;
+            // Use an unsigned type to prevent overflow when called from mdspan::size.
+            typename _Extents::size_type _Result = 1;
             for (size_t _Dim = 0; _Dim < _Idx; ++_Dim) {
-                _Result *= _Exts.extent(_Dim);
+                _Result *= static_cast<_Extents::size_type>(_Exts.extent(_Dim));
             }
-            return _Result;
+
+            return static_cast<_Extents::index_type>(_Result);
```
MattStephanson, line 357: This may result in a negative number if `index_type` is signed, but I think the only way that can happen without a precondition violation is in the call from `mdspan::size`. This is a well-defined conversion from C++20 on, and casting back to `size_type`, which `size()` does, recovers the original product.

This is the reason that the specialization below, for the `rank_dynamic() == 0` case, works as-is.

CaseyCarter (approved):

tests/std/tests/GH_004023_mdspan_fwd_prod_overflow/test.compile.pass.cpp
```diff
@@ -0,0 +1,79 @@
...
+using namespace std;
+
+struct layout_packed_upper { // LAPACK packed storage, VERY INCOMPLETE
+    template <class Extents>
+    class mapping {
+    public:
+        using extents_type = Extents;
+        using index_type   = typename extents_type::index_type;
+        using size_type    = typename extents_type::size_type;
+        using rank_type    = typename extents_type::rank_type;
```
StephanTLavavej, line 18: We can omit these `typename`s.

tests/std/tests/GH_004023_mdspan_fwd_prod_overflow/test.compile.pass.cpp
```diff
@@ -0,0 +1,79 @@
...
+    constexpr int32_t dim        = 47'000; // sqrt(2^32) > dim > sqrt(2^31), dim assumed even below
+    constexpr auto expected_size = [] {
+        int32_t unused;
+        bool overflow = _Mul_overflow(dim, dim, unused);
+        assert(overflow);
+
+        using UType     = remove_cv_t<make_unsigned_t<decltype(dim)>>;
+        const auto udim = static_cast<UType>(dim);
+        UType result{};
+        overflow = _Mul_overflow<UType>(udim, udim, result);
```
StephanTLavavej, line 52: The explicit template argument here is unnecessary, and the signature `(const _Int _Left, const _Int _Right, _Int& _Out)` leaves no room for doubt.

tests/std/tests/GH_004023_mdspan_fwd_prod_overflow/test.compile.pass.cpp
```diff
@@ -0,0 +1,79 @@
...
+        assert(!overflow);
+        return result;
+    }();
+
+    constexpr auto expected_req_size = [] {
+        int32_t result{};
+        bool overflow = _Mul_overflow(dim / 2, dim + 1, result);
```
StephanTLavavej, line 59: This `overflow` could be `const`.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! FYI @CaseyCarter I pushed trivial test changes after you approved.

CaseyCarter (approved):

tests/std/tests/GH_004023_mdspan_fwd_prod_overflow/test.compile.pass.cpp
MattStephanson: Ah, yes, I didn't notice this cleanup. Previously I was getting an "unable to deduce template parameter" error, but that was because I lacked the `remove_cv_t` so `UType` was `const uint_32`.

StephanTLavavej:

StephanTLavavej:
Thanks for fixing this bug before it shipped to users! :lady_beetle: :tada: :rocket:

PR #4039, opened 2023-09-20
Require VS 2022 17.8, remove workaround

StephanTLavavej:
The MSVC-internal toolset (i.e. the checked-in compiler binaries that build the compiler) has finally been updated, so we can begin requiring VS 2022 17.8, and we can remove a workaround for a compiler bug that was fixed in 17.7 Preview 1.


CaseyCarter (approved):

StephanTLavavej:

Issue #3273, opened 2022-12-07
SSO breaks backwards compatibility of ABI


RIscRIpt:
STL shipped with MSVC 14.34 introduced SSO (1ea95df). SSO implementation has changed `_STD::_String_val::_Bxty::_Bxty()` constructor: previously it zero-initialized `_Ptr` field (in MSVC 14.33), now it initializes `_Buf` field (in MSVC 14.34). SSO implementation relies on the fact that `_Buf` is initialized with zeroes, however this assumption breaks when `std::string` comes from TU compiled using older STL, where `_Ptr` field was initialized instead. When `std::string` in old TU is initialized using a small (>8 and <16) string literal, `std::string::c_str()` would return pointer to buffer which is not zero terminated properly.

**Command-line test case**

Compiling TU using MSVC 14.33 without SSO
```
//projects/msvc-sso-14_33-34_bug/lib_1433
❯ msvc-env x64 14.33

//projects/msvc-sso-14_33-34_bug/lib_1433
❯ type .\lib.cpp
#include <string>
static_assert(_MSC_VER == 1933);
std::string GetStringOld(const char* p) {
    return std::string(p);
}

//projects/msvc-sso-14_33-34_bug/lib_1433
❯ cl /c /O2 /EHsc lib.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.33.31631 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

lib.cpp
```

Compiling TU using MSVC 14.34 with SSO
```
//projects/msvc-sso-14_33-34_bug/exe_1434
❯ msvc-env x64 14.34

//projects/msvc-sso-14_33-34_bug/exe_1434
❯ type exe.cpp
#include <iostream>
#include <string>

static_assert(_MSC_VER == 1934);

std::string GetStringOld(const char* p);

__declspec(noinline) std::string GetStringNew(const char* p) {
    return std::string(p);
}

__declspec(noinline) void dirty_stack() {
    char trash[32];
    memset(trash, '@', sizeof(trash) - 1);
    trash[sizeof(trash) - 1] = '\0';
    std::cout << trash << '\n';
}

__declspec(noinline) void nobug_new() {
    std::cout << GetStringNew("01234567#$").c_str() << '\n';
}

__declspec(noinline) void bug_old() {
    std::cout << GetStringOld("01234567#$").c_str() << '\n';
}

int main() {
    dirty_stack();
    bug_old();
    dirty_stack();
[...]

PR #3235, opened 2022-11-16
Don't rely on SSO zeroing in `basic_string::_Construct`

CaseyCarter:
... or we will have horrible failures when linking with object files from 17.3. `_Construct` must null-terminate the string itself on all control paths. I've added a comment to the `_String_val::_Bxty` constructor directing us not to rely on the semantic changes introduced in 17.4 to avoid regressing.

PR #4047, opened 2023-09-22
Lifetime cleanups for `basic_string`

achabense:
Resolves the comments [(1)](https://github.com/microsoft/STL/pull/4031#discussion_r1331063123) [(2)](https://github.com/microsoft/STL/pull/4031#issuecomment-1727571761) in pr 4031. Hope that is the last lifetime bug of `basic_string` using fancy pointers (`3.1.`).
<hr>

Sometimes the pointer's destruction and buffer's activation are dangerously separated into different functions, in an obscure way (mainly relevant to the usage of `_Tidy_init`).
This pr introduces a helper function `_String_val::_Bxty::_Switch_to_buf` to make the transition state (where both `_Ptr` and `_Buf` are inactive) invisible to `basic_string`'s ctors and methods.
<hr>

Remaining issue after this pr:
https://github.com/microsoft/STL/blob/48eedd369d6c88130fff4d15afb88f016fa6e648/stl/inc/xstring#L2282-L2284
As `_String_val` already initializes the buffer on construction, and every switch back to small mode is now protected by the helper function, in `basic_string`'s ctor and methods, when we can decide a string is in small mode, we can also assume that elements in `_Buf` are activated. *Therefore, every remaining usage of `_Activate_SSO_buffer` (including the one in newly added `_Construct_empty`) becomes redundant*...
Or, `Don't rely on the new behavior`? Does this mean we should not assume that `_Buf` is initialized in ctor, even if it does so now? Can we remove the remaining `_Activate_SSO_buffer` then? I'm highly horrified by this; especially, what's its relation with the lifetime issue of fancy pointers?

Related: #1735 #3235 and #3273.

### DIFF
--- a/stl/inc/xstring
+++ b/stl/inc/xstring
@@ -2968,10 +2968,14 @@ public:
         _Released_buffer _Result;
         auto& _My_data = _Mypair._Myval2;
         _Result._Size  = _My_data._Mysize;
+        _My_data._Orphan_all();
         _ASAN_STRING_REMOVE(*this);
         if (_My_data._Large_mode_engaged()) {
             _Result._Ptr                    = _My_data._Bx._Ptr;
             _Result._Actual_allocation_size = _My_data._Myres + 1;
+
+            _Destroy_in_place(_My_data._Bx._Ptr);
+            _My_data._Activate_SSO_buffer();
         } else {
             // use _Least_allocation_size to avoid small mode, if the buffer is assigned back
             size_type _Allocated = _Least_allocation_size;
@@ -2979,8 +2983,9 @@ public:
             _Traits::copy(_Unfancy(_Result._Ptr), _My_data._Bx._Buf, _BUF_SIZE);
             _Result._Actual_allocation_size = _Allocated;
         }
-        _My_data._Orphan_all();
-        _Tidy_init();
+        _My_data._Mysize = 0;
+        _My_data._Myres  = _Small_string_capacity;
+        _Traits::assign(_My_data._Bx._Buf[0], _Elem());
         return _Result;
     }
 #endif // _HAS_CXX20
@@ -4831,17 +4836,6 @@ private:
     _CONSTEXPR20 void _Eos(const size_type _New_size) noexcept { // set new length and null terminator
         _ASAN_STRING_MODIFY(*this, _Mypair._Myval2._Mysize, _New_size);
         _Traits::assign(_Mypair._Myval2._Myptr()[_Mypair._Myval2._Mysize = _New_size], _Elem());
-    }
-
-    _CONSTEXPR20 void _Tidy_init() noexcept {
-        // initialize basic_string data members
-        auto& _My_data   = _Mypair._Myval2;
-        _My_data._Mysize = 0;
-        _My_data._Myres  = _Small_string_capacity;
-        _My_data._Activate_SSO_buffer();
-
-        // the _Traits::assign is last so the codegen doesn't think the char write can alias this
-        _Traits::assign(_My_data._Bx._Buf[0], _Elem());
     }
 
     _CONSTEXPR20 void _Tidy_deallocate() noexcept { // initialize buffer, deallocating any storage
--- a/stl/inc/xstring
+++ b/stl/inc/xstring
@@ -2287,6 +2287,19 @@ public:
         value_type _Buf[_BUF_SIZE];
         pointer _Ptr;
         char _Alias[_BUF_SIZE]; // TRANSITION, ABI: _Alias is preserved for binary compatibility (especially /clr)
+
+        _CONSTEXPR20 void _Switch_to_buf() noexcept {
+            _STD _Destroy_in_place(_Ptr);
+
+#if _HAS_CXX20
+            // start the lifetime of the array elements
+            if (_STD is_constant_evaluated()) {
+                for (size_type _Idx = 0; _Idx < _BUF_SIZE; ++_Idx) {
+                    _Buf[_Idx] = value_type();
+                }
+            }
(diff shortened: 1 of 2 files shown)


stl/inc/xstring
```diff
@@ -4815,26 +4831,13 @@ private:
...
         _My_data._Orphan_all();
         if (_My_data._Large_mode_engaged()) {
             _ASAN_STRING_REMOVE(*this);
-            const pointer _Ptr = _My_data._Bx._Ptr;
+            const pointer _Ptr = _My_data._Release_ptr_and_activate_buffer();
             auto& _Al          = _Getal();
-            _Destroy_in_place(_My_data._Bx._Ptr);
-            _My_data._Activate_SSO_buffer();
             _Deallocate_for_capacity(_Al, _Ptr, _My_data._Myres);
         }
```
achabense, line 4843: Slightly irrelevant; it also seems that the block in this function is actually useless (`_Tidy_deallocate` seems always followed by methods that just rewrite these values).
```
        _My_data._Mysize = 0;
        _My_data._Myres  = _Small_string_capacity;
        // the _Traits::assign is last so the codegen doesn't think the char write can alias this
        _Traits::assign(_My_data._Bx._Buf[0], _Elem());
```

tests/std/tests/VSO_0000000_allocator_propagation/test.cpp
```diff
@@ -946,6 +946,20 @@ _CONSTEXPR20 void test_string_swap(const size_t id1, const size_t id2) {
     assert(dst.get_allocator().id() == id1);
 }
 
+#if _HAS_CXX20
```
achabense, line 949: This don't have to be guarded by `_HAS_CXX20`, as `strbuf(move(str))` will still call copy constructor in pre-cpp20 mode. However, I think adding `_HAS_CXX20` will make it more obvious that the leak is relevant to `basic_stringbuf(basic_string&&,...)` introduced in C++20.

achabense:
https://github.com/microsoft/STL/blob/48eedd369d6c88130fff4d15afb88f016fa6e648/stl/inc/xstring#L2282-L2284
`Don't rely on the new behavior` Does this mean we should not assume that `_Buf` is initialized on ctor, even if it does so?

stl/inc/xstring
```diff
@@ -4826,6 +4822,20 @@ private:
         _Traits::assign(_My_data._Bx._Buf[0], _Elem());
     }
 
+    _CONSTEXPR20 void _Construct_empty() noexcept {
+        // this function needs to be noexcept for conformance to the standard.
```
achabense: I'm not too sure whether it's good to add `noexcept` for this function.

stl/inc/xstring
```diff
@@ -2630,6 +2640,20 @@ private:
         _Al.deallocate(_Old_ptr, _Capacity + 1); // +1 for null terminator
     }
 
+    _CONSTEXPR20 void _Construct_empty() noexcept {
+        // this function needs to be noexcept for conformance to the standard.
```
achabense, line 2644: I'm not too sure whether it's good to add `noexcept` for this function.

StephanTLavavej:
Regarding the "Don't rely on the new behavior" comment - if you look at `git blame`, this was introduced by #3235 (which you noticed was related), because we found an ODR mix-and-match problem between 17.3 and 17.4. The problem was a case of "don't make coordinated behavioral changes between two functions" because the linker might mix old `a()` with new `b()` or new `a()` with old `b()`. This is pretty rare for us, so we aren't good at noticing when it's about to happen (unlike representation changes).

I'm pretty nervous about changes in this area as a result. The idea behind `_String_val::_Bxty::_Switch_to_buf` appears highly desirable, but we need to be careful about assuming that another member function (or a constructor) has left the object in a known state, if that could have been different in 17.3 and earlier.

It might be time to rename `_String_val` now, even though that will make this PR bigger, to reduce the risk of introducing a damaged state. We'll still need to maintain layout compatibility (and keep the visualizers working), but we won't need to worry about mix-and-match behavior. (The linker could still pick old and new `basic_string` member functions, but those all leave the object in a guaranteed state.)

stl/inc/xstring
StephanTLavavej: It doesn't *need* to be noexcept - any (non-conformant) exceptions thrown by the proxy allocation will slam into the `basic_string` constructor's annotation. Arguably, this shouldn't actually be marked - note that `basic_string() noexcept(is_nothrow_default_constructible_v<_Alty>)` permits exceptions to be emitted (although this is from a surprising source). I think we should remove this marking and the comment.

StephanTLavavej: It's also a behavioral change compared to the previous direct allocation of a proxy.

StephanTLavavej (approved):

StephanTLavavej:
Thanks for these fine-grained commits, they made reviewing very straightforward! After thinking about each one, I believe that this PR has no ABI impact, and we should proceed as-is since it fixes the lifetime bug and generally makes it harder to introduce such mistakes in the future. Each commit is a transformation that doesn't introduce new inter-function assumptions (as far as I can tell).

This PR is definitely complicated enough and changing a highly-used, risky area of the codebase that we'll need a second maintainer signoff.

It definitely seems like any followup PR to remove `_Activate_SSO_buffer` needs to also rename `_String_val`, otherwise it'll be making the same ABI break as we suffered previously.

CaseyCarter (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for making `basic_string` more maintainable! :tada: :heart_eyes_cat: :yarn:

PR #3206, opened 2022-11-12
Update `source_location` to use `__builtin_FUNCSIG`

SuperWig:
Fixes #3063 

* Should there be an escape hatch?
* The column number is strangely different in the lambda test. `x1` is different to all others
   ```
   x1 = source_location::current()
                                ^
  x2 = source_location::current()
                         ^
   ```
   Edit: scratch that, it already has a transition comment I didn't merge in.

PR #4055, opened 2023-09-27
`<source_location>`: Light up detailed `function_name()` for Clang 17, add escape hatch

### LINKED ISSUE #3729: `<source_location>`: Use `__builtin_FUNCSIG` when Clang 17 is available
I've implemented `__builtin_FUNCSIG` in Clang recently (https://github.com/llvm/llvm-project/commit/78d8312acea96f58f13a876f98d241590c16d530):
```c++
const char* __builtin_FUNCSIG();
```

We should use this builtin in `source_location` implementation after VS gets Clang 17:
https://github.com/microsoft/STL/blob/a62109595b6d89e08172fdf4beb75a2670fe0cc9/stl/inc/source_location#L28-L32

Comparison:
```c++
// https://godbolt.org/z/zq6v9v87q
#include <iostream>

void f(const char* old_ = __builtin_FUNCTION(), const char* new_ = __builtin_FUNCSIG()) {
    std::cout << "old: " << old_ << "\nnew: " << new_ << '\n';
}

int main() {
    f();
}
```

```console
old: main
new: int __cdecl main(void)
```

StephanTLavavej:
Fixes #3729.

After #3206 in VS 2022 17.6 made `source_location::function_name()` more detailed, some users experienced significant increases in binary size (see microsoft/cppwinrt#1337), because heavy usage generated lots of long strings which couldn't be folded away. microsoft/cppwinrt#1260 added an escape hatch for that project, but we should add one too, which fixes VSO-1893424 / AB#1893424 .

This uses [Clang's `__has_builtin`](https://clang.llvm.org/docs/LanguageExtensions.html#has-builtin) to detect the upcoming support. When supported, enabling the detailed function name by default is still the right design - it's *much* more useful.

```
C:\Temp>type meow.cpp
```
```cpp
#include <print>
#include <source_location>
#include <vector>
using namespace std;

void purr(source_location loc = source_location::current()) {
    println("{}", loc.function_name());
}

template <typename T>
void cats() {
    purr();
}

int main() {
#ifdef __clang__
    println("Clang {}", __clang_major__);
#else
    println("MSVC");
#endif
    cats<vector<int>>();
}
```
```
C:\Temp>cl /EHsc /nologo /W4 /std:c++latest meow.cpp && meow
meow.cpp
MSVC
void __cdecl cats<class std::vector<int,class std::allocator<int> >>(void)

C:\Temp>cl /EHsc /nologo /W4 /std:c++latest /D_USE_DETAILED_FUNCTION_NAME_IN_SOURCE_LOCATION=0 meow.cpp && meow
meow.cpp
MSVC
cats

C:\Temp>clang-cl /EHsc /nologo /W4 /std:c++latest meow.cpp && meow
Clang 16
cats

C:\Temp>set PATH=C:\Users\stl\Downloads\LLVM-17.0.1-win64\bin;%PATH%

C:\Temp>clang-cl /EHsc /nologo /W4 /std:c++latest meow.cpp && meow
Clang 17
void __cdecl cats(void) [T = std::vector<int>]

C:\Temp>clang-cl /EHsc /nologo /W4 /std:c++latest /D_USE_DETAILED_FUNCTION_NAME_IN_SOURCE_LOCATION=0 meow.cpp && meow
Clang 17
cats
```

### DIFF
--- a/stl/inc/source_location
+++ b/stl/inc/source_location
@@ -20,15 +20,25 @@ _STL_DISABLE_CLANG_WARNINGS
 #pragma push_macro("new")
 #undef new
 
+#ifndef _USE_DETAILED_FUNCTION_NAME_IN_SOURCE_LOCATION
+#ifdef __EDG__ // TRANSITION, DevCom-10199227
+#define _USE_DETAILED_FUNCTION_NAME_IN_SOURCE_LOCATION 0
+#elif defined(__clang__) // TRANSITION, Clang 17 has this builtin
+#define _USE_DETAILED_FUNCTION_NAME_IN_SOURCE_LOCATION __has_builtin(__builtin_FUNCSIG)
+#else // ^^^ Clang / MSVC vvv
+#define _USE_DETAILED_FUNCTION_NAME_IN_SOURCE_LOCATION 1
+#endif // ^^^ MSVC ^^^
+#endif // ^^^ !defined(_USE_DETAILED_FUNCTION_NAME_IN_SOURCE_LOCATION) ^^^
+
 _STD_BEGIN
 _EXPORT_STD struct source_location {
     _NODISCARD static consteval source_location current(const uint_least32_t _Line_ = __builtin_LINE(),
         const uint_least32_t _Column_ = __builtin_COLUMN(), const char* const _File_ = __builtin_FILE(),
-#if defined(__clang__) || defined(__EDG__) // TRANSITION, DevCom-10199227 and LLVM-58951
-        const char* const _Function_ = __builtin_FUNCTION()
-#else // ^^^ workaround / no workaround vvv
+#if _USE_DETAILED_FUNCTION_NAME_IN_SOURCE_LOCATION
         const char* const _Function_ = __builtin_FUNCSIG()
-#endif // TRANSITION, DevCom-10199227 and LLVM-58951
+#else
+        const char* const _Function_ = __builtin_FUNCTION()
+#endif
             ) noexcept {
         source_location _Result{};
         _Result._Line     = _Line_;
--- a/stl/inc/source_location
+++ b/stl/inc/source_location
@@ -36,9 +36,9 @@ _EXPORT_STD struct source_location {
         const uint_least32_t _Column_ = __builtin_COLUMN(), const char* const _File_ = __builtin_FILE(),
 #if _USE_DETAILED_FUNCTION_NAME_IN_SOURCE_LOCATION
         const char* const _Function_ = __builtin_FUNCSIG()
-#else
+#else // ^^^ detailed / basic vvv
         const char* const _Function_ = __builtin_FUNCTION()
-#endif
+#endif // ^^^ basic ^^^
             ) noexcept {
         source_location _Result{};
         _Result._Line     = _Line_;
--- a/tests/std/tests/P1208R6_source_location/test.cpp
+++ b/tests/std/tests/P1208R6_source_location/test.cpp
@@ -61,11 +61,11 @@ constexpr void local_test() {
 #else // ^^^ EDG / C1XX vvv
     assert(x.column() == 37);
 #endif // ^^^ C1XX ^^^
-#if defined(__clang__) || defined(__EDG__) // TRANSITION, DevCom-10199227 and LLVM-58951
-    assert(x.function_name() == "local_test"sv);
-#else // ^^^ workaround / no workaround vvv
+#if _USE_DETAILED_FUNCTION_NAME_IN_SOURCE_LOCATION
     assert(x.function_name() == "void __cdecl local_test(void)"sv);
-#endif // TRANSITION, DevCom-10199227 and LLVM-58951
+#else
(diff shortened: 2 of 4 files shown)


stl/inc/source_location
```diff
@@ -20,15 +20,25 @@ _STL_DISABLE_CLANG_WARNINGS
...
 _EXPORT_STD struct source_location {
     _NODISCARD static consteval source_location current(const uint_least32_t _Line_ = __builtin_LINE(),
         const uint_least32_t _Column_ = __builtin_COLUMN(), const char* const _File_ = __builtin_FILE(),
-#if defined(__clang__) || defined(__EDG__) // TRANSITION, DevCom-10199227 and LLVM-58951
-        const char* const _Function_ = __builtin_FUNCTION()
-#else // ^^^ workaround / no workaround vvv
+#if _USE_DETAILED_FUNCTION_NAME_IN_SOURCE_LOCATION
         const char* const _Function_ = __builtin_FUNCSIG()
-#endif // TRANSITION, DevCom-10199227 and LLVM-58951
+#else
```
CaseyCarter, line 39: Why did this `#else` and `#endif` need comments before, but not now?

CaseyCarter:
I note that the test fails tragically and horribly with Clang 17. This certainly isn't a problem for our CI that uses Clang 16, but (1) defers a bunch of work to the poor person who has to update the CI for Clang 17 and (2) makes it hard to validate this change. Should we update the test so it's compatible with both Clang 16 and 17 for now?

stl/inc/source_location
StephanTLavavej: `workaround / no workaround` is a nice reminder when we plan to remove one of the branches in the future. Now, both branches are permanent, each of them is 1 line with no surrounding preprocessor logic, and "use detailed function name" is clearly readable so there's no real chance for confusion about which side is doing what.

I'd like to keep this as-is, using our usual convention that we can skip comments in this 1-line scenario.

StephanTLavavej:
I'll update the test for Clang 17, thanks. I validated it by hand but not in the test harness, good catch.

stl/inc/source_location
CaseyCarter: Great! I didn't think through the fact that this was a workaround but is no more.

StephanTLavavej:
Let me rework this one more time - the x86 vs. non-x86 logic is bloated.

stl/inc/source_location
StephanTLavavej: The tests have enough dense preprocessor logic that I'm finding I want "detailed / basic" in them - I think I'll do that in the product code for consistency. So you'll get comments after all :joy_cat:

StephanTLavavej:
Ok, I've validated and pushed changes. The test now passes with MSVC, Clang 16, and Clang 17, for both x64 and x86. I also extracted the calling convention variation (using `constexpr basic_string`!) because that logic was getting unwieldy.

The changes to `test_header_units_and_modules.hpp` are "speculative" because neither Clang nor EDG are running that yet, but I verified (with classic includes) that Clang will produce that output. Because the named modules don't emit macros, I determine "by hand" whether we're using detailed info.

tests/std/tests/P1208R6_source_location/test.cpp
```diff
@@ -5,10 +5,17 @@
 #include <cassert>
 #include <functional>
 #include <source_location>
+#include <string>
 #include <string_view>
 #include <type_traits>
 using namespace std;
 
+#ifdef _M_IX86
+#define THISCALL_OR_CDECL_STR "__thiscall"s
```
CaseyCarter, line 14: I note that all uses of `THISCALL_OR_CDECL_STR` concatenate it with string literals. We _could_ define this as a string literal and let the parser do the concatenation for us (replace `"meow" + THISCALL_OR_CDECL_STR + "woof"` with `"meow" THISCALL_OR_CDECL_STR "woof"`).

Absent that, there seems to be no reason to use a macro when 
```c++
constexpr string thiscall_or_cdecl{
#ifdef _M_IX86
    "__thiscall"
#else
    "__cdecl"
#endif
};
```
would work as well.

No change requested, because (1) we don't really have a convention, this is just style, and (2) I feel like I've already asked for enough work in this PR.

CaseyCarter (approved):

tests/std/tests/P1208R6_source_location/test.cpp
StephanTLavavej: As I mentioned on Discord:

* We can't just say `"meow" MACRO "woof"` because `function_name()` returns `const char*`, so we need a `string_view` or a `string` to compare against.
* We can't have a top-level `constexpr string` because `constexpr` dynamic allocations can't survive until runtime.

However, `"meow" MACRO "woof"sv` works. I went ahead and pushed a change to do that, because `constexpr string` is a very high-powered feature (read: lots of compiler bugs historically) and this test is already stressing the compiler enough.

CaseyCarter (approved):

StephanTLavavej:

PR #4058, opened 2023-09-28
[ASan] Disable annotations on unsupported platforms

### LINKED ISSUE #4035: MSVC 14.37.32822 does not ship with stl_asan.lib for arm64
# Describe the bug
When trying to build compiler rt test cases on windows I am getting the following linker error:

```bash
LINK : fatal error LNK1104: cannot open file 'stl_asan.lib'
```
I reported the bug to llvm, but the fix likely needs to come from MSVC to provide a `stl_asan.lib` for arm64.
https://github.com/llvm/llvm-project/issues/66600

Without `stl_asan.lib` you can't build compiler_rt test cases using `COMPILER_RT_INCLUDE_TESTS`. 

These test cases are important to validate a port of sanitzers for winodws arm64.

# Command-line test case
```
cmake -GNinja -DLLVM_ENABLE_PROJECTS=llvm;compiler-rt -DCMAKE_BUILD_TYPE=Release \
 -DLLVM_ENABLE_PDB=ON -DLLVM_ENABLE_ASSERTIONS=ON -DLLVM_TARGETS_TO_BUILD=AArch64 \
-DCMAKE_C_COMPILER=clang-cl.exe -DCMAKE_CXX_COMPILER=clang-cl.exe \
 -DCOMPILER_RT_INCLUDE_TESTS=TRUE -DCMAKE_LINKER=lld-link.exe  \
-S llvm-project\llvm -B compiler-rt_build
cd llvm_build
ninja 
ninja check-all
```
Possible solutions
for arm64 set: `_DISABLE_STRING_ANNOTATION` and `_DISABLE_VECTOR_ANNOTATION`
https://github.com/microsoft/STL/blob/8f67ece4a654625220642c92fa029a5cfc77afa8/stl/inc/__msvc_sanitizer_annotate_container.hpp#L18
or remove `defined(_M_ARM64)` so that we don't try to invoke the linker for asan on windows arm64
https://github.com/microsoft/STL/blob/8f67ece4a654625220642c92fa029a5cfc77afa8/stl/inc/__msvc_sanitizer_annotate_container.hpp#L115

strega-nil-ms:
Add a new define, `_DISABLE_ASAN_ANNOTATIONS`, which disables all ASan annotations in the standard library as opposed to requiring a person to define both `_DISABLE_STRING_ANNOTATION` and
`_DISABLE_VECTOR_ANNOTATION`.

Disable ASan annotations on unsupported platforms like ARM64. If someone is working on adding address sanitizer support on those platforms (* wink wink nudge nudge *) they can define the
`_ENABLE_ASAN_ANNOTATIONS_ON_UNSUPPORTED_PLATFORMS` escape hatch to get STL annotations anyways, though they will need to provide their own stl_asan.lib.

Additionally, add two new defines:

- `_DISABLE_STL_ANNOTATION`
- `_ANNOTATE_STL`

which do the job of `_DISABLE_STRING_ANNOTATION` and `_DISABLE_VECTOR_ANNOTATION`, and `_ANNOTATE_STRING` and `_ANNOTATE_VECTOR` respectively. This allows users to pass these defines without worrying we'll add new ones for other containers.

Also, fix a bug where `annotate_string` (and `annotate_vector`) were not getting set to `0` if you passed both `_DISABLE_STRING_ANNOTATION` and `_DISABLE_VECTOR_ANNOTATION`.

Fixes #4035


stl/inc/__msvc_sanitizer_annotate_container.hpp
```diff
@@ -15,7 +15,28 @@ _STL_DISABLE_CLANG_WARNINGS
 #pragma push_macro("new")
 #undef new
 
-#if !defined(_M_CEE_PURE) && !(defined(_DISABLE_STRING_ANNOTATION) && defined(_DISABLE_VECTOR_ANNOTATION))
+#ifdef _DISABLE_ASAN_ANNOTATIONS
+
+#ifdef _ENABLE_ASAN_ANNOTATIONS_ON_UNSUPPORTED_PLATFORMS
+#error \
+    "Invalid set of defines - one cannot define both _ENABLE_ASAN_ANNOTATIONS_ON_UNSUPPORTED_PLATFORMS and _DISABLE_ASAN_ANNOTATIONS at the same time"
```
CaseyCarter, line 22: Mega-nit: Obviously one _can_ define both - one just did!
```suggestion
    "Invalid set of defines: _ENABLE_ASAN_ANNOTATIONS_ON_UNSUPPORTED_PLATFORMS and _DISABLE_ASAN_ANNOTATIONS are mutually exclusive"
```

stl/inc/__msvc_sanitizer_annotate_container.hpp
```diff
@@ -15,7 +15,28 @@ _STL_DISABLE_CLANG_WARNINGS
 #pragma push_macro("new")
 #undef new
 
-#if !defined(_M_CEE_PURE) && !(defined(_DISABLE_STRING_ANNOTATION) && defined(_DISABLE_VECTOR_ANNOTATION))
+#ifdef _DISABLE_ASAN_ANNOTATIONS
+
+#ifdef _ENABLE_ASAN_ANNOTATIONS_ON_UNSUPPORTED_PLATFORMS
+#error \
+    "Invalid set of defines - one cannot define both _ENABLE_ASAN_ANNOTATIONS_ON_UNSUPPORTED_PLATFORMS and _DISABLE_ASAN_ANNOTATIONS at the same time"
+#endif
```
CaseyCarter, line 23:
```suggestion
#endif // ^^^ defined(_ENABLE_ASAN_ANNOTATIONS_ON_UNSUPPORTED_PLATFORMS) ^^^
```

stl/inc/__msvc_sanitizer_annotate_container.hpp
```diff
@@ -15,7 +15,28 @@ _STL_DISABLE_CLANG_WARNINGS
...
+
+#ifndef _ENABLE_ASAN_ANNOTATIONS_ON_UNSUPPORTED_PLATFORMS
+#define _DISABLE_ASAN_ANNOTATIONS
+#endif
+
+#elif defined(_M_CEE_PURE)
+#define _DISABLE_ASAN_ANNOTATIONS
+#endif
+
+#endif // ^^^ !_DISABLE_ASAN_ANNOTATIONS
```
CaseyCarter, line 39:
```suggestion
#endif // ^^^ !defined(_DISABLE_ASAN_ANNOTATIONS) ^^^
```

stl/inc/__msvc_sanitizer_annotate_container.hpp
```diff
@@ -45,11 +66,11 @@ _STL_DISABLE_CLANG_WARNINGS
 
 #endif // ^^^ !defined(__clang__) && !defined(__SANITIZE_ADDRESS__) ^^^
 
-#ifdef _DISABLE_STRING_ANNOTATION
+#if defined(_DISABLE_ASAN_ANNOTATION) || defined(_DISABLE_STRING_ANNOTATION)
 #undef _ACTIVATE_STRING_ANNOTATION
 #undef _INSERT_STRING_ANNOTATION
 #endif // defined(_DISABLE_STRING_ANNOTATION)
```
CaseyCarter, line 72:
```suggestion
#endif // defined(_DISABLE_ASAN_ANNOTATION) || defined(_DISABLE_STRING_ANNOTATION)
```

stl/inc/__msvc_sanitizer_annotate_container.hpp
```diff
@@ -45,11 +66,11 @@ _STL_DISABLE_CLANG_WARNINGS
...
-#ifdef _DISABLE_STRING_ANNOTATION
+#if defined(_DISABLE_ASAN_ANNOTATION) || defined(_DISABLE_STRING_ANNOTATION)
 #undef _ACTIVATE_STRING_ANNOTATION
 #undef _INSERT_STRING_ANNOTATION
 #endif // defined(_DISABLE_STRING_ANNOTATION)
-#ifdef _DISABLE_VECTOR_ANNOTATION
+#if defined(_DISABLE_ASAN_ANNOTATION) || defined(_DISABLE_VECTOR_ANNOTATION)
 #undef _ACTIVATE_VECTOR_ANNOTATION
 #undef _INSERT_VECTOR_ANNOTATION
 #endif // defined(_DISABLE_VECTOR_ANNOTATION)
```
CaseyCarter, line 76:
```suggestion
#endif // defined(_DISABLE_ASAN_ANNOTATION) || defined(_DISABLE_VECTOR_ANNOTATION)
```

CaseyCarter (changes requested):

stl/inc/__msvc_sanitizer_annotate_container.hpp
CaseyCarter: Maybe strike "Invalid set of defines: " as well, to get this under 120 columns.

StephanTLavavej: FYI, this is an indication of editor misconfiguration; we should all be using format-on-save to avoid failures like this.

CaseyCarter: I believe this is clang-format output. Putting just:
```c++
#error "Invalid set of defines - one cannot define both _ENABLE_ASAN_ANNOTATIONS_ON_UNSUPPORTED_PLATFORMS and _DISABLE_ASAN_ANNOTATIONS at the same time"
```
in a file and clang-formatting produces the still overlong:
```c++
#error \
    "Invalid set of defines - one cannot define both _ENABLE_ASAN_ANNOTATIONS_ON_UNSUPPORTED_PLATFORMS and _DISABLE_ASAN_ANNOTATIONS at the same time"
```

strega-nil-ms: I'm fine with everything else, but I do not believe in passing non-strings to `#error` :P

benchmarks/google-benchmark
CaseyCarter, line 1: I assume this reversion to 1.7.0 is unintentional?

stl/inc/__msvc_sanitizer_annotate_container.hpp
```diff
@@ -15,7 +15,42 @@ _STL_DISABLE_CLANG_WARNINGS
...
+//     (this will be auto-defined on unsupported platforms)
+//     + _DISABLE_STRING_ANNOTATION: same, but for only `basic_string`
+//     + _DISABLE_VECTOR_ANNOTATION: same, but for only `vector`
+//   - _ENABLE_STL_ANNOTATION_ON_UNSUPPORTED_PLATFORMS: Don't auto-disable ASan annotations
+//   - _ANNOTATE_STL: Even when ASan annotations are disabled, insert the code for annotating into the STL anyways;
+//     this is useful when building static libraries which may be linked against both ASan and non-ASan binaries.
+//     + _ANNOTATE_STRING: same, but only for `basic_string`
+//     + _ANNOTATE_VECTOR: same, but only for `vector`
+
+// if our user hasn't defined `_DISABLE_STL_ANNOTATION`, then we may disable it anyways
```
CaseyCarter, line 29: Line 20 just told me this. Do I need a reminder? (No change requested.)

stl/inc/__msvc_sanitizer_annotate_container.hpp
```diff
@@ -15,7 +15,42 @@ _STL_DISABLE_CLANG_WARNINGS
...
+#if !defined(_DISABLE_STL_ANNOTATION) && !defined(_ENABLE_STL_ANNOTATION_ON_UNSUPPORTED_PLATFORMS)
+
+#if defined(_M_ARM64EC) || defined(_M_ARM64) || defined(_M_ARM) || defined(_M_CEE_PURE)
+#define _DISABLE_STL_ANNOTATION
+#endif // ^^^ unsupported platform
+
+#endif // ^^^ !defined(_DISABLE_STL_ANNOTATION) && !defined(_ENABLE_STL_ANNOTATION_ON_UNSUPPORTED_PLATFORM)
+
+// this may be user-defined, or we may be on an unsupported platform
```
CaseyCarter, line 38: Ditto "Line 20 told me this"

stl/inc/__msvc_sanitizer_annotate_container.hpp
```diff
@@ -121,9 +156,7 @@ void __cdecl __sanitizer_annotate_contiguous_container(
 #error Unknown architecture
 #endif // ^^^ unknown architecture ^^^
 
-#endif // insert asan annotations
-
-#endif // !defined(_M_CEE_PURE) && asan not disabled
+#endif // ^^^ insert asan annotations
```
CaseyCarter, line 159: Not for this PR, but we really should consistently capitalize "ASan" in docs and comments to avoid confusion. (No change requested.)

CaseyCarter (changes requested):

CaseyCarter (approved):

stl/inc/__msvc_sanitizer_annotate_container.hpp
CaseyCarter: That's fine for `#pragma message`, but `#error` is more formal; it _hates_ contractions. =)

stl/inc/__msvc_sanitizer_annotate_container.hpp
```diff
@@ -15,7 +15,40 @@ _STL_DISABLE_CLANG_WARNINGS
...
+//     + _ANNOTATE_STRING: same, but only for `basic_string`
+//     + _ANNOTATE_VECTOR: same, but only for `vector`
+
+#if !defined(_DISABLE_STL_ANNOTATION) && !defined(_ENABLE_STL_ANNOTATION_ON_UNSUPPORTED_PLATFORMS)
+
+#if defined(_M_ARM64EC) || defined(_M_ARM64) || defined(_M_ARM) || defined(_M_CEE_PURE)
+#define _DISABLE_STL_ANNOTATION
+#endif // ^^^ unsupported platform
+
+#endif // ^^^ !defined(_DISABLE_STL_ANNOTATION) && !defined(_ENABLE_STL_ANNOTATION_ON_UNSUPPORTED_PLATFORM)
```
StephanTLavavej, line 35: Typo.

stl/inc/__msvc_sanitizer_annotate_container.hpp
```diff
@@ -15,7 +15,40 @@ _STL_DISABLE_CLANG_WARNINGS
...
+
+#define _DISABLE_STRING_ANNOTATION
+#define _DISABLE_VECTOR_ANNOTATION
+
+#endif // ^^^ defined(_DISABLE_STL_ANNOTATION)
+
+#ifdef _ANNOTATE_STL
+#define _ANNOTATE_STRING
+#define _ANNOTATE_VECTOR
+#endif // ^^^ defined(_ANNOTATE_STL)
```
StephanTLavavej, line 51: This will warn if the user has accumulated both `/D_ANNOTATE_STL` and either `/D_ANNOTATE_STRING` or `/D_ANNOTATE_VECTOR` in their build system:

```
C:\Temp>type cats.cpp
#ifdef ANNOTATE_STL
#define ANNOTATE_VECTOR
#endif

C:\Temp>cl /EHsc /nologo /W4 /c cats.cpp
cats.cpp

C:\Temp>cl /EHsc /nologo /W4 /c /DANNOTATE_STL /DANNOTATE_VECTOR cats.cpp
cats.cpp
cats.cpp(2): warning C4005: 'ANNOTATE_VECTOR': macro redefinition
cats.cpp(2): note: 'ANNOTATE_VECTOR' previously declared on the command line
```

It seems harmless to enable both the coarse-grained and fine-grained options here, so I recommend wrapping each of these in `#ifndef`s.

stl/inc/__msvc_sanitizer_annotate_container.hpp
```diff
@@ -15,7 +15,40 @@ _STL_DISABLE_CLANG_WARNINGS
...
+#endif // ^^^ !defined(_DISABLE_STL_ANNOTATION) && !defined(_ENABLE_STL_ANNOTATION_ON_UNSUPPORTED_PLATFORM)
+
+#ifdef _DISABLE_STL_ANNOTATION
+
+#ifdef _ENABLE_STL_ANNOTATION_ON_UNSUPPORTED_PLATFORMS
+#error _ENABLE_STL_ANNOTATION_ON_UNSUPPORTED_PLATFORMS and _DISABLE_STL_ANNOTATION are mutually exclusive
+#endif // ^^^ defined(_ENABLE_STL_ANNOTATION_ON_UNSUPPORTED_PLATFORMS)
+
+#define _DISABLE_STRING_ANNOTATION
+#define _DISABLE_VECTOR_ANNOTATION
```
StephanTLavavej, line 44: Similarly (see comment below), should we wrap each of these in `#ifndef` to tolerate both the coarse-grained and fine-grained disable macros being used together?

StephanTLavavej (changes requested):

stl/inc/__msvc_sanitizer_annotate_container.hpp
```diff
@@ -34,41 +78,41 @@ _STL_DISABLE_CLANG_WARNINGS
...
 #undef _INSERT_VECTOR_ANNOTATION
-#endif // defined(_DISABLE_VECTOR_ANNOTATION)
+#endif // ^^^ defined(_DISABLE_VECTOR_ANNOTATION)
+
+#ifdef _ANNOTATE_STRING
+#define _INSERT_STRING_ANNOTATION
+#endif // ^^^ defined(_ANNOTATE_STRING)
+#ifdef _ANNOTATE_VECTOR
+#define _INSERT_VECTOR_ANNOTATION
+#endif // ^^^ defined(_ANNOTATE_VECTOR)
```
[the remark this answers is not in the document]
StephanTLavavej: Oops this was the wrong region to comment on, disregard.

stl/inc/__msvc_sanitizer_annotate_container.hpp
```diff
@@ -15,7 +15,51 @@ _STL_DISABLE_CLANG_WARNINGS
...
+
+#ifdef _ANNOTATE_STL
+
+#ifdef _ANNOTATE_STRING
+#define _ANNOTATE_STRING
+#endif // ^^^ defined(_ANNOTATE_STRING)
+
+#ifdef _ANNOTATE_VECTOR
+#define _ANNOTATE_VECTOR
+#endif // ^^^ defined(_ANNOTATE_VECTOR)
```
StephanTLavavej, line 60: This says `ifdef` but you want `ifndef`. (Comments too.)

StephanTLavavej (changes requested):

StephanTLavavej (approved):

CaseyCarter (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for improving the ASan experience! :tada: :chart_with_upwards_trend: :gear:

PR #4063, opened 2023-09-30
Avoid branching with using negation to automatically reach the correct value for _Size

AtariDreams:
I was able to save an instruction in the generated asm for armv8.

Proof: https://godbolt.org/z/sEYnvbodz


stl/inc/functional
```diff
@@ -2448,11 +2448,9 @@ struct uses_allocator<function<_Fty>, _Alloc> : true_type {}; // true_type if co
...
     // moves _Size up to account for storing an object of size _Size_added with alignment _Alignment in a memory buffer
     // assumes that the memory buffer will be allocated for the "worst" alignment used in the resulting buffer
-    const size_t _Align_masked = _Size & (_Alignment - 1u);
-    _Size += _Size_added;
-    if (_Align_masked != 0) {
-        _Size += _Alignment - _Align_masked;
-    }
+    // note that -_Size causes the compiler to emit a warning, as _Size is unsigned, but luckily, when optimization
+    // is enabled, "~_Size + 1u" generates the exact same assembler as -_Size
+    _Size += ((~_Size + 1u) & (_Alignment - 1u)) + _Size_added;
```
AlexGuteniev, line 2453: I'd prefer `_Size_added + ((~_Size + 1u) & (_Alignment - 1u))`. First operand then correction looks more natural.

stl/inc/functional
```diff
@@ -2448,11 +2448,9 @@ struct uses_allocator<function<_Fty>, _Alloc> : true_type {}; // true_type if co
 inline void _Add_alloc_size(size_t& _Size, const size_t _Size_added, const size_t _Alignment) {
     // moves _Size up to account for storing an object of size _Size_added with alignment _Alignment in a memory buffer
     // assumes that the memory buffer will be allocated for the "worst" alignment used in the resulting buffer
-    const size_t _Align_masked = _Size & (_Alignment - 1u);
-    _Size += _Size_added;
-    if (_Align_masked != 0) {
-        _Size += _Alignment - _Align_masked;
-    }
+    // note that -_Size causes the compiler to emit a warning, as _Size is unsigned, but luckily, when optimization
+    // is enabled, "~_Size + 1u" generates the exact same assembler as -_Size
```
CaseyCarter, line 2452: We usually use `0 - x` to silence warnings about `-x` when `x` is unsigned. Can we do so here as well? (and remove the comment, which we don't need since the compiler will warn if someone does change this to `-_Size`?)

AtariDreams: @CaseyCarter Done!

stl/inc/functional
AtariDreams: I'm visualizing it as: align to the end and then add, as that is what we are doing.

AlexGuteniev (approved):

StephanTLavavej:
Thanks for looking into this code. We talked about it at the weekly maintainer meeting and have decided to close this PR without merging. We believe that this change heavily obfuscates what the code is trying to do (even with retaining the comment, as it no longer clearly maps to the code below). Additionally, this is not a performance-sensitive area - this is called during allocation of Boyer-Moore data structures, whose whole purpose in life is that they're allocated once before a large number of string searches, so this is even less performance-sensitive than most code that happens per allocation (which are typically quite expensive, so additional costs are minor in comparison). Finally, the code size impact is irrelevant, as this function is not heavily instantiated.

PR #4064, opened 2023-09-30
Cleanups for `<flat_set>`

achabense:
Contents in this pr:
- Drop `_Compressed_pair` optimization, using separate container and comparer.
~ I think it will be ok to drop this optimization, as the spatial gain is not significant, and it greatly complicates constructors and brings indirection when accessing members. The spatial optimization will become easily achievable (and without the `final` constraint) when `[[no_unique_address]]` becomes available. I notice that `mdspan` has already adopted this approach.
- Various refactorings, including some renamings, and giving separate definitions for `_Emplace[_hint]:_Multi/!_Multi`, for better clarity.
- Record some remaining issues in the current codebase (`// TRANSITION` ones). Some of them are really tough :|
- Fix a bug introduced by previous pr (my fault :( The problem is that even a successful noexcept `move` may not empty the container.

### DIFF
--- a/stl/inc/flat_set
+++ b/stl/inc/flat_set
@@ -483,83 +483,93 @@ private:
     }
 
     template <class _Ty>
-    auto _Emplace(_Ty&& _Val) {
-        if constexpr (_Multi) {
-            _STL_INTERNAL_STATIC_ASSERT(is_same_v<remove_cvref_t<_Ty>, _Kty>);
-            return _Mycont.emplace(upper_bound(_Val), _STD forward<_Ty>(_Val));
-        } else {
-            const iterator _End   = end();
-            const iterator _Where = lower_bound(_Val);
-            if (_Where != _End && !_Compare(_Val, *_Where)) {
-                return pair{_Where, false};
-            }
+        requires _Multi // flat_multiset
+    iterator _Emplace(_Ty&& _Val) {
+        _STL_INTERNAL_STATIC_ASSERT(is_same_v<remove_cvref_t<_Ty>, _Kty>);
+        return _Mycont.emplace(upper_bound(_Val), _STD forward<_Ty>(_Val));
+    }
 
-            if constexpr (is_same_v<remove_cvref_t<_Ty>, _Kty>) {
-                return pair{_Mycont.emplace(_Where, _STD forward<_Ty>(_Val)), true};
-            } else {
-                // for flat_set::insert(auto&&)
-                _STL_INTERNAL_STATIC_ASSERT(_Keylt_transparent && is_constructible_v<_Kty, _Ty>);
-                _Kty _Keyval(_STD forward<_Ty>(_Val));
-                _STL_ASSERT(_Can_insert(_Where, _Keyval), "The input type was not equivalent to key_type!");
-                return pair{_Mycont.emplace(_Where, _STD move(_Keyval)), true};
-            }
+    template <class _Ty>
+        requires (!_Multi) // flat_set
+    pair<iterator, bool> _Emplace(_Ty&& _Val) {
+        const iterator _Where = lower_bound(_Val);
+        if (_Where != end() && !_Compare(_Val, *_Where)) {
+            return pair{_Where, false};
+        }
+
+        if constexpr (is_same_v<remove_cvref_t<_Ty>, _Kty>) {
+            return pair{_Mycont.emplace(_Where, _STD forward<_Ty>(_Val)), true};
+        } else {
+            // heterogeneous insertion
+            _STL_INTERNAL_STATIC_ASSERT(_Keylt_transparent && is_constructible_v<_Kty, _Ty>);
+            _Kty _Keyval(_STD forward<_Ty>(_Val));
+            _STL_ASSERT(_Can_insert(_Where, _Keyval), "The conversion from the heterogeneous key to key_type should "
+                                                      "be consistent with the heterogeneous lookup!");
+            return pair{_Mycont.emplace(_Where, _STD move(_Keyval)), true};
         }
     }
 
     template <class _Ty>
+        requires _Multi // flat_multiset
+    iterator _Emplace_hint(const_iterator _Where, _Ty&& _Val) {
+        _STL_INTERNAL_STATIC_ASSERT(is_same_v<remove_cvref_t<_Ty>, _Kty>);
+
+        const const_iterator _Begin = cbegin();
+        const const_iterator _End   = cend();
+
+        // look for upper_bound(_Val)
+        if (_Where == _End || _Compare(_Val, *_Where)) {
+            // _Val < *_Where
+            if (_Where == _Begin || !_Compare(_Val, *(_Where - 1))) {
+                // _Val >= *(_Where-1) ~ upper_bound is _Where
+            } else {
+                // _Val < *(_Where-1) ~ upper_bound is in [_Begin,_Where-1]
+                _Where = _STD upper_bound(_Begin, _Where - 1, _Val, _Pass_comp());
+            }
+        } else {
+            // _Val >= *_Where ~ upper_bound is in [_Where+1,_End]
+            _Where = _STD upper_bound(_Where + 1, _End, _Val, _Pass_comp());
+        }
+
+        _STL_INTERNAL_CHECK(_Can_insert(_Where, _Val));
+        return _Mycont.emplace(_Where, _STD forward<_Ty>(_Val));
+    }
+
+    template <class _Ty>
+        requires (!_Multi) // flat_set
     iterator _Emplace_hint(const_iterator _Where, _Ty&& _Val) {
         const const_iterator _Begin = cbegin();
         const const_iterator _End   = cend();
 
-        if constexpr (_Multi) {
-            // look for the upper_bound for flat_multiset
-            if (_Where == _End || _Compare(_Val, *_Where)) {
-                // _Val < *_Where
-                if (_Where == _Begin || !_Compare(_Val, *(_Where - 1))) {
-                    // _Val >= *(_Where-1) ~ upper_bound is _Where
-                } else {
-                    // _Val < *(_Where-1) ~ upper_bound is in [_Begin,_Where-1]
-                    _Where = _STD upper_bound(_Begin, _Where - 1, _Val, _Pass_comp());
-                }
+        // look for lower_bound(_Val)
+        if (_Where == _End || !_Compare(*_Where, _Val)) {
+            // _Val <= *_Where
+            if (_Where == _Begin || _Compare(*(_Where - 1), _Val)) {
+                // _Val > *(_Where-1) ~ lower_bound is _Where
             } else {
-                // _Val >= *_Where ~ upper_bound is in [_Where+1,_End]
-                _Where = _STD upper_bound(_Where + 1, _End, _Val, _Pass_comp());
+                // _Val <= *(_Where-1) ~ lower_bound is in [_Begin,_Where-1]
+                _Where = _STD lower_bound(_Begin, _Where - 1, _Val, _Pass_comp());
             }
         } else {
-            // look for the lower_bound for flat_set
-            if (_Where == _End || !_Compare(*_Where, _Val)) {
-                // _Val <= *_Where
-                if (_Where == _Begin || _Compare(*(_Where - 1), _Val)) {
-                    // _Val > *(_Where-1) ~ lower_bound is _Where
-                } else {
-                    // _Val <= *(_Where-1) ~ lower_bound is in [_Begin,_Where-1]
-                    _Where = _STD lower_bound(_Begin, _Where - 1, _Val, _Pass_comp());
-                }
-            } else {
-                // _Val > *_Where ~ lower_bound is in [_Where+1,_End]
-                _Where = _STD lower_bound(_Where + 1, _End, _Val, _Pass_comp());
-            }
+            // _Val > *_Where ~ lower_bound is in [_Where+1,_End]
+            _Where = _STD lower_bound(_Where + 1, _End, _Val, _Pass_comp());
         }
 
-        if constexpr (_Multi) {
-            _STL_INTERNAL_STATIC_ASSERT(is_same_v<remove_cvref_t<_Ty>, _Kty>);
+        if (_Where != _End && !_Compare(_Val, *_Where)) {
+            // convert const_iterator to iterator
+            return _Mycont.begin() + (_Where - _Begin);
+        }
+
+        if constexpr (is_same_v<remove_cvref_t<_Ty>, _Kty>) {
             _STL_INTERNAL_CHECK(_Can_insert(_Where, _Val));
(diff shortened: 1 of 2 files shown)


achabense:
I'd have to leave for some time. I will give some detailed explanations for each `TRANSITION` issue tomorrow.

stl/inc/flat_set
```diff
@@ -273,9 +272,7 @@ public:
         if constexpr (requires { _Mycont.append_range(_STD forward<_Rng>(_Range)); }) {
             _Mycont.append_range(_STD forward<_Rng>(_Range));
         } else {
-            for (const auto& _Val : _Range) {
-                _Mycont.insert(_Mycont.end(), _Val);
-            }
+            _Mycont.insert_range(_Mycont.end(), _STD forward<_Rng>(_Range));
```
achabense, line 276: > However, I think we don't have to give an extra check for that chance, as it is likely that a container that doesn't support append_range will not support insert_range either.

I've changed my mind now, as I think there is no reason not to utilize a function that is required for every container by the standard.

stl/inc/flat_set
```diff
@@ -100,12 +95,13 @@ public:
...
-    explicit _Base_flat_set(const key_compare& _Comp) : _My_pair(_One_then_variadic_args_t{}, _Comp) {}
+    explicit _Base_flat_set(const key_compare& _Comp) : _Mycont(), _Mycomp(_Comp) {}
     template <_Allocator_for<container_type> _Alloc>
-    _Base_flat_set(const key_compare& _Comp, const _Alloc& _Al) : _My_pair(_One_then_variadic_args_t{}, _Comp, _Al) {}
+    _Base_flat_set(const key_compare& _Comp, const _Alloc& _Al) : _Mycont(_Al), _Mycomp(_Comp) {}
     template <_Allocator_for<container_type> _Alloc>
-    explicit _Base_flat_set(const _Alloc& _Al) : _My_pair(_Zero_then_variadic_args_t{}, _Al) {}
+    explicit _Base_flat_set(const _Alloc& _Al) : _Mycont(_Al), _Mycomp() {}
 
+    // TRANSITION, an allocator-aware container may not support "C(_First, _Last, _Al)".
```
achabense, line 104: I think there is no need to use `make_obj_using_allocator` for `C(another,al) and C(al)`, as these expressions are required for [allocator-aware containers](https://eel.is/c++draft/container.alloc.reqmts).

For `C(first,last,al)`, `C(from_range...,al)` and `C(ilist,al)`, however, I realize that `make_obj_using_allocator` is not capable. As a failure to construct with these arguments are considered ill-formed by `make_obj_using_allocator`.
> https://eel.is/c++draft/allocator.uses.construction#5.4.sentence-1
> Otherwise, the program is ill-formed.

However, in `flat_set` these constructors only say the container should be constructed with the allocator, then behave as if inserting the range. So these constructors should be valid even if the container doesn't support these constructions.

(The most conservative fix would be fall back to constructing the container using only `_Al`, then doing insertion in the function body.)

stl/inc/flat_set
```diff
@@ -206,16 +203,18 @@ public:
...
+        return _Mycont.size();
     }
     _NODISCARD size_type max_size() const noexcept {
-        return _Get_cont().max_size();
+        return _Mycont.max_size();
     }
 
     // modifiers
+    // TRANSITION, the "insert" and "erase" methods may not be able to restore the invariant, if the underlying
+    // container is unable to provide strong guarantee for "erase" and "insert" methods.
```
achabense, line 217: These are related to #4059. Aside from exception guarantee, I think the standard is also not clear enough about how should the invariant be restored.
>If any member function in [[flat.set.defn]](https://eel.is/c++draft/flat.set#defn) exits via an exception, the invariant is restored[.](https://eel.is/c++draft/flat.set#overview-6.sentence-1)
[Note [2](https://eel.is/c++draft/flat.set#overview-note-2): This can result in the flat_set's being emptied[.](https://eel.is/c++draft/flat.set#overview-6.sentence-2) — end note]

Should `flat_meow` be unchanged, or should `flat_meow` be emptied? I think the standard should really give some specifications. At least for erasion and single-element insertion, strong-guarantee is possible if the underlying container provide strong-guarantee too.

stl/inc/flat_set
```diff
@@ -621,32 +628,30 @@ private:
...
+            _Mycont.erase(_STD unique(begin(), _End, _Equivalent), _End);
         }
     }
 
     template <bool _Presorted>
     void _Restore_invariants_after_insert(const size_type _Old_size) {
-        auto _Comp              = _Get_comp_v();
         const iterator _Begin   = begin();
         const iterator _Old_end = _Begin + static_cast<difference_type>(_Old_size);
-        const iterator _New_end = end();
```
achabense, line 638: The two `_New_end` were sharing different meanings in `_Erase_dupes_if_not_multi` and here. Renamed to common `_End` here to match with other methods.

stl/inc/flat_set
```diff
@@ -621,32 +628,30 @@ private:
...
-            _STL_ASSERT(_Check_sorted(_Old_end, _New_end), _Msg_not_sorted);
+            _STL_ASSERT(_Is_sorted(_Old_end, _End), _Msg_not_sorted);
         }
 
-        _STD inplace_merge(_Begin, _Old_end, _New_end, _Comp);
+        _STD inplace_merge(_Begin, _Old_end, _End, _Pass_comp());
         _Erase_dupes_if_not_multi();
 
-        _STL_INTERNAL_CHECK(_Check_sorted(cbegin(), cend()));
+        _STL_INTERNAL_CHECK(_Is_sorted(cbegin(), cend()));
```
achabense, line 654: regaining `cend` is necessary, as `_Erase_dupes_if_not_multi` may invalidate `end`. I think for better clarity we'd better use `_Is_sorted(_Mycont.cbegin(), _Mycont.cend())` when the check is container-scale.

stl/inc/flat_set
```diff
@@ -70,17 +65,17 @@ public:
     static_assert(random_access_iterator<iterator>, "The C++ Standard forbids containers without random "
                                                     "access iterators from being adapted. See [flatset.overview].");
```
achabense, line 66: `See [flatset.overview]` is slightly problematic, as this is also for `flat_multiset`; not dealt with in this pr. (Is it ok to remove these references?)

stl/inc/flat_set
```diff
@@ -482,94 +481,102 @@ private:
...
+        if constexpr (is_same_v<remove_cvref_t<_Ty>, _Kty>) {
+            _STL_INTERNAL_CHECK(_Can_insert(_Where, _Val));
+            return _Mycont.emplace(_Where, _STD forward<_Ty>(_Val));
+        } else {
+            // heterogeneous insertion
+            // TRANSITION, the requirements may be stricter than needed by the standard.
+            _STL_INTERNAL_STATIC_ASSERT(_Keylt_transparent && is_constructible_v<_Kty, _Ty>);
+            _Kty _Keyval(_STD forward<_Ty>(_Val));
+            _STL_ASSERT(_Can_insert(_Where, _Keyval), "The conversion from the heterogeneous key to key_type should "
+                                                      "be consistent with the heterogeneous lookup!");
```
achabense, line 571: >But we believe that the most common use-case for both heterogeneous lookup and insertion is the case when key_type and heterogeneous key represents the same thing like for std::string and std::string_view, which gives correct and unambiguous conversion from the heterogeneous key to key_type.

https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2363r2.html#dec_conv_comp_consistency

The current implementation works well when `key_type and heterogeneous key represents the same thing`, and the assertion implies this.

It's possible to support weirder (and much less useful) cases (and relax the assertion to only `!contains(_Keyval)`), without breaking the invariant at release mode, by forwarding the `_Keyval` to another `_Emplace_hint` function. This is, however, unfair for common cases, and will make the performance a lot worse. 

What does the standard say about `Conversion and comparison consistency`?

tests/std/tests/P1222R4_flat_set/test.cpp
```diff
@@ -395,14 +395,75 @@ void test_non_static_comparer() {
...
+        always_copy& operator=(const always_copy&) = default;
+
+        // the move ctor & assignment will not empty the container.
+        always_copy(always_copy&& other) noexcept /* intentional */ : always_copy(as_const(other)) {}
+        always_copy& operator=(always_copy&& other) noexcept /* intentional */ {
+            return operator=(as_const(other));
+        }
+    };
+
+    // constexpr int elements[]{1, 2, 3, 4};
```
achabense, line 460: hmm, forgot to remove this

stl/inc/flat_set
```diff
@@ -271,13 +270,10 @@ public:
...
+        if constexpr (requires { _Mycont.append_range(_STD forward<_Rng>(_Range)); }) {
+            _Mycont.append_range(_STD forward<_Rng>(_Range));
         } else {
-            for (const auto& _Val : _Range) {
-                _Cont.insert(_Cont.end(), _Val);
-            }
+            _Mycont.insert_range(_Mycont.end(), _STD forward<_Rng>(_Range));
         }
         _Restore_invariants_after_insert<false>(_Old_size);
     }
```
achabense, line 279:
```
    void insert(initializer_list<_Kty> _Ilist) {
        _Insert_range<false>(_Ilist.begin(), _Ilist.end());
    }

    void insert(_Tsorted, initializer_list<_Kty> _Ilist) {
        _Insert_range<true>(_Ilist.begin(), _Ilist.end());
    }
```
These methods can use `cont.insert(iter, ilist)` for insertion; temporarily kept unchanged for tidiness.

stl/inc/flat_set
```diff
@@ -447,9 +443,9 @@ public:
     }
 
 private:
-    _NODISCARD bool _Check_sorted(const_iterator _It, const const_iterator _End) const {
+    _NODISCARD bool _Is_sorted(const_iterator _It, const const_iterator _End) const {
```
achabense, line 446: It might be better to name this function as `_Is_Tsorted`, if `T` means "type-dependent".
`_Tsorted->_Tsorted_tag`
`_Is_sorted->_Is_Tsorted`
`_Msg_not_sorted->_Msg_not_Tsorted`

However I'm not very sure about the original meaning of `T` in `_Tsorted`...

stl/inc/flat_set
```diff
@@ -291,21 +287,21 @@ public:
...
     _NODISCARD container_type extract() && noexcept(is_nothrow_move_constructible_v<_Container>) /* strengthened */ {
         // always clears the container (N4950 [flat.set.modifiers]/14 and [flat.multiset.modifiers]/10)
-        _Clear_guard<_Base_flat_set, is_nothrow_move_constructible_v<_Container>> _Guard{this};
-        return _STD move(_Get_cont());
+        _Clear_guard<_Base_flat_set> _Guard{this};
+        return _STD move(_Mycont);
     }
     void replace(container_type&& _Cont) {
-        _STL_ASSERT(_Check_sorted(_Cont.cbegin(), _Cont.cend()), _Msg_not_sorted);
-        _Clear_guard<_Base_flat_set, is_nothrow_move_assignable_v<_Container>> _Guard{this};
```
achabense, line 299: This optimization was valid, but I think we'd better put this off after solving the remaining invariant-safety issues in insert/erase methods. (And `"bool _Noexcept = false"` might not be the most suitable form.)

stl/inc/flat_set
```diff
@@ -192,10 +189,10 @@ public:
     }
 
     _NODISCARD const_iterator cbegin() const noexcept {
-        return begin();
+        return _Mycont.cbegin();
     }
     _NODISCARD const_iterator cend() const noexcept {
-        return end();
+        return _Mycont.cend();
```
achabense, line 195: (I begin to believe it a good idea to use a container method as long as it is required in [container.reqmts](https://eel.is/c++draft/container.reqmts) or [sequence.reqmts](https://eel.is/c++draft/sequence.reqmts))

stl/inc/flat_set
```diff
@@ -70,17 +65,17 @@ public:
     static_assert(random_access_iterator<iterator>, "The C++ Standard forbids containers without random "
                                                     "access iterators from being adapted. See [flatset.overview].");
 
-    _Base_flat_set() : _My_pair(_Zero_then_variadic_args_t{}) {}
+    _Base_flat_set() : _Mycont(), _Mycomp() {}
 
+    // TRANSITION, "_My_comp" may need to be copied, even in move construction / assignment.
```
StephanTLavavej, line 70: `_My_comp` is a typo.

StephanTLavavej, line 70: We use TRANSITION for things that need to ship in `main`, but that we hope to improve in the future. For things that must be addressed before this feature is officially merged, it seems that we should use FIXME. The distinction is that we never allow "FIXME" to reach `main`.

tests/std/tests/P1222R4_flat_set/test.cpp
```diff
@@ -408,14 +395,73 @@ void test_non_static_comparer() {
...
+    class always_copy : public vector<int> {
+        using base = vector<int>;
+
+    public:
+        using base::base;
+        always_copy(const always_copy&)            = default;
+        always_copy& operator=(const always_copy&) = default;
+
+        // the move ctor & assignment will not empty the container.
+        always_copy(always_copy&& other) noexcept /* intentional */ : always_copy(as_const(other)) {}
```
StephanTLavavej, line 454: Need `<utility>` for `as_const()`.

tests/std/tests/P1222R4_flat_set/test.cpp
```diff
@@ -408,14 +395,73 @@ void test_non_static_comparer() {
...
+
+    class always_copy : public vector<int> {
+        using base = vector<int>;
+
+    public:
+        using base::base;
+        always_copy(const always_copy&)            = default;
+        always_copy& operator=(const always_copy&) = default;
+
+        // the move ctor & assignment will not empty the container.
```
StephanTLavavej, line 453: Because `&` has a special meaning, we generally avoid using it in comments. We should spell out "and" here.

tests/std/tests/P1222R4_flat_set/test.cpp
```diff
@@ -408,14 +395,73 @@ void test_non_static_comparer() {
...
-void test_extract() {
-    constexpr int elements[]{1, 2, 3, 4};
-    C fs{1, 2, 3, 4};
-    auto cont = std::move(fs).extract();
-    assert(fs.empty());
-    assert(ranges::equal(cont, elements));
+    static bool will_throw = false;
+
+    class test_exception : public vector<int> {
+        using base = vector<int>;
```
StephanTLavavej, line 405: This should explicitly say `private:` before `base`, to follow our usual conventions. Occurs below for `class always_copy`.

StephanTLavavej (approved):

stl/inc/flat_set
StephanTLavavej: (Mentioning this for the future, but I didn't push changes here.)

StephanTLavavej:
:pancakes: :broom: :tada:
===

PR #4067, opened 2023-10-03
Add C4868 to _STL_DISABLED_WARNINGS

tiagomacarios:
[warning C4868: compiler may not enforce left-to-right evaluation order in braced initializer list](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-c4868?view=msvc-170)

Office is seeing instances of C4868 being emitted inside of the STL. Example:
`xutility(5637) : warning C4868: compiler may not enforce left-to-right evaluation order in braced initializer list`

After discussing the issue with Stephan he suggested adding the supression to the STL itself:
```
You can non-intrusively add this to _STL_EXTRA_DISABLED_WARNINGS.
We would also consider a PR to add it to _STL_DISABLED_WARNINGS for
everyone (as we do globally suppress a number of /Wall warnings there).
We basically never have such ordering assumptions in the STL, so this
warning provides no value to us.
```

### DIFF
--- a/stl/inc/yvals_core.h
+++ b/stl/inc/yvals_core.h
@@ -793,6 +793,7 @@
 // warning C4702: unreachable code
 // warning C4793: function compiled as native
 // warning C4820: 'N' bytes padding added after data member 'meow' (/Wall)
+// warning C4868: compiler may not enforce left-to-right evaluation order in braced initializer list
 // warning C4988: variable declared outside class/function scope (/Wall /d1WarnOnGlobals)
 // warning C5026: move constructor was implicitly defined as deleted (/Wall)
 // warning C5027: move assignment operator was implicitly defined as deleted (/Wall)
@@ -805,8 +806,8 @@
 // clang-format off
 #define _STL_DISABLED_WARNINGS                        \
     4180 4412 4455 4494 4514 4574 4582 4583 4587 4588 \
-    4619 4623 4625 4626 4643 4648 4702 4793 4820 4988 \
-    5026 5027 5045 5220 6294                          \
+    4619 4623 4625 4626 4643 4648 4702 4793 4820 4868 \
+    4988 5026 5027 5045 5220 6294                     \
     _STL_DISABLED_WARNING_C4577                       \
     _STL_DISABLED_WARNING_C4984                       \
     _STL_DISABLED_WARNING_C5053                       \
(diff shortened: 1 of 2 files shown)


strega-nil-ms (approved):

stl/inc/yvals_core.h
```diff
@@ -793,6 +793,7 @@
 // warning C4702: unreachable code
 // warning C4793: function compiled as native
 // warning C4820: 'N' bytes padding added after data member 'meow' (/Wall)
+// warning C4868: compiler may not enforce left-to-right evaluation order in braced initializer list
```
StephanTLavavej, line 796: These comments say "(/Wall)" for off-by-default warnings to remind us of their nature, so we should follow that pattern here.

stl/inc/yvals_core.h
```diff
@@ -805,8 +806,8 @@
 // clang-format off
 #define _STL_DISABLED_WARNINGS                        \
     4180 4412 4455 4494 4514 4574 4582 4583 4587 4588 \
-    4619 4623 4625 4626 4643 4648 4702 4793 4820 4988 \
-    5026 5027 5045 5220 6294                          \
+    4619 4623 4625 4626 4643 4648 4702 4793 4820 4868 \
```
StephanTLavavej, line 809: There are 3 suppressions in `<execution>` which can now be removed, further reinforcing the goodness of this change.

StephanTLavavej (approved):

StephanTLavavej:
Thanks!

@CaseyCarter @strega-nil-ms I pushed changes after you approved. I had forgotten that `<execution>` locally disabled this warning, but I have a heuristic "does anything else mention this" when making changes of this nature, so searching the codebase turned up these mentions of `4868`. (Ctrl+Shift+F is incredibly powerful :magic_wand:) At that point I noticed that `<execution>`'s comments said "(/Wall)", which in theory I could have noticed by looking at `<yvals_core.h>` alone but I missed that on first glance too. Just explaining my thought process here!

CaseyCarter (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for silencing this annoying warning and congratulations on your first microsoft/STL commit! :tada: :rocket: :smile_cat:

PR #4068, opened 2023-10-03
Add `*` tags to all non-tagged configs

CaseyCarter:
... in our test matrices.

`RUNALL_CROSSLIST` forms the cross product of all lines from above with all lines from below, and from each pairing whose tag sets intersect produces a new line tagged with the intersection that defines all variables defined on either line, with values that are the concatenation of the values from `x` and `y`. Non-tagged lines are treated as if they are all tagged with the same value, and `*` is the identity element for tag set intersection. For example:

```
	MEOW=A
X	MEOW=B
Y	MEOW=C
*	MEOW=D
RUNALL_CROSSLIST
	WOOF=1
X,Y	WOOF=2
X	WOOF=3
Y	MEOW=E WOOF=4
```
is equivalent to:
```
	MEOW=A WOOF=1
X	MEOW=B WOOF=2
X	MEOW=B WOOF=3
Y	MEOW=C WOOF=2
Y	MEOW="C E" WOOF=4
	MEOW=D WOOF=1
X,Y	MEOW=D WOOF=2
X	MEOW=D WOOF=3
Y	MEOW="D E" WOOF=4
```

[GH-4052](https://github.com/microsoft/STL/pull/4052) added `*` tags to all non-leaf lines so they would properly combine with leaf configurations that have either an `ASAN` or empty tag. There's a danger of "losing" `ASAN` configurations if we add a crosslist that inadvertently matches untagged lines with `ASAN` lines. We decided that all lines should be tagged with either `ASAN` or `*` to lessen the likelihood of such a mistake.


StephanTLavavej (approved):

StephanTLavavej:
I searched carefully for any missed lines and didn't find them.

StephanTLavavej:
There are no stealth conflicts with the Ready To Merge PRs (a few are adding `env.lst`s that just `RUNALL_INCLUDE` the normal ones). As this is test-only I can add it to the batch and get this out of the way. Thanks!

CaseyCarter:
I agree, this should not be problematic. I manually queued an ASAN run for good measure (https://dev.azure.com/vclibs/STL/_build/results?buildId=15321&view=results) and it is nicely green.

StephanTLavavej:
:star: :star2: :star_struck:
===

PR #2795, opened 2022-06-16
Fix non-deterministic test sharding with `--order=lexical`

StephanTLavavej:
Fixes #2794, which silently degraded our test coverage.

Thanks to @CaseyCarter for investigating the bizarre symptoms and finding [this documented option](https://llvm.org/docs/CommandGuide/lit.html#cmdoption-lit-order).

PR #4069, opened 2023-10-04
build and test infra cleanup

CaseyCarter:
* Extract common pipeline config into a new `azure-devops/config.yml` variable template so we don't need to repeat pool info and target directories in both pipelines.
* Improve the `displayName`s of all the pipeline stages.
* Use `trigger: none` so ASan pipeline runs don't trigger for pushes to _any_ branch. We _do_ want to schedule a daily build, so add `schedules`.
* `benchmarkBuildOutputLocationVar` was actually used to control building the benchmarks; replace with a boolean `buildBenchmarks`.
* We need to explicitly remove `asan-pipeline`'s second stage's dependency on the first with `dependsOn: []`.
* We never set `buildOutputLocationVar` to a non-default value; get rid of it.
* Make `litFlags` a defaulted parameter of `cmake-configure-build.yml` so we don't need to duplicate it.
  + Also remove `--shuffle`, which was overridden by `--order=lexical`; this should have been done in #2795.
* Don't try to build the benchmarks if configuring the benchmarks fails.
* Consistently quote paths in `azure-devops/cmake-configure-build.yml`.

## Reminder: We'll need to update https://github.com/microsoft/STL/wiki/Checklist-for-Toolset-Updates after this merges, the pool name now appears in `azure-devops/config.yml`.

### DIFF
--- a/azure-devops/asan-pipeline.yml
+++ b/azure-devops/asan-pipeline.yml
@@ -4,36 +4,46 @@
 # Build STL targeting x86 and x64, and run extra ASan testing
 
 variables:
-  tmpDir: 'D:\Temp'
-  buildOutputLocation: 'D:\build'
-  # Restrict to "stlasan" test
-  testSelection: '-R stlasan'
-
-pool:
-  name: 'StlBuild-2023-09-14T1251-Pool'
-  demands: EnableSpotVM -equals true
+  - template: config.yml
+  - name: testSelection
+    value: '-R stlasan' # Restrict to "stlasan" test
 
 pr: none
+trigger: none
+
+schedules:
+- cron: '0 2 * * *'
+  displayName: ASan-Daily-CI
+  branches:
+    include:
+    - main
 
 stages:
   - stage: Build_And_Test_x64
-    displayName: 'Build and Test'
+    displayName: 'Build and Test x64'
+    pool:
+      name: ${{ variables.poolName }}
+      demands: ${{ variables.poolDemands }}
     jobs:
       - template: native-build-test.yml
         parameters:
           targetPlatform: x64
           vsDevCmdArch: amd64
-          benchmarkBuildOutputLocationVar: ''
+          buildBenchmarks: false
           testSelection: ${{ variables.testSelection }}
 
   - stage: Build_And_Test_x86
-    displayName: 'Build and Test'
+    displayName: 'Build and Test x86'
+    dependsOn: []
+    pool:
+      name: ${{ variables.poolName }}
+      demands: ${{ variables.poolDemands }}
     jobs:
       - template: native-build-test.yml
         parameters:
           targetPlatform: x86
           vsDevCmdArch: x86
-          benchmarkBuildOutputLocationVar: ''
+          buildBenchmarks: false
           testSelection: ${{ variables.testSelection }}
 
   # no coverage for ARM and ARM64
--- a/azure-devops/config.yml
+++ b/azure-devops/config.yml
@@ -1,0 +1,18 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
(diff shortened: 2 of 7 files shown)


azure-devops/cross-build.yml
```diff
@@ -9,23 +9,14 @@ parameters:
...
+- name: buildBenchmarks
+  type: boolean
+  default: true
 - name: numShards
   type: number
   default: 8
 jobs:
 - job: '${{ parameters.targetPlatform }}'
-  variables:
-    fixedFlags: '--timeout=240;--shuffle'
```
CaseyCarter, line 24: I forgot to mention in the commit log that I removed this `--shuffle` flag that is overridden by the later `--order=lexical`.

azure-devops/asan-pipeline.yml
```diff
@@ -4,36 +4,46 @@
...
+    include:
+    - main
 
 stages:
   - stage: Build_And_Test_x64
-    displayName: 'Build and Test'
+    displayName: 'Build and Test x64'
+    pool:
+      name: ${{ variables.poolName }}
+      demands: ${{ variables.poolDemands }}
```
CaseyCarter, line 26: Yes, the `pool` info must be repeated in each stage like this: ["Expressions are only expanded for stages, jobs, steps, and containers (inside resources)."](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/template-expressions?view=azure-devops).

azure-devops/asan-pipeline.yml
```diff
@@ -4,36 +4,46 @@
...
-  demands: EnableSpotVM -equals true
+  - template: config.yml
+  - name: testSelection
+    value: '-R stlasan' # Restrict to "stlasan" test
 
 pr: none
+trigger: none
+
+schedules:
+- cron: '0 2 * * *'
```
CaseyCarter, line 15: This starts a run every day at 02:00 UTC if there's been a change since the last scheduled run and no run is currently in progress (which should never matter: our runs never take 24 hours). I'm not married to 02:00, I chose it arbitrarily.

azure-devops/config.yml
```diff
@@ -0,0 +1,18 @@
...
+  readonly: true
+- name: poolDemands
+  value: 'EnableSpotVM -equals true'
+  readonly: true
+- name: tmpDir
+  value: 'D:\Temp'
+- name: buildOutputLocation
+  value: 'D:\build'
+- name: benchmarkBuildOutputLocation
+  value: 'D:\benchmark'
```
StephanTLavavej, line 18: It appears that we never overwrite `tmpDir`, `buildOutputLocation`, and `benchmarkBuildOutputLocation`; should they also be `readonly: true`? For variables transferred between YAML files, I think there's extra value in being "const correct".

I'll push changes for this, since I think it's simple enough to not reset your extensive manual validation.

azure-devops/cross-build.yml
StephanTLavavej: Ooh, good cleanup, this was potentially hazardous if never overridden.

StephanTLavavej: Updated the PR description to mention this. Also mentioned the improved path quoting that I noticed elsewhere.

StephanTLavavej (approved):

StephanTLavavej:

azure-devops/cross-build.yml
CaseyCarter: > Ooh, good cleanup, this was potentially hazardous if never overridden.

Yes, I don't know what we were thinking in https://github.com/microsoft/STL/pull/2795.

StephanTLavavej:
Thanks for setting up this daily ASan CI! I've updated the toolset checklist. :white_check_mark: :calendar: :tada:

CaseyCarter:
> I've updated the toolset checklist. ✅ 📆 🎉

Great, thanks.

PR #4075, opened 2023-10-06
Fix ASAN CI failure caused by stealth merge conflict

StephanTLavavej:
Originally, these `ctad.static.compile.pass.cpp` tests were marked as failing for all compilers. After #4053, they were marked as failing for MSVC only (configuration 0). After #4069 finished setting up the ASan-Daily-CI pipeline, these tests additionally need to be marked as failing for MSVC with ASAN (configuration 1).


CaseyCarter (approved):

CaseyCarter:
As a general rule, we should have a fail for `:1` if we have a fail for `:0` unless the failure is a bug in MSVC's static analyzer, which isn't run in `:1` configs.

StephanTLavavej:

PR #4083, opened 2023-10-11
Toolset update: VS 2022 17.8 Preview 3

StephanTLavavej:
* Updated dependencies.
    + Updated build compiler to VS 2022 17.8 Preview 3.
    + Updated Python to 3.12.0 (now required).
* Remove compiler bug workarounds:
  + Remove workaround for VSO-1826196 "False positive C26454 in charconv" as the internal toolset has been updated. Validated internally, we can resolve this bug when merging.
  + Drop TRANSITION comment about VSO-251998 "File issue for `cbegin` interaction with other std containers with only non-member `begin()` functions". This wasn't a compiler or STL bug; it was a reminder to maybe file an LWG issue someday, which never happened and isn't worth pursuing now.
  + VSO-1281038 "False C6287 for C++20 synthesized operators" was rightfully resolved as By Design. Permanently avoid it by separately asserting each expression, which is our preferred style anyways.
    - Although we generally avoid cleaning up `tr1`, this is worth it.
  + Remove workaround for VSO-1538014 "\[Feedback\] ARM64 `memchr` implementation may not be standard conformant". It was fixed on 2023-06-02 by MSVC-PR-468038 in `prod/be` and is present in both `rel/17.7` and `rel/17.8`. It was marked as fixed in 17.8 Preview 1.
  + Replace coarse-grained workaround for VSO-1026729 "Conformant Copy Elision" (a User Story that's been Proposed since 2019-11-22) with fine-grained workarounds for VSO-1899423 "C1XX mishandles C++17 deferred temporary materialization in `is_invocable_r_v` and `is_nothrow_invocable_r_v`" (newly reported with this exact code).
* Use PowerShell 7.3.8 to create the pool.

### DIFF
--- a/tests/std/tests/VSO_0000000_instantiate_containers/test.compile.pass.cpp
+++ b/tests/std/tests/VSO_0000000_instantiate_containers/test.compile.pass.cpp
@@ -702,7 +702,6 @@ void extended_math_functions_test(T value) {
     (void) tanh(value);
 }
 
-// TRANSITION, VSO-251998
 // cbegin() calls begin() fully qualified, so cannot
 // find valarray non-member begin().
 
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ Just try to follow these rules, so we can spend more time fixing bugs and implem
 
 # How To Build With The Visual Studio IDE
 
-1. Install Visual Studio 2022 17.8 Preview 2 or later.
+1. Install Visual Studio 2022 17.8 Preview 3 or later.
     * Select "Windows 11 SDK (10.0.22000.0)" in the VS Installer.
     * We recommend selecting "C++ CMake tools for Windows" in the VS Installer.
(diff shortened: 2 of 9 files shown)


tests/std/tests/VSO_0000000_instantiate_containers/test.compile.pass.cpp
```diff
@@ -702,7 +702,6 @@ void extended_math_functions_test(T value) {
     (void) tanh(value);
 }
 
-// TRANSITION, VSO-251998
```
CaseyCarter, line 705: I've reported this to the LWG and LEWG reflectors. I personally don't think it's worth the trouble to fix, but now I don't have to feel guilty about not doing so.

CaseyCarter (approved):

StephanTLavavej:

PR #1356, opened 2020-10-08
_Execute_once bug fix.

jrabet:
The original XFG enlightening for _Execute_once tried to be clever by using Context to pass the final callback to the trampoline. Unfortunately, Context is used by InitOnceExecuteOnce itself and having the wrong bits set can result in an error being returned. This PR addresses this by defining a struct which contains the parameter value and callback pointer to pass those to the trampoline.

PR #1318, opened 2020-09-25
XFG: made _Execute_once XFG-compatible.

jrabet:
_Execute_once_fp_t and PINIT_ONCE_FN differ in type signature, therefore we introduce _Xfg_trampoline which has PINIT_ONCE_FN's type signature and calls into _Callback as an _Execute_once_fp_t for XFG compatibility.

PR #4087, opened 2023-10-12
Cleanup `_Execute_once()`

StephanTLavavej:
* Revert #1318, #1356, and #1398 (`xonce.cpp` only) while preserving #1194's change to directly call `InitOnceExecuteOnce()`.
  + XFG is being removed from the toolset (see internal VSO-1726733, MSVC-PR-462377). Some things like ASan still need to handle XFG hashes, but things like the STL no longer need unusual contortions. I checked with @amyw-msft and she said this should be okay to remove.
* Drop `_Execute_once()` usage in `ppltasks.cpp`, and as an associated change, simplify `dllmain.cpp`.
  + See below.
* Use `call_once()` instead of `_Execute_once()` in `excptptr.cpp`.
  + After #3861, they're both declared by `xcall_once.h`, so we don't need to use the internal `_Execute_once()` (with a worse interface) anymore.
  + We're `using namespace std;` so we don't need qualification.
  + Because `call_once()` supports stateful lambdas, we can simply capture `_Storage` by reference.
  + We're in `stl/src` so we don't really need to worry about True Placement New being hijacked, but let's `static_cast<void*>` to preserve the original code and make it clear that we are indeed invoking True Placement New.
* Remove declarations from `xcall_once.h`.
  + They were guarded by `_CRTBLD` after #3935, so this doesn't affect users.
  + The definition of `_Execute_once()` exactly repeated the declaration, except for `extern "C++"` which is fine to drop because `stl/src` is always built classically.
* Update comment now that `_Execute_once()` is unused.
  + Also drop pointless "`_Execute_once` function" comment.

The code removal in `ppltasks.cpp` looks like I'm just thoughtlessly annihilating a bunch of weird logic, but I actually performed it carefully step-by-step. This code was `_CRT_APP`-only, for "async causality tracing". As far as I can tell (looking up TFS changesets 923210 on 2013-04-02 and 1175213 on 2014-02-07, as this code was introduced and moved around), this was implemented in the hopes of improving debugging, back when ppltasks was envisioned as a top-level feature and the new major model for concurrency, instead of the minor implementation detail of `<future>` that it eventually ended up as. The async causality logging mechanism never existed for Desktop, and was conditionally available for App (it depended on OS version). That is, it's not necessary for any of `<future>`'s work, and this was so undocumented, I strongly doubt that anyone aside from the original developer in 2013-2014 ever used it (I certainly haven't).

Attempting to use real `std::call_once()` here ran into problems, because App didn't have access to the necessary Windows API functions that power it instead of `_Execute_once()`. In theory that could have been solvable with enough MSBuild effort, but I think that we can simply remove this code, and I'm willing to take the risk and bear any potential servicing burden if I'm wrong.

This doesn't affect bincompat because it's internal to `stl/src` and There's Only One STL (`stl/src` is always consistent with itself). I also verified that the exports of `binaries\x86chk\bin\i386\app\msvcp140d_app.dll` are unchanged.

The steps that I followed started by considering what would happen if the causality APIs weren't detected as present, and then performing logical transformations after that.

* Step 1: Suppose we can never get the causality APIs.
  + In `AsyncCausalityTracer`, `m_causalityAPIs` and `m_isSupported` are private data members, initialized to `nullptr` and `false`. Let's pretend that `GetActivationFactory()` fails. In that case, the data members remain `nullptr` and `false` forever. Accordingly, we can hardcode the return values of `get()` and `isCausalitySupported()`. Also, `release()` becomes a no-op.
* Step 2: There are no causality static factories to cleanup.
  + Now that `asyncCausalityTracer.release()` is a no-op, `__crtCleanupCausalityStaticFactories()` does nothing. Note that it wasn't exported - it was used only for cross-TU interaction between `ppltasks.cpp` and `dllmain.cpp` in `stl/src`, so we can remove this function outright. Then, `dllmain.cpp` becomes a stub.
* Step 3: `asyncCausalityTracer.isCausalitySupported()` is always `false` now.
  + Accordingly, we can drop all of these branches. Note that this drops `_M_scheduled = true;`.
* Step 4: `_M_scheduled` is always `false` now.
  + In `<ppltasks.h>`, `_TaskEventLogger`'s constructor assigned `_M_scheduled = false;`, and we previously removed the only way for it to be set to `true`. Accordingly, we can drop this branch.
* Step 5: Remove unused code.
  + From bottom to top:
  + `const GUID PPLTaskCausalityPlatformID` wasn't exported. It had internal linkage, so we can definitely drop it.
  + Nothing mentions the `AsyncCausalityTracer` type or the `asyncCausalityTracer` global variable anymore.
  + Finally, we don't need the using-directives here.
* Step 6: Unify Vulcan and Romulus.
  + The app implementation of `_TaskEventLogger` is now identical to the desktop implementation (after we drop the `_IsContinuation` parameter name). Unify them.
* Pre-merge with #4106 by marking `extern "C"` `DllMain()` as `noexcept`.
  + Also drop the commented-out parameter names to avoid wrapping; they aren't useful when the function does nothing.

### DIFF
--- a/stl/inc/xcall_once.h
+++ b/stl/inc/xcall_once.h
@@ -28,15 +28,6 @@ _EXPORT_STD struct once_flag { // opaque data structure for call_once()
     void* _Opaque;
 };
 
-#ifdef _CRTBLD
-// Returns BOOL, nonzero to indicate success, zero for failure
-using _Execute_once_fp_t = int(__stdcall*)(void*, void*, void**);
-
-// Returns BOOL, nonzero to indicate success, zero for failure
-extern "C++" _CRTIMP2_PURE int __CLRCALL_PURE_OR_CDECL _Execute_once(
-    once_flag& _Flag, _Execute_once_fp_t _Callback, void* _Pv) noexcept;
-#endif // _CRTBLD
-
 template <class _Ty>
 union _Immortalizer_impl { // constructs _Ty, never destroys
     constexpr _Immortalizer_impl() noexcept : _Storage{} {}
--- a/stl/src/dllmain.cpp
+++ b/stl/src/dllmain.cpp
@@ -5,18 +5,6 @@
 
 #include <Windows.h>
 
-#ifdef _CRT_APP
-// free static resource used by causality
-extern "C" void __cdecl __crtCleanupCausalityStaticFactories();
-#endif // defined(_CRT_APP)
-
-extern "C" BOOL APIENTRY DllMain(HMODULE /* hModule */, DWORD ul_reason_for_call, [[maybe_unused]] LPVOID lpReserved) {
-    if (ul_reason_for_call == DLL_PROCESS_DETACH) {
-#ifdef _CRT_APP
-        if (lpReserved == nullptr) { // only when the process is not terminating
-            __crtCleanupCausalityStaticFactories();
-        }
-#endif // defined(_CRT_APP)
-    }
+extern "C" BOOL APIENTRY DllMain(HMODULE /* hModule */, DWORD /* ul_reason_for_call */, LPVOID /* lpReserved */) {
     return TRUE;
 }
--- a/stl/src/dllmain.cpp
+++ b/stl/src/dllmain.cpp
@@ -5,6 +5,6 @@
 
 #include <Windows.h>
 
-extern "C" BOOL APIENTRY DllMain(HMODULE /* hModule */, DWORD /* ul_reason_for_call */, LPVOID /* lpReserved */) {
+extern "C" BOOL APIENTRY DllMain(HMODULE, DWORD, LPVOID) noexcept {
     return TRUE;
 }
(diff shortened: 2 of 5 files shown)


AlexGuteniev:
What is XFG?

StephanTLavavej:
According to my limited understanding, it was a variant of the control flow guard feature, but was abandoned because it proved to be unworkable in practice.

AlexGuteniev (approved):

CaseyCarter (approved):

CaseyCarter:
>   * Also drop pointless "`_Execute_once` function" comment.

🔥🔥🔥 Yes, YES! 🔥🔥🔥

StephanTLavavej:

StephanTLavavej:
Pulling this back for rework, I need to deal with the app-only logic in `ppltasks.cpp`.

CaseyCarter (approved):

StephanTLavavej:

PR #4094, opened 2023-10-16
Use same alignment idiom as rest of STL codebase

AtariDreams:
In addition to being faster, I wrote comments making it easy to follow.


StephanTLavavej:
We don't feel that this change is beneficial enough to justify the work needed to review it for correctness, and the inherent risk that comes from codebase churn.

The current comment seems sufficiently clear to us, and as mentioned in https://github.com/microsoft/STL/pull/4063#issuecomment-1747647304 the performance of this function is not important.

PR #4098, opened 2023-10-17
Turn niebloids into function objects

### LINKED ISSUE #4097: `_Not_quite_object` has outlived its usefulness
The Standard denotes some overload sets of function templates as having special properties which make it possible to implement them as function objects. For example [algorithms.requirements]/2:

> The entities defined in the `std::ranges` namespace in this Clause are not found by argument-dependent name lookup ([basic.lookup.argdep]). When found by unqualified ([basic.lookup.unqual]) name lookup for the postfix-expression in a function call ([expr.call]), they inhibit argument-dependent name lookup. [ed: Example omitted for brevity.]

together with [algorithms.requirements]/15:

> The well-formedness and behavior of a call to an algorithm with an explicitly-specified template argument list is unspecified, except where explicitly stated otherwise.

provides such properties for the algorithms implemented in namespace `std::ranges`. The Standard doesn't make these explicitly function objects in the hopes that C++ will eventually provide core language support for overload sets as first-class objects, mechanisms to forbid finding selected function( template)s via ADL, and mechanisms for selected function( template)s to inhibit ADL when found by normal unqualified name lookup. The fear is that allowing users to take advantage of the fact that these algorithms are implemented as function objects today may create breakage if and when they transition into overload sets of function templates using future tech to provide the desired properties. 

These overload sets not guaranteed to be implemented via function objects but in practice implemented exactly so are colloquially known as niebloids, much to the chagrin of Eric Niebler who wanted them to simply be function objects until being swayed by my starry-eyed naivete (as a relatively new WG21 member) into believing we could change the language in a reasonable timeframe. The STL uses an internal type `_Not_quite_object`:

https://github.com/microsoft/STL/blob/1c59a205787f4fdcd9c20decc8fa6dfa9096d394/stl/inc/xutility#L3220-L3243

to help define niebloids with as little object-like behavior as possible to avoid users depending on their object-ness.

Fast-forward to 2023. No core language support for niebloids has materialized, nor is likely to do so in time for C++26. Both libstdc++ and libc++ have implemented the niebloids as simple function objects. C++20 code in the wild has begun to depend on that object-ness, and the STL is beginning to receive bug reports like DevCom-10450794 for code that compiles fine with GCC and Clang (well, it would if libc++ had shipped `chunk_by`). Niebloids have failed both to encourage C++ core language support, and to prevent users from depending on implementation as function objects. 

There seems to be no reason to keep our `_Not_quite_object` around; I propose that we remove it and the resulting implementation divergence.

CaseyCarter:
The core language doesn't seem to be any closer to providing first-class overload sets, and we're starting to get "bug" reports (DevCom-10450794) since the other standard library implementations simply implement these as (semiregular?) function objects.

This change removes the type `std::_Not_quite_object`, promoting its member `_Construct_tag` to namespace scope since it is also used in the implementation of `join_view` and `join_with_view`.

Fixes #4097

### DIFF
--- a/stl/inc/algorithm
+++ b/stl/inc/algorithm
@@ -292,10 +292,8 @@ namespace ranges {
     _EXPORT_STD template <class _In, class _Fun>
     using for_each_result = in_fun_result<_In, _Fun>;
 
-    class _For_each_fn : private _Not_quite_object {
+    class _For_each_fn {
     public:
-        using _Not_quite_object::_Not_quite_object;
-
         template <input_iterator _It, sentinel_for<_It> _Se, class _Pj = identity,
             indirectly_unary_invocable<projected<_It, _Pj>> _Fn>
         constexpr for_each_result<_It, _Fn> operator()(_It _First, _Se _Last, _Fn _Func, _Pj _Proj = {}) const {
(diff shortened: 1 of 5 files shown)


fsb4000:
Am I correct and this change also fixes #3995 ?

CaseyCarter:
> Am I correct and this change also fixes #3995 ?

I don't think we can "fix" something that's not a bug, but I can confirm that the repro from #3995 compiles and runs just fine with this change.

StephanTLavavej:

strega-nil-ms (approved):
I'm very happy about this haha

StephanTLavavej:
Thanks for increasing global implementation consistency! :earth_africa: :earth_americas: :earth_asia:

PR #4103, opened 2023-10-19
Fix and simplify the feature-test macro test

StephanTLavavej:
* Drop extra spaces in `#error` messages.
* Mechanically simplify.
  + This test originally covered both the compiler and library feature-test macros, and was therefore extremely paranoid about checking the behavior of the preprocessor. We gave the compiler test coverage to the compiler team a long time ago, so we can simplify the library test coverage now. This commit is a 100% regex search-and-replace, from:
    ```cpp
    #ifndef (__cpp_lib_\w+)
    #error \1 is not defined
    #elif \1 != (\d{6}L)
    #error \1 is not \2
    #else
    STATIC_ASSERT\(\1 == \2\);
    #endif
    ```
    to:
    ```cpp
    #ifdef $1
    STATIC_ASSERT($1 == $2);
    #else
    #error $1 is not defined
    #endif
    ```
    except with `\n` between each line of the "from" and "to" (presented here on separate lines for clarity).
* Simplify `__cpp_lib_chrono` and `__cpp_lib_shared_ptr_arrays`.
  + These are always defined, but to varying values, so they use a slightly different pattern.
* Simplify `__cpp_lib_shift`.
  + This was trying to be extra helpful with a message saying "is OLD_VALUE when it should be NEW_VALUE", but that made the code verbose and unsystematic. This test is passing, so we don't need elaborate error messages in case of failure.
* Fix damaged `__cpp_lib_containers_ranges` test.
  + This would never have detected `__cpp_lib_containers_ranges` being defined outside of the relevant mode.
* Mechanically simplify again (100% regex).
  + From:
    ```cpp
    #ifdef (__cpp_lib_\w+)
    (STATIC_ASSERT\(\1 == \d{6}L\);)
    #else
    #error \1 is not defined
    #endif
    ```
    to:
    ```
    $2
    ```
* Manually simplify `__cpp_lib_chrono` and `__cpp_lib_shared_ptr_arrays` again.
  + This removes all occurrences of "is not defined".
* Mechanically simplify yet again (100% regex).
  + From:
    ```cpp
    #else
    #ifdef (__cpp_lib_\w+)
    #error \1 is defined
    #endif
    #endif
    ```
    to:
    ```cpp
    #elif defined($1)
    #error $1 is defined
    #endif
    ```
* Manually simplify `__cpp_lib_shared_timed_mutex`.
  + Testing for non-`/clr:pure` mode allows this to follow the consistent pattern.
* Fix damaged `__cpp_lib_is_layout_compatible` and `__cpp_lib_is_pointer_interconvertible` tests.
  + These tests weren't doing anything for C++14/17 mode. Now they'll detect the macros being present when they shouldn't be.
* DROP SHOUTY COMMENT.

### DIFF
--- a/tests/std/tests/VSO_0157762_feature_test_macros/test.compile.pass.cpp
+++ b/tests/std/tests/VSO_0157762_feature_test_macros/test.compile.pass.cpp
@@ -5,7 +5,6 @@
 
 #define STATIC_ASSERT(...) static_assert(__VA_ARGS__, #__VA_ARGS__)
 
-// LIBRARY FEATURE-TEST MACROS
 #if _HAS_CXX23
 STATIC_ASSERT(__cpp_lib_adaptor_iterator_pair_constructor == 202106L);
 #elif defined(__cpp_lib_adaptor_iterator_pair_constructor)
--- a/tests/std/tests/VSO_0157762_feature_test_macros/test.compile.pass.cpp
+++ b/tests/std/tests/VSO_0157762_feature_test_macros/test.compile.pass.cpp
@@ -7,12 +7,10 @@
 
 // LIBRARY FEATURE-TEST MACROS
 #if _HAS_CXX23
-#ifndef __cpp_lib_adaptor_iterator_pair_constructor
-#error __cpp_lib_adaptor_iterator_pair_constructor is not defined
-#elif __cpp_lib_adaptor_iterator_pair_constructor != 202106L
-#error __cpp_lib_adaptor_iterator_pair_constructor is not 202106L
-#else
+#ifdef __cpp_lib_adaptor_iterator_pair_constructor
 STATIC_ASSERT(__cpp_lib_adaptor_iterator_pair_constructor == 202106L);
+#else
+#error __cpp_lib_adaptor_iterator_pair_constructor is not defined
 #endif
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
I suppose we could further simplify
```cpp
#ifdef $1
STATIC_ASSERT($1 == $2);
#else
#error $1 is not defined
#endif
```
to:
```cpp
STATIC_ASSERT($1 == $2);
```
and accept that if the feature-test macro isn't defined, we'll be mentioning an unknown identifier that will fail to compile. Thoughts on such a further simplification?

CaseyCarter:
> Thoughts on such a further simplification?

It wouldn't have the most user-friendly error messages, but failures are so rare in practice that I don't think it would be a problem. I'd be happy with such a change. I wouldn't _demand_ it, but it sure would be nice to trim this down to one line per test case.

CaseyCarter:
The fact that my bug sat here latent for 14 months suggests that simplifying the structure is a good idea.

StephanTLavavej:
Cool, I'll make the change. One moment...

CaseyCarter (approved):
This is so much nicer - thanks!

StephanTLavavej:

PR #4106, opened 2023-10-19
make `extern "C"` functions explicitly `noexcept`

CaseyCarter:
Functions declared `extern "C"` are implicitly `noexcept(true)` when compiling with `/EHc`, and implicitly `noexcept(false)` when compiling without. To ensure consistency regardless of the compiler's command line options, all `extern "C"` functions should have explicit *noexcept-specifier*s.

This change excludes (with explanatory comment) the redeclarations of `locale.h` functions in `xlocinfo` to avoid clashing with the primary declarations in the UCRT.

Drive-by:
* In `mutex.cpp`, indicate that `_Thrd_abort` is exported for ABI only; there are no callers outside this file.
* In `winapisupp.cpp`, the dynamic initializer `initialize_pointers` never throws exceptions and should therefore be `noexcept`.
* In `vector_algorithms.cpp`, the APIs are `noexcept` so the helpers must be as well. Add annotations to make that contractual obligation explicit.

### DIFF
--- a/stl/inc/execution
+++ b/stl/inc/execution
@@ -45,7 +45,7 @@ using __std_PTP_CALLBACK_ENVIRON  = __std_TP_CALLBACK_ENVIRON*;
 _NODISCARD unsigned int __stdcall __std_parallel_algorithms_hw_threads() noexcept;
 
 using __std_PTP_WORK_CALLBACK = void(__stdcall*)(
-    _Inout_ __std_PTP_CALLBACK_INSTANCE, _Inout_opt_ void*, _Inout_ __std_PTP_WORK);
+    _Inout_ __std_PTP_CALLBACK_INSTANCE, _Inout_opt_ void*, _Inout_ __std_PTP_WORK) _NOEXCEPT_FNPTR;
 
 _NODISCARD __std_PTP_WORK __stdcall __std_create_threadpool_work(
     _In_ __std_PTP_WORK_CALLBACK, _Inout_opt_ void*, _In_opt_ __std_PTP_CALLBACK_ENVIRON) noexcept;
--- a/stl/inc/execution
+++ b/stl/inc/execution
@@ -45,7 +45,7 @@ using __std_PTP_CALLBACK_ENVIRON  = __std_TP_CALLBACK_ENVIRON*;
 _NODISCARD unsigned int __stdcall __std_parallel_algorithms_hw_threads() noexcept;
 
 using __std_PTP_WORK_CALLBACK = void(__stdcall*)(
-    _Inout_ __std_PTP_CALLBACK_INSTANCE, _Inout_opt_ void*, _Inout_ __std_PTP_WORK) _NOEXCEPT_FNPTR;
+    _Inout_ __std_PTP_CALLBACK_INSTANCE, _Inout_opt_ void*, _Inout_ __std_PTP_WORK);
 
 _NODISCARD __std_PTP_WORK __stdcall __std_create_threadpool_work(
     _In_ __std_PTP_WORK_CALLBACK, _Inout_opt_ void*, _In_opt_ __std_PTP_CALLBACK_ENVIRON) noexcept;
--- a/stl/src/dllmain.cpp
+++ b/stl/src/dllmain.cpp
@@ -7,10 +7,11 @@
 
 #ifdef _CRT_APP
 // free static resource used by causality
-extern "C" void __cdecl __crtCleanupCausalityStaticFactories();
+extern "C" void __cdecl __crtCleanupCausalityStaticFactories() noexcept;
 #endif // defined(_CRT_APP)
 
-extern "C" BOOL APIENTRY DllMain(HMODULE /* hModule */, DWORD ul_reason_for_call, [[maybe_unused]] LPVOID lpReserved) {
+extern "C" BOOL APIENTRY DllMain(
+    HMODULE /* hModule */, DWORD ul_reason_for_call, [[maybe_unused]] LPVOID lpReserved) noexcept {
     if (ul_reason_for_call == DLL_PROCESS_DETACH) {
 #ifdef _CRT_APP
         if (lpReserved == nullptr) { // only when the process is not terminating
--- a/stl/src/dllmain.cpp
+++ b/stl/src/dllmain.cpp
@@ -7,11 +7,10 @@
 
 #ifdef _CRT_APP
 // free static resource used by causality
-extern "C" void __cdecl __crtCleanupCausalityStaticFactories() noexcept;
+extern "C" void __cdecl __crtCleanupCausalityStaticFactories();
 #endif // defined(_CRT_APP)
 
-extern "C" BOOL APIENTRY DllMain(
-    HMODULE /* hModule */, DWORD ul_reason_for_call, [[maybe_unused]] LPVOID lpReserved) noexcept {
+extern "C" BOOL APIENTRY DllMain(HMODULE /* hModule */, DWORD ul_reason_for_call, [[maybe_unused]] LPVOID lpReserved) {
     if (ul_reason_for_call == DLL_PROCESS_DETACH) {
 #ifdef _CRT_APP
         if (lpReserved == nullptr) { // only when the process is not terminating
--- a/stl/src/excptptr.cpp
+++ b/stl/src/excptptr.cpp
@@ -32,7 +32,7 @@
 // V4 and later managed exception code
 #define MANAGED_EXCEPTION_CODE_V4 0XE0434352
 
-extern "C" _CRTIMP2 void* __cdecl __AdjustPointer(void*, const PMD&) noexcept; // defined in frame.cpp
+extern "C" _CRTIMP2 void* __cdecl __AdjustPointer(void*, const PMD&); // defined in frame.cpp
 
 using namespace std;
 
--- a/stl/src/excptptr.cpp
+++ b/stl/src/excptptr.cpp
@@ -32,7 +32,7 @@
 // V4 and later managed exception code
 #define MANAGED_EXCEPTION_CODE_V4 0XE0434352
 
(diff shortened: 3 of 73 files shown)


stl/src/mutex.cpp
```diff
@@ -172,20 +173,21 @@ _CRTIMP2_PURE _Thrd_result __cdecl _Mtx_trylock(_Mtx_t mtx) { // attempt to lock
 }
 
 // TRANSITION, ABI: preserved for binary compatibility
-_CRTIMP2_PURE _Thrd_result __cdecl _Mtx_timedlock(_Mtx_t mtx, const _timespec64* xt) { // attempt to lock timed mutex
+_CRTIMP2_PURE _Thrd_result __cdecl _Mtx_timedlock(
+    _Mtx_t mtx, const _timespec64* xt) noexcept { // attempt to lock timed mutex
```
StephanTLavavej, line 177: Detach comment?

stl/src/pplerror.cpp
```diff
@@ -11,7 +11,7 @@
 #include <roerrorapi.h>
 #include <trnsctrl.h>
 
-extern "C" void* __GetPlatformExceptionInfo(int*);
+extern "C" void* __GetPlatformExceptionInfo(int*) noexcept;
```
StephanTLavavej, line 14: This is another `src/vctools/crt/vcruntime/src/eh/ehhelpers.cpp` function - should we be changing our declaration?

stl/src/ppltasks.cpp
```diff
@@ -29,7 +29,7 @@ static GUID const Local_IID_ICallbackWithNoReentrancyToApplicationSTA = {
 #if defined(_CRT_APP) && defined(_DEBUG)
 extern "C" NTSYSAPI _Success_(return != 0) WORD NTAPI
     RtlCaptureStackBackTrace(_In_ DWORD FramesToSkip, _In_ DWORD FramesToCapture,
-        _Out_writes_to_(FramesToCapture, return) PVOID* BackTrace, _Out_opt_ PDWORD BackTraceHash);
+        _Out_writes_to_(FramesToCapture, return) PVOID* BackTrace, _Out_opt_ PDWORD BackTraceHash) noexcept;
```
StephanTLavavej, line 32: This function is definitely not ours - should we be changing our handcrafted declaration?

stl/src/stdhndlr.cpp
```diff
@@ -7,7 +7,7 @@
 
 using new_hand = int(__cdecl*)(size_t);
 
-extern "C" new_hand __cdecl _set_new_handler(new_hand);
+extern "C" new_hand __cdecl _set_new_handler(new_hand) noexcept;
```
StephanTLavavej, line 10: This is a UCRT function; I have a separate cleanup (PR forthcoming) to simply include `<new.h>` to get this. I recommend reverting this change.

stl/src/vector_algorithms.cpp
```diff
@@ -800,7 +800,7 @@ namespace {
 #ifndef _M_ARM64EC
         static constexpr bool _Has_portion_max = false;
 
-        static __m128i _Sign_correction(const __m128i _Val, const bool _Sign) {
+        static __m128i _Sign_correction(const __m128i _Val, const bool _Sign) noexcept {
```
StephanTLavavej, line 803: This is outside `extern "C"` as far as I can tell (it ended on line 483). This appears to be a good change for consistency with `_Minmax_traits_1/2/4`, but at a minimum it should be called out as a drive-by change.

Similarly for the following changes in this file (where an `extern "C"` block ends on 1125 immediately before them).

stl/src/winapisupp.cpp
```diff
@@ -305,7 +305,7 @@ extern "C" _Success_(return > 0 && return < BufferLength) DWORD
 
 #else // ^^^ defined(_ONECORE) / !defined(_ONECORE) vvv
 
-static int __cdecl initialize_pointers() {
+static int __cdecl initialize_pointers() noexcept {
```
StephanTLavavej, line 308: This does not appear to be `extern "C"`. I think it's a good change but it's a drive-by and deserves mention in the PR description (I am increasingly uncomfortable with mixing these drive-by `noexcept` additions into this large PR, whereas the drive-by comment addition for `_Thrd_abort` is fine).

StephanTLavavej (changes requested):

stl/inc/execution
```diff
@@ -45,7 +45,7 @@ using __std_PTP_CALLBACK_ENVIRON  = __std_TP_CALLBACK_ENVIRON*;
 _NODISCARD unsigned int __stdcall __std_parallel_algorithms_hw_threads() noexcept;
 
 using __std_PTP_WORK_CALLBACK = void(__stdcall*)(
-    _Inout_ __std_PTP_CALLBACK_INSTANCE, _Inout_opt_ void*, _Inout_ __std_PTP_WORK);
+    _Inout_ __std_PTP_CALLBACK_INSTANCE, _Inout_opt_ void*, _Inout_ __std_PTP_WORK) _NOEXCEPT_FNPTR;
 
 _NODISCARD __std_PTP_WORK __stdcall __std_create_threadpool_work(
     _In_ __std_PTP_WORK_CALLBACK, _Inout_opt_ void*, _In_opt_ __std_PTP_CALLBACK_ENVIRON) noexcept;
```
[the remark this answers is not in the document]
CaseyCarter: I'll change the definition to take `__std_PTP_WORK_CALLBACK` (which is implicitly convertible to `PTP_WORK_CALLBACK`) as well.

stl/src/ppltasks.cpp
CaseyCarter: Ack, no.

stl/src/excptptr.cpp
```diff
@@ -32,7 +32,7 @@
 // V4 and later managed exception code
 #define MANAGED_EXCEPTION_CODE_V4 0XE0434352
 
-extern "C" _CRTIMP2 void* __cdecl __AdjustPointer(void*, const PMD&); // defined in frame.cpp
+extern "C" _CRTIMP2 void* __cdecl __AdjustPointer(void*, const PMD&) noexcept; // defined in frame.cpp
```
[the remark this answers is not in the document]
CaseyCarter: It _should_, but I think not in this PR: VCRuntime changes are a bit scarier and should be done separately.

stl/src/pplerror.cpp
CaseyCarter: Ditto "yes, but not in this PR"

stl/src/vector_algorithms.cpp
CaseyCarter: Yes, these are functions called from `extern "C"` functions - without wrapping in `catch(...)` - which consequently need to be `noexcept`. Not exactly drive-by, but deserving of calling out in the commit message.

stl/inc/execution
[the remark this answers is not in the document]
CaseyCarter: Ugh: `parallel_algorithms.cpp` doesn't include `<execution>`, and can't because it's linked into the import lib and `<execution>` is non-core. I'll revert this and consider a separate change.

stl/src/vector_algorithms.cpp
```diff
@@ -1136,7 +1136,7 @@ _Min_max_element_t __stdcall __std_minmax_element_8(
 
 namespace {
     template <class _Ty>
-    const void* _Find_trivial_unsized_fallback(const void* _First, _Ty _Val) {
+    const void* _Find_trivial_unsized_fallback(const void* _First, _Ty _Val) noexcept {
```
StephanTLavavej, line 1139: #4113 is ready to merge and is removing these 4 functions (by manually inlining them). I'll push a change to pre-emptively resolve this merge conflict by dropping the `noexcept`s here. (The `_Minmax_traits_8::_Sign_correction` change above is still righteous.)

stl/src/ppltasks.cpp
```diff
@@ -350,7 +350,7 @@ namespace Concurrency {
 } // namespace Concurrency
 
 #ifdef _CRT_APP
-extern "C" void __cdecl __crtCleanupCausalityStaticFactories() {
+extern "C" void __cdecl __crtCleanupCausalityStaticFactories() noexcept {
```
StephanTLavavej, line 353: #4087 is removing this function; I'll pre-emptively resolve the conflict by dropping this change.

stl/src/dllmain.cpp
```diff
@@ -7,10 +7,11 @@
 
 #ifdef _CRT_APP
 // free static resource used by causality
-extern "C" void __cdecl __crtCleanupCausalityStaticFactories();
+extern "C" void __cdecl __crtCleanupCausalityStaticFactories() noexcept;
 #endif // defined(_CRT_APP)
 
-extern "C" BOOL APIENTRY DllMain(HMODULE /* hModule */, DWORD ul_reason_for_call, [[maybe_unused]] LPVOID lpReserved) {
+extern "C" BOOL APIENTRY DllMain(
+    HMODULE /* hModule */, DWORD ul_reason_for_call, [[maybe_unused]] LPVOID lpReserved) noexcept {
```
StephanTLavavej, line 14: #4087 is removing `__crtCleanupCausalityStaticFactories()` and changing `DllMain()` to be a stub. I'll pre-emptively resolve this conflict by dropping this PR's changes here, and adding `noexcept` to `DllMain()` in #4087.

StephanTLavavej: I got confused and mentioned this PR itself in the commit message instead of #4087, oopsy. The change itself is fine though.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for this big consistency improvement! :tada: :white_check_mark: :heart_eyes_cat:

PR #4111, opened 2023-10-20
Add a Standard mode check to `<__msvc_print.hpp>`

StephanTLavavej:
`<__msvc_print.hpp>` was missing the Standard mode check that the rest of our internal headers perform. This one's a little special because it's included by `<format>` in C++20 Concepts mode, and by `<print>` in C++23 Concepts mode.

Without this check, the header is still rejected in C++14/17 mode, but with a less clear error about the usage of `consteval`:

https://github.com/microsoft/STL/blob/1c59a205787f4fdcd9c20decc8fa6dfa9096d394/stl/inc/__msvc_print.hpp#L58

Also, this moves the other Standard mode checks in `<xmeow.h>` as early as possible. The mode is available after including `<yvals_core.h>`, and if we're doomed, we should diagnose doom immediately. `<xcharconv_tables.h>` was already doing this.

Next, `<ostream>` was guarding its inclusion of `<__msvc_print.hpp>` with `#if _HAS_CXX23`, even though its usage was guarded with `#ifdef __cpp_lib_print`:

https://github.com/microsoft/STL/blob/adea8d5ae280cafb91ae69b8dfaecd1c37a847d9/stl/inc/ostream#L1085

This meant that in our test suite, EDG in C++23 non-concepts mode had `<ostream>` including `<__msvc_print.hpp>` and then doing nothing with it. That was previously harmless but is now an error, so we need to make the guards consistent. (I don't think `<ostream>`'s dependency needs to be mentioned in the `<__msvc_print.hpp>` error message, that would just be confusing and it wouldn't be why users were trying to include the thing.)

Finally, `GH_001411_core_headers` was guarding with `#if _HAS_CXX23`, which was both too restrictive (the header is used by `<format>` in C++20) and too permissive (the header is used only in Concepts mode). We need to change this to `#ifdef __cpp_lib_concepts`.

Thanks to @ThomasFWise for the report!

### DIFF
--- a/stl/inc/__msvc_print.hpp
+++ b/stl/inc/__msvc_print.hpp
@@ -8,9 +8,8 @@
 #include <yvals_core.h>
 #if _STL_COMPILER_PREPROCESSOR
 
-#ifndef __cpp_lib_concepts
-#error The contents of <format> are available only with C++20 or later, and the contents of <print> \
-are available only with C++23 or later. (Also, you should not include this internal header.)
+#ifndef __cpp_lib_concepts // note: <format> includes this header in C++20 mode
+#error The contents of <print> are available only with C++23. (Also, you should not include this internal header.)
 #endif // ^^^ !defined(__cpp_lib_concepts) ^^^
 
 #include <cstdio>
--- a/stl/inc/__msvc_print.hpp
+++ b/stl/inc/__msvc_print.hpp
@@ -8,6 +8,11 @@
 #include <yvals_core.h>
 #if _STL_COMPILER_PREPROCESSOR
 
+#ifndef __cpp_lib_concepts
+#error The contents of <format> are available only with C++20 or later, and the contents of <print> \
+are available only with C++23 or later. (Also, you should not include this internal header.)
+#endif // ^^^ !defined(__cpp_lib_concepts) ^^^
+
 #include <cstdio>
 #include <xfilesystem_abi.h>
 
--- a/stl/inc/ostream
+++ b/stl/inc/ostream
@@ -9,11 +9,11 @@
 #if _STL_COMPILER_PREPROCESSOR
 #include <ios>
 
-#if _HAS_CXX23
+#ifdef __cpp_lib_print
 #include <__msvc_filebuf.hpp>
 #include <__msvc_print.hpp>
 #include <format>
-#endif // _HAS_CXX23
+#endif // ^^^ defined(__cpp_lib_print) ^^^
 
 #pragma pack(push, _CRT_PACKING)
 #pragma warning(push, _STL_WARNING_LEVEL)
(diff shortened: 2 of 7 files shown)


stl/inc/__msvc_print.hpp
```diff
@@ -8,6 +8,11 @@
 #include <yvals_core.h>
 #if _STL_COMPILER_PREPROCESSOR
 
+#ifndef __cpp_lib_concepts
+#error The contents of <format> are available only with C++20 or later, and the contents of <print> \
+are available only with C++23 or later. (Also, you should not include this internal header.)
```
CaseyCarter, line 13: I think these duplicate error messages from internal headers only confuse users. `<format>` presumably tells me when I can't include it, so I don't need `__msvc_meow.hpp` to tell me repeatedly. Why can't we consistently change the internal headers to just "You should not include this internal header."?

CaseyCarter (changes requested):

stl/inc/__msvc_print.hpp
StephanTLavavej: There are no duplicate error messages - can you take another look?

The Standard headers emit `#pragma message` warnings and expand to nothing, so they bypass any errors from internal headers. Here's what it looks like:

```
D:\GitHub\STL\out\x64>type meow1.cpp
#include <format>

D:\GitHub\STL\out\x64>cl /EHsc /nologo /W4 /std:c++17 /c meow1.cpp
meow1.cpp
D:\GitHub\STL\out\x64\out\inc\format(42): warning STL4038: The contents of <format> are available only with C++20 or later.

D:\GitHub\STL\out\x64>type meow2.cpp
#include <print>

D:\GitHub\STL\out\x64>cl /EHsc /nologo /W4 /std:c++17 /c meow2.cpp
meow2.cpp
D:\GitHub\STL\out\x64\out\inc\print(11): warning STL4038: The contents of <print> are available only with C++23 or later.

D:\GitHub\STL\out\x64>type meow3.cpp
#include <__msvc_print.hpp>

D:\GitHub\STL\out\x64>cl /EHsc /nologo /W4 /std:c++17 /c meow3.cpp
meow3.cpp
D:\GitHub\STL\out\x64\out\inc\__msvc_print.hpp(13): fatal error C1189: #error:  The contents of <format> are available only with C++20 or later, and the contents of <print> are available only with C++23 or later. (Also, you should not include this internal header.)
```

https://github.com/microsoft/STL/blob/adea8d5ae280cafb91ae69b8dfaecd1c37a847d9/stl/inc/print#L6-L14

CaseyCarter: I stand corrected, and it now makes even _less_ sense to me that e.g. `__msvc_print.hpp` mentions `<format>`. Why is this message informing me of the requirements of headers I didn't even include?

StephanTLavavej: I've added a comment to the guard, because otherwise it looks strange that we're checking for 20-concepts when `<print>` needs 23-concepts, and otherwise rephrased the message to follow the others (dropping "or later", as the others do, which makes it fit on one line).

CaseyCarter (approved):

StephanTLavavej:

PR #4113, opened 2023-10-21
inline internal function that are called only once in `vector_algorithms.cpp`

### LINKED ISSUE #3963: `vector_algorithms.cpp`: turn tails
There are some scalar tails extracted in `vector_algorithms.cpp`,
Whereas some are used multiple times, some are used just one. A few examples:
https://github.com/microsoft/STL/blob/8674b3d732a2bae43105fb7654d3b8c1c57adb56/stl/src/vector_algorithms.cpp#L1057-L1058
https://github.com/microsoft/STL/blob/8674b3d732a2bae43105fb7654d3b8c1c57adb56/stl/src/vector_algorithms.cpp#L1396
https://github.com/microsoft/STL/blob/8674b3d732a2bae43105fb7654d3b8c1c57adb56/stl/src/vector_algorithms.cpp#L1345

The tail extraction does not serve any purpose in such cases, it only provides an obstacle for debugging of optimized code (as tails are expectedly inlined. so the debugger does not navigate the lines).

AlexGuteniev:
Resolves #3963 

Also remove some stray newlines @CaseyCarter complained about.
Also drop some superfluous `typename`s.

Will conflict with #3928 , which I'll have to fix after either is merged.

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -1474,6 +1474,7 @@ namespace {
             } while (_First != _Stop_at);
         }
 #endif // !_M_ARM64EC
+        _Result >>= _Traits::_Shift;
         auto _Ptr = static_cast<const _Ty*>(_First);
         for (; _Ptr != _Last; ++_Ptr) {
             if (*_Ptr == _Val) {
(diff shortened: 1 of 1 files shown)


CaseyCarter (approved):

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for simplifying this code! :sparkles: :magic_wand: :tada:

PR #4117, opened 2023-10-23
Various cleanups: Comments and strings

StephanTLavavej:
* Comments: Drop extra space.
* Comments: Add them to `#endif`.
* Comments: Drop "TRANSITION, requires expression support".
  + We use `_Resettable_pointer` twice; *requires-expression*s would be more verbose.
* Comments: Fix capitalization typo.
* Comments: Strengthen comment about `ole32.dll`.
  + See #3607 - we can't depend on `ole32.dll` anywhere.
* String literals: Space before wrapping.
  + There were only a few places that weren't following our usual convention.
* String literals: Unwrap.
  + This one was easier to read on a single line.


CaseyCarter (approved):

StephanTLavavej:

PR #4120, opened 2023-10-23
`<random>`: Fix minor conformance issues

StephanTLavavej:
* Conformance: `base_type` in `<random>` is a non-Standard typedef.
  + Remove it outright for Standard `discard_block_engine` (via the new base class `_Discard_block_base` introduced by #4066), `independent_bits_engine`, and `shuffle_order_engine`.
  + For `discard_block`, which is usually the base class for Standard `discard_block_engine`, we can guard the typedef with TR1's availability, and deprecate it. We should then avoid using it ourselves.
* Conformance: Add `_Engine&&` constructors to `discard_block` and `_Discard_block_base`.
  + `discard_block_engine` had an `_Engine&&` constructor but its base classes didn't, so it was always copying.

### DIFF
--- a/stl/inc/random
+++ b/stl/inc/random
@@ -1362,6 +1362,8 @@ public:
 
     explicit discard_block(const _Engine& _Ex) : _Eng(_Ex), _Nx(0) {}
 
+    explicit discard_block(_Engine&& _Ex) : _Eng(_STD move(_Ex)), _Nx(0) {}
+
     explicit discard_block(result_type _Seed) : _Eng(_Seed), _Nx(0) {}
 
     template <class _Seed_seq, _Enable_if_seed_seq_t<_Seed_seq, discard_block, _Engine> = 0>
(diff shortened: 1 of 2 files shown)


stl/inc/random
```diff
@@ -1349,7 +1349,10 @@ public:
 template <class _Engine, int _Px, int _Rx>
 class discard_block { // discard_block compound engine
 public:
-    using base_type   = _Engine;
+#if _HAS_TR1_NAMESPACE
+    using base_type _DEPRECATE_TR1_NAMESPACE = _Engine; // TR1-only typedef
```
CaseyCarter, line 1353: `discard_block` is entirely TR1-era machinery. Should we remove `discard_block_engine`'s dependence on it and wrap the entire class in `_HAS_TR1_NAMESPACE`? (No change requested.)

CaseyCarter (approved):

stl/inc/random
StephanTLavavej: Unfortunately, we need to keep `discard_block_engine`'s data member the wrong size (`int`) to preserve ABI, except when the class would have been utterly broken, which is why #4066 uses `_Discard_block_engine` as the base class conditionally.

We could remove the inheritance and switch the data member's type directly, to preserve the layout while altering the inheritance, but we haven't (yet) done that for `uniform_int` and `uniform_int_distribution` etc. I suppose if vNext is delayed enough, excising more TR1 over time would be worth it.

CaseyCarter: Yeah, this would be nice if we ever have time, but certainly is more unrelated work than would be reasonable in this PR. I don't think we need a tracking issue either since we're well aware that we need to remove everything TR1-related.

StephanTLavavej:

PR #4121, opened 2023-10-23
Various cleanups: Simplify product code

StephanTLavavej:
* `<exception>`: `throw_with_nested()` no longer needs `_Glued` to defend against the preprocessor.
  + After #3973, `_With_nested_v2<_Uty>` doesn't contain commas.
* Simplify: `ios_base::_Iostate{}` => `ios_base::goodbit`
  + It's just zero.
* Simplify `_Copy_unchecked()`'s `#ifdef __cpp_lib_concepts` logic.
  + Here's a case where a negative test allows us to considerably simplify the code by avoiding the need for two `#ifdef __cpp_lib_concepts` regions (which was hard to follow). Now the code reads much more naturally - in concepts mode, we need to handle different iterator/sentinel types.
* `<random>`: `discard_block_engine` can use its `result_type` typedef for `min` and `max`.
  + A constructor above is already using this typedef.
* Style: Avoid using an assignment expression as an index in `basic_string::_Eos()`.
  + We prefer to avoid expressions with lots of side effects, so we should update `_Mysize` before indexing. I'm using `_New_size` for the index since it's shorter.

### DIFF
--- a/stl/inc/exception
+++ b/stl/inc/exception
@@ -361,8 +361,7 @@ _EXPORT_STD template <class _Ty>
 
     if constexpr (is_class_v<_Uty> && !is_base_of_v<nested_exception, _Uty> && !is_final_v<_Uty>) {
         // throw user exception glued to nested_exception
-        using _Glued = _With_nested_v2<_Uty>;
-        _THROW(_Glued(_STD forward<_Ty>(_Arg)));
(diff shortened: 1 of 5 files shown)


CaseyCarter (approved):

StephanTLavavej:

PR #4122, opened 2023-10-23
Various cleanups: Test code

StephanTLavavej:
Usual disclaimer about how we dislike grab-bag PRs as a general rule, *and* how we avoid large-scale cleanups of test code (low value, potentially disrupts the weird subtle things that the tests were trying to cover). The existence of this PR means that I think these cases are worth it for codebase consistency and that they're low-risk.

* `format.py`: `import copy` was unused.
  + Thanks to Pylance for finding this.
* Style: `constexpr friend` => `friend constexpr`
  + This was the only place that wasn't following our convention.
* Style: Nameless value parameter `const choose` => `choose`
* Drop unnecessary `static` for namespace-scope `permissive()`.
  + This one was easy to miss, because it's followed by a `static` member function.
* Test cleanup: Catch by const ref, even when unnamed.
  + Catching by modifiable reference is weird; there are a couple of places elsewhere where we're intentionally doing so, but here we certainly don't need to. As usual, following the normal pattern in most places helps to draw attention to the weird occurrences.
* Test cleanup: Catch by const ref.
  + Here we're just observing the `filesystem_error`.
* `P0811R3_midpoint_lerp`: `static inline constexpr` => `static constexpr`
  + I suspect that the `inline` was accumulated before a compiler bug workaround was added.
* `VSO_0000000_vector_algorithms`: Drop `inline` after `template`.
  + We conventionally never say `inline` here.
* Tests: Add bug databases to function names.
  + It's very important to cite bug databases and not just plain numbers (as we've gone through so many databases over the years). These occurrences were at least properly commented, but we conventionally name functions with bug databases too.
  + Example internal URLs (one must be VPNed to access these): [Dev10-391805](https://vstfdevdiv.redmond.corp.microsoft.com/DevDiv/Dev10/_workitems/edit/391805) and [DevDiv-1184701](https://vstfdevdiv.redmond.corp.microsoft.com/DevDiv2/DevDiv/_workitems/edit/1184701).
* Tests: Drop unnecessary `std::` qualification.
  + I consider these to be "slam dunks". On the other hand, we accumulated a bunch of `std::move` and `std::forward` during an annoying era where Clang was warning about unqualified calls before we silenced that; I don't think it's worth the effort to go clean them up.
* `P0083R3_splicing_maps_and_sets`: `#ifdef __cpp_constinit` => `#if _HAS_CXX20`
  + Once compiler divergence has subsided and there are no escape hatches, we prefer to use the coarse-grained modes to guard things.

### DIFF
--- a/tests/std/include/range_algorithm_support.hpp
+++ b/tests/std/include/range_algorithm_support.hpp
@@ -20,7 +20,7 @@ template <class>
 inline constexpr bool always_false = false;
 
 namespace detail {
-    static constexpr bool permissive() {
+    constexpr bool permissive() {
         return false;
     }
 
--- a/tests/std/include/test_mdspan_support.hpp
+++ b/tests/std/include/test_mdspan_support.hpp
@@ -235,7 +235,7 @@ constexpr void check_members_with_various_extents(Fn&& fn) {
 }
 
 namespace details {
-    static constexpr bool permissive() {
+    constexpr bool permissive() {
         return false;
     }
 
--- a/tests/std/tests/Dev09_174589_tr1_function_storing_pmf_called_with_reference_or_pointer/test.cpp
+++ b/tests/std/tests/Dev09_174589_tr1_function_storing_pmf_called_with_reference_or_pointer/test.cpp
@@ -69,7 +69,7 @@ struct B {
 
 struct X : public B {};
 
-void test_294051() {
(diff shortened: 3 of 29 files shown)


CaseyCarter (approved):

StephanTLavavej:

PR #4124, opened 2023-10-23
Various cleanups: Use unnamed namespaces

StephanTLavavej:
(I recommend "ignore whitespace" while reviewing this.)

Over time, we've developed a convention of using unnamed namespaces in `stl/src` for internal machinery. This clearly separates TU-local code from functions that are accessed from other `stl/src` TUs, injected into the import lib, or dllexported. I noticed a couple of older TUs that could benefit from being cleaned up.

* `xtime.cpp`: The `_Epoch` and `_Nsec100_per_sec` constants can be function-local.
* `xtime.cpp`: Use an unnamed namespace, following our modern convention.
  + The `constexpr` variables and `static` functions had internal linkage, so this is unobservable to other TUs. We can drop `static` now.
* `stdhndlr.cpp`: Use an unnamed namespace, following our modern convention.
  + UCRT `<new.h>` declares `_set_new_handler`, which is better than trying to declare it ourselves. We can then drop the `new_hand` typedef which was being used for that declaration.
  + The `_New_handler` variable was `static`, so we can clearly move it into an unnamed namespace and drop the `static`. We do need to qualify the `_STD new_handler` type now.
  + The `_New_handler_interface` helper function was never dllexported nor referred to by other `stl/src` TUs. We can move it out of `namespace std` and into the unnamed namespace.
  + The resulting code has our desired distinction between TU-local and dllexported machinery.

### DIFF
--- a/stl/src/xtime.cpp
+++ b/stl/src/xtime.cpp
@@ -8,39 +8,41 @@
 
 #include "awint.hpp"
 
-constexpr long _Nsec_per_sec  = 1000000000L;
-constexpr long _Nsec_per_msec = 1000000L;
-constexpr int _Msec_per_sec   = 1000;
+namespace {
+    constexpr long _Nsec_per_sec  = 1000000000L;
+    constexpr long _Nsec_per_msec = 1000000L;
+    constexpr int _Msec_per_sec   = 1000;
 
-static void _timespec64_normalize(_timespec64* xt) { // adjust so that 0 <= tv_nsec < 1 000 000 000
-    while (xt->tv_nsec < 0) { // normalize target time
-        xt->tv_sec -= 1;
-        xt->tv_nsec += _Nsec_per_sec;
-    }
-    while (_Nsec_per_sec <= xt->tv_nsec) { // normalize target time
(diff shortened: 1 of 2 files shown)


CaseyCarter (approved):

StephanTLavavej:

StephanTLavavej:
I've resolved the simple merge conflict in `xtime.cpp` where #4106 added `noexcept` and this PR moved constants to be local.

stl/src/xtime.cpp
```diff
@@ -46,7 +46,8 @@ namespace {
 
 _EXTERN_C
 
-_CRTIMP2_PURE long long __cdecl _Xtime_get_ticks() { // get system time in 100-nanosecond intervals since the epoch
+_CRTIMP2_PURE long long __cdecl _Xtime_get_ticks() noexcept {
+    // get system time in 100-nanosecond intervals since the epoch
```
CaseyCarter, line 50: It's been hard work causing all the merge conflicts on my own since @BillyONeal left the team.

CaseyCarter (approved):

PR #4125, opened 2023-10-23
Various cleanups: Simplify `ios_base` and `locale` constants

StephanTLavavej:
# Background
Using `static constexpr` is what allows the major improvement of dropping out-of-line definitions - see #3381 which made that change to `ios_base`'s base class `_Iosb`.

# `<xiosbase>`
* The DLL's exports don't mention any of `_Dummy_enum`, `_Dummy_enum_val`, `_Stdio`, `_Openmode`, `_Openmask`, `_Seekdir`, `_Seekbeg`, `_Seekcur`, `_Seekend`, `_Openprot`. 
* Move retained enums to the bottom of `_Iosb` with an improved comment.
  + There's an internal MaxMPDCompat tool that notices the enum types and enumerators. With significant effort I think I could update the checked-in CompatibilityData baseline, but for now this is the easiest cleanup.
* We can drop `_Stdio` entirely (it's not an enumerator).
* We can simplify the definitions of the `meowfield` constants (I verified that this doesn't change them).
* The `_Seekdir` enumerators were being used exactly once; we can simply hardcode their values.

# `<xlocale>`
Simplify `locale`'s category constants.

We need to retain `_Locbase` and comment it as TRANSITION, ABI because it affects `sizeof(locale)` due to MSVC's deficient EBCO. (On x64, `sizeof(locale)` is 16, but removing `_Locbase` would shrink it to 8.)

However, there appears to be nothing stopping us from moving the constants into `locale`, as depicted in the Standard. I've verified that the DLL's exports are unaffected, and they don't mention `_Locbase` at all. (They also don't mention the names of these constants, excluding `std::collate`, `std::ctype`, and `std::messages` which are different.)

I don't believe we need `_PGLOBAL` - it's rarely used in the STL, and never for new code. `crtdefs.h` defines it to be `__declspec(process)` for `_M_CEE` (i.e. `/clr` and `/clr:pure`). https://learn.microsoft.com/en-us/cpp/cpp/process?view=msvc-170 explains "When compiling with `/clr`, global and static variables are per-process by default and do not need to use `__declspec(process)`." so this doesn't affect `/clr` at all, which is what matters. (As for `/clr:pure`, it's vanishingly unlikely to matter - these are *constants*, so if multiple application domains start getting multiple copies, who cares?)

`using category = int;` is mandated by the Standard. We previously said `int` for the constants, but the Standard depicts `category`, so I'm following that now that the typedef is in scope.

Finally, I'm upgrading `const` to `constexpr` - this is not depicted in the Standard, but I believe it's more consistent with the `static constexpr` variables we use everywhere these days. Note that \[constexpr.functions\] forbids constexpr-strengthening functions, but it doesn't care about variables.

### DIFF
--- a/stl/inc/xiosbase
+++ b/stl/inc/xiosbase
@@ -45,9 +45,9 @@ public:
     static constexpr int hexfloat = 0x3000; // TRANSITION, ABI, GH-3296
 
     static constexpr int boolalpha   = 0x4000;
-    static constexpr int adjustfield = 0x01C0; // left | right | internal
-    static constexpr int basefield   = 0x0E00; // dec | oct | hex
-    static constexpr int floatfield  = 0x3000; // scientific | fixed
+    static constexpr int adjustfield = left | right | internal;
+    static constexpr int basefield   = dec | oct | hex;
+    static constexpr int floatfield  = scientific | fixed;
 
     enum _Iostate { // constants for stream states
         _Statmask = 0x17
--- a/stl/inc/xiosbase
+++ b/stl/inc/xiosbase
@@ -75,6 +75,13 @@ public:
     static constexpr int end = 2;
 
     static constexpr int _Default_open_prot = _SH_DENYNO; // constant for default file opening protection
+
+    // TRANSITION, ABI: These enums don't appear in the STL DLL's export surface, but MaxMPDCompat
+    // in the MSVC-internal build detects that they appear in the CompatibilityData baseline.
+    enum _Dummy_enum { _Dummy_enum_val = 1 };
+    enum _Openmode { _Openmask = 0xff };
+    enum _Seekdir { _Seekbeg, _Seekcur, _Seekend };
+    enum { _Openprot = _SH_DENYNO };
 };
 
 _EXPORT_STD extern "C++" class _CRTIMP2_PURE_IMPORT ios_base : public _Iosb<int> { // base class for ios
--- a/stl/inc/xiosbase
+++ b/stl/inc/xiosbase
@@ -22,7 +22,6 @@ _STD_BEGIN
 template <class _Dummy>
 class _Iosb { // define templatized bitmask/enumerated types, instantiate on demand
 public:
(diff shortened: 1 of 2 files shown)


stl/inc/xiosbase
```diff
@@ -46,7 +45,6 @@ public:
     static constexpr int hexfloat = 0x3000; // TRANSITION, ABI, GH-3296
 
     static constexpr int boolalpha   = 0x4000;
-    static constexpr int _Stdio      = 0x8000;
     static constexpr int adjustfield = 0x01C0; // left | right | internal
     static constexpr int basefield   = 0x0E00; // dec | oct | hex
     static constexpr int floatfield  = 0x3000; // scientific | fixed
```
CaseyCarter, line 50: No change requested, but since we're touching these anyway, could we:
```suggestion
    static constexpr int adjustfield = left | right | internal;
    static constexpr int basefield   = dec | oct | hex;
    static constexpr int floatfield  = scientific | fixed;
```

CaseyCarter (approved):

stl/inc/xiosbase
StephanTLavavej: Done. I verified that this compiles before and after:

```
```
```cpp
#include <ios>
using namespace std;

static_assert(ios_base::adjustfield == (ios_base::left | ios_base::right | ios_base::internal));
static_assert(ios_base::basefield == (ios_base::dec | ios_base::oct | ios_base::hex));
static_assert(ios_base::floatfield == (ios_base::scientific | ios_base::fixed));
```
```
D:\GitHub\STL\out\x64>cl /EHsc /nologo /W4 /c meow.cpp
meow.cpp

D:\GitHub\STL\out\x64>
```

StephanTLavavej:
@CaseyCarter I pushed the change you requested and checks have passed, FYI.

CaseyCarter (approved):

StephanTLavavej:

StephanTLavavej:
I had to push a change to retain the enums in `_Iosb`, but at least I understand why now.

CaseyCarter (approved):

PR #4132, opened 2023-10-27
`<flat_set>`: `flat_(multi)set`'s `iterator` type should be constant iterator

achabense:
Citation:
>For associative containers where the value type is the same as the key type, both iterator and const_iterator are constant iterators[.](https://eel.is/c++draft/container.requirements#associative.reqmts.general-6.sentence-2) It is unspecified whether or not iterator and const_iterator are the same type[.](https://eel.is/c++draft/container.requirements#associative.reqmts.general-6.sentence-3)

This pr makes `flat_(multi)set`'s `iterator` an alias for `const_iterator`.

### DIFF
--- a/stl/inc/flat_set
+++ b/stl/inc/flat_set
@@ -56,8 +56,8 @@ public:
     using const_reference        = const value_type&;
     using size_type              = _Container::size_type;
     using difference_type        = _Container::difference_type;
-    using iterator               = _Container::iterator;
     using const_iterator         = _Container::const_iterator;
+    using iterator               = const_iterator;
     using reverse_iterator       = _STD reverse_iterator<iterator>;
     using const_reverse_iterator = _STD reverse_iterator<const_iterator>;
     using container_type         = _Container;
@@ -324,7 +324,9 @@ public:
         _Guard._Target = nullptr;
     }
 
-    iterator erase(iterator _Where) {
+    iterator erase(iterator _Where)
+        requires (!is_same_v<iterator, const_iterator>)
+    {
         return _Mycont.erase(_Where);
     }
     iterator erase(const_iterator _Where) {
@@ -357,6 +359,10 @@ public:
     }
     _NODISCARD value_compare value_comp() const {
         return _Mycomp;
+    }
+
+    _NODISCARD const container_type& _Get_container() const {
+        return _Mycont;
     }
 
     // set operations
@@ -666,16 +672,16 @@ private:
             const auto _Equivalent = [this](const _Kty& _Lhs, const _Kty& _Rhs) {
                 return !_Compare(_Lhs, _Rhs) && !_Compare(_Rhs, _Lhs);
             };
-            const iterator _End = end();
-            _Mycont.erase(_STD unique(begin(), _End, _Equivalent), _End);
+            const auto _End = _Mycont.end();
+            _Mycont.erase(_STD unique(_Mycont.begin(), _End, _Equivalent), _End);
         }
     }
 
     template <bool _Presorted>
     void _Restore_invariants_after_insert(const size_type _Old_size) {
-        const iterator _Begin   = begin();
-        const iterator _Old_end = _Begin + static_cast<difference_type>(_Old_size);
-        const iterator _End     = end();
+        const auto _Begin   = _Mycont.begin();
+        const auto _Old_end = _Begin + static_cast<difference_type>(_Old_size);
+        const auto _End     = _Mycont.end();
 
         if constexpr (!_Presorted) {
             _STD sort(_Old_end, _End, _Pass_comp());
@@ -690,15 +696,15 @@ private:
     }
 
     void _Make_invariants_fulfilled() {
-        const iterator _Begin = begin();
-        const iterator _End   = end();
+        const auto _Begin = _Mycont.begin();
+        const auto _End   = _Mycont.end();
 
         if (_Begin == _End) {
             return;
         }
 
         // O(N) if already sorted.
-        const iterator _Begin_unsorted = _STD is_sorted_until(_Begin, _End, _Pass_comp());
+        const auto _Begin_unsorted = _STD is_sorted_until(_Begin, _End, _Pass_comp());
 
         _STD sort(_Begin_unsorted, _End, _Pass_comp());
         _STD inplace_merge(_Begin, _Begin_unsorted, _End, _Pass_comp());
@@ -769,8 +775,9 @@ _EXPORT_STD template <class _Kty, class _Keylt, class _Container, class _Pred>
 _Container::size_type erase_if(flat_set<_Kty, _Keylt, _Container>& _Val, _Pred _Predicate) {
     // clears the container to maintain the invariants when an exception is thrown (N4950 [flat.set.erasure]/5)
     _Clear_guard<flat_set<_Kty, _Keylt, _Container>> _Guard{_STD addressof(_Val)};
-    const auto _Erased_count = _STD _Erase_remove_if(_Val, _STD _Pass_fn(_Predicate));
-    _Guard._Target           = nullptr;
+    const auto _Erased_count =
+        _STD _Erase_remove_if(const_cast<_Container&>(_Val._Get_container()), _STD _Pass_fn(_Predicate));
+    _Guard._Target = nullptr;
     return _Erased_count;
 }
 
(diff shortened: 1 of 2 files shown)


stl/inc/flat_set
```diff
@@ -56,8 +56,8 @@ public:
     using const_reference        = const value_type&;
     using size_type              = _Container::size_type;
     using difference_type        = _Container::difference_type;
-    using iterator               = _Container::iterator;
     using const_iterator         = _Container::const_iterator;
+    using iterator               = const_iterator;
     using reverse_iterator       = _STD reverse_iterator<iterator>;
     using const_reverse_iterator = _STD reverse_iterator<const_iterator>;
```
CaseyCarter, line 62: For consistency (and to keep things simple; there's no need to "hide" the relationship) should we:
```suggestion
    using const_reverse_iterator = _STD reverse_iterator<const_iterator>;
    using reverse_iterator       = const_reverse_iterator;
```

stl/inc/flat_set
```diff
@@ -180,25 +180,27 @@ public:
...
         _Mycont.assign(_Ilist);
         _Make_invariants_fulfilled();
         _Guard._Target = nullptr;
         return static_cast<_Deriv&>(*this);
     }
 
     // iterators
+    // for flat_meow, `const_iterator` is required to be convertible to `iterator` (per N4958
+    // [associative.reqmts.general]/6)
     _NODISCARD iterator begin() noexcept {
```
CaseyCarter, line 193: Why not remove this overload, leaving only `begin() const`? (Ditto for `end()` on line 199.)

stl/inc/flat_set
```diff
@@ -314,17 +316,19 @@ public:
...
         _STL_ASSERT(_Is_sorted(_Cont), _Msg_not_sorted);
-        _Clear_guard<_Base_flat_set> _Guard{this};
+        _Clear_guard<container_type> _Guard{_STD addressof(_Mycont)};
         _Mycont        = _STD move(_Cont);
         _Guard._Target = nullptr;
     }
 
-    iterator erase(iterator _Where) {
+    iterator erase(iterator _Where)
+        requires (!is_same_v<iterator, const_iterator>)
```
CaseyCarter, line 330: It looks like this will never be satisfied; why not simply remove this overload of `erase`?

stl/inc/flat_set
```diff
@@ -589,8 +597,9 @@ private:
         }
 
         if (_Where != _End && !_Compare(_Val, *_Where)) {
-            // *_Where is equivalent to _Val; convert _Where to iterator type.
-            return _Mycont.begin() + (_Where - _Begin);
+            // *_Where is equivalent to _Val.
+            // for flat_meow, _Where should be convertible to iterator type.
```
CaseyCarter, line 601: These are a bit too "add one to x":
```suggestion
```

stl/inc/flat_set
```diff
@@ -768,17 +764,19 @@ public:
 _EXPORT_STD template <class _Kty, class _Keylt, class _Container, class _Pred>
 _Container::size_type erase_if(flat_set<_Kty, _Keylt, _Container>& _Val, _Pred _Predicate) {
     // clears the container to maintain the invariants when an exception is thrown (N4950 [flat.set.erasure]/5)
-    _Clear_guard<flat_set<_Kty, _Keylt, _Container>> _Guard{_STD addressof(_Val)};
-    const auto _Erased_count = _STD _Erase_remove_if(_Val, _STD _Pass_fn(_Predicate));
+    _Container& _Cont = const_cast<_Container&>(_Val._Get_container());
```
CaseyCarter, line 767: Why `const_cast` every use of `_Get_container()` instead of making it non-`const`?

stl/inc/flat_set
```diff
@@ -180,25 +180,27 @@ public:
...
+        _Clear_guard<container_type> _Guard{_STD addressof(_Mycont)};
         _Mycont.assign(_Ilist);
         _Make_invariants_fulfilled();
         _Guard._Target = nullptr;
         return static_cast<_Deriv&>(*this);
     }
 
     // iterators
+    // for flat_meow, `const_iterator` is required to be convertible to `iterator` (per N4958
+    // [associative.reqmts.general]/6)
```
CaseyCarter, line 192: This citation is incorrect - [container.reqmts]/6 requires the iterator conversion for all containers - and not IMO useful here.
```suggestion
```

stl/inc/flat_set
```diff
@@ -180,25 +180,27 @@ public:
...
     _NODISCARD iterator begin() noexcept {
-        return _Mycont.begin();
+        return cbegin();
```
CaseyCarter, line 194: I think we should avoid adding a layer of indirection here to simplify debug codegen and the stepping experience. I don't much care if it's `_Mycont.begin()` or `_Mycont.cbegin()`, but I do think we should call directly rather than via `flat_set::cbegin`. (Four occurrences.)

tests/std/tests/P1222R4_flat_set/test.cpp
```diff
@@ -73,6 +73,22 @@ void assert_reversible_container_requirements(const T& s) {
...
 }
 
+template <class T>
+void assert_set_requirements() {
+    using iterator       = T::iterator;
+    using const_iterator = T::const_iterator;
+    using key_type       = T::key_type;
+    using value_type     = T::value_type;
+
+    static_assert(_Constant_iterator<const_iterator>);
```
CaseyCarter, line 83: We could avoid directly using internal machinery here by instead:
```suggestion
    static_assert(same_as<std::const_iterator<const_iterator>, const_iterator>);
```
(and similarly on line 88)

CaseyCarter:
FWIW, I usually hear "flat_meow" used to collectively refer to `flat_map`, `flat_multimap`, `flat_set`, and `flat_multiset`.

stl/inc/flat_set
achabense: Ideally, every method related to `iterator` can be deleted without behavior change. This includes non-const `begin/...` functions, `erase(iterator)` and non-const searching functions like `lower_bound`.
>set doesn't attempt to remove the non-const overloads of lower_bound etc., so I recommend following that pattern.

However, as talk about early on discord, this pr tries to avoid removing any of these functions directly (relying on convertibility instead of const_iter and iterator being the same type; `_Find` is removed as it's internal method). I agree that the code is unnecessarily complicated though...

(As to `erase` here, it is mimicking the current practice in `_Tree`)
https://github.com/microsoft/STL/blob/f392449fb72d1a387ac5025d508e8442914477f9/stl/inc/xtree#L1310-L1311

stl/inc/flat_set
achabense: yes this looks awkward... I thought if `flat_set` want to expose its container, it should always be const to avoid outside modification... `erase_if` will be only exception, but it's unfortunately also the only user of this function now... Is it ok to remove `_Get_container` and befriend `erase_if` instead?

stl/inc/flat_set
StephanTLavavej: After thinking about this again, I think the difference here is that most of `std::set`'s machinery is shared with `std::map`, so we don't know if `iterator` is `const_iterator`. `<flat_set>` is different because it lacks such sharing, so I think avoiding the duplicated member functions would be ultimately less squirrelly. Apologies for not thinking more clearly earlier.

achabense: Thanks! Does this mean we can remove all methods related to non-const `iterator`, or instead a subset of these methods?

stl/inc/flat_set
achabense: Is it allowed to define erase_if friend in the class?

StephanTLavavej: I believe there are observable differences between hidden friends and namespace-scope functions, so we can't make such a transformation.

achabense: I've made `_Get_container` non-const, but renamed it to `_Get_container_for_erase_if`, as I'm concerned about exposing the non-const container under a too general name.

stl/inc/flat_set
```diff
@@ -324,20 +313,15 @@ public:
...
     iterator erase(const_iterator _Where) {
         return _Mycont.erase(_Where);
     }
     size_type erase(const _Kty& _Val) {
         return _Erase(_Val);
     }
     template <_Different_from<_Kty> _Other>
-        requires (
-            _Keylt_transparent && !is_convertible_v<_Other, iterator> && !is_convertible_v<_Other, const_iterator>)
+        requires (_Keylt_transparent && !is_convertible_v<_Other, const_iterator>)
```
achabense, line 324: Is it ok not to add comment for this one? Or am I adding too many repetitive comments?

StephanTLavavej:
Thanks for `const`antly improving `<flat_set>`! :joy_cat: :tada: :heart_eyes_cat:

PR #4140, opened 2023-11-01
Use clz instead of clzs

AtariDreams:
Why am I suggesting this?

The answer is because for some reason, __builtin_clzs is generating worse code than __builtin_clz minus the difference in bits. This seems like a clang bug, but until then, let's change the definition of _Clang_arm_arm64_countl_zero for now as a workaround.

Source: https://godbolt.org/z/TfP7hGcv7

### DIFF
--- a/stl/inc/__msvc_bit_utils.hpp
+++ b/stl/inc/__msvc_bit_utils.hpp
@@ -140,9 +140,25 @@ _NODISCARD int _Checked_x86_x64_countl_zero(const _Ty _Val) noexcept {
 
 #if defined(_M_ARM) || defined(_M_ARM64)
 #ifdef __clang__ // TRANSITION, GH-1586
+#ifdef _M_ARM64
+_NODISCARD constexpr int _Clang_arm_arm64_countl_zero(const unsigned char _Val) {
+    // TODO: Revert back to __builtin_clzs once clang longer generates worse code than __builtin_clz
+    return __builtin_clz(_Val) - (_Unsigned_integer_digits<unsigned int> - _Unsigned_integer_digits<unsigned char>);
+}
+
+_NODISCARD constexpr int _Clang_arm_arm64_countl_zero(const unsigned short _Val) {
+    // TODO: Revert back to __builtin_clzs once clang longer generates worse code than __builtin_clz
+    return __builtin_clz(_Val) - (_Unsigned_integer_digits<unsigned int> - _Unsigned_integer_digits<unsigned short>);
+}
+#else
+_NODISCARD constexpr int _Clang_arm_arm64_countl_zero(const unsigned char _Val) {
+    return __builtin_clzs(_Val) - (_Unsigned_integer_digits<unsigned short> - _Unsigned_integer_digits<unsigned char>);
+}
+
 _NODISCARD constexpr int _Clang_arm_arm64_countl_zero(const unsigned short _Val) {
     return __builtin_clzs(_Val);
 }
+#endif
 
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
Is there a confirmed https://github.com/llvm/llvm-project issue that we can cite? When we work around compiler issues, we cite bugs with `// TRANSITION, LLVM-nnnn` so we can check when it's time to remove the workaround.

stl/inc/__msvc_bit_utils.hpp
```diff
@@ -140,9 +140,26 @@ _NODISCARD int _Checked_x86_x64_countl_zero(const _Ty _Val) noexcept {
...
+_NODISCARD constexpr int _Clang_arm_arm64_countl_zero(const unsigned char _Val) {
+    // TODO: Revert back to __builtin_clzs once clang longer generates worse code than __builtin_clz
+    return __builtin_clz(_Val) - (_Unsigned_integer_digits<unsigned int> - _Unsigned_integer_digits<unsigned char>);
+}
+
+_NODISCARD constexpr int _Clang_arm_arm64_countl_zero(const unsigned short _Val) {
+    // TODO: Revert back to __builtin_clzs once clang longer generates worse code than __builtin_clz
+    return __builtin_clz(_Val) - (_Unsigned_integer_digits<unsigned int> - _Unsigned_integer_digits<unsigned short>);
+}
+#else
```
StephanTLavavej, line 153: We don't care about ARM32 performance, as long as it compiles and runs properly, so it isn't worth having a distinct codepath for it here.

stl/inc/__msvc_bit_utils.hpp
```diff
@@ -140,9 +140,26 @@ _NODISCARD int _Checked_x86_x64_countl_zero(const _Ty _Val) noexcept {
 
 #if defined(_M_ARM) || defined(_M_ARM64)
 #ifdef __clang__ // TRANSITION, GH-1586
+#ifdef _M_ARM64
+_NODISCARD constexpr int _Clang_arm_arm64_countl_zero(const unsigned char _Val) {
+    // TODO: Revert back to __builtin_clzs once clang longer generates worse code than __builtin_clz
```
StephanTLavavej, line 145: After there's an upstream llvm-project issue that has been reported, these comments should say `// TRANSITION, LLVM-nnnn` citing the bug number. (We don't say TODO in the product code because IDEs pick that up and tell users that they have TODOs to take care of, which is confusing for them.)

stl/inc/__msvc_bit_utils.hpp
```diff
@@ -140,9 +140,26 @@ _NODISCARD int _Checked_x86_x64_countl_zero(const _Ty _Val) noexcept {
...
+#else
+_NODISCARD constexpr int _Clang_arm_arm64_countl_zero(const unsigned char _Val) {
+    // TODO: Revert back to __builtin_clzs once clang longer generates worse code than __builtin_clz
+    return __builtin_clzs(_Val) - (_Unsigned_integer_digits<unsigned short> - _Unsigned_integer_digits<unsigned char>);
+}
+
 _NODISCARD constexpr int _Clang_arm_arm64_countl_zero(const unsigned short _Val) {
     return __builtin_clzs(_Val);
 }
+#endif
```
StephanTLavavej, line 162: This `#endif` should be commented. (We add preprocessor comments for everything beyond 1-2 lines to help keep track of what region we're in.)

StephanTLavavej (changes requested):

StephanTLavavej:
Closing this as there's been no response to our questions (and there's been force-pushing, which we've repeatedly asked to avoid).

Issue #4141, opened 2023-11-02
`<ranges>`: `ranges::to<array<T, N>>()` strikes fear into the hearts of compilers


StephanTLavavej:
The following code causes MSVC to run out of memory and crash (when x86-native; x64-native can get away with allocating more memory but it still appears to be infinite). Interestingly, Clang also gets trapped here, although it doesn't run out of memory. I suspect that this leads to infinite constraint recursion, which may be something we need to fix either in the Standard or our implementaton.

<details><summary>Click to expand environment.</summary>

```
C:\Temp>"C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Auxiliary\Build\vcvarsall.bat" x86
**********************************************************************
** Visual Studio 2022 Developer Command Prompt v17.8.0-pre.6.0
** Copyright (c) 2022 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'x86'

C:\Temp>cl
Microsoft (R) C/C++ Optimizing Compiler Version 19.38.33129 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.

usage: cl [ option... ] filename... [ /link linkoption... ]

C:\Temp>clang-cl -v
clang version 16.0.5
Target: i686-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Tools\Llvm\bin
```
</details>

```
C:\Temp>type meow.cpp
```
```cpp
#include <array>
#include <ranges>
#include <span>
#include <string>
using namespace std;

void TestFn(span<const wchar_t> str) {
    auto split = str | views::split('.');

    auto vals = split | ranges::to<array<wstring, 4>>();
}
```
```
C:\Temp>cl /EHsc /nologo /W4 /std:c++latest /MTd /Od /c meow.cpp
meow.cpp
C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Tools\MSVC\14.38.33129\include\type_traits(1430): fatal error C1060: compiler is out of heap space

C:\Temp>clang-cl /EHsc /nologo /W4 /std:c++latest /MTd /Od /c meow.cpp
[*** RUNS FOREVER ***]
```

This code works with `vector<wstring>` as an argument for `ranges::to`.

VS 2022 17.8 Preview 6 `type_traits` 1430 is this line in `main` right now:

[...]

PR #4142, opened 2023-11-02
`<ranges>`: Avoid infinite recursion in `ranges::to` when the inner ranges match but the destination container is still not constructible

cpplearner:
Fixes GH-4141.

The standard says ([\[range.utility.conv.to\]](http://eel.is/c++draft/range.utility.conv.to)):

> - (2.1) If `C` does not satisfy `input_range` or `convertible_to<range_reference_t<R>, range_value_t<C>>` is `true`:
>     - [...]
> - (2.2) Otherwise, if `input_range<range_reference_t<R>>` is `true`:
>      ```
>      to<C>(r | views::transform([](auto&& elem) {
>          return to<range_value_t<C>>(std::forward<decltype(elem)>(elem));
>      }), std::forward<Args>(args)...);
>      ```

So `views::transform` shouldn't be used if `convertible_to<range_reference_t<R>, range_value_t<C>>` is `true`.

Notably, when step (2.2) recursively calls `ranges::to<C>`, it passes a `transform_view` whose `range_reference_t` is the same as `range_value_t<C>`. So in this case `convertible_to<range_reference_t<R>, range_value_t<C>>` is `convertible_to<range_value_t<C>, range_value_t<C>>`, which should usually be true, which prevents infinite recursion.

***But***, `convertible_to<range_value_t<C>, range_value_t<C>>` may be false (i.e. `range_value_t<C>` may not be move-constructible):

```c++
#include <ranges>

struct ImmovableRange {
    ImmovableRange(int*, int*);
    ImmovableRange(ImmovableRange&&) = delete;

    int* begin();
    int* end();
};

struct C {
    ImmovableRange* begin();
    ImmovableRange* end();
};

using R = int[1][2];

void test() {
    (void)std::ranges::to<C>(R{});  // infinite recursion
}
```

This PR does *not* fix this case, as I can't find where the standard addresses this. Maybe it really is a defect in the standard after all.

### DIFF
--- a/stl/inc/ranges
+++ b/stl/inc/ranges
@@ -10383,7 +10383,7 @@ namespace ranges {
             }
             _RANGES copy(_Range, _Container_inserter<range_reference_t<_Rng>>(_Cont));
             return _Cont;
-        } else if constexpr (input_range<range_reference_t<_Rng>>) {
+        } else if constexpr (!_Ref_converts<_Rng, _Container> && input_range<range_reference_t<_Rng>>) {
             const auto _Xform = [](auto&& _Elem) {
                 return _RANGES to<range_value_t<_Container>>(_STD forward<decltype(_Elem)>(_Elem));
             };


StephanTLavavej (approved):

StephanTLavavej:
Thanks for the fix and the detailed explanation!

hewillk:
Hum, the nonmovable  range always seems problematic to me (although `format_parse_context` is necessary for its existence).

```cpp
#include <ranges>

struct ImmovableRange {
  using value_type = int;
  ImmovableRange();
  ImmovableRange(ImmovableRange&&) = delete;

  int* begin();
  int* end();
  void push_back(value_type);
};

using R = int[1];

void test() {
  (void)std::ranges::to<ImmovableRange>(R{}); // hard error
}
```

CaseyCarter:
> Hum, the nonmovable range always seems problematic to me (although `format_parse_context` is necessary for its existence).

I don't understand what this comment is saying about the change in this PR.

hewillk:
> I don't understand what this comment is saying about the change in this PR.

Not sure why I made this comment in the first place.
Here are the exact comments about the PR:
Why don't we follow the wording and use more nested if conditions, for example:
```cpp
if constexpr (_Ref_converts<_Rng, _Container>) {
    if constexpr (/* */) {
        // ...
    } else if constexpr (/* */) {
        // ...
    } else if constexpr (/* */) {
        // ...
    } else if constexpr (/* */) {
        // ...
    } else {
        static_assert(_Always_false<_Container>, "the program is ill-formed per N4950 [range.utility.conv.to]/2.3");
    }
    // ...
} else if constexpr (input_range<range_reference_t<_Rng>>) {
    // ...
} else {
    static_assert(_Always_false<_Container>, "the program is ill-formed per N4950 [range.utility.conv.to]/2.3");
}
```
instead use a less-nested but *repeatedly-checked* structure?

```cpp
if constexpr (_Ref_converts<_Rng, _Container> && /* */) {
    // ...
} else if constexpr (_Ref_converts<_Rng, _Container> && /* */) {
    // ...
} else if constexpr (_Ref_converts<_Rng, _Container> && /* */) {
    // ...
} else if constexpr (_Ref_converts<_Rng, _Container> && /* */) {
    // ...
} else if constexpr (!_Ref_converts<_Rng, _Container> && input_range<range_reference_t<_Rng>>) {
    // ...
} else {
    static_assert(_Always_false<_Container>, "the program is ill-formed per N4950 [range.utility.conv.to]/2.3");
}
```
I think it is more appropriate to strictly adhere to the first structure especially after the editorial enhancements [#6568](https://github.com/cplusplus/draft/pull/6568)  (Although both forms are _currently_ equivalent).

StephanTLavavej:

CaseyCarter:
> Here are the exact comments about the PR: Why don't we follow the wording and use more nested if conditions, for example:
> [snip]

I don't recall now why I used this structure, maybe it was simply to avoid repeating the `static_assert`? In any case, I agree it would be clearer if we more closely followed the structure of the Standard wording. I'd be happy to approve a PR making that change.

CaseyCarter (approved):

StephanTLavavej:
Thanks for fixing this bug where for your PR I commented, Thanks for fixing this bug where for your PR I commented, Thanks for fixing this bug where for your PR I commented, :joy_cat: :tada: :zany_face:

PR #4143, opened 2023-11-02
Various cleanups: `extern "C"`

StephanTLavavej:
We've accumulated inconsistency between using our `_EXTERN_C` macro and directly saying `extern "C"`:

https://github.com/microsoft/STL/blob/f392449fb72d1a387ac5025d508e8442914477f9/stl/inc/memory#L3720-L3723
https://github.com/microsoft/STL/blob/f392449fb72d1a387ac5025d508e8442914477f9/stl/inc/__msvc_sanitizer_annotate_container.hpp#L131-L134

The macro used to be important when some of our headers were dual C and C++. Now that they're always C++, and that we don't expect to play special games with `extern "C"` in the foreseeable future, we should resolve this inconsistency and favor dropping the macro.

(In contrast, we're quite disciplined about `_STD_BEGIN` in product code, and I do want to play special games with it soon.)

* Replace: `_EXTERN_C ... _END_EXTERN_C` => `extern "C" { ... } // extern "C"`
* Add `// extern "C"` comments.
* Wrap `stacktrace`'s function pointer typedefs in `extern "C"`.
  + It doesn't really matter because typedefs are "transparent" (they don't create new types), but I believe it's desirable for the signatures of `extern "C"` functions to restrict themselves to mentioning only things that are within the `extern "C"` universe. During the initial modularization effort, I overhauled `extern "C"` functions to mention only `extern "C"` structs, where it did matter.
* `sharedmutex.cpp`: Move the `_Smtx_t` typedef within `extern "C"`.
  + Again, it doesn't really matter for typedefs, but this is now consistent with `stl/inc/xthreads.h`: https://github.com/microsoft/STL/blob/f392449fb72d1a387ac5025d508e8442914477f9/stl/inc/xthreads.h#L21-L28
* Mark `_Atomic_memory_order_meow` as `extern "C"` to match vcruntime.
* Mark `_C_meow_complex` as `extern "C"` to match the UCRT.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -271,8 +271,6 @@ StatementMacros:
   - _STDEXT_END
   - _FMT_P2286_BEGIN
   - _FMT_P2286_END
-  - _EXTERN_C
-  - _END_EXTERN_C
   - _EXTERN_C_UNLESS_PURE
   - _END_EXTERN_C_UNLESS_PURE
 # TabWidth:        8
--- a/stl/inc/__msvc_print.hpp
+++ b/stl/inc/__msvc_print.hpp
@@ -22,7 +22,7 @@ _STL_DISABLE_CLANG_WARNINGS
 #pragma push_macro("new")
 #undef new
 
-_EXTERN_C
+extern "C" {
 
 enum class __std_unicode_console_handle : intptr_t { _Invalid = -1 };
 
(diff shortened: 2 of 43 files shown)


StephanTLavavej:

CaseyCarter (approved):

StephanTLavavej:
I've pushed a merge with `main` to resolve a trivial adjacent-edit conflict with #4150 in `__msvc_sanitizer_annotate_container.hpp`.

PR #4145, opened 2023-11-03
Various cleanups: Use `_STDEXT_BEGIN` and `_STDEXT_END`

StephanTLavavej:
Select "Hide whitespace" to see how simple this change is.

We usually use our namespace macros when defining `stdext` machinery:

https://github.com/microsoft/STL/blob/f392449fb72d1a387ac5025d508e8442914477f9/stl/inc/iterator#L1468-L1470

This PR fixes the one inconsistency in our headers. (Right now, there's no difference between the macros and directly mentioning the namespace, but in the future the macros might be trickier.)


StephanTLavavej:

CaseyCarter (approved):

PR #4147, opened 2023-11-03
Various cleanups: `<xatomic_wait.h>` constants

StephanTLavavej:
One more round of cleanups resulting from my audit of things in the global namespace that aren't `extern "C"`.

* Move `_Atomic_wait_no_deadline` from `<xatomic_wait.h>` to `atomic_wait.cpp`.
  + Nothing else mentions this, obviously.
  + It's moving into the unnamed namespace for isolation, and dropping `_INLINE_VAR`.
* Rename `_Atomic_wait_no_timeout` => `__std_atomic_wait_no_timeout`
  + This is moving into `extern "C"` like `<xfilesystem_abi.h>`'s constants: https://github.com/microsoft/STL/blob/f392449fb72d1a387ac5025d508e8442914477f9/stl/inc/xfilesystem_abi.h#L21-L23
  + `<xatomic_wait.h>` can assume C++20 mode, so we can use `inline` variables unconditionally.
  + There will be a trivial merge conflict with #4143.

### DIFF
--- a/stl/inc/atomic
+++ b/stl/inc/atomic
@@ -514,7 +514,7 @@ void _Atomic_wait_direct(
             return;
         }
 
-        __std_atomic_wait_direct(_Storage_ptr, &_Expected_bytes, sizeof(_Value_type), _Atomic_wait_no_timeout);
+        __std_atomic_wait_direct(_Storage_ptr, &_Expected_bytes, sizeof(_Value_type), __std_atomic_wait_no_timeout);
     }
 }
 #endif // _HAS_CXX20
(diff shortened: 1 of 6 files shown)


AlexGuteniev (approved):

StephanTLavavej:

CaseyCarter (approved):

StephanTLavavej:
I've pushed a merge with `main` to resolve a trivial adjacent-edit conflict with #4143 in `xatomic_wait.h`.

PR #4150, opened 2023-11-06
ASan: Remove `noexcept` from `__sanitizer_annotate_contiguous_container`

StephanTLavavej:
#4106 marked `__sanitizer_annotate_contiguous_container` as `noexcept` because it's `extern "C"`. However, it has to match ASan's primary declaration, so we can't mark it.

This PR reverts this declaration to its original form, and adds a comment for future maintainers.


strega-nil-ms (approved):

AlexGuteniev:
If it throws, should be `noexcept(false)` to opt out of implicit noexcept, no?

CaseyCarter (approved):

StephanTLavavej:

PR #1131, opened 2020-08-02
Specialize find and count for vector<bool>

miscco:
This builds uppon the recently merged specialization of `fill` for `vector<bool>`.

Note that the specializations require `__cpp_lib_bitops`

PR #879, opened 2020-06-05
Specialize fill and fill_n for vector<bool>

miscco:
This partially addresses #625 by specializing fill and fill_n

This is essentially a subpart of #750 with one of the algorithms where the implementation is both simple and a clear win.

PR #4151, opened 2023-11-07
Various cleanups: Declare `_Meow_vbool` to avoid ADL

StephanTLavavej:
I noticed this a while ago, but it's increasingly important due to modules. (This PR avoids the likely compiler bug VSO-1914077 "Modules: `extern "C++"` interferes with ADL".)

The `vector<bool>` algorithm optimizations (#879 #1131 #3353) detect `vector<bool>` iterators via an internal type trait, then dispatch to dedicated functions. We were defining those functions in `<vector>`, but were never declaring them before calling them in `<xutility>` (which is "higher up" in the dependency graph), so our calls were actually using ADL. That was squirrelly :chipmunk: since we never rely on ADL like this for internal functions.

This PR adds declarations (dropping `const` on value parameters as usual), then qualifies the calls for good measure (also aligning with our developing practice of qualifying internal calls, which personally hasn't filled me with delight elsewhere but is definitely reassuring here).

### DIFF
--- a/stl/inc/xutility
+++ b/stl/inc/xutility
@@ -4518,6 +4518,18 @@ _OutCtgIt _Copy_memmove_n(_CtgIt _First, const size_t _Count, _OutCtgIt _Dest) {
 template <class _It, bool _RequiresMutable = false>
 _INLINE_VAR constexpr bool _Is_vb_iterator = false;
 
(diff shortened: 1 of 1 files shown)


CaseyCarter (approved):

StephanTLavavej:

PR #4228, opened 2023-12-02
`<sstream>`: Avoid truncation warnings in `basic_stringbuf`'s constructor

StephanTLavavej:
Test coverage will be provided by the upcoming libcxx update, which has allocators with custom size types. #4219 fixed outright compiler errors, but there's still a compiler warning:

```
sstream(57): warning C4267: 'argument': conversion from 'size_t' to 'const unsigned int', possible loss of data
```

`_Convert_size` is our machinery for dealing with this.


StephanTLavavej:

PR #4229, opened 2023-12-02
`<chrono>`: Make `tzdb_list`'s internal ctor harder to unintentionally use

StephanTLavavej:
`tzdb_list` has "unspecified additional constructors", WG21-N4964 \[time.zone.db.list\]. #4119 added defenses to `time_zone` and `time_zone_link`, so we should update `tzdb_list` similarly.

This was found by `llvm-project/libcxx/test/std/time/time.zone/time.zone.db/time.zone.db.list/types.compile.pass.cpp` in the upcoming libcxx update that I'm working on. While this isn't a conformance issue, it's better to strengthen our product than weaken their test here.

### DIFF
--- a/stl/inc/chrono
+++ b/stl/inc/chrono
@@ -2202,6 +2202,10 @@ namespace chrono {
         return _STD move(_Icu_version) + "." + _STD to_string(_Num_leap_seconds);
     }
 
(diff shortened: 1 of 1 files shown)


StephanTLavavej:

PR #4234, opened 2023-12-04
`<random>`: Optimize range adjustment in `uniform_int`

horenmar:
Neither Clang nor MSVC is smart enough to optimize the previous code into the simple non-branching non-conditional move variant. (GCC actually is, but GCC does not matter in this context)

See
https://godbolt.org/z/T9GT7s5rG
https://godbolt.org/z/eE7q9G9W8

---------

**Comment on the PR instructions:**

I am working on custom distributions for Catch2, and implemented the range adjustment like this. Afterwards I went to check how the STLs do it and quick test showed that MSVC is better served with my implementation. Since this started as my original work, even though for different repo, I think that's fine.

### DIFF
--- a/stl/inc/random
+++ b/stl/inc/random
@@ -2089,12 +2089,7 @@ private:
     static _Uty _Adjust(_Uty _Uval) noexcept { // convert signed ranges to unsigned ranges and vice versa
         if constexpr (is_signed_v<_Ty>) {
             constexpr _Uty _Adjuster = (static_cast<_Uty>(-1) >> 1) + 1; // 2^(N-1)
-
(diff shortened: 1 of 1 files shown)


CaseyCarter (approved):

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for improving the performance of everyone's favorite distribution, and congratulations on your first microsoft/STL commit! :heart_eyes_cat: :rocket: :tada:

This will ship in VS 2022 17.10 Preview 1.

PR #4236, opened 2023-12-04
`<mdspan>`: Construct mappings from `extents_type` rvalues when necessary

StephanTLavavej:
WG21-N4964 \[mdspan.mdspan.cons]\/10.2:

> direct-non-list-initializes *`map_`* with `extents_type(exts)`

We were correctly passing `extents_type` rvalues up to the `_Mdspan_mapping_base` layer, but that was incorrectly always copying.

Test coverage will be provided by `std/containers/views/mdspan/mdspan/ctor.dh_span.pass.cpp` in the upcoming libcxx update that I'm working on.


CaseyCarter (approved):

StephanTLavavej:

PR #4240, opened 2023-12-05
Don't `*` every test matrix line

CaseyCarter:
The internal test runner runs `*`-tagged lines when asked to run lines with any particular tag, including `ASAN`. Putting `*` on everything is not the clever hack we thought it was, it results in the internal `ASAN` pipeline running everything. We don't need this redundant coverage so we need to revert #4068.

This change includes three commits for ease of review:
1. the reversion of #4068 (this was purely mechanical, so feel free to skim),
2. restoration of some `*` tags that #4068 added to non-leaf lines, which were actually necessary bug fixes,
3. a fix to add leaf lines to `P2465R3_standard_library_modules` so all fourteen configurations aren't run in both ASan and non-ASan pipelines.

I validated these changes by running each of `std`, `tr1`, `libcxx` both internally and externally with and without the change to verify that the correct number of tests are run in every case.


StephanTLavavej (approved):

StephanTLavavej:
0 out of 5 stars, would not merge again! :joy_cat: :zany_face: :no_entry_sign: :star:

CaseyCarter:
> 0 out of 5 stars, would not merge again! 😹 🤪 🚫 ⭐

Oh no, my Yelp!

PR #4252, opened 2023-12-10
Fix truncation warnings in `allocate_shared()`

StephanTLavavej:
Found by `std/utilities/memory/util.smartptr/util.smartptr.shared/util.smartptr.shared.create/allocate_shared_for_overwrite.pass.cpp` in the upcoming libcxx update.

```
memory(2858): warning C4267: 'argument': conversion from 'size_t' to 'test_allocator<U>::size_type', possible loss of data
```

We need to `_Convert_size` when allocating, but if that succeeds, we can simply `static_cast` when deallocating.

### DIFF
--- a/stl/inc/memory
+++ b/stl/inc/memory
@@ -2670,7 +2670,8 @@ private:
 
         this->~_Ref_count_unbounded_array_alloc();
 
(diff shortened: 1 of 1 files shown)


StephanTLavavej:

CaseyCarter (approved):

PR #4253, opened 2023-12-10
Fix `sub_match <=> basic_string` with custom traits/allocators

StephanTLavavej:
Found by `std/re/re.submatch/re.submatch.op/compare.pass.cpp` in the upcoming libcxx update.

This follows WG21-N4964 \[re.submatch.op\]/5:

```cpp
template<class BiIter, class ST, class SA>
  auto operator<=>(
      const sub_match<BiIter>& lhs,
      const basic_string<typename iterator_traits<BiIter>::value_type, ST, SA>& rhs);
```

*Returns:*
```cpp
static_cast<SM-CAT(BiIter)>(lhs.compare(
    typename sub_match<BiIter>::string_type(rhs.data(), rhs.size()))
      <=> 0
    )
```

without constructing a temporary. Our equality operators are already correct.


fsb4000 (approved):

StephanTLavavej:

CaseyCarter (approved):

PR #4272, opened 2023-12-16
Mark ASAN failures in libcxx

StephanTLavavej:
Followup to #4263.

* For some MSVC-specific failures, I forgot to mark both :0 (MSVC plain) and :1 (MSVC ASAN).
  + I ran ASAN several times while working on the llvm-project update, but forgot to run it at the very end.
* Mark x86-specific "AddressSanitizer: allocator is out of memory" issues as `:1 SKIPPED`.
  + The STL-ASan-CI had to find these, because I was unable to run whole x86 ASAN test runs (they would hang). However, I'm able to reliably reproduce these failures locally by running the individual directories.
* Comment why 2 tests in mem.poly.allocator.mem are marked for :0 (MSVC) only - it's static analysis warning C6011.
* Consistently comment when we're marking only :0 (MSVC) because :1 (ASAN) doesn't run static analysis.
* Enhance a comment to mention that :1 means ASAN.

In the comments, I thought ":0 (MSVC)" and ":1 (ASAN)" were sufficiently clear, but I could say ":0 (MSVC plain)" and ":1 (MSVC ASAN)" to clarify at the cost of some verbosity.

### DIFF
--- a/tests/libcxx/expected_results.txt
+++ b/tests/libcxx/expected_results.txt
@@ -370,6 +370,7 @@ std/utilities/meta/meta.unary/meta.unary.prop/is_constructible.pass.cpp:0 FAIL
 std/utilities/meta/meta.unary/meta.unary.prop/is_constructible.pass.cpp:1 FAIL
 
 # VSO-1271673 "static analyzer doesn't know about short-circuiting"
+# Note: The :1 (ASAN) configuration doesn't run static analysis.
 std/algorithms/alg.sorting/alg.sort/partial.sort/partial_sort.pass.cpp:0 FAIL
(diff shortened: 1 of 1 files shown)


AlexGuteniev:
Maybe "mark ASAN failures" or "expect ASAN failures", not really "fix" ?

StephanTLavavej:
Done, thanks! :heart_eyes_cat:

CaseyCarter (approved):

CaseyCarter:
> /azp run STL-ASan-CI

This wasn't enabled for various reasons, mostly because I hate things that can't be configured in YAML. See #4295.

StephanTLavavej:

PR #4283, opened 2023-12-22
`<chrono>`: Fix `formatter` ignoring dynamically provided width

### LINKED ISSUE #4201: `<chrono>` `<format>`: `duration` formatting ignores dynamic width
```
C:\Temp>type meow.cpp
```
```cpp
#include <chrono>
#include <format>
#include <iostream>
#include <string>
#include <thread>
using namespace std;

int main() {
#ifdef __GLIBCXX__
    cout << "libstdc++\n";
#elif defined(_LIBCPP_VERSION)
    cout << "libc++\n";
#elif defined(_MSVC_STL_UPDATE)
    cout << "microsoft/STL\n";
#endif
    const auto id = this_thread::get_id();
    cout << format("       int {{:20}}: [{:20}]\n", 1729);
    cout << format("       int {{:{{}}}}: [{:{}}]\n", 1729, 20);
    cout << format("thread::id {{:20}}: [{:20}]\n", id);
    cout << format("thread::id {{:{{}}}}: [{:{}}]\n", id, 20);
    cout << format(" seconds {{:20%T}}: [{:20%T}]\n", 314159s);
    cout << format(" seconds {{:{{}}%T}}: [{:{}%T}]\n", 314159s, 20);
}
```
```
meow.cpp
microsoft/STL
       int {:20}: [                1729]
       int {:{}}: [                1729]
thread::id {:20}: [               34680]
thread::id {:{}}: [               34680]
 seconds {:20%T}: [87:15:59            ]
 seconds {:{}%T}: [87:15:59]
```
https://godbolt.org/z/YbM9qnocf

When formatting a `duration`, the fact that `{:20%T}` respects the statically provided width but `{:{}%T}` ignores the dynamically provided width is *super* buggy, and neither libstdc++ nor libc++ exhibit that behavior.

(There's a separate question of whether `duration` should default to being left-aligned or right-aligned. MSVC and libstdc++ chose left, libc++ chose right.)

(libc++ is also printing the hours mod 24, unlike MSVC and libstdc++.)

mt3d:
Fixes #4201.

As has @cpplearner noted, `_Chrono_formatter::_Write()` calls and returns `_Write_aligned` before trying to check if a width is provided dynamically. The updated behavior is now similar to that of `_Formatter_base::format()`.

### DIFF
--- a/stl/inc/chrono
+++ b/stl/inc/chrono
@@ -5477,7 +5477,14 @@ namespace chrono {
 
             int _Estimated_width = -1;
             (void) _Measure_string_prefix(_Stream.view(), _Estimated_width);
-            return _Write_aligned(_STD move(_FormatCtx.out()), _Estimated_width, _Specs, _Fmt_align::_Left,
+
+            auto _FormatSpecs = _Specs;
+            if (_Specs._Dynamic_width_index >= 0) {
+                _FormatSpecs._Width = _Get_dynamic_specs<_Width_checker>(
+                    _FormatCtx.arg(static_cast<size_t>(_Specs._Dynamic_width_index)));
+            }
+
+            return _Write_aligned(_STD move(_FormatCtx.out()), _Estimated_width, _FormatSpecs, _Fmt_align::_Left,
                 [&](auto _Out) { return _Fmt_write(_STD move(_Out), _Stream.view()); });
         }
 
--- a/stl/inc/chrono
+++ b/stl/inc/chrono
@@ -5478,13 +5478,13 @@ namespace chrono {
             int _Estimated_width = -1;
             (void) _Measure_string_prefix(_Stream.view(), _Estimated_width);
 
-            auto _FormatSpecs = _Specs;
+            auto _Format_specs = _Specs;
             if (_Specs._Dynamic_width_index >= 0) {
-                _FormatSpecs._Width = _Get_dynamic_specs<_Width_checker>(
+                _Format_specs._Width = _Get_dynamic_specs<_Width_checker>(
                     _FormatCtx.arg(static_cast<size_t>(_Specs._Dynamic_width_index)));
             }
 
-            return _Write_aligned(_STD move(_FormatCtx.out()), _Estimated_width, _FormatSpecs, _Fmt_align::_Left,
+            return _Write_aligned(_STD move(_FormatCtx.out()), _Estimated_width, _Format_specs, _Fmt_align::_Left,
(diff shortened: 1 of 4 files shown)


stl/inc/chrono
```diff
@@ -5477,7 +5477,14 @@ namespace chrono {
 
             int _Estimated_width = -1;
             (void) _Measure_string_prefix(_Stream.view(), _Estimated_width);
-            return _Write_aligned(_STD move(_FormatCtx.out()), _Estimated_width, _Specs, _Fmt_align::_Left,
+
+            auto _FormatSpecs = _Specs;
```
StephanTLavavej, line 5481: Naming nitpick: We conventionally use `_Ugly_snake_case` for variables, and `_UglyCamelCase` for template parameters. `<format>` and its related code are somewhat inconsistent (e.g. `_FormatCtx` violates this convention) but we should try to follow it in new code. This will improve consistency with the existing code you mentioned, where the name is `_Format_specs`:

https://github.com/microsoft/STL/blob/87580ca91c1a34645d22f79f7640a13f68bb31c5/stl/inc/format#L3639-L3643

tests/std/tests/GH_004201_chrono_formatter/test.cpp
```diff
@@ -0,0 +1,15 @@
+#include <cassert>
```
StephanTLavavej, line 1: This newly added file needs our copyright/license banner.

tests/std/tests/GH_004201_chrono_formatter/test.cpp
```diff
@@ -0,0 +1,15 @@
+#include <cassert>
+#include <chrono>
+#include <format>
+#include <iostream>
+#include <string>
+#include <thread>
```
StephanTLavavej, line 6: `<iostream>` and `<thread>` were used in #4201, but not in this test, so they should be dropped here.

StephanTLavavej (approved):

StephanTLavavej:
Thanks for fixing this bug and adding a test! :heart_eyes_cat: I pushed a commit to fix minor nitpicks.

We simultaneously merge PRs to GitHub and our MSVC-internal repo; this is a semi-manual process so we batch up PRs to save time. Your PR will be part of the next batch, likely in the next couple of days.

StephanTLavavej:

StephanTLavavej (approved):

StephanTLavavej:
I've pushed a merge with `main` to resolve a trivial adjacent-add conflict in `tests/std/test.lst` with #4275.

PR #4294, opened 2024-01-04
Fix an issue in the special snowflake windows version of the STL

barcharcraz:
We change the size of critical sections based on `_CRT_WINDOWS`, but that macro is defined at build time only, and the size is relevant in our headers. We need to change the size based on `UNDOCKED_WINDOWS_UCRT` as well, which is defined in consumers. 

I've added a pragma detect mismatch as well, but as its own commit. If we don't want it we don't have to accept that commit.

### DIFF
--- a/stl/inc/xthreads.h
+++ b/stl/inc/xthreads.h
@@ -34,6 +34,7 @@ struct _Stl_critical_section {
 
 struct _Mtx_internal_imp_t {
 #if defined(_CRT_WINDOWS) || defined(UNDOCKED_WINDOWS_UCRT)
+#pragma detect_mismatch("_CRT_WINDOWS_mutex_size", "1")
 #ifdef _WIN64
     static constexpr size_t _Critical_section_size = 16;
 #else // ^^^ defined(_WIN64) / !defined(_WIN64) vvv
@@ -41,12 +42,14 @@ struct _Mtx_internal_imp_t {
 #endif // ^^^ !defined(_WIN64) ^^^
 #else // ^^^ defined(_CRT_WINDOWS) || defined(UNDOCKED_WINDOWS_UCRT) / !defined(_CRT_WINDOWS) ||
       // defined(UNDOCKED_WINDOWS_UCRT) vvv
+#pragma detect_mismatch("_CRT_WINDOWS_mutex_size", "0")
 #ifdef _WIN64
     static constexpr size_t _Critical_section_size = 64;
 #else // ^^^ defined(_WIN64) / !defined(_WIN64) vvv
     static constexpr size_t _Critical_section_size = 36;
 #endif // ^^^ !defined(_WIN64) ^^^
 #endif // ^^^ !defined(_CRT_WINDOWS) || defined(UNDOCKED_WINDOWS_UCRT) ^^^
+
 
     static constexpr size_t _Critical_section_align = alignof(void*);
 
--- a/stl/inc/xthreads.h
+++ b/stl/inc/xthreads.h
@@ -32,23 +32,26 @@ struct _Stl_critical_section {
     _Smtx_t _M_srw_lock = nullptr;
 };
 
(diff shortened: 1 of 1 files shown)


stl/inc/xthreads.h
```diff
@@ -33,19 +33,23 @@ struct _Stl_critical_section {
 };
 
 struct _Mtx_internal_imp_t {
-#ifdef _CRT_WINDOWS
+#if defined(_CRT_WINDOWS) || defined(UNDOCKED_WINDOWS_UCRT)
+#pragma detect_mismatch("_CRT_WINDOWS_mutex_size", "1")
```
CaseyCarter, line 37: `_CRT_WINDOWS_mutex_size` seems a bit hefty. How about `windows_mutex_abi`, or just `windows_mutex` given that all detect_mismatch uses are fundamentally about ABI?

stl/inc/xthreads.h
```diff
@@ -33,19 +33,23 @@ struct _Stl_critical_section {
...
+#if defined(_CRT_WINDOWS) || defined(UNDOCKED_WINDOWS_UCRT)
+#pragma detect_mismatch("_CRT_WINDOWS_mutex_size", "1")
 #ifdef _WIN64
     static constexpr size_t _Critical_section_size = 16;
 #else // ^^^ defined(_WIN64) / !defined(_WIN64) vvv
     static constexpr size_t _Critical_section_size = 8;
 #endif // ^^^ !defined(_WIN64) ^^^
-#else // ^^^ defined(_CRT_WINDOWS) / !defined(_CRT_WINDOWS) vvv
+#else // ^^^ defined(_CRT_WINDOWS) || defined(UNDOCKED_WINDOWS_UCRT) / !defined(_CRT_WINDOWS) ||
+      // defined(UNDOCKED_WINDOWS_UCRT) vvv
```
CaseyCarter, line 44: The complement here is improperly De Morgan-ed. Given how ridiculously long it is, I think we should just change this to "Windows private STL" and "public STL".

stl/inc/xthreads.h
```diff
@@ -33,19 +33,23 @@ struct _Stl_critical_section {
...
+#pragma detect_mismatch("_CRT_WINDOWS_mutex_size", "0")
 #ifdef _WIN64
     static constexpr size_t _Critical_section_size = 64;
 #else // ^^^ defined(_WIN64) / !defined(_WIN64) vvv
     static constexpr size_t _Critical_section_size = 36;
 #endif // ^^^ !defined(_WIN64) ^^^
-#endif // ^^^ !defined(_CRT_WINDOWS) ^^^
+#endif // ^^^ !defined(_CRT_WINDOWS) || defined(UNDOCKED_WINDOWS_UCRT) ^^^
```
CaseyCarter, line 51: Ditto "`!(a || b)` is not `!a || !b` but just make this 'public STL' anyway"

CaseyCarter (dismissed):

CaseyCarter:
The build is failing with "C4384: pragma `detect_mismatch` should should only be used at global scope" which seems bogus to me. I suggest that you suppress the warning, verify that the detect_mismatch is correctly emitted into object files, and report the bogus warning to the compiler folks. Failing that, I suppose we'll have to have a separate conditional outside of the class scope just for the detect_mismatch.

barcharcraz:
> The build is failing with "C4384: pragma `detect_mismatch` should should only be used at global scope" which seems bogus to me. I suggest that you suppress the warning, verify that the detect_mismatch is correctly emitted into object files, and report the bogus warning to the compiler folks. Failing that, I suppose we'll have to have a separate conditional outside of the class scope just for the detect_mismatch.

eagh, it is a bit shifty to have pragma detect mismatch in the middle of a class definition, I'll just move it

StephanTLavavej (approved):

PR #4295, opened 2024-01-04
Enable STL-ASAN-CI on PRs

CaseyCarter:
.... but set the UI to require a maintainer comment to trigger.

### DIFF
--- a/azure-devops/asan-pipeline.yml
+++ b/azure-devops/asan-pipeline.yml
@@ -8,7 +8,6 @@ variables:
   - name: testSelection
     value: '-R stlasan' # Restrict to "stlasan" test
 
-pr: none
 trigger: none
 
 schedules:


CaseyCarter:
If and when we merge this change, we need to update the Wiki page that describes how to recreate the pipelines: there's a manual UI step needed to require a maintainer comment before running a pipeline. Without that setting, STL-ASan-CI will now run automatically on all PRs.

As far as a can tell, there's no way to do this purely in YAML - which is why I didn't do so previously. I think the value of triggering the optional pipeline outweighs the annoyance of hand configuration.

StephanTLavavej (approved):

StephanTLavavej:
Thanks for improving the ASAN experience! :chart_with_upwards_trend: :tada: :smile_cat:

CaseyCarter:
> If and when we merge this change, we need to update the Wiki page that describes how to recreate the pipelines: there's a manual UI step needed to require a maintainer comment before running a pipeline. Without that setting, STL-ASan-CI will now run automatically on all PRs.

I have [updated the Wiki instructions](https://github.com/microsoft/STL/wiki/Setting-Up-Azure-Pipelines/_compare/8a92d5db4ee096f281f4298747aa6eeb5037345f...d14b70d31b67322139c17f9d564fb24dde17b922).

Issue #4320, opened 2024-01-16
<format>: crash when formatting a floating-point value with `#`/`L` and a large precision


cpplearner:
# Describe the bug
`<format>` uses a loop to find the beginning of the exponent part in order to insert trailing zeroes. But when `_Specs._Alt || _Specs._Localized` is true, the position has already been stored in `_Exponent_start`. The loop actually moves `_Exponent_start` away from the correct position, and eventually causes a crash due to out-of-bound access.

https://github.com/microsoft/STL/blob/3eac329d1f614ecf138d96c22a3b02f87076bc4a/stl/inc/format#L3108-L3112

https://github.com/microsoft/STL/blob/3eac329d1f614ecf138d96c22a3b02f87076bc4a/stl/inc/format#L3087-L3089

# Command-line test case
```
D:\test>type test-format-3.cpp
#include <format>
#include <iostream>

int main() {
    std::cout << std::format("{:#.1075e}\n", 1.0);
}

D:\test>cl /std:c++20 /EHs /utf-8 test-format-3.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.39.33321 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

test-format-3.cpp
Microsoft (R) Incremental Linker Version 14.39.33321.0
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:test-format-3.exe
test-format-3.obj

D:\test>test-format-3

D:\test>echo %errorlevel%
-1073741819
```

# Expected behavior
no crash

# STL version

    https://github.com/microsoft/STL/commit/3eac329d1f614ecf138d96c22a3b02f87076bc4a

PR #4321, opened 2024-01-16
Issue #4320 Fixes handling of exponent positioning in <format> to pre…

anirudhrana800:
…vent out-of-bound access and crashes. When _Specs._Alt || _Specs._Localized is true, the correct position is already stored in _Exponent_start, eliminating the need for unnecessary loop execution. This modification ensures code stability and prevents issues related to trailing zeroes insertion.

### DIFF
--- a/stl/inc/format
+++ b/stl/inc/format
@@ -3106,9 +3106,11 @@ _NODISCARD _OutputIt _Fmt_write(
         case chars_format::hex:
         case chars_format::scientific:
             if (_Extra_precision != 0) {
+                 if (!(_Specs._Alt || _Specs._Localized)){
                 // Trailing zeroes are in front of the exponent
                 while (*--_Exponent_start != _Exponent) {
                 }
+             }     
             }
             [[fallthrough]];
         case chars_format::fixed:


StephanTLavavej:
Thanks for working on this!

* The CI is failing because your code needs to be clang-formatted. It's best to actually use clang-format (which can be obtained via the VS Installer), but for very small changes you can manually adjust your code. If you follow enough links, you can see where our build logs report what clang-format expects:
  + Click on "Details" after "STL-CI (Code Format Validation)" in the failing checks.
  + Click on "View more details on Azure Pipelines".
  + There's a red X next to the Validation job, and another red X next to the Create Diff task. Click "Create Diff".
  + This log explains how to run clang-format if you have it installed, or how to apply the `format.diff` that we've saved. You can also click on "Expected formatting - click to expand diff" to see the diff.
* Instead of mentioning the issue number in the PR title (where it has no effect), you should edit your PR description to use [GitHub's close/fix/resolve syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword). That is, if you say `Fixes #4320` in your PR description, it'll create a special link from this PR to that issue, such that merging this PR will close that issue. (If an issue is mentioned in an ordinary comment like mine, then a link is created for the reference, but without the special auto-closing behavior.) As a maintainer I have the ability to edit your PR title and description but I thought I'd explain how to do it yourself and why. :smile_cat:
* Ideally we should add test coverage, to ensure that this stays fixed.

fsb4000:
You can just run from "Developer Command Prompt for VS 2022 Preview":
```
clang-format -i some_cpp_file
```
for every cpp file you have changed.
In your case it would be:
```
clang-format -i stl/inc/format
```
and your files will be formatted correctly.

PR #4323, opened 2024-01-16
Report as UNRESOLVED when exception is thrown in `execute`

CaseyCarter:
This doesn't solve the underlying problem causing the failure but will hopefully provide enough info for us to analyze it.

Addresses #4308


StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for figuring out how to make these mysterious failures more actionable! :bulb: :tada: :heart_eyes_cat:

PR #4328, opened 2024-01-18
Move libc++'s `ascii` tests to "LIKELY BOGUS TESTS"

cpplearner:
libc++'s "ascii" mode (`_LIBCPP_HAS_NO_UNICODE`) means "every code unit outside ASCII is treated as a valid printable character". AFAIK we don't support such a mode.

For `char`, MSVC STL uses the code page specified by `_MSVC_EXECUTION_CHARACTER_SET`. Some code pages might consider most code units printable, but even Windows 1252 has a couple of non-printable characters.

For `wchar_t`, we don't have a way to opt-out UTF-16, and I don't see a need to do so.

### DIFF
--- a/tests/libcxx/expected_results.txt
+++ b/tests/libcxx/expected_results.txt
@@ -712,6 +712,11 @@ std/iterators/iterator.requirements/iterator.concepts/iterator.concept.random.ac
 # Bogus test expects to_address() to SFINAE away for int.
 std/utilities/memory/pointer.conversion/to_address_without_pointer_traits.pass.cpp FAIL
 
+# libc++'s "ascii" mode considers every code unit to represent a valid printable narrow character.
+# We don't support such a mode.
+std/utilities/format/format.functions/ascii.pass.cpp FAIL
+std/utilities/format/format.functions/escaped_output.ascii.pass.cpp FAIL
+
 
 # *** LIKELY STL BUGS ***
 # Not analyzed, likely STL bugs. Various assertions.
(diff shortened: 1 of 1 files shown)


tests/libcxx/expected_results.txt
```diff
@@ -712,6 +712,11 @@ std/iterators/iterator.requirements/iterator.concepts/iterator.concept.random.ac
 # Bogus test expects to_address() to SFINAE away for int.
 std/utilities/memory/pointer.conversion/to_address_without_pointer_traits.pass.cpp FAIL
 
+# libc++'s "ascii" mode considers every code unit to represent a valid printable narrow character.
+# We don't support such a mode.
```
CaseyCarter, line 716:
```suggestion
# These test libc++'s non-standard "ascii" mode. We should submit a PR to relocate them from tests/std to tests/libcxx.
```

tests/libcxx/expected_results.txt
```diff
@@ -712,6 +712,11 @@ std/iterators/iterator.requirements/iterator.concepts/iterator.concept.random.ac
 # Bogus test expects to_address() to SFINAE away for int.
 std/utilities/memory/pointer.conversion/to_address_without_pointer_traits.pass.cpp FAIL
 
+# libc++'s "ascii" mode considers every code unit to represent a valid printable narrow character.
+# We don't support such a mode.
+std/utilities/format/format.functions/ascii.pass.cpp FAIL
+std/utilities/format/format.functions/escaped_output.ascii.pass.cpp FAIL
```
CaseyCarter, line 718: `FAIL` feels wrong for this: I don't care that these are failing and wouldn't care if they start passing. Let's `SKIPPED` them instead.

CaseyCarter (changes requested):

tests/libcxx/expected_results.txt
StephanTLavavej: I've created such a PR and will push a change here.

StephanTLavavej (approved):

CaseyCarter (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for investigating what was happening here! :detective: :bulb: :heart_eyes_cat:

PR #4335, opened 2024-01-21
Run CTest tests serially to avoid oversubscribing cores

horenmar:
The lit tests are already internally parallelized, so using CTest to run them in parallel ends up with the CPU cores hopelessly oversubscribed.


fsb4000 (approved):
I've checked that the CI time doesn't change after the update, and I also consulted the CMake documentation to ensure there were no misspellings:
https://cmake.org/cmake/help/latest/command/set_tests_properties.html
https://cmake.org/cmake/help/latest/prop_test/RUN_SERIAL.html

StephanTLavavej (approved):

StephanTLavavej:
Thanks! I never run `ctest` (instead I always invoke `stl-lit.py` directly), so I didn't realize it was trying to run `stl` and `stlasan` simultaneously with separate Python invocations. Great catch :heart_eyes_cat:

CaseyCarter (approved):

StephanTLavavej:

StephanTLavavej:
Thanks again! :tada: :smile_cat: :rocket:

PR #3950, opened 2023-08-11
Update preprocessor `#else` and `#endif` comment

AlexGuteniev:
More towards #351. Product files (non-test).

Skipped in the previous #3900
 * `&&` / `||`, de-morganed
 * lone `#endif`, not specifically targeted in the previous PR
 * `yals_core.h` / `yvals.h` macro introduction
 * `#elif`-containing blocks
 * Missed for some other reasons

PR #4339, opened 2024-01-24
Revert "Make the constexpr mutex constructor opt-in (#4000)"

### LINKED ISSUE #4338: `<mutex>` : enable `constexpr` default constructor by, err, default
Some background:

#3824 fixed a decade-old bug by making the default constructor of `std::mutex` be `constexpr`. There was much rejoicing. There was also some wailing and gnashing of teeth from folks in Microsoft who were using current headers and old libraries - an unsupported scenario - who had to fix their broken builds/processes/pipelines. We had anticipated such problems, and added an opt-out macro specifically so folks could transition on their own timelines.

At the 11th hour before release, we realized that MSVC supports the UWP framework libraries from the Windows store with the live headers (:boggle:). While we were happy to stick to our guns and force misbehaving internals to fix their stuff in order to ship a fix for a ten-year-old bug, breaking well-behaved users was a non-starter. #4000 made the feature opt-in instead of opt-out to avoid breaking folks in 17.8 until we could get the UWP framework libraries updated.

The UWP library update process is incredibly slow but should be complete in time for us to re-enable the `constexpr` constructor by default (flipping back from opt-in to opt-out) in the 17.10 timeframe. This issue is a reminder to do so.

CaseyCarter:
This reverts commit 782dcd536a57fd5c6a215d1b9a3014ba4c25282f.

There were some minor merge conflicts (we added `^^^` to the `#endif` comment, and ASan changed the `env.lst` file) so this isn't a pure reversion.

Fixes #4338


StephanTLavavej:
This is possible now that @amyw-msft merged her MSVC-PR-519388 "Update crt140 nuget package to version 33321 (17.9p2)" today. Thanks for taking care of this @CaseyCarter, it was on my list of things to do today and I didn't know it would take extra effort beyond a plain revert :heart_eyes_cat:

stl/inc/mutex
```diff
@@ -32,18 +32,18 @@ _EXPORT_STD class condition_variable_any;
 
 class _Mutex_base { // base class for all mutex types
 public:
-#ifdef _ENABLE_CONSTEXPR_MUTEX_CONSTRUCTOR
+#ifdef _DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR
+    _Mutex_base(int _Flags = 0) noexcept {
+        _Mtx_init_in_situ(_Mymtx(), _Flags | _Mtx_try);
+    }
+#else // ^^^ _DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR / !_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR vvv
```
StephanTLavavej, line 39: After #4000 merged on 2023-08-30, #3950 merged on 2023-09-21 did more than add arrows - it added discipline regarding the difference between `#if` and `#ifdef`. That is, `!_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR` is not really correct - the region should be properly identified as `!defined(_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR)`.

tests/std/tests/VSO_0226079_mutex/test.cpp
```diff
@@ -478,7 +478,7 @@ test_constexpr_ctor obj;
 #if _HAS_CXX20 && !defined(_M_CEE)
 constinit test_constexpr_ctor obj2;
 #endif // _HAS_CXX20 && !defined(_M_CEE)
-#endif // _ENABLE_CONSTEXPR_MUTEX_CONSTRUCTOR
+#endif // !_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR
```
StephanTLavavej, line 481: Ditto `defined`.

CaseyCarter:
> Thanks for taking care of this @CaseyCarter, it was on my list of things to do today and I didn't know it would take extra effort beyond a plain revert 😻

#4000 was me, so I felt obliged to fix the damage I did to the STL =)

StephanTLavavej (approved):

StephanTLavavej:
Pushed comment changes!

StephanTLavavej:

StephanTLavavej:
Thanks for undoing your undo! :repeat: :heart_eyes_cat: :tada:

PR #4340, opened 2024-01-24
Remove a `/clr` workaround in test code

StephanTLavavej:
@Codiferous's MSVC-PR-522996 fixed VSO-1078742 DevCom-939490 "Calling a pointer to member function at `constexpr` time in `/clr` triggers internal compiler error" reported on 2020-03-04, so we can finally remove this workaround. As this is `/clr`, we can do so immediately, no need to wait for a toolset update. I've verified this internally.


StephanTLavavej:

CaseyCarter (approved):

PR #4341, opened 2024-01-25
ARM64 test improvements

CaseyCarter:
This is a dual of MSVC-PR-407828, which enables lots of ARM64 CI in the internal MSVC repository.


tests/std/tests/P0226R1_math_special_functions/test.cpp
```diff
@@ -260,10 +260,10 @@ void test_cyl_neumann() {
         assert(isclose(cyl_neumannf(0.5f, 0.333f), -1.30'671'255'810'072'199f));
         assert(isclose(cyl_neumannl(0.5L, 0.333L), -1.30'671'255'810'072'199L, 2));
     }
-    {
-        assert(isclose(cyl_neumann(0.5, 0.666), -0.768'760'134'281'402'040));
+    {
```
CaseyCarter, line 263: Ick, didn't notice this trailing whitespace when reviewing the internal PR.

CaseyCarter: I've removed the trailing whitespace here so we can merge, but it's still present in the internal PR. (I didn't want to reset testing for this enormous PR that's already up to 110 updates.) We'll "fix it in post" (remove the whitespace internally in a followup PR) to avoid divergence.

StephanTLavavej (approved):

tests/std/tests/P0226R1_math_special_functions/test.cpp
StephanTLavavej: This was fixed internally so we have one less thing to worry about :tada:

StephanTLavavej:
:tada: :smile_cat: :rocket:
===

PR #4348, opened 2024-01-29
llvm-project Mini-Update

StephanTLavavej:
I started by adding a script `check_libcxx_paths.py` to make LLVM updates easier. Then I figured I'd see how easy it was, and it only took me a day instead of a month.

llvm/llvm-project#79793 is still in flight, but the affected test is blocked by a suspected MSVC compiler bug with `constexpr`, so we don't need to wait.

### DIFF
--- a/tests/libcxx/expected_results.txt
+++ b/tests/libcxx/expected_results.txt
@@ -1341,6 +1341,25 @@ std/input.output/filesystems/fs.op.funcs/fs.op.symlink_status/symlink_status.pas
 std/input.output/filesystems/fs.op.funcs/fs.op.temp_dir_path/temp_directory_path.pass.cpp SKIPPED
 std/input.output/filesystems/fs.op.funcs/fs.op.weakly_canonical/weakly_canonical.pass.cpp SKIPPED
 
+# Not analyzed.
+# MSVC error C2719: '_Fun_': formal parameter with requested alignment of 128 won't be aligned
+# Clang error: unknown attribute 'no_unique_address' ignored [-Werror,-Wunknown-attributes]
+std/ranges/ranges_robust_against_no_unique_address.pass.cpp FAIL
+
+# Not analyzed. Runtime assertion, we appear to be ignoring stream width.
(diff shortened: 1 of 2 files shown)


cpplearner:
> [llvm/llvm-project#79793](https://github.com/llvm/llvm-project/pull/79793) is still in flight, but the affected test is blocked by a suspected MSVC compiler bug with `constexpr`, so we don't need to wait.

It's probably the same bug as DevCom-10439137 "Discarded id-expression causes unnecessary reading".

StephanTLavavej:
Yep, I agree! Great catch, I didn't know about that bug report. I've updated the entry.

I also noted in VSO-1869865 (the internal mirror of DevCom-10439137) that the ICE I just reduced, VSO-1948221 "x86chk ICE with `constexpr type_info`: Assertion failed: `isIndirection()`", is likely related - it involved `constexpr` and a `(void)` cast, which I found curious at the time.

CaseyCarter (approved):

StephanTLavavej:
I had to push an additional commit because new tests appeared that require Unix headers, and the MSVC-internal test harness doesn't understand this.

cpplearner:
Should `UNSUPPORTED: !has-unix-headers` be added to [`magic_comments.txt`](https://github.com/microsoft/STL/blob/main/tests/libcxx/magic_comments.txt)? I don't understand what this file does, but apparently it's used by the MSVC-internal test harness.

CaseyCarter (approved):

StephanTLavavej:
@cpplearner That's a good idea. `magic_comments.txt` is indeed used by the MSVC-internal test harness - it's an extremely primitive mechanism that I wrote a while ago (later adapted by Casey) where if a test mentions any of those "magic comments", it's skipped. This imitates the real test harness's implementation of "feature" detection.

We already have `// REQUIRES: has-unix-headers` there; arguably upstream should be changed to say that, avoiding the current double negation. (The comment matching we do is also whole-line, as I recall.)

Since the PR and its mirror have passed checks, I'll explore doing this the next time I update LLVM. Thanks!

PR #4351, opened 2024-01-30
`<generator>`: Mark strengthened noexcept (plus tiny drive-by changes)

cpplearner:
Mostly tiny adjustments to better match the standard.

Also implements LWG-3894 "`generator::promise_type::yield_value(ranges::elements_of<Rng, Alloc>)` should not be `noexcept`".

Works towards #2936.

### DIFF
--- a/stl/inc/generator
+++ b/stl/inc/generator
@@ -362,7 +362,7 @@ public:
         return *this;
     }
 
-    _NODISCARD _Ref operator*() const noexcept(is_nothrow_copy_constructible_v<_Ref>) {
+    _NODISCARD _Ref operator*() const noexcept(noexcept(static_cast<_Ref>(*_Coro.promise()._Top.promise()._Ptr))) {
         _STL_ASSERT(!_Coro.done(), "Can't dereference generator end iterator");
         return static_cast<_Ref>(*_Coro.promise()._Top.promise()._Ptr);
     }
--- a/stl/inc/generator
+++ b/stl/inc/generator
@@ -210,7 +210,7 @@ public:
     }
 
     _NODISCARD auto yield_value(const remove_reference_t<_Yielded>& _Val) noexcept(
-        is_nothrow_constructible_v<remove_cvref_t<_Yielded>, const remove_reference_t<_Yielded>&>)
+        is_nothrow_constructible_v<remove_cvref_t<_Yielded>, const remove_reference_t<_Yielded>&>) /* strengthened */
         requires (is_rvalue_reference_v<_Yielded>
                   && constructible_from<remove_cvref_t<_Yielded>, const remove_reference_t<_Yielded>&>)
     {
@@ -225,7 +225,7 @@ public:
 
     template <_RANGES input_range _Rng, class _Alloc>
         requires convertible_to<_RANGES range_reference_t<_Rng>, _Yielded>
-    _NODISCARD auto yield_value(_RANGES elements_of<_Rng, _Alloc> _Elem) noexcept {
+    _NODISCARD auto yield_value(_RANGES elements_of<_Rng, _Alloc> _Elem) {
         using _Vty = _RANGES range_value_t<_Rng>;
         return _Nested_awaitable<_Yielded, _Vty, _Alloc>{
             [](allocator_arg_t, _Alloc, _RANGES iterator_t<_Rng> _It,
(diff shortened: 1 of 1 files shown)


stl/inc/generator
```diff
@@ -362,7 +362,7 @@ public:
         return *this;
     }
 
-    _NODISCARD _Ref operator*() const noexcept {
+    _NODISCARD _Ref operator*() const noexcept(is_nothrow_copy_constructible_v<_Ref>) {
```
CaseyCarter, line 365: This is constructing, but IIRC not necessarily copy constructing, a `_Ref`. Should this be `is_nothrow_constructible_v<_Ref, decltype(*_Coro.promise()._Top.promise()._Ptr)>`?

cpplearner: I think `is_nothrow_copy_constructible_v` is correct.

- `_Ptr` has type `add_pointer_t<_Yielded>`.
- `_Yielded` is either `_Ref` (if it is a reference) or `const _Ref&`.
- so `decltype(*_Ptr)` is either `_Ref&` or `const _Ref&`.
- If `_Ref` is a reference, then the construction from `_Ref&` is non-throwing, and `is_nothrow_copy_constructible_v<_Ref>` happens to be `true` as well.
- If `_Ref` is not a reference, then `decltype(*_Ptr)` must be `const _Ref&`, and the condition is equivalent to `is_nothrow_copy_constructible_v<_Ref>`.

The standard [also says `noexcept(is_nothrow_copy_constructible_v<reference>)`](https://eel.is/c++draft/coro.generator.iterator#lib:operator*,generator::iterator).

cpplearner: I think the condition needs to be `noexcept(static_cast<_Ref>(*_Coro.promise()._Top.promise()._Ptr))`.

If `_Ref` is an rvalue reference, this is a cast from lvalue to rvalue reference. Such a cast is non-throwing IIUC, but `is_nothrow_constructible_v` is `false`.

The standard says [`noexcept(is_nothrow_copy_constructible_v<reference>)`](https://eel.is/c++draft/coro.generator.iterator#lib:operator*,generator::iterator), which also needs to be fixed.

stl/inc/generator
```diff
@@ -210,7 +210,7 @@ public:
     }
 
     _NODISCARD auto yield_value(const remove_reference_t<_Yielded>& _Val) noexcept(
-        is_nothrow_constructible_v<remove_cvref_t<_Yielded>, const remove_reference_t<_Yielded>&>)
+        is_nothrow_constructible_v<remove_cvref_t<_Yielded>, const remove_reference_t<_Yielded>&>) /* strengthened */
```
StephanTLavavej, line 213: Stylistic nitpick: We use `/* strengthened */` when it's embedded in a line (especially before an opening brace), but the following `requires` will always be clang-formatted on a new line, so this should ideally be a C++ comment `// strengthened`. Not worth resetting testing though.

stl/inc/generator
```diff
@@ -377,8 +377,9 @@ public:
         ++*this;
     }
 
-    _NODISCARD bool operator==(default_sentinel_t) const noexcept {
-        return _Coro.done();
+    _NODISCARD_FRIEND bool operator==(const _Gen_iter& _It, default_sentinel_t) noexcept /* strengthened */
+    {
```
StephanTLavavej, line 381: Another stylistic nitpick: There's enough room for the brace here to be attached; clang-format exhibits "sticky" behavior and will preserve it. Also not worth resetting testing for this PR.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! Had a couple of formatting nitpicks that can be addressed in followup PRs, but the functional changes all look great. :heart_eyes_cat:

stl/inc/generator
CaseyCarter: On further thought, I believe the logic here is "If `_Ref` is a reference type, this function doesn't throw, which is consistent with `is_nothrow_copy_constructible_v<_Ref>`. If it's not a reference type, it must be a (cv-unqualified?) object type, and `_Ptr` is a pointer to (const?) `_Ref` so the `static_cast` expression in fact does copy a `_Ref`."

What we've merged here is correct, but we should nonetheless change it in a followup to obviously agree with the Standard wording.

PR #4356, opened 2024-01-31
Enable warnings for the benchmarks, and fix them

StephanTLavavej:
Discovered while reviewing #3928, but this has been lurking since the addition of the benchmarks in #2780. No stealth merge conflicts - #3928's new benchmark has been fixed to be warning-clean.

* Enable warnings when building the benchmarks.
  + Start with the usual: `/diagnostics:caret /W4 /WX`
  + I'm enabling the same off-by-default warnings as the STL's build (C4265 (non-virtual dtor), C5038 (data member init order)) and adding C5262 (implicit fallthrough).
  + Finally, let's add `/utf-8` in case we ever benchmark `<format>`.
* Fix x86 `size_t` truncation warnings in the vector.bool benchmarks.
* `udt.hpp`: Rename template parameters from `Contained` to `Data`, avoiding REALLY confusing diagnostics.
* `utility.hpp`: Silence truncation warnings with a big comment.

### DIFF
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -61,6 +61,9 @@ endif()
 
 set(CMAKE_BUILD_TYPE RelWithDebInfo)
 
+# /utf-8 affects <format>.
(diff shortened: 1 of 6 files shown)


StephanTLavavej:

CaseyCarter (approved):

PR #4370, opened 2024-02-06
Merge `main` to `feature/generator`

StephanTLavavej:

* Merge `main` to `feature/generator`
  + Resolved conflict in `yvals_core.h`, removing all guards around `__cpp_lib_generator` except `_HAS_CXX23`.
    - `__cpp_lib_concepts` is now always available in C++20 and later.
    - `<generator>` doesn't care about `std::byte` now.
    - C1XX, Clang, and EDG all define `__cpp_impl_coroutine` in C++20 and later, which wasn't the case earlier.
* `concepts_latest_matrix.lst` => `usual_latest_matrix.lst`
* Adjust `VSO_0157762_feature_test_macros` to match `yvals_core.h`.
* Update `<generator>`'s `_HAS_CXX23` guard and message.
  + Also add a missing semicolon - this emits a `_Pragma`.
* Work around VSO-1951821 "EDG instantiates the wrong overload of `promise_type::operator new` for `generator` machinery" by skipping `stateful_alloc` coverage for EDG.
  + Thanks to @cpplearner for reporting this bug.
* `std.ixx`: Guard `<generator>` with `_HAS_CXX23`.
* Guard `import <generator>;` with `#if TEST_STANDARD >= 23`.
* Use `^^^ no workaround ^^^` comments around good code.


stl/inc/generator
```diff
@@ -9,9 +9,9 @@
 #include <yvals_core.h>
 #if _STL_COMPILER_PREPROCESSOR
 
-#if !_HAS_CXX23 || !defined(__cpp_impl_coroutine) || !defined(__cpp_lib_concepts) // TRANSITION, GH-395
-_EMIT_STL_WARNING(STL4038, "The contents of <generator> are available only with C++23 or later coroutine support.")
-#else // ^^^ no coroutine support / coroutines vvv
+#if !_HAS_CXX23
+_EMIT_STL_WARNING(STL4038, "The contents of <generator> are available only with C++23 or later.");
```
CaseyCarter, line 13: I led you astray here: we need the `!defined(__cpp_impl_coroutine)` (and the vague "coroutine support" in the error message, which could probably be improved) to disable the header when `/await` is on the command line. `/await` tells MSVC to disable the standard coroutine ABI and use the pre-standard ABI instead. We do _not_ want to deal with this potential ODR fiasco, so I think we should revert the changes to lines 12, 13 and 471.

tests/std/tests/P1502R1_standard_library_header_units/test.cpp
```diff
@@ -36,7 +36,9 @@ import <forward_list>;
 import <fstream>;
 import <functional>;
 import <future>;
+#if TEST_STANDARD >= 23
```
CaseyCarter, line 39: I don't think we need any additional guards here, since we don't have (and IMO should never have) `/await` coverage of this standard library header unit test.

CaseyCarter (changes requested):

stl/inc/generator
CaseyCarter: I'm mistaken: `<coroutine>` (which `<generator>` includes) will `_EMIT_STL_ERROR` when included with `/await`, so `<generator>` need not guard against `/await`.

tests/std/tests/P1502R1_standard_library_header_units/test.cpp
CaseyCarter: This is still true, but not needed to resolve any confusion since https://github.com/microsoft/STL/pull/4370#discussion_r1480749847 is invalid.

CaseyCarter (approved):

stl/inc/generator
StephanTLavavej: It's actually `_EMIT_STL_WARNING` followed by expanding to nothing, which will presumably cause `<generator>` to catastrophically fail to compile:

https://github.com/microsoft/STL/blob/192a84008a59ac4d2e55681e1ffac73535788674/stl/inc/coroutine#L11-L14

This is just a quality of diagnostic issue, so I'll proceed with this merge and we can do something about it later if we want. Thanks for looking into this.

PR #4372, opened 2024-02-07
Fix rethrowing the caught exception from streambuf

### LINKED ISSUE #4322: `<ostream>`: Exception from streambuf should be caught and not rethrown
Originally reported as DevCom-10555755 / VSO-1939195 / AB#1939195 . I've cleaned up the repro and especially the Standardese citations, so I believe this is a bug.

```
```
```cpp
#include <cassert>
#include <iostream>
#include <sstream>
#include <stdexcept>
using namespace std;

class MyStreamBuf : public streambuf {
protected:
    // N4971 [streambuf.virt.get]/1-2:
    // "streamsize showmanyc();
    // Returns: An estimate of the number of characters available in the sequence, or -1.
    // If it returns a positive value, then successive calls to underflow() will not return
    // traits::eof() until at least that number of characters have been extracted from the stream.
    // If showmanyc() returns -1, then calls to underflow() or uflow() will fail. [Note 277]
    // Default behavior: Returns zero."

    // Note 277:
    // "underflow or uflow can fail by throwing an exception prematurely. The intention
    // is not only that the calls will not return eof() but that they will return "immediately"."

    int_type underflow() override {
        throw runtime_error("WOOF");
    }
};

int main() {
    MyStreamBuf rdbuf;
    ostringstream oss;
    assert(oss.exceptions() == ios_base::goodbit);

    try {
        // N4971 [ostream.inserters]/8-9:
        // "basic_ostream& operator<<(basic_streambuf<charT, traits>* sb);
        // Gets characters from sb and inserts them in *this. Characters are
        // read from sb and inserted until any of the following occurs:
        // [...]
        // - an exception occurs while getting a character from sb.
        // If the function inserts no characters, it calls setstate(failbit) (which
        // may throw ios_base::failure ([iostate.flags])). If an exception was thrown
        // while extracting a character, the function sets failbit in the error state,
        // and if failbit is set in exceptions() the caught exception is rethrown."
        oss << &rdbuf;
        cout << "No exception observed in main().\n";
    } catch (const runtime_error&) {
        cout << "Caught runtime_error in main().\n";
    }

    assert(oss.rdstate() == ios_base::failbit);
}
```
```
D:\GitHub\STL\out\x64>cl /EHsc /nologo /W4 /MTd /Od meow.cpp && meow
meow.cpp
Caught runtime_error in main().
```

libstdc++ and libc++ print "No exception observed in main()." as shown by https://godbolt.org/z/Yq3noKn8s .

grcm10:
Fixes #4322

### DIFF
--- a/stl/inc/ostream
+++ b/stl/inc/ostream
@@ -493,8 +493,10 @@ public:
                 _TRY_BEGIN
                 _Meta = _Traits::eq_int_type(_Traits::eof(), _Meta) ? _Strbuf->sgetc() : _Strbuf->snextc();
                 _CATCH_ALL
-                _Myios::setstate(ios_base::failbit);
-                _RERAISE;
+                // N4971 [ostream.inserters]/8-9: "...If an exception was thrown
+                // while extracting a character, the function sets failbit in the error state,
+                // and if failbit is set in exceptions() the caught exception is rethrown."
+                _Myios::setstate(ios_base::failbit, (_Myios::exceptions() == ios_base::failbit));
                 _CATCH_END
 
                 if (_Traits::eq_int_type(_Traits::eof(), _Meta)) {
--- a/stl/inc/ostream
+++ b/stl/inc/ostream
@@ -493,10 +493,10 @@ public:
                 _TRY_BEGIN
                 _Meta = _Traits::eq_int_type(_Traits::eof(), _Meta) ? _Strbuf->sgetc() : _Strbuf->snextc();
                 _CATCH_ALL
-                // N4971 [ostream.inserters]/8-9: "...If an exception was thrown
+                // N4971 [ostream.inserters]/9: "If an exception was thrown
                 // while extracting a character, the function sets failbit in the error state,
                 // and if failbit is set in exceptions() the caught exception is rethrown."
-                _Myios::setstate(ios_base::failbit, (_Myios::exceptions() == ios_base::failbit));
+                _Myios::setstate(ios_base::failbit, _Myios::exceptions() == ios_base::failbit);
                 _CATCH_END
 
                 if (_Traits::eq_int_type(_Traits::eof(), _Meta)) {
--- a/tests/std/tests/GH_001858_iostream_exception/test.cpp
+++ b/tests/std/tests/GH_001858_iostream_exception/test.cpp
@@ -36,6 +36,8 @@ struct test_exception {};
 template <class CharT>
 class throwing_buffer : public basic_streambuf<CharT> {
 public:
+    using typename basic_streambuf<CharT>::int_type;
+
     streampos seekoff(streamoff, ios_base::seekdir, ios_base::openmode = ios_base::in | ios_base::out) override {
         throw test_exception{};
     }
@@ -45,6 +47,10 @@ public:
     }
 
     int sync() override {
+        throw test_exception{};
+    }
+
+    int_type underflow() override {
         throw test_exception{};
     }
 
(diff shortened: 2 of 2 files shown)


kirankokare09 (approved):

stl/inc/ostream
```diff
@@ -493,8 +493,10 @@ public:
                 _TRY_BEGIN
                 _Meta = _Traits::eq_int_type(_Traits::eof(), _Meta) ? _Strbuf->sgetc() : _Strbuf->snextc();
                 _CATCH_ALL
-                _Myios::setstate(ios_base::failbit);
-                _RERAISE;
+                // N4971 [ostream.inserters]/8-9: "...If an exception was thrown
```
StephanTLavavej, line 496: Reducing this citation from the longer one in #4322 is good, but now this is quoting part of paragraph 9 only, so it should be adjusted accordingly.

stl/inc/ostream
```diff
@@ -493,8 +493,10 @@ public:
                 _TRY_BEGIN
                 _Meta = _Traits::eq_int_type(_Traits::eof(), _Meta) ? _Strbuf->sgetc() : _Strbuf->snextc();
                 _CATCH_ALL
-                _Myios::setstate(ios_base::failbit);
-                _RERAISE;
+                // N4971 [ostream.inserters]/8-9: "...If an exception was thrown
+                // while extracting a character, the function sets failbit in the error state,
+                // and if failbit is set in exceptions() the caught exception is rethrown."
+                _Myios::setstate(ios_base::failbit, (_Myios::exceptions() == ios_base::failbit));
```
StephanTLavavej, line 499: The parentheses here are unnecessary. We usually omit unnecessary parentheses, except in certain contexts where precedence is notoriously confusing for humans (e.g. around bitwise operators) or to silence compiler warnings (e.g. when mixing `&&` and `||`).

tests/std/tests/GH_001858_iostream_exception/test.cpp
```diff
@@ -229,6 +235,33 @@ void test_ostream_exceptions() {
...
+        try {
+            os << &buffer;
+        } catch (const ios_base::failure&) {
+            assert(false);
+        } catch (const test_exception&) {
+            assert(false);
+        }
+    }
+
+    { // operator<< rethrow the caught exception if failbit is set in exceptions()
```
StephanTLavavej, line 252: Grammar: `operator<<` is singular, so it `rethrows`.

tests/std/tests/GH_001858_iostream_exception/test.cpp
```diff
@@ -229,6 +235,33 @@ void test_ostream_exceptions() {
             // Expected case
         }
     }
+
+    { // operator<< with exceptions
+        basic_ostream<CharT> os(buffer.to_buf());
+        os.exceptions(ios_base::goodbit);
```
StephanTLavavej, line 241: This is inconsistent. `exceptions()` defaults to `goodbit`, so we don't need to say that here. Above, there are pairs of test cases that say `// meow` (using the default behavior) and then `// meow with exceptions` where `exceptions()` is called. I think we should follow that convention here.

For extra clarity, we could cite #4322.

StephanTLavavej (approved):

StephanTLavavej:
Thanks for fixing this bug and especially for adding test coverage, this looks great! :heart_eyes_cat: I pushed a few nitpicky changes for consistency, nothing major. This should be part of our next merge batch (not sure when, maybe this week, possibly next week).

StephanTLavavej:

StephanTLavavej:
Thanks for fixing this ancient bug! :lady_beetle: :hammer_and_wrench: :heart_eyes_cat:

PR #4375, opened 2024-02-08
Export VCRuntime entities properly

### LINKED ISSUE #4350: Exported declaration of `std::exception` follows non-exported declaration
In the header `<exception>` firstly `<vcruntime_exception.h>` is [included](https://github.com/microsoft/STL/blob/32e65ac3e54a7150202cc9b3cc4de7330d58d24e/stl/inc/exception#L34) and then `class exception` is redeclared [as exported](https://github.com/microsoft/STL/blob/32e65ac3e54a7150202cc9b3cc4de7330d58d24e/stl/inc/exception#L39). This is incorrect by the standard (https://eel.is/c++draft/module.interface#6), namely it repeats https://eel.is/c++draft/module.interface#example-4, where `struct S` is replaced with `class exception`.

StephanTLavavej:
Fixes #4350. Mirrored as MSVC-PR-527804 with VCRuntime changes.

# :lady_beetle: The Problem
The Standard has a rule WG21-N4971 \[module.interface\]/6: "A redeclaration of an entity *X* is implicitly exported if *X* was introduced by an exported declaration; otherwise it shall not be exported." followed by an example:

```cpp
export module M;
struct S { int n; };
typedef S S;
export typedef S S; // OK, does not redeclare an entity
export struct S;    // error: exported declaration follows non-exported declaration
```

[Clang enforces this rule](https://godbolt.org/z/bcG9b4hzv), but MSVC let me be a bad kitty :smirk_cat:, jumping up on the table where I'm not allowed, reported as VSO-1953157 "Modules: MSVC should emit an error when an exported declaration follows a non-exported declaration".

# :adhesive_bandage: The Short-Term Fix
For types, we can export aliases, which also avoids the need to mention `extern "C++"`. This technique looks unusual, so we'll comment it with Standard citations.

The `std::nothrow` object and `::operator new/new[]/delete/delete[]` functions are subject to the same rule, but they're trickier to handle because we can't just alias them. So, I'm fusing `<new>`'s `#ifdef _BUILD_STD_MODULE` machinery directly into `std.ixx` (it expanded to nothing elsewhere).

In `std.ixx`, we have to wait for the exact right moment. We begin with the Global Module Fragment, we define `_BUILD_STD_MODULE` immediately after, and then include CRT headers. (These won't drag in VCRuntime machinery.) Then we say `export module std;` to start the party, and we silence the warning about including headers within a named module.

Now, before we include any STL headers (which will drag in VCRuntime headers), we include `<yvals_core.h>`. This will observe the definition of `_BUILD_STD_MODULE` above (we're assuming classic headers, as commented), and drags in `<vcruntime.h>` (but not any `<vcruntime_MEOW.h>`).

We check whether that emits `_VCRT_EXPORT_STD` (see the long-term fix below). If not, we're working with an older VCRuntime that won't export its own machinery, so we need to introduce exported declarations first.

For laziness, I'm using the full set of push/pop guards, even though `std.ixx` isn't a header and all we really need is the packing. Then I can declare the `nothrow_t` type, the `nothrow` object, and the `align_val_t` type. Finally, I need to replicate the (incredibly verbose) declarations of all of the global allocation/deallocation operators, complete with SAL annotations (we're the first declarations, so we can't omit them). They mention `NULL` which I need to preserve :vomiting_face: because VCRuntime says that and they need to match. I adjusted the declarations to use our `_STD` and west `const`.

Because this is `std.ixx` itself, I'm saying `export` directly, there's no reason to `_EXPORT_STD` here.

# :hammer_and_wrench: The Long-Term Fix
This is an unusual PR because it takes two forms simultaneously. The GitHub portion is the short-term fix, but it'll be simultaneously merged with VCRuntime changes that are the long-term fix. The VCRuntime changes will make `<vcruntime.h>` detect `_BUILD_STD_MODULE` and define `_VCRT_EXPORT_STD`, so it'll export its own machinery. When this happens, the presence of the `_VCRT_EXPORT_STD` macro will automatically disable the STL's workarounds, so it won't try to redundantly `export` things. In `std.ixx`, we'll be left with an unguarded inclusion of `<yvals_core.h>` which is harmless (it's immediately followed by including `<algorithm>`).

After this PR is merged, and ships in (expected) VS 2022 17.10 Preview 3, I'll be able to remove all of the workarounds in the STL. Having VCRuntime export its own machinery is what I should have done in the first place, and this will finally get us there.

### DIFF
--- a/stl/inc/exception
+++ b/stl/inc/exception
@@ -35,8 +35,8 @@ _STD_END
 
 _STD_BEGIN
 
-_EXPORT_STD class exception;
-_EXPORT_STD class bad_exception;
+_EXPORT_STD using exception     = exception;
+_EXPORT_STD using bad_exception = bad_exception;
 
 _EXPORT_STD using ::terminate;
 
--- a/stl/inc/exception
+++ b/stl/inc/exception
@@ -35,9 +35,11 @@ _STD_END
 
 _STD_BEGIN
 
+#ifndef _VCRT_EXPORT_STD // TRANSITION, VCRuntime update expected in 17.10 Preview 3
 // Follow N4971 [module.interface]/6 by exporting aliases (a type alias is not an entity, N4971 [basic.pre]/3):
 _EXPORT_STD using exception     = exception;
 _EXPORT_STD using bad_exception = bad_exception;
+#endif // ^^^ workaround ^^^
 
 _EXPORT_STD using ::terminate;
 
--- a/stl/inc/exception
+++ b/stl/inc/exception
@@ -35,7 +35,7 @@ _STD_END
 
 _STD_BEGIN
 
-// Follow N4971 [module.interface]/6 by exporting aliases:
+// Follow N4971 [module.interface]/6 by exporting aliases (a type alias is not an entity, N4971 [basic.pre]/3):
 _EXPORT_STD using exception     = exception;
 _EXPORT_STD using bad_exception = bad_exception;
 
--- a/stl/inc/exception
+++ b/stl/inc/exception
@@ -35,6 +35,7 @@ _STD_END
 
 _STD_BEGIN
 
+// Follow N4971 [module.interface]/6 by exporting aliases:
 _EXPORT_STD using exception     = exception;
 _EXPORT_STD using bad_exception = bad_exception;
 
--- a/stl/inc/new
+++ b/stl/inc/new
@@ -23,11 +23,6 @@ _STL_DISABLE_CLANG_WARNINGS
 
 // Mirroring <vcruntime_new.h> declarations:
 
-_STD_BEGIN
-_EXPORT_STD using nothrow_t = nothrow_t;
-_EXPORT_STD extern "C++" const nothrow_t nothrow;
-_STD_END
-
 _EXPORT_STD extern "C++" _NODISCARD _VCRT_ALLOCATOR void* __CRTDECL operator new(size_t);
 _EXPORT_STD extern "C++" _NODISCARD _VCRT_ALLOCATOR void* __CRTDECL operator new(
     size_t, const _STD nothrow_t&) noexcept;
--- a/stl/inc/new
+++ b/stl/inc/new
@@ -24,7 +24,7 @@ _STL_DISABLE_CLANG_WARNINGS
 // Mirroring <vcruntime_new.h> declarations:
 
 _STD_BEGIN
-_EXPORT_STD extern "C++" struct nothrow_t;
+_EXPORT_STD using nothrow_t = nothrow_t;
 _EXPORT_STD extern "C++" const nothrow_t nothrow;
 _STD_END
 
@@ -76,8 +76,8 @@ _EXPORT_STD extern "C++" void __CRTDECL operator delete[](void*, void*) noexcept
 
 _STD_BEGIN
 #if _HAS_EXCEPTIONS
-_EXPORT_STD class bad_alloc;
-_EXPORT_STD class bad_array_new_length;
+_EXPORT_STD using bad_alloc            = bad_alloc;
+_EXPORT_STD using bad_array_new_length = bad_array_new_length;
 #else // ^^^ _HAS_EXCEPTIONS / !_HAS_EXCEPTIONS vvv
 // <exception> exports bad_alloc and bad_array_new_length.
 #endif // ^^^ !_HAS_EXCEPTIONS ^^^
--- a/stl/inc/new
+++ b/stl/inc/new
@@ -43,13 +43,7 @@ _EXPORT_STD extern "C++" void __CRTDECL operator delete[](void*, const _STD noth
 
 #ifdef __cpp_aligned_new
 _STD_BEGIN
-#ifdef __EDG__ // TRANSITION, VSO-1618988
-extern "C++" {
-_EXPORT_STD enum class align_val_t : size_t;
-}
-#else // ^^^ workaround / no workaround vvv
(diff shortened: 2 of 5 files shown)


stl/inc/exception
```diff
@@ -35,8 +35,8 @@ _STD_END
 
 _STD_BEGIN
 
-_EXPORT_STD class exception;
-_EXPORT_STD class bad_exception;
+_EXPORT_STD using exception     = exception;
+_EXPORT_STD using bad_exception = bad_exception;
```
CaseyCarter, line 39: I think we need a comment explaining this bit of cleverness to future us. How about "We can't export types that VCRuntime has already declared, but we can export identically-named aliases."?

stl/inc/typeinfo
```diff
@@ -23,7 +23,7 @@ _STL_DISABLE_CLANG_WARNINGS
 #include <vcruntime_typeinfo.h>
 #pragma pop_macro("raw_name")
 
-_EXPORT_STD extern "C++" class type_info; // for typeid, MSVC looks for type_info in the global namespace
+_EXPORT_STD using type_info = type_info; // for typeid, MSVC looks for type_info in the global namespace
```
CaseyCarter, line 26: Ditto "weird enough we should explain" (also on 36-37).

tests/std/include/test_header_units_and_modules.hpp
```diff
@@ -471,6 +475,12 @@ void test_mutex() {
 void test_new() {
     using namespace std;
     puts("Testing <new>.");
+
+    static_assert(is_class_v<bad_alloc>);
+    static_assert(is_class_v<bad_array_new_length>);
+    static_assert(is_same_v<underlying_type_t<align_val_t>, size_t>);
+    static_assert(is_class_v<nothrow_t>);
```
CaseyCarter, line 482: Should we `(void)nothrow;` here (or something) to esnure `nothrow` works as well?

stl/inc/new
```diff
@@ -43,13 +38,7 @@ _EXPORT_STD extern "C++" void __CRTDECL operator delete[](void*, const _STD noth
...
 #ifdef __cpp_aligned_new
 _STD_BEGIN
-#ifdef __EDG__ // TRANSITION, VSO-1618988
-extern "C++" {
-_EXPORT_STD enum class align_val_t : size_t;
-}
-#else // ^^^ workaround / no workaround vvv
-_EXPORT_STD extern "C++" enum class align_val_t : size_t;
-#endif // ^^^ no workaround ^^^
+_EXPORT_STD using align_val_t = align_val_t;
```
CaseyCarter, line 41: Ditto "weird enough we should explain" (also lines 68-69).

CaseyCarter (changes requested):

CaseyCarter:
In the long term, should we have VCRuntime itself export these declarations with something like:
```c++
#ifdef _EXPORT_STD
#define _VCRT_EXPORT_STD _EXPORT_STD
#else
#define _VCRT_EXPORT_STD
#endif
// ...
_VCRT_EXPORT_STD struct nothrow_t {
// ...
#undef _VCRT_EXPORT_STD
```
or do you have no concerns about exporting aliases being potentially brittle? (It feels like a corner case to me, but you're the modules expert.)

StephanTLavavej:
I'm not concerned that exporting aliases will be brittle - this technique is depicted in a Standard example as being OK. However, in the long run, having VCRuntime export its own machinery would indeed be less repetitive.

Now that I look at it yet again, I've solved the problem for types and objects, but there are still the *functions* `operator new` etc. and I think they might be problematic too (life is difficult without compiler enforcement).

I think what I'd like to do is proceed with this PR's changes which are a strict improvement (after I wordsmith comments as you've requested), but also add disabled-by-default VCRuntime changes to export its own machinery. Once that ships in a VS Preview, then we can update the STL to activate it in `std.ixx`, so we don't need this trickery, and `operator new` will be handled too.

tests/std/include/test_header_units_and_modules.hpp
CaseyCarter: I only see red and green =)

stl/modules/std.ixx
```diff
@@ -36,6 +36,84 @@ export module std;
 #pragma warning(push)
 #pragma warning(disable : 5244) // '#include <meow>' in the purview of module 'std' appears erroneous.
 
+#include <yvals_core.h>
+#ifndef _VCRT_EXPORT_STD // TRANSITION, VCRuntime update expected in 17.10 Preview 3
+
+// N4971 [module.interface]/6: "A redeclaration of an entity X is implicitly exported
+// if X was introduced by an exported declaration; otherwise it shall not be exported."
+
+// Therefore, we'll need to introduce exported declarations of <vcruntime_new.h> machinery before including it.
```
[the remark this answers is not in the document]
StephanTLavavej: Sure, I can handle this in a followup. This code is destined to be removed after the 17.10p3 toolset update anyways.

stl/modules/std.ixx
```diff
@@ -36,6 +36,84 @@ export module std;
...
+#pragma push_macro("new")
+#undef new
+
+_STD_BEGIN
+export extern "C++" struct nothrow_t;
+
+export extern "C++" const nothrow_t nothrow;
+
+#ifdef __cpp_aligned_new
+export extern "C++" enum class align_val_t : size_t;
```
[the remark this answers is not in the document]
StephanTLavavej: That's correct.

PR #4382, opened 2024-02-09
Return after SSE2 code

### LINKED ISSUE #4371: `vector_algorithms.cpp`: `return;` missing in `__std_bitset_to_string_1` / `_2`
The fallback is only needed for ARM64EC and `_Use_sse2() == false` modes.
It should not be active code path normally.
https://github.com/microsoft/STL/blob/192a84008a59ac4d2e55681e1ffac73535788674/stl/src/vector_algorithms.cpp#L1918-L1928

AlexGuteniev:
Fixes #4371


StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for adding these important keywords! :joy_cat: :key: :tada:

PR #4387, opened 2024-02-12
Add `find` / `count` benchmark

AlexGuteniev:
Ported from #2434 description, dropping unnecessary parts.

### DIFF
--- a/benchmarks/src/find_and_count.cpp
+++ b/benchmarks/src/find_and_count.cpp
@@ -1,0 +1,73 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <algorithm>
+#include <benchmark/benchmark.h>
+#include <cstddef>
+#include <cstdint>
+#include <ranges>
+#include <type_traits>
+
+enum class Op {
+    FindSized,
+    FindUnsized,
+    Count,
+};
+
+using namespace std;
+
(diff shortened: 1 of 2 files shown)


benchmarks/src/find_and_count.cpp
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <algorithm>
+#include <benchmark/benchmark.h>
+#include <cstddef>
+#include <cstdint>
+#include <ranges>
+#include <type_traits>
```
StephanTLavavej, line 9: `<type_traits>` appears to be unused here.

benchmarks/src/find_and_count.cpp
```diff
@@ -0,0 +1,73 @@
...
+    Count,
+};
+
+using namespace std;
+
+template <class T, size_t Size, size_t Pos, Op Operation>
+void bm(benchmark::State& state) {
+    T a[Size];
+
+    std::fill_n(a, Size, T{'0'});
```
StephanTLavavej, line 23: This file is already `using namespace std;`. Occurs below.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for making it easier to monitor the performance of the repo! :rocket: :stopwatch: :heart_eyes_cat:

PR #4390, opened 2024-02-13
Make `~mutex` trivial

CaseyCarter:
By simply removing `~_Mutex_base`. This destructor has been only a debug check since GH-3770. Losing the debug check is a small price to pay to elide the destructor call, doubly so for static storage duration mutexes that now won't need dynamic initializers to register with `at_exit`.

`_Mtx_destroy` and `_Mtx_destroy_in_situ` are only called by APIs preserved for binary compatibility, so mark them as also "preserved for binary compatibility".

### DIFF
--- a/stl/inc/mutex
+++ b/stl/inc/mutex
@@ -45,10 +45,6 @@ public:
     }
 #endif // ^^^ !defined(_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR) ^^^
 
-    ~_Mutex_base() noexcept {
-        _Mtx_destroy_in_situ(_Mymtx());
-    }
-
     _Mutex_base(const _Mutex_base&)            = delete;
     _Mutex_base& operator=(const _Mutex_base&) = delete;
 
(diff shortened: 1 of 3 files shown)


stl/inc/mutex
```diff
@@ -45,10 +45,6 @@ public:
     }
 #endif // ^^^ !defined(_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR) ^^^
 
-    ~_Mutex_base() noexcept {
-        _Mtx_destroy_in_situ(_Mymtx());
```
CaseyCarter, line 49: I could be convinced to retain this debug check for `~recursive_mutex`, which feels less performance-sensitive to me.

AlexGuteniev (approved):

stl/inc/mutex
StephanTLavavej: I'm fine with dropping it.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for this trivial PR with nontrivial impact! :joy_cat: :rocket: :tada:

PR #4406, opened 2024-02-18
Equality for unordered_set and unordered_map should be based on equality of elements

### LINKED ISSUE #4388: `<unordered_set>`, `<unordered_map>`: `operator==` is incorrect with custom equivalence functor
# Describe the bug
According to the [C++ standard](https://timsong-cpp.github.io/cppwp/n4861/unord.req#12), two unordered_sets should only compare equal if they have elements that compare equal per value_type’s operator==, regardless of the key_equal equivalence functor. When using unordered_set with a custom key_equal equivalence functor (and compatible hasher functor), operator== erroneously returns true when the sets contain values that are equivalent but not equal.

The same issue occurs for unordered_map, but not for unordered_multimap, unordered_multiset or any of the associative containers.

See also https://godbolt.org/z/99Mnf9hqv 

# Command-line test case
```
C:\temp>type unordered.cpp
#include <algorithm>
#include <cassert>
#include <map>
#include <set>
#include <unordered_map>
#include <unordered_set>


namespace modulo
{
// Functors for comparison and hashing
// based on modular arithmetic equivalence classes
template<int N>
struct compare
{
    bool operator()(int lhs, int rhs) const
    {
        return (lhs % N) < (rhs % N);
    }
};

template<int N>
struct equals
{
    bool operator()(int lhs, int rhs) const
    {
        return (lhs % N) == (rhs % N);
    }
};

template<int N>
struct hash
{
    std::size_t operator()(int value) const
    {
        return value % N;
    }
};
}

// Overloaded function to get a key from a set / map's value_type
int const& get_key(int const& value)
{
    return value;
}
int const& get_key(std::pair<int const, int> const& pair)
{
    return pair.first;
}

// Equality comparison for unordered sets and maps as per the C++ standard
// It also gives the correct result for associative containers (though is sub-optimal)
template<typename Container>
bool std_equal(Container const& lhs, Container const& rhs)
{
    if (lhs.size() != rhs.size())
        return false;

    for(auto it = lhs.cbegin(); it != lhs.cend(); ) {
        auto const& key = get_key(*it);
        auto l_range = lhs.equal_range(key);
        auto r_range = rhs.equal_range(key);

        if (!std::is_permutation(l_range.first, l_range.second, r_range.first, r_range.second))
            return false;
        it = l_range.second;
    }
    return true;
}

template<typename Map>
bool test_maps()
{
    // In all cases, these sets should compare not equal, even though their elements are equivalent
    Map map1 = {{1, 1}, {2, 2}};
    Map map2 = {{21, 1}, {12, 2}};

    return (map1 == map2) == std_equal(map1, map2);
}

template<typename Set>
bool test_sets()
{
    // In all cases, these sets should compare not equal, even though their elements are equivalent
    Set set1 = {1, 2};
    Set set2 = {21, 12};

    return (set1 == set2) == std_equal(set1, set2);
}

int main()
{
    using Set = std::set<int, modulo::compare<10>>;
    using MultiSet = std::multiset<int, modulo::compare<10>>;
    using UnorderedSet = std::unordered_set<int, modulo::hash<10>, modulo::equals<10>>;
    using UnorderedMultiSet = std::unordered_multiset<int, modulo::hash<10>, modulo::equals<10>>;

    assert(test_sets<Set>()); // ok
    assert(test_sets<MultiSet>()); // ok
    assert(test_sets<UnorderedSet>()); // FAILURE!
    assert(test_sets<UnorderedMultiSet>()); // ok

    using Map = std::map<int, int, modulo::compare<10>>;
    using MultiMap = std::multimap<int, int, modulo::compare<10>>;
    using UnorderedMap = std::unordered_map<int, int, modulo::hash<10>, modulo::equals<10>>;
    using UnorderedMultiMap = std::unordered_multimap<int, int, modulo::hash<10>, modulo::equals<10>>;

    assert(test_maps<Map>()); // ok
    assert(test_maps<MultiMap>()); // ok
    assert(test_maps<UnorderedMap>()); // FAILURE!
    assert(test_maps<UnorderedMultiMap>()); // ok

    return 0;
}

C:\temp>cl /EHsc /W4 /WX /std:c++latest .\unordered.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.38.33134 for x64

unordered.cpp
Microsoft (R) Incremental Linker Version 14.38.33134.0

/out:unordered.exe
unordered.obj

C:\temp>.\unordered.exe
Assertion failed: test_sets<UnorderedSet>(), file .\unordered.cpp, line 105
```

# Expected behavior
Equality for unordered_set and unordered_map should be based on equality of elements, not just equivalence, just as it is for the other containers.

# STL version
* Option 1: Visual Studio version
Microsoft Visual Studio Community 2022 (64-bit) - Current
Version 17.8.4

# Additional context
I have previously raised this issue at DevCom-1392782 (internal VSO-1307889 / AB#1307889 ), but didn't seem to get much traction there.  See also https://godbolt.org/z/99Mnf9hqv

fsb4000:
Fixes #4388


tests/std/test.lst
```diff
@@ -239,6 +239,7 @@ tests\GH_004040_container_nonmember_functions
 tests\GH_004109_iter_value_t_direct_initialization
 tests\GH_004201_chrono_formatter
 tests\GH_004275_seeking_fancy_iterators
+tests\GH_0004388_unordered_meow_operator_equal
```
MattStephanson, line 242: Sorry if I missed it, but did we start using 7-digit numbers for GH issues? If not, this has an extra leading "0" that makes it not lexical-sorted.

stl/inc/xhash
```diff
@@ -2015,7 +2015,7 @@ _NODISCARD bool _Hash_equal(const _Hash<_Traits>& _Left, const _Hash<_Traits>& _
             // look for element with equivalent key
             const auto& _Keyval = _Traits::_Kfn(_LVal);
             const auto _Next2   = _Right._Find_last(_Keyval, _Right._Traitsobj(_Keyval))._Duplicate;
-            if (!(static_cast<bool>(_Next2) && _Traits::_Nonkfn(_LVal) == _Traits::_Nonkfn(_Next2->_Myval))) {
+            if (!(static_cast<bool>(_Next2) && _LVal == _Next2->_Myval)) {
```
MattStephanson, line 2018: A [GH search](https://github.com/search?q=repo%3Amicrosoft%2FSTL%20_Nonkfn&type=code) suggests that this is the only place that `_Nonkfn` was used. Maybe we should remove it entirely?

StephanTLavavej: This can definitely be dropped, no ABI impact. Good catch!

tests/std/tests/GH_004388_unordered_meow_operator_equal/test.cpp
```diff
@@ -0,0 +1,110 @@
...
+            return static_cast<size_t>(value % N);
+        }
+    };
+} // namespace modulo
+
+// Overloaded function to get a key from a set / map's value_type
+int const& get_key(int const& value) {
+    return value;
+}
+int const& get_key(pair<int const, int> const& pair) {
```
StephanTLavavej, line 42: This parameter name shadows the type name.

tests/std/tests/GH_004388_unordered_meow_operator_equal/test.cpp
```diff
@@ -0,0 +1,110 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <algorithm>
+#include <cassert>
+#include <map>
+#include <set>
+#include <unordered_map>
+#include <unordered_set>
```
StephanTLavavej, line 9: Should include `<cstddef>` for `size_t` and `<utility>` for `pair`.

tests/std/tests/GH_004388_unordered_meow_operator_equal/test.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+    template <int N>
+    struct equals {
+        bool operator()(int lhs, int rhs) const {
+            return (lhs % N) == (rhs % N);
+        }
+    };
+
+    template <int N>
+    struct hash {
```
StephanTLavavej, line 31: This is really, really nitpicking, but even though this is not technically shadowing due to the `modulo` namespace, having the lowercase `hash` floating around makes me a bit nervous. I think it's best for test code to be visually distinct from product code, and I don't see the namespace as serving an important purpose here. These types would be clearer as `ModLess`, `ModEqual`, and `ModHash`.

tests/std/tests/GH_004388_unordered_meow_operator_equal/test.cpp
```diff
@@ -0,0 +1,110 @@
...
+    return true;
+}
+
+template <typename Map>
+bool test_maps() {
+    // In all cases, these sets should compare not equal, even though their elements are equivalent
+    Map map1 = {{1, 1}, {2, 2}};
+    Map map2 = {{21, 1}, {12, 2}};
+
+    return (map1 == map2) == std_equal(map1, map2);
```
StephanTLavavej, line 73: Substantial issue: When writing tests, it's always important to cover both positive and negative cases. This guards against problems, in both the product code and the test code, where a mistake causes the answer to be "always true" or "always false" regardless of the input. Here, we should have test coverage where the containers are actually `==`, to verify that both `operator==` and the test's `std_equal` are getting the right answer.

tests/std/tests/GH_004388_unordered_meow_operator_equal/test.cpp
```diff
@@ -0,0 +1,110 @@
...
+        }
+        it = l_range.second;
+    }
+    return true;
+}
+
+template <typename Map>
+bool test_maps() {
+    // In all cases, these sets should compare not equal, even though their elements are equivalent
+    Map map1 = {{1, 1}, {2, 2}};
```
StephanTLavavej, line 70: Nitpick: This copy-pasted comment says "sets" when these are maps.

tests/std/tests/GH_004388_unordered_meow_operator_equal/test.cpp
```diff
@@ -0,0 +1,110 @@
...
+template <typename Container>
+bool std_equal(Container const& lhs, Container const& rhs) {
+    if (lhs.size() != rhs.size()) {
+        return false;
+    }
+
+    for (auto it = lhs.cbegin(); it != lhs.cend();) {
+        auto const& key = get_key(*it);
+        auto l_range    = lhs.equal_range(key);
+        auto r_range    = rhs.equal_range(key);
```
StephanTLavavej, line 57: Suggestion: These pairs could be `const`.

StephanTLavavej (approved):

StephanTLavavej:
Thanks for fixing this long-standing bug (and removing my confused `_Nonkfn` machinery)! I pushed some changes to the test.

StephanTLavavej:

StephanTLavavej:
Thanks for fixing this long-standing bug! :lady_beetle: :hammer_and_wrench: :tada:

PR #4407, opened 2024-02-18
`<new>`: `SRWLOCK` to protect the `new` handler instead of an indexed `CRITICAL_SECTION`

AlexGuteniev:
Towards #3521

After finally making both constructor of `std::mutex` `constexpr` and its destructor trivial, it is should be fine to use usual mutexes everywhere!

### DIFF
--- a/stl/src/stdhndlr.cpp
+++ b/stl/src/stdhndlr.cpp
@@ -20,7 +20,7 @@ namespace {
 _STD_BEGIN
 
 _CRTIMP2 new_handler __cdecl set_new_handler(_In_opt_ new_handler pnew) noexcept { // remove current handler
-    unique_lock _Lock{_New_handler_mutex};
+    lock_guard _Lock{_New_handler_mutex};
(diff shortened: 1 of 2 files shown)


stl/src/stdhndlr.cpp
```diff
@@ -3,11 +3,13 @@
 
 // set_new_handler
 
+#include <mutex>
 #include <new.h>
 #include <new>
 
 namespace {
     _STD new_handler _New_handler;
+    constinit std::mutex _New_handler_mutex;
```
StephanTLavavej, line 12: Style nitpick: `stl/src` is somewhat inconsistent, but `_STD` appears immediately above, so let's use that here.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for noticing and improving the performance here! :eagle: :eye: :rocket:

PR #4422, opened 2024-02-24
AVX2 vectorization for very large bitsets

AlexGuteniev:
Not sure if it worth merging due to complexity growth, and noticeable improvement only for larger bitsets

Turned out that the existing vectorization is fine for bitsets beyond 256 bits (32 bytes), and AVX2 upgrade is harmful.
For very large bitsets the improvement is noticeable.

## Results
```
Before:
BM_bitset_to_string<15, char>                    211 ns          206 ns      2800000
BM_bitset_to_string<64, char>                   2662 ns         2668 ns       263529
BM_bitset_to_string<512, char>                  3730 ns         3767 ns       186667
BM_bitset_to_string_large_single<char>           230 ns          230 ns      2986667
BM_bitset_to_string<7, wchar_t>                  131 ns          131 ns      5600000
BM_bitset_to_string<64, wchar_t>                2488 ns         2511 ns       298667
BM_bitset_to_string<512, wchar_t>               4360 ns         4349 ns       154483
BM_bitset_to_string_large_single<wchar_t>        306 ns          300 ns      2240000

After:
BM_bitset_to_string<15, char>                    224 ns          223 ns      2800000
BM_bitset_to_string<64, char>                   2696 ns         2668 ns       263529
BM_bitset_to_string<512, char>                  3313 ns         3278 ns       224000
BM_bitset_to_string_large_single<char>           157 ns          157 ns      4072727
BM_bitset_to_string<7, wchar_t>                  128 ns          126 ns      4977778
BM_bitset_to_string<64, wchar_t>                2421 ns         2400 ns       280000
BM_bitset_to_string<512, wchar_t>               3245 ns         3223 ns       203636
BM_bitset_to_string_large_single<wchar_t>        209 ns          209 ns      3446154
```

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -2237,8 +2237,8 @@ __declspec(noalias) void __stdcall __std_bitset_to_string_1(
 
         if (_Size_bits > 0) {
             __assume(_Size_bits < 32);
-            uint32_t _Val;
-            memcpy(&_Val, _Src, _Size_bits / 4);
+            uint32_t _Val = 0;
+            memcpy(&_Val, _Src, (_Size_bits + 7) / 8);
             const __m256i _Elems = _Bitset_to_string_1_step_avx(_Val, _Px0, _Px1);
             char _Tmp[32];
             _mm256_storeu_si256(reinterpret_cast<__m256i*>(_Tmp), _Elems);
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -2166,6 +2166,17 @@ __declspec(noalias) size_t
 
 #ifndef _M_ARM64EC
 namespace {
+    __m256i __forceinline _Bitset_to_string_1_step_avx(const uint32_t _Val, const __m256i _Px0, const __m256i _Px1) {
+        const __m128i _Vx0 = _mm_cvtsi32_si128(_Val);
+        const __m128i _Vx1 = _mm_shuffle_epi8(_Vx0, _mm_set_epi32(0x00000000, 0x01010101, 0x02020202, 0x03030303));
+        const __m256i _Vx2 = _mm256_castsi128_si256(_Vx1);
+        const __m256i _Vx3 = _mm256_permutevar8x32_epi32(_Vx2, _mm256_set_epi32(3, 3, 2, 2, 1, 1, 0, 0));
+        const __m256i _Msk = _mm256_and_si256(_Vx3, _mm256_set1_epi64x(0x0102040810204080));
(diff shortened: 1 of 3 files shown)


stl/src/vector_algorithms.cpp
```diff
@@ -2192,6 +2215,38 @@ extern "C" {
 __declspec(noalias) void __stdcall __std_bitset_to_string_1(
     char* const _Dest, const void* _Src, size_t _Size_bits, const char _Elem0, const char _Elem1) noexcept {
 #ifndef _M_ARM64EC
+    if (_Use_avx2() && _Size_bits >= 256) {
```
StephanTLavavej, line 2218: `VSO_0000000_vector_algorithms` should be extended with correctness coverage - right now we max out at 75 bits. We already have the logic to test AVX2 enabled and disabled.

StephanTLavavej (changes requested):

stl/src/vector_algorithms.cpp
AlexGuteniev: There's randomized test up to 2048 bits below.
I'm not sure if it is needed to test with literals as well.

AlexGuteniev:
Created DevCom-10601346 based on suboptimal AVX2 codegen. 
This change might give better results if the issue is fixed.

StephanTLavavej (approved):

StephanTLavavej:

stl/src/vector_algorithms.cpp
```diff
@@ -2192,6 +2215,38 @@ extern "C" {
...
+
+        if (_Size_bits > 0) {
+            __assume(_Size_bits < 32);
+            uint32_t _Val;
+            memcpy(&_Val, _Src, _Size_bits / 4);
+            const __m256i _Elems = _Bitset_to_string_1_step_avx(_Val, _Px0, _Px1);
+            char _Tmp[32];
+            _mm256_storeu_si256(reinterpret_cast<__m256i*>(_Tmp), _Elems);
+            const char* const _Tmpd = _Tmp + (32 - _Size_bits);
+            memcpy(_Dest, _Tmpd, _Size_bits);
```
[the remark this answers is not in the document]
AlexGuteniev: Correction: this applies to the above `memcpy` only, To write more here we should over-reserve string, which though seems also feasible.

StephanTLavavej:
Thanks for keeping those vector units busy! :chart_with_upwards_trend: :tada: :joy_cat:

AlexGuteniev:
> Created DevCom-10601346 based on suboptimal AVX2 codegen.
> This change might give better results if the issue is fixed.

No, DevCom-10601346 fix didn't make the noticeable improvement overall.

PR #4426, opened 2024-02-26
Silence warning C4324: structure was padded due to alignment specifier

### LINKED ISSUE #4425: `<ranges>`: `views::cartesian_product` and `views::filter` emit spurious warning C4324 'structure was padded due to alignment specifier'
```
```
```cpp
#include <print>
#include <ranges>
#include <vector>
using namespace std;

int main() {
    const vector<int> vec{11, 22, 33};
    auto always_true = [](const auto&) { return true; };

    for (const auto& [e0, e1] : views::cartesian_product(vec, vec) | views::filter(always_true)) {
        println("{} {}", e0, e1);
    }
}
```
```
D:\GitHub\STL\out\x64>cl /EHsc /nologo /W4 /std:c++latest /MTd /Od meow.cpp
meow.cpp
D:\GitHub\STL\out\x64\out\inc\ranges(2403): warning C4324: 'std::ranges::filter_view<std::ranges::cartesian_product_view<std::ranges::ref_view<const std::vector<int,std::allocator<int>>>,std::ranges::ref_view<const std::vector<int,std::allocator<int>>>>,main::<lambda_1>>': structure was padded due to alignment specifier
D:\GitHub\STL\out\x64\out\inc\ranges(2403): note: the template instantiation context (the oldest one first) is
D:\GitHub\STL\out\x64\out\inc\ranges(198): note: while compiling class template member function 'decltype(auto) std::ranges::_Range_closure<std::ranges::views::_Filter_fn,main::<lambda_1>>::operator ()(_Ty &&) noexcept(<expr>) &'
D:\GitHub\STL\out\x64\out\inc\ranges(930): note: while evaluating concept 'invocable<std::ranges::views::_Filter_fn,std::ranges::cartesian_product_view<std::ranges::ref_view<std::vector<int,std::allocator<int> > const >,std::ranges::ref_view<std::vector<int,std::allocator<int> > const > >,`main'::`2'::<lambda_1> &>'
D:\GitHub\STL\out\x64\out\inc\ranges(2418): note: see reference to class template instantiation 'std::ranges::filter_view<std::ranges::cartesian_product_view<std::ranges::ref_view<const std::vector<int,std::allocator<int>>>,std::ranges::ref_view<const std::vector<int,std::allocator<int>>>>,main::<lambda_1>>' being compiled

D:\GitHub\STL\out\x64>meow
11 11
11 22
11 33
22 11
22 22
22 33
33 11
33 22
33 33
```

StephanTLavavej:
Fixes #4425.

This globally silences the warning because I don't think it's ever useful for the STL to emit it. If the user gives us a highly-aligned type, and we have to build another type with it, warning about padding is unlikely to be productive - presumably they had reasons for making their type highly-aligned, and it's not like they can change the layout of our type. (Yeah, in theory maybe they could request a different order of `tuple` elements or something, but most of the time it's not going to be actionable.) And if the warning is emitted purely from STL code, as in my test case, it's extremely annoying. Finally, this is just a performance warning - it has no impact on correctness.

### DIFF
--- a/stl/inc/execution
+++ b/stl/inc/execution
@@ -542,8 +542,6 @@ struct alignas(_Ty) alignas(size_t) alignas(_Atomic_counter_t) _Circular_buffer 
     }
 };
 
-#pragma warning(push)
-#pragma warning(disable : 4324) // structure was padded due to alignment specifier
 template <class _Ty>
 class alignas(hardware_destructive_interference_size) _Work_stealing_deque {
     // thread-local work-stealing deque, which allows efficient access from a single owner thread at the "bottom"
@@ -657,7 +655,6 @@ private:
     _Guarded_by_(_Segment_lock) _Circular_buffer<_Ty>* _Segment{_Circular_buffer<_Ty>::_New_circular_buffer()};
     mutex _Segment_lock{};
 };
-#pragma warning(pop)
 
 enum class _Steal_result { _Success, _Abort, _Done };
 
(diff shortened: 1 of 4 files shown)


AlexGuteniev (changes requested):
There are four occurrences of 
```c++
#pragma warning(disable : 4324)
```
just in headers, plus two in `atomic_wait.cpp`.

Should local suppression be removed?

StephanTLavavej:
Great catch, I forgot to scan for existing suppressions! I'll remove the header occurrences - the source file occurrences aren't subject to our header suppressions though.

AlexGuteniev (approved):

CaseyCarter (approved):

StephanTLavavej:

PR #4431, opened 2024-03-02
Auto vectorize `replace_copy`, `replace_copy_if`

AlexGuteniev:
Work around DevCom-10606350

Benchmark result before:
```
----------------------------------------------------------
Benchmark                Time             CPU   Iterations
----------------------------------------------------------
rc<uint8_t>           1458 ns         1465 ns       448000
rc<uint16_t>          1647 ns         1650 ns       407273
rc<uint32_t>          1680 ns         1650 ns       407273
rc<uint64_t>          1666 ns         1674 ns       448000
rc_if<uint8_t>        1644 ns         1650 ns       407273
rc_if<uint16_t>       1686 ns         1688 ns       407273
rc_if<uint32_t>       1648 ns         1674 ns       448000
rc_if<uint64_t>       1750 ns         1726 ns       407273
```
After:
```
----------------------------------------------------------
Benchmark                Time             CPU   Iterations
----------------------------------------------------------
rc<uint8_t>            159 ns          160 ns      4977778
rc<uint16_t>           272 ns          270 ns      2488889
rc<uint32_t>           505 ns          502 ns      1120000
rc<uint64_t>          1008 ns         1004 ns       746667
rc_if<uint8_t>         166 ns          165 ns      4072727
rc_if<uint16_t>        348 ns          339 ns      2488889
rc_if<uint32_t>        566 ns          572 ns      1120000
rc_if<uint64_t>       1474 ns         1465 ns       448000
```
After with `/arch:AVX2`:
```
----------------------------------------------------------
Benchmark                Time             CPU   Iterations
----------------------------------------------------------
rc<uint8_t>           58.5 ns         59.4 ns     10000000
rc<uint16_t>           102 ns          103 ns      6400000
rc<uint32_t>           187 ns          188 ns      3733333
rc<uint64_t>           559 ns          558 ns      1120000
rc_if<uint8_t>        76.8 ns         76.7 ns     11200000
rc_if<uint16_t>        131 ns          132 ns      4977778
rc_if<uint32_t>        230 ns          230 ns      2986667
rc_if<uint64_t>       1691 ns         1688 ns       407273
```

The anomality for `rc_if<uint64_t>` AVX2 case it that compiler generates AVX512 code and checks for AVX512 in `__isa_available` and I don't have that, so it is scalar, still a bit faster due to branchless. Applies specifically to unsigned 64 bit only.


stl/inc/algorithm
```diff
@@ -3851,10 +3851,17 @@ _CONSTEXPR20 _OutIt replace_copy(_InIt _First, _InIt _Last, _OutIt _Dest, const
     const auto _ULast = _STD _Get_unwrapped(_Last);
     auto _UDest       = _STD _Get_unwrapped_n(_Dest, _STD _Idl_distance<_InIt>(_UFirst, _ULast));
     for (; _UFirst != _ULast; ++_UFirst, (void) ++_UDest) {
-        if (*_UFirst == _Oldval) {
-            *_UDest = _Newval;
+        if constexpr ((is_integral_v<_Ty> && is_integral_v<_Iter_value_t<_InIt>>)
+                      || (is_pointer_v<_Ty> && is_pointer_v<_Iter_value_t<_InIt>>) ) {
+            // TRANSITION, DevCom-10606350: help the compiler auto-vectorizing for simple types
+            // by making transformation that can be done for simple types
+            *_UDest = (*_UFirst == _Oldval) ? _Newval : *_UFirst;
```
StephanTLavavej, line 3858: I don't believe that this is an allowable transformation. WG21-N4971 \[alg.replace\]/7 "*Mandates:* The results of the expressions `*first` and `new_value` are writable (\[iterator.requirements.general\]) to `result`." If `_Ty` is `unsigned char` and `*_UFirst` is `unsigned short`, then the conditional operator will perform the usual arithmetic conversions and we'll try to write `int` to the output iterator, which might not compile.

For pointers, we'll get their composite pointer type, which can similarly synthesize a third type.

Similarly for the ranges algorithm below, we have `output_iterator<_Out, const _Ty2&>` and `indirectly_copyable<_It, _Out>` but not whatever the usual arithmetic conversions might try to do.

We could add even more metaprogramming to detect output iterators which won't care about the difference, but I think that's too much library complexity to attempt to compensate for an optimizer deficiency. Thanks for investigating this (and reporting the optimizer issue!), but I don't think we should make any changes here.

AlexGuteniev: We already do some complex metaprogramming elsewhere.
Consider `_Equal_memcmp_is_safe`, `_Lex_compare_three_way_memcmp_classify` and `_Vector_alg_in_find_is_safe`.
In those cases we direct optimized path to `memcmp` function or our own vector algorithm, so ultimately doing more complex thing. I think directing to auto-vectorization is more favorable.

StephanTLavavej: I'm not super eager to take on additional maintenance burden and regression risk to compensate for C2's current deficiency (as I observe that LLVM happily autovectorizes this), but I think I'd be okay with allowing a limited set of easy-to-reason-about scenarios through. I think we need (in addition to the current criteria):

* The sources are the same type (so the conditional operator won't synthesize a third type)
* The unwrapped output iterator is a raw pointer to an integer/pointer (practically always necessary for autovectorization anyways, and this ensures that the output iterator won't be playing squirrelly games with inspecting constness/value category)

Given those restrictions, I don't think we need to require the output type to be the same as the source types.

AlexGuteniev: One more thing to consider here is #4433 . If it is resolved to be desirable optimization, we would basically share the metaprogramming.

Otherwise I'd want the compiler to fix deficiency, but whereas such issues ocassionally fixed, and sometimes even quickly, most of such issues created by me are either still open, or closed without fix.

StephanTLavavej: We talked about this at the weekly maintainer meeting and @barcharcraz says: "This is gross, but grr, fine" :joy_cat: :zany_face:

AlexGuteniev:
> I don't think we need to require the output type to be the same as the source types.

Curiously, the AVX2 auto vectorization does indeed work even for different source and destination element size, this is achieved using `vpmovsx*` or `vpmovzx*` to widen and `vpshufb` to compress.

Surprising the compiles goes that much far, and misses the final piece to vectorize without this PR

benchmarks/src/replace.cpp
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <algorithm>
+#include <benchmark/benchmark.h>
+
+const char src[] =
+    "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam mollis imperdiet massa, at dapibus elit interdum "
```
StephanTLavavej, line 8: What's the exact source of this lorem ipsum text?

benchmarks/src/replace.cpp
```diff
@@ -0,0 +1,73 @@
...
+    "vitae auctor a, cursus et sem. Cras vehicula ante in consequat fermentum. Praesent at massa nisi. Mauris pretium "
+    "euismod eros, ut posuere ligula ullamcorper id. Nullam aliquam malesuada est at dignissim. Pellentesque finibus "
+    "sagittis libero nec bibendum. Phasellus dolor ipsum, finibus quis turpis quis, mollis interdum felis.";
+
+template <class T>
+void rc(benchmark::State& state) {
+    std::vector<T> a(std::size(src));
+    std::vector<T> b(std::size(src));
+
+    std::copy(std::begin(src), std::end(src), std::begin(a));
```
StephanTLavavej, line 43: Suggestion: It would be simpler to range-construct `a` and then make it `const`. Occurs below.

benchmarks/src/replace.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+template <class T>
+void rc_if(benchmark::State& state) {
+    std::vector<T> a(std::size(src));
+    std::vector<T> b(std::size(src));
+
+    std::copy(std::begin(src), std::end(src), std::begin(a));
+
+    for (auto _ : state) {
+        std::replace_copy_if(
```
StephanTLavavej, line 58: While it's not strictly necessary, I'd recommend void-casting `std::replace_copy_if()` here, as a reminder that we don't care if we're partially overwriting `b`. (We haven't marked the algorithm as `[[nodiscard]]` because calling it with a `back_insert_iterator` etc. makes the return value irrelevant, but in general if you're writing to a plain iterator and you don't know how many values are going to satisfy the predicate, you really should be interested in the return value.)

benchmarks/src/replace.cpp
```diff
@@ -0,0 +1,73 @@
...
+            std::begin(a), std::end(a), std::begin(b), [](auto x) { return x <= 'Z'; }, T{'X'});
```
StephanTLavavej, line 59: It's a literal, but `T{'Z'}` would make the comparison homogeneous and consistent with the other literals.

stl/inc/algorithm
```diff
@@ -3843,6 +3843,16 @@ namespace ranges {
 } // namespace ranges
 #endif // _HAS_CXX20
 
+// TRANSITION, DevCom-10606350: help the compiler auto-vectorizing for simple types
+// by making transformation that can be done for simple types
+template <class _UOutIt, class _InTy, class _NewTy, class _OutTy = remove_pointer_t<_UOutIt>>
+constexpr bool _Can_vectorize_replace_copy = conjunction_v<is_pointer<_UOutIt>, is_same<_InTy, _NewTy>,
```
StephanTLavavej, line 3849: We should consistently use `_INLINE_VAR`.

stl/inc/algorithm
```diff
@@ -3843,6 +3843,16 @@ namespace ranges {
 } // namespace ranges
 #endif // _HAS_CXX20
 
+// TRANSITION, DevCom-10606350: help the compiler auto-vectorizing for simple types
+// by making transformation that can be done for simple types
```
StephanTLavavej, line 3847: This is somewhat repetitive, I think we can shorten this comment.

stl/inc/algorithm
```diff
@@ -3851,10 +3861,14 @@ _CONSTEXPR20 _OutIt replace_copy(_InIt _First, _InIt _Last, _OutIt _Dest, const
     const auto _ULast = _STD _Get_unwrapped(_Last);
     auto _UDest       = _STD _Get_unwrapped_n(_Dest, _STD _Idl_distance<_InIt>(_UFirst, _ULast));
     for (; _UFirst != _ULast; ++_UFirst, (void) ++_UDest) {
-        if (*_UFirst == _Oldval) {
-            *_UDest = _Newval;
+        if constexpr (_Can_vectorize_replace_copy<decltype(_UDest), _Iter_value_t<_InIt>, _Ty>) {
+            *_UDest = (*_UFirst == _Oldval) ? _Newval : *_UFirst;
```
StephanTLavavej, line 3865: The parentheses here are unnecessary; I don't *strongly* object to them, but they don't appear in the other expressions, and this one isn't very complicated.

StephanTLavavej (approved):

benchmarks/src/replace.cpp
AlexGuteniev: Generated via this one https://www.lipsum.com/

AlexGuteniev:
Noticed a bug here, will push a change after CI passes.
I didn't realize initially that projection can change type, so may break the ternary.
We should check it for being `identity`.

StephanTLavavej:
Good idea to think about the projection - I hadn't explicitly done so. However, does it matter? The condition is being moved monolithically from the `if`-statement to the ternary operator, so my understanding is that the projection (and predicate) can be arbitrarily squirrelly without damaging the semantics of this transformation.

benchmarks/src/replace.cpp
StephanTLavavej: Seems low risk to me.

AlexGuteniev:
> However, does it matter? The condition is being moved monolithically from the `if`-statement to the ternary operator, so my understanding is that the projection (and predicate) can be arbitrarily squirrelly without damaging the semantics of this transformation

Yes, you are right. It doesn't matter what type is in the conditional of ternary, only what types are in resulting expressions.

Initially I event tested that simple projections still vectorize along with the whole algorithm.

StephanTLavavej:

StephanTLavavej:
Thanks for helping the compiler, said the developer who once gave a talk titled Don't Help The Compiler! :joy_cat: :zany_face: :rocket:

AlexGuteniev:
This no longer works in latest Preview. 
Not sure when exactly it was broken, but works in latest Production.
This is compiler change, not STL change.

DevCom-10895463

PR #4435, opened 2024-03-03
Tools: add a script to download Unicode data files, and rename `grapheme_break_property_data_gen.py` to `unicode_properties_data_gen.py`

cpplearner:
These are attempts to make it easier to regenerate source files from the Unicode data.

SG16 recommends specifying Unicode 15.1.0 as the minimum Unicode version for C++23 (as a DR) and C++26. See https://github.com/cplusplus/papers/issues/1736 and https://github.com/sg16-unicode/sg16-meetings.

Currently, `__msvc_format_ucd_tables.hpp` is based on Unicode 15.0.0. In order to support 15.1.0, this file will need to be regenerated.

(Other changes are also needed. For example, `_Grapheme_break_property_iterator` might need to account for the changes in [UAX 29](https://www.unicode.org/reports/tr29/tr29-43.html). I haven't taken a close look, though.)


tools/unicode_properties_parse/download_unicode_data_files.py
```diff
@@ -0,0 +1,20 @@
+from urllib.request import urlretrieve
```
StephanTLavavej, line 1: This new file needs our copyright/license banner.

tools/unicode_properties_parse/download_unicode_data_files.py
```diff
@@ -0,0 +1,20 @@
+from urllib.request import urlretrieve
+
+
+Unicode_data_files = {
+    "DerivedCoreProperties.txt": "https://www.unicode.org/Public/UCD/latest/ucd/DerivedCoreProperties.txt",
+    "DerivedGeneralCategory.txt": "https://www.unicode.org/Public/UCD/latest/ucd/extracted/DerivedGeneralCategory.txt",
+    "EastAsianWidth.txt": "https://www.unicode.org/Public/UCD/latest/ucd/EastAsianWidth.txt",
+    "GraphemeBreakProperty.txt": "https://www.unicode.org/Public/UCD/latest/ucd/auxiliary/GraphemeBreakProperty.txt",
+    "GraphemeBreakText.txt": "https://www.unicode.org/Public/UCD/latest/ucd/auxiliary/GraphemeBreakTest.txt",
```
StephanTLavavej, line 9: Heh, there's a pre-existing typo in our `.gitignore` which we should fix now.

StephanTLavavej (approved):

tools/unicode_properties_parse/download_unicode_data_files.py
cpplearner: `.gitignore` is correct. The typo is in `grapheme_break_test_data_gen.py`, which I erroneously copied over.

https://github.com/microsoft/STL/blob/8b081e26ba016970ce3338cb483ff10f2ade30a5/tools/unicode_properties_parse/grapheme_break_test_data_gen.py#L94

StephanTLavavej: Ha, thanks for noticing the mistake in my mistake correction! :repeat: :joy_cat:

StephanTLavavej:

StephanTLavavej:
Thanks for simplifying this process - I've wanted a download script for a while! :magic_wand: :heart_eyes_cat: :tada:

tools/unicode_properties_parse/download_unicode_data_files.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+Unicode_data_files = {
+    "DerivedCoreProperties.txt": "https://www.unicode.org/Public/UCD/latest/ucd/DerivedCoreProperties.txt",
+    "DerivedGeneralCategory.txt": "https://www.unicode.org/Public/UCD/latest/ucd/extracted/DerivedGeneralCategory.txt",
+    "EastAsianWidth.txt": "https://www.unicode.org/Public/UCD/latest/ucd/EastAsianWidth.txt",
+    "GraphemeBreakProperty.txt": "https://www.unicode.org/Public/UCD/latest/ucd/auxiliary/GraphemeBreakProperty.txt",
+    "GraphemeBreakTest.txt": "https://www.unicode.org/Public/UCD/latest/ucd/auxiliary/GraphemeBreakTest.txt",
+    "emoji-data.txt": "https://www.unicode.org/Public/UCD/latest/ucd/emoji/emoji-data.txt",
+}
```
tahonermann, line 14: Is it desirable that these pull data from the latest available UCD files? That complicates rerunning the scripts with the intention to match what might have been downloaded previously. If it is intentional, perhaps rename the script to `download_latest_unicode_data_files.py`. Otherwise, the files specific to Unicode 15.1.0 (for example) are available at https://unicode.org/Public/15.1.0/ucd. Another alternative would be to allow a base Unicode version URL to be supplied on the command line.

PR #4437, opened 2024-03-03
Enforce `std::time_base::mdy` for "C" locale

### LINKED ISSUE #3346: Tests failing for DMY locales
@rogerorr reported to me that some of the STL's tests are failing when Windows 11 is set to an English (United Kingdom) locale with `dd/MM/yyyy` short date format and `dd MMMM yyyy` long date format (and Monday as the first day of the week, although that is probably unrelated).

The failing tests were:

* `libcxx` `std/localization/locale.categories/category.time/locale.time.get/locale.time.get.members/date_order.pass.cpp`
  + [Line 33](https://github.com/llvm/llvm-project/blob/cb2f23a9b0de9c6747afb1efdc6e05b4691e0c50/libcxx/test/std/localization/locale.categories/category.time/locale.time.get/locale.time.get.members/date_order.pass.cpp#L33) says `assert(f.date_order() == std::time_base::mdy)` so failure is unsurprising.
* `tr1/tests/locale3` is seeing the day and month being swapped:
  + https://github.com/microsoft/STL/blob/fbed7eaec0df4fefec2e447604e12d39d9eab857/tests/tr1/tests/locale3/test.cpp#L439-L440
* `tr1/tests/locale4` has the same problem:
  + https://github.com/microsoft/STL/blob/fbed7eaec0df4fefec2e447604e12d39d9eab857/tests/tr1/tests/locale4/test.cpp#L411-L412

fsb4000:
Fixes #3346

1) I found out that the tests work on libc++. 
2) I looked how libc++ detect date_order. It has a some date and run `std::strftime(mbstr, sizeof(mbstr), "%x", std::localtime(&t));` and analyse its output.
3) I looked at `strftime` implementation. It creates `_LocaleUpdate` object. and passes `nullptr` to it.
`_LocaleUpdate` constructor looks like that (pseudo-code)
```C
if (locale)
{
    _locale_pointers = *locale;
}
else if (is_it_C_locale())
{
    _locale_pointers = some_predefined_values;
}
else
{
   query runtime data to fill _locale_pointers from the current locale.
}
```
4) So  `std::strftime` uses `mdy` for "C" locale and we can do too. I found another places where we did a special case for "C" locale.
https://github.com/microsoft/STL/blob/8b081e26ba016970ce3338cb483ff10f2ade30a5/stl/src/xstrcoll.cpp#L57-L69
https://github.com/microsoft/STL/blob/8b081e26ba016970ce3338cb483ff10f2ade30a5/stl/src/_tolower.cpp#L52-L58
https://github.com/microsoft/STL/blob/8b081e26ba016970ce3338cb483ff10f2ade30a5/stl/src/_toupper.cpp#L51-L57

### DIFF
--- a/stl/src/xdateord.cpp
+++ b/stl/src/xdateord.cpp
@@ -10,14 +10,14 @@
 _EXTERN_C_UNLESS_PURE
 
 _CRTIMP2_PURE int __CLRCALL_PURE_OR_CDECL _Getdateorder() noexcept { // return date order for current locale
-    wchar_t buf[2]                   = {0};
     const wchar_t* const locale_name = ___lc_locale_name_func()[LC_TIME];
     if (locale_name == nullptr) {
         // Enforce std::time_base::mdy for "C" locale
         return std::time_base::mdy;
-    } else {
-        GetLocaleInfoEx(locale_name, LOCALE_ILDATE, buf, static_cast<int>(std::size(buf)));
     }
+
+    wchar_t buf[2] = {0};
+    GetLocaleInfoEx(locale_name, LOCALE_ILDATE, buf, static_cast<int>(std::size(buf)));
 
     switch (buf[0]) {
     case L'0':
--- a/stl/src/xdateord.cpp
+++ b/stl/src/xdateord.cpp
@@ -10,8 +10,14 @@
 _EXTERN_C_UNLESS_PURE
 
 _CRTIMP2_PURE int __CLRCALL_PURE_OR_CDECL _Getdateorder() noexcept { // return date order for current locale
-    wchar_t buf[2] = {0};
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
I believe this behavioral change is desirable - we'll properly detect when we should be using the "C" locale, instead of picking up `LOCALE_NAME_USER_DEFAULT` for the date order - but as it affects legacy code in my magical weak point, I'd like a second maintainer review here.

StephanTLavavej:

StephanTLavavej:
Thanks for investigating and fixing this longstanding bug that was affecting several contributors! :earth_americas: :earth_africa: :earth_asia:

PR #4439, opened 2024-03-04
Comment cleanups

StephanTLavavej:
* Fix comment typos.
* Drop comment `// inspected by GSL, do not remove`.
  + This is no longer necessary after microsoft/GSL#822.
* Comments: `C++23 / C++20` => `_HAS_CXX23 / !_HAS_CXX23`
  + This is more conventional and more accurately descriptive.
* Comments: Replace weird `_HAS_CXX23` comments with conventional ones.
* Comments: Replace weird `^^ real test / vv workaround` comments with conventional ones.
* Comments: In `yvals_core.h`, drop most comments around "macros with language mode sensitivity".
  + This eliminates one last occurrence of `^^^ C++23 / C++20 vvv`. We were pretty inconsistent about the patterns here, and these comments were distracting from the actually useful paper citations. The regions here are all single line, so it's okay to skip preprocessor comments. The only comment I retained was `#endif // language mode` because it's next to `#endif // !defined(_M_CEE_PURE)`.

### DIFF
--- a/stl/inc/compare
+++ b/stl/inc/compare
@@ -305,9 +305,9 @@ concept three_way_comparable_with =
     three_way_comparable<_Ty1, _Cat> && three_way_comparable<_Ty2, _Cat>
 #if _HAS_CXX23
     && _Comparison_common_type_with<_Ty1, _Ty2>
-#else // ^^^ C++23 / C++20 vvv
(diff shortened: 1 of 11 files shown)


StephanTLavavej:

PR #4297, opened 2024-01-04
Enable `__cpp_lib_concepts` for EDG, part 2

cpplearner:
After #4296, `concepts_20_matrix.lst` and `concepts_latest_matrix.lst` will be near-identical to their usual counterparts. This PR changes the tests to include the usual matrices instead of the concepts ones, and then removes the concepts matrices.

Blocked on #4296.

PR #4440, opened 2024-03-04
Tests: Rename the strict matrices

StephanTLavavej:
This renames the "strict" (`/permissive-`) matrices, now that our C++20 mode always supports concepts. Followup to #4297.

* `strict_concepts_20_matrix.lst` => `strict_20_matrix.lst`
* `strict_concepts_latest_matrix.lst` => `strict_latest_matrix.lst`


StephanTLavavej:

PR #4441, opened 2024-03-04
Enable more test coverage for warning C4668 ('MEOW' is not defined as a preprocessor macro)

StephanTLavavej:
Warning C4668 is off-by-default, but quite useful because (1) we conventionally never want to rely on the Standard's behavior of expanding not-defined macros to `0`, and (2) it can detect typos. We already have some test coverage for it:

https://github.com/microsoft/STL/blob/8b081e26ba016970ce3338cb483ff10f2ade30a5/stl/inc/__msvc_all_public_headers.hpp#L13

However, we haven't applied this warning to our entire test suite, so I've had to catch other PRs unintentionally expanding not-defined macros.

* C4668: In `std` and `tr1`, enable warning C4668 `'MEOW'` is not defined as a preprocessor macro, replacing with `'0'` for `'#if/#elif'`.
  + I'm excluding `libcxx` because it doesn't follow our convention.
* C4668: Disable warning C4668 around `<Windows.h>`.
  + This works around OS-23694920 "Windows SDK uses undefined `_WIN32_WINNT_*` version macros (which may subtly break code)".
* C4668: Remove unnecessary inclusion of `<Windows.h>`.
* C4668: `#if WIDE` => `#ifdef WIDE`
  + These occurrences were inconsistent.
* C4668: Activate C99 logic in `tr1/include/tdefs.h`.
  + This is a behavioral change/enhancement (the code was previously inactive). It fixes "warning C4668: `'__STDC_VERSION__'` is not defined as a preprocessor macro, replacing with `'0'` for `'#if/#elif'`".
  + We don't usually mess with `tr1`, but this looks like a straightforward improvement, and it only takes effect when the "approximately equal" test is already going to fail.
* C4668: Disable warning C4668 around `<limits.h>`.
  + This is the only product code change, to work around VSO-1985428 "`<limits.h>` emits warning C4668: `'__STDC_WANT_SECURE_LIB__'` is not defined as a preprocessor macro, replacing with `'0'` for `'#if/#elif'`".


StephanTLavavej:

PR #4442, opened 2024-03-04
Tests: Extract `is_permissive.hpp`

StephanTLavavej:
We've accumulated a bunch of test files copy-pasting the machinery used to detect `/permissive` mode. Let's centralize it.

This is based on `test_mdspan_support.hpp`, with `namespace details` renamed to `namespace detail` (see below). Its `PermissiveTestBase` and `PermissiveTest` were nicely named, and it was using `inline constexpr bool`. Also add `is_permissive_v<T>`, when we need it to be dependent on a template parameter. We need to include `<yvals_core.h>` for `_INLINE_VAR`, since there's some C++14 usage.

`test_mdspan_support.hpp`: Drive-by change, this file had both `namespace details` and `namespace detail`. Standardize on the latter, as it's used everywhere else.

`P2136R3_invoke_r`: `// TRANSITION, DevCom-1457457` is repeated where `is_permissive` is used, so we can drop the one next to the definition.

### DIFF
--- a/tests/std/include/is_permissive.hpp
+++ b/tests/std/include/is_permissive.hpp
@@ -1,0 +1,31 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#pragma once
+
+#include <yvals_core.h>
+
+namespace detail {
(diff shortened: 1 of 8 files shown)


StephanTLavavej:

PR #4443, opened 2024-03-04
Expand old test coverage

StephanTLavavej:
* Overhaul FDIS citation and extend test coverage for `std::function` PMFs/PMDs.
  + FDIS citations were so old, they escaped our scans for outdated WPs. We don't need to quote `INVOKE`'s Standardese at length, we just need to summarize it. It used to be really confusing, now it's easy mode.
  + A couple of things have changed: derived values are now covered by the Standardese, and `reference_wrapper` support was added. This is tested elsewhere, e.g. in `Dev11_0535636_functional_overhaul`, but we can easily add comprehensive test coverage here.
  + I chose `f1x` and `g1x` to avoid renumbering everything.
* Update FDIS citations and add test coverage for `mem_fn` typedefs.
  + These typedefs were deprecated in C++17 and removed in C++20, so I'm citing WG21-N4659, the C++17 WP.
  + We were also missing test coverage for when `mem_fn` provides only `result_type`. Even though this is deprecated/removed, we may as well be comprehensive.
* `P0088R3_variant`: Activate EDG test coverage.
  + Now that concepts are available.
  + The preprocessor comments aren't being changed here because #4439 is cleaning them up better.
* Drive-by cleanup: In `benchmarks/inc`, include `<cstdint>` and qualify `std::`.

### DIFF
--- a/benchmarks/inc/utility.hpp
+++ b/benchmarks/inc/utility.hpp
@@ -14,7 +14,7 @@
 template <class Contained>
 std::vector<Contained> random_vector(size_t n) {
     std::random_device rd;
-    std::uniform_int_distribution<uint64_t> id64;
+    std::uniform_int_distribution<std::uint64_t> id64;
     xoshiro256ss prng{id64(rd), id64(rd), id64(rd), id64(rd)};
 
     std::vector<Contained> res(n);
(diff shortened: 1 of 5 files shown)


StephanTLavavej:

PR #4444, opened 2024-03-04
Various cleanups: product code

StephanTLavavej:
* In `_Popcount_fallback()`, after `if constexpr ... return` add `else` to avoid dead code.
  + We always follow this convention for `if constexpr`.
* Remove unnecessary `_STD` qualification on types, variable templates, and alias templates.
  + This became possible after my #4146 moved these functions into `namespace std`.
* `yvals_core.h`: Improve "error STL1003: Unexpected compiler, expected C++ compiler." like #4020.
  + Then we can move `_STL_STRINGIZE` down to its point of use, near our other macros.
  + Result:
    ```
    D:\GitHub\STL\out\x64>type woof.c
    #include <vector>
    
    D:\GitHub\STL\out\x64>cl /EHsc /nologo /W4 /c woof.c
    woof.c
    D:\GitHub\STL\out\x64\out\inc\yvals_core.h(23): fatal error C1189: #error:  error STL1003: Unexpected compiler, expected C++ compiler.
    ```


StephanTLavavej:

PR #4445, opened 2024-03-04
Various cleanups: `ppltasks.cpp`

StephanTLavavej:
I've validated this with an internal build and test run. I know there's not a lot of value in cleaning up old `src` files, but this eliminates some complexity and I believe it's safe.

* `RtlCaptureStackBackTrace` is universally available.
  + It might not have been when this was originally implemented, but it's now [documented](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-rtlcapturestackbacktrace) as "Target Platform: Universal", and this aligns with its actual declaration.
* Move `GUID` into `Concurrency::details` within `#if defined(_CRT_APP) || defined(UNDOCKED_WINDOWS_UCRT)`.
  + It's `static`, so there's no ABI impact.
* We don't need `<mutex>` - only `<functional>` for `std::function` and `<stdexcept>` for `std::runtime_error`.
* Drop "ASYNC TODO" comment.
  + We're never going to change this legacy behavior.

### DIFF
--- a/stl/src/ppltasks.cpp
+++ b/stl/src/ppltasks.cpp
@@ -52,10 +52,6 @@ namespace Concurrency {
             static unsigned int s_asyncId = 0;
 
             _CRTIMP2 unsigned int __cdecl GetNextAsyncId() {
-                //
-                // ASYNC TODO: Determine the requirements on the domain uniqueness of this value.  C++ / C# / WRL are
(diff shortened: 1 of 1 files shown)


StephanTLavavej:

PR #4450, opened 2024-03-05
Fix vectorized `ranges::find` with `unreachable_sentinel` to properly mask the beginning and handle unaligned pointers

### LINKED ISSUE #4449: `<xutility>`: `ranges::find` with `unreachable_sentinel` / `__std_find_trivial_unsized_1` gives wrong result
# Describe the bug
When using ranges::find on a range of integers with unreachable_sentinel as end iterator, sometimes it finds a value before the begin iterator. It only happens if _USE_STD_VECTOR_ALGORITHMS is 1, when it's using __std_find_trivial_unsized_1.

# Command-line test case
```
>type repro.cpp
#include <cstdint>
#include <array>
#include <algorithm>
#include <print>

int main()
{
    std::array<std::uint8_t, 0x10000> arr;

    std::ranges::fill(arr, std::uint8_t{1});
    arr.back() = 0;

    std::println("begin: {}", static_cast<void*>(std::to_address(std::begin(arr))));
    std::println("end:   {}", static_cast<void*>(std::to_address(std::end(arr))));

    for (auto it = std::begin(arr); it != std::end(arr); ++it) {
        auto where = std::ranges::find(it, std::unreachable_sentinel, 0);
        auto dist = std::distance(it, where);
        if (dist < 0)
            std::println("{:p} = {}", static_cast<void*>(std::to_address(where)), dist);
    }
}

>cl /EHsc /W4 /WX /std:c++latest .\repro.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.40.33521 for x64

repro.cpp
Microsoft (R) Incremental Linker Version 14.40.33521.0

/out:repro.exe
repro.obj

>.\repro.exe
begin: 0x6802effbb0
end:   0x6802f0fbb0
0x6802effba0 = -16
0x6802effba6 = -11
0x6802effba6 = -12
0x6802effba6 = -13
0x6802effba6 = -14
0x6802effba6 = -15
0x6802effba6 = -16
0x6802effba6 = -17
0x6802effba6 = -18
0x6802effba6 = -19
0x6802effba6 = -20
0x6802effba6 = -21
0x6802effba6 = -22
0x6802effba6 = -23
0x6802effba6 = -24
0x6802effba6 = -25
```

# Expected behavior
Expected to find the 0 at the end of the range instead of getting results before the start of the range.

# STL version
 ```
Version 17.9.2
 ```

# Additional context
The bug only seems to occur at certain memory addresses, so with the test array placed on the stack, the code might not show wrong results every single run. Similar problem seems to occur with different integer sizes.

### LINKED ISSUE #4454: `vector_algorithms.cpp`: `__std_find_trivial_unsized_impl` assumes N-byte elements are N-aligned
This initially appeared to be x86 `/clr` specific - but it wasn't, see below.

With #4450 and an MSVC-internal build:

```
C:\Temp>cl /clr /nologo /W4 /std:c++20 /MDd /Od D:\msvc\src\vctools\crt\github\tests\std\tests\VSO_0000000_vector_algorithms\test.cpp /I D:\msvc\src\vctools\crt\github\tests\std\include
test.cpp

C:\Temp>test
This is a randomized test.
DO NOT IGNORE/RERUN ANY FAILURES.
You must report them to the STL maintainers.

Seed vector: [...]

Unhandled Exception: System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at __std_find_trivial_unsized_8(Void* , UInt64 )
   at std.ranges._Find_unchecked<unsigned __int64 *,struct std::unreachable_sentinel_t,unsigned __int64,struct std::identity>(UInt64* _First, unreachable_sentinel_t _Last, UInt64* _Val, identity _Proj)
   at std.ranges._Find_fn.operator()<unsigned __int64 *,struct std::unreachable_sentinel_t,unsigned __int64,struct std::identity>(_Find_fn* , UInt64* _First, unreachable_sentinel_t _Last, UInt64* _Val, identity _Proj)
   at test_gh_4449<unsigned __int64>()
   at test_vector_algorithms(mersenne_twister_engine<unsigned __int64\,64\,312\,156\,31\,-5403634167711393303\,29\,6148914691236517205\,17\,8202884508482404352\,37\,-2270628950310912\,43\,6364136223846793005>* gen)
   at main()
   at _mainCRTStartup()
```

I can also repro this with plain VS 2022 17.10 Preview 1 x86, if I take #4450's test and alter it to disable the `assert(where == arr + mid2)`, so this doesn't appear to be a regression caused by #4450.

StephanTLavavej:
Fixes #4449.

For unsized `ranges::find`, `vector_algorithms.cpp` reads elements "before the beginning", although (hopefully) not in a way that annoys memory page protection or ASAN. Then we "we mask out matches that don't belong to the range", pretending that they weren't there. This allows us to use vectorized loads for the entire unbounded loop, instead of starting with classic scalar code until we get to a nicely-aligned boundary.

However, we had a control flow bug. We started with "load" a vector chunk, "mask" away matches before the beginning, "check" if we found anything (and return if so). Then we started our infinite loop: `for (;;) { load; check; advance; }`. But this repeated the initial load and check, *without the mask!* It was also unnecessary extra work.

A minimal correctness fix would be to cycle around the "advance" step: `load; mask; check; for (;;) { advance; load; check; }`. But we can do even better - the "check" steps are exactly identical between the first part and the infinite loop. So we can cycle the loop a bit more: `load; mask; for (;;) { check; advance; load; }` avoids having to repeat the "check". (The "load" steps are also exactly identical, but there's no easy way to fuse them, given that we want to mask only the first one.)

Also fixes #4454.

Unlike the other vectorized algorithms, find-unsized uses aligned loads, so it requires that its N-byte elements are N-aligned. This is notoriously untrue on x86, where 8-byte elements can appear on a 4-aligned stack. Packed structs can also subvert this assumption.

We can simply test whether the pointer is properly aligned. I was able to fix this with the following control flow:

```cpp
#ifndef _M_ARM64EC
if (unaligned) {
    // use the scalar fallback below
} else if (_Use_avx2()) {
    always return AVX2 result;
} else if (_Traits::_Sse_available()) {
    always return SSE-n result;
}
#endif // !_M_ARM64EC
return scalar fallback;
```

Note that (unlike some other vectorized algorithms) our AVX2 and SSE-n codepaths here are always-return, so chaining them with `else if` is fine. I thought this was less disruptive than increasing the level of control flow nesting.

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -1862,16 +1862,8 @@ namespace {
             unsigned int _Bingo = static_cast<unsigned int>(_mm256_movemask_epi8(_Traits::_Cmp_avx(_Data, _Comparand)));
 
             _Bingo &= _Mask;
-            if (_Bingo != 0) {
-                unsigned long _Offset = _tzcnt_u32(_Bingo);
-                _Advance_bytes(_First, _Offset);
-                return _First;
-            }
 
             for (;;) {
-                _Data  = _mm256_load_si256(static_cast<const __m256i*>(_First));
-                _Bingo = static_cast<unsigned int>(_mm256_movemask_epi8(_Traits::_Cmp_avx(_Data, _Comparand)));
-
                 if (_Bingo != 0) {
                     unsigned long _Offset = _tzcnt_u32(_Bingo);
                     _Advance_bytes(_First, _Offset);
@@ -1879,6 +1871,9 @@ namespace {
                 }
 
                 _Advance_bytes(_First, 32);
+
+                _Data  = _mm256_load_si256(static_cast<const __m256i*>(_First));
+                _Bingo = static_cast<unsigned int>(_mm256_movemask_epi8(_Traits::_Cmp_avx(_Data, _Comparand)));
             }
         }
 
@@ -1898,17 +1893,8 @@ namespace {
             unsigned int _Bingo = static_cast<unsigned int>(_mm_movemask_epi8(_Traits::_Cmp_sse(_Data, _Comparand)));
 
             _Bingo &= _Mask;
-            if (_Bingo != 0) {
-                unsigned long _Offset;
-                _BitScanForward(&_Offset, _Bingo); // lgtm [cpp/conditionallyuninitializedvariable]
-                _Advance_bytes(_First, _Offset);
-                return _First;
-            }
 
             for (;;) {
-                _Data  = _mm_load_si128(static_cast<const __m128i*>(_First));
-                _Bingo = static_cast<unsigned int>(_mm_movemask_epi8(_Traits::_Cmp_sse(_Data, _Comparand)));
-
                 if (_Bingo != 0) {
                     unsigned long _Offset;
                     _BitScanForward(&_Offset, _Bingo); // lgtm [cpp/conditionallyuninitializedvariable]
@@ -1917,6 +1903,9 @@ namespace {
                 }
 
                 _Advance_bytes(_First, 16);
+
+                _Data  = _mm_load_si128(static_cast<const __m128i*>(_First));
+                _Bingo = static_cast<unsigned int>(_mm_movemask_epi8(_Traits::_Cmp_sse(_Data, _Comparand)));
             }
         }
 #endif // !_M_ARM64EC
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -1844,7 +1844,10 @@ namespace {
     template <class _Traits, class _Ty>
     const void* __stdcall __std_find_trivial_unsized_impl(const void* _First, const _Ty _Val) noexcept {
 #ifndef _M_ARM64EC
-        if (_Use_avx2()) {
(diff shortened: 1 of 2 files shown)


AlexGuteniev (approved):

StephanTLavavej:

barcharcraz (approved):

stl/src/vector_algorithms.cpp
```diff
@@ -1862,27 +1865,20 @@ namespace {
...
-            }
 
             for (;;) {
-                _Data  = _mm256_load_si256(static_cast<const __m256i*>(_First));
-                _Bingo = static_cast<unsigned int>(_mm256_movemask_epi8(_Traits::_Cmp_avx(_Data, _Comparand)));
-
                 if (_Bingo != 0) {
                     unsigned long _Offset = _tzcnt_u32(_Bingo);
                     _Advance_bytes(_First, _Offset);
                     return _First;
```
AlexGuteniev, line 1873: I observe that since we have only one return from AVX2 code path, we can remove ` _Zeroupper_on_exit _Guard;` and put explicit `_mm256_zeroupper();` here. Sure even if it is worth doing, can be done as a follow-up

Issue #997, opened 2020-07-05
<regex>: Grouping within repetition causes regex stack error


AlexGuteniev:
**Describe the bug**
`std::regex` with satisfactory large input and certain compiler options overflows the stack

**Command-line test case**
```

d:\Temp2>type repro.cpp
#include <Windows.h>
#include <iostream>
#include <regex>
#include <string>

void regex_test()
{
    try
    {
        auto str = std::string(1000, 'a');
        auto match = std::smatch{};
        std::regex_match(str, match, std::regex("a+"));     // OK
        std::regex_match(str, match, std::regex("(?:a)+")); // stack error
    }
    catch (std::exception& ex)
    {
        std::cout << ex.what() << '\n';
    }
}

int main()
{
    bool stack_overflow_observed = false;
    __try
    {
        regex_test();
    }
    __except (GetExceptionCode() == EXCEPTION_STACK_OVERFLOW)
    {
        stack_overflow_observed = true;
    }
    if (stack_overflow_observed)
    {
        _resetstkoflw();
        std::cout << "Stack overflow was observed\n";
    }
    else
    {
        std::cout << "Stack overflow was NOT observed\n";
    }
    return 0;
}

d:\Temp2>cl /EHa /RTC1 /MDd /std:c++17 .\repro.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.27.29009.1 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

repro.cpp
Microsoft (R) Incremental Linker Version 14.27.29009.1
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:repro.exe
repro.obj

d:\Temp2>.\repro.exe
Stack overflow was observed
```

**Expected behavior**
Successful parsing or C++ exception, but not stack overflow

**STL version**
```
Microsoft Visual Studio Professional 2019 Preview
Version 16.7.0 Preview 3.1
```

**Additional context**
This item is also tracked on Developer Community as DevCom-885115  and by Microsoft-internal VSO-1054746 / [AB#1054746](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1054746).

See also #405

**vNext note**: Resolving this issue will require breaking binary compatibility. We won't be able to accept pull requests for this issue until the vNext branch is available. See https://github.com/microsoft/STL/issues/169 for more information.

AlexGuteniev:
[...]

PR #4451, opened 2024-03-05
`<regex>`: Limit capture groups

StephanTLavavej:
In general, the STL cannot avoid all stack overflows. (We don't know how much stack has been consumed when the user calls us, so thinking *any* thoughts might be too much.) However, `<regex>` tries to throw exceptions instead of stack overflowing due to excessive recursion. We've had a long-standing struggle with exactly where to set this threshold during matching; right now our threshold detects most potential overflows before they happen, but not all.

However, we don't do anything during regex *construction*. It turns out that requesting an excessive number of capture groups can easily trigger a stack overflow. We even had commented-out legacy code, indicating that our predecessors thought about this scenario somewhat. This PR adds a hardcoded limit arbitrarily set at 1000, which is more than generous enough for realistic code, yet comfortably far away from where I begin to observe stack overflows.

Alternatively, we could have more comprehensive logic to track the number of nested function calls during regex construction (which I believe would be entirely ABI-safe), but I'd like to go with this targeted change for now.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -4153,9 +4153,13 @@ void _Parser<_FwdIt, _Elem, _RxTraits>::_CharacterClass() { // add bracket expre
 
 template <class _FwdIt, class _Elem, class _RxTraits>
 void _Parser<_FwdIt, _Elem, _RxTraits>::_Do_capture_group() { // add capture group
-    // if (_MAX_GRP <= ++_Grp_idx)
-    //     _Error(regex_constants::error_complexity);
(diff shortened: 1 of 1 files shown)


fsb4000:
Will your patch fix the issue: https://github.com/microsoft/STL/issues/997 ?

StephanTLavavej:
No, that occurs during matching.

StephanTLavavej:

barcharcraz (approved):
This seems fine, although I find checking for stack overflows in this way somewhat dubious.

PR #4466, opened 2024-03-10
Vectorize `find_first_of` for 8 and 16 bit elements with SSE4.2 `pcmpestri`

AlexGuteniev:
Suggested by @Alcaro in https://github.com/microsoft/STL/issues/4415#issuecomment-1954703449

For element set that fits SSE register, that is length of "needle" is up to 16 for 8-bit element, up to 8 for 16-bit element

Possible future work:
 * Use `find` instead for 1-element cases as @Alcaro suggested
 * `basic_string::find_first_of`. Certainly not the general case with user-provided char traits, but maybe for standard cases;
 * Expand to more that 16 bytes "needle". Will add inner loop, make it square algorithmic complexity, and more implementation complexity, but may be useful.

**Benchmark result**
 
The first non-type template parameter in the benchmark results is the position where the value is found, "haystack" length is twice that.
The second non-type template parameter in the benchmark results is the "needle" size.

Before:
```
bm<uint8_t, 2, 3>            6.33 ns         6.42 ns    112000000
bm<uint16_t, 2, 3>           6.47 ns         6.45 ns     89600000
bm<uint8_t, 7, 4>            22.8 ns         22.9 ns     32000000
bm<uint16_t, 7, 4>           15.1 ns         15.3 ns     44800000
bm<uint8_t, 9, 3>            17.7 ns         17.6 ns     37333333
bm<uint16_t, 9, 3>           16.6 ns         16.7 ns     44800000
bm<uint8_t, 22, 5>           73.7 ns         73.2 ns      8960000
bm<uint16_t, 22, 5>          69.3 ns         69.8 ns     11200000
bm<uint8_t, 3056, 7>         7991 ns         8022 ns        89600
bm<uint16_t, 3056, 7>        7314 ns         7324 ns        89600
bm<uint8_t, 1011, 11>        3709 ns         3767 ns       186667
bm<uint16_t, 1011, 11>       3662 ns         3599 ns       186667
```
After:
```
bm<uint8_t, 2, 3>            6.22 ns         6.14 ns    112000000
bm<uint16_t, 2, 3>           6.95 ns         6.98 ns     89600000
bm<uint8_t, 7, 4>            23.9 ns         24.1 ns     29866667
bm<uint16_t, 7, 4>           23.1 ns         22.9 ns     32000000
bm<uint8_t, 9, 3>            13.1 ns         13.2 ns     49777778
bm<uint16_t, 9, 3>           14.1 ns         14.1 ns     49777778
bm<uint8_t, 22, 5>           14.3 ns         14.3 ns     44800000
bm<uint16_t, 22, 5>          15.0 ns         15.0 ns     44800000
bm<uint8_t, 3056, 7>          314 ns          314 ns      2240000
bm<uint16_t, 3056, 7>         611 ns          614 ns      1120000
bm<uint8_t, 1011, 11>         107 ns          107 ns      6400000
bm<uint16_t, 1011, 11>       3478 ns         3516 ns       213333
```

Explanation:
 * Starting with `bm<uint8_t, 9, 3>` row, the vectorization is engaged;
 * `bm<uint8_t, 22, 5>` already shows significant improvement;
 * The exact amount of the improvement is hard to determine, as it is comparison of O(N**2) vs O(N) algorithm, the longer "needle" is, the more improvement is observed;
 * `bm<uint16_t, 1011, 11>`  falls back to the scalar algorithm due to "needle" length not fitting SSE register.


Alcaro:
As we discovered on Discord a while ago, some people call std::find_first_of with a single-element needle. This should be forwarded to std::find (or memchr or something), even if vectorization is otherwise disabled.

AlexGuteniev:
> As we discovered on Discord a while ago, some people call std::find_first_of with a single-element needle. This should be forwarded to std::find (or memchr or something), even if vectorization is otherwise disabled.

I remember this, but it os so much unrelated, that alI think it better fits a separate PR

stl/inc/algorithm
```diff
@@ -3321,6 +3358,26 @@ _NODISCARD _CONSTEXPR20 _FwdIt1 find_first_of(
...
+    if constexpr (_Vector_alg_in_find_first_of_is_safe<_FwdIt1, _FwdIt2, _Pr>) {
+        if (!_STD _Is_constant_evaluated() && _ULast1 - _UFirst1 >= _Threshold) {
+            const auto _First1_ptr = _STD _To_address(_UFirst1);
+            const auto _Result     = _STD __std_find_first_of_trivial(
+                    _First1_ptr, _STD _To_address(_ULast1), _STD _To_address(_UFirst2), _STD _To_address(_ULast2));
+
+            if constexpr (is_pointer_v<decltype(_UFirst1)>) {
+                _UFirst1 = _Result;
+            } else {
+                _UFirst1 += (_Result - _First1_ptr);
```
StephanTLavavej, line 3373: The parentheses here are unnecessary (compound assignment has very low precedence).

benchmarks/src/find_first_of.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+template <class T, size_t Pos, size_t NSize, size_t HSize = Pos * 2, size_t Which = 0>
+void bm(benchmark::State& state) {
+    vector<T> h(HSize, T{'.'});
+    vector<T> n(NSize);
+    iota(n.begin(), n.end(), T{'a'});
+
+    static_assert(Pos < HSize);
+    static_assert(Which < NSize);
+    h.at(Pos) = n.at(Which);
```
StephanTLavavej, line 20: After the `static_assert`s, `at()` is super unnecessary, it should just be square brackets.

benchmarks/src/find_first_of.cpp
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <algorithm>
+#include <benchmark/benchmark.h>
+#include <cstddef>
+#include <cstdint>
+#include <numeric>
+
```
StephanTLavavej, line 9: Should include `<vector>`.

stl/inc/algorithm
```diff
@@ -160,6 +165,38 @@ _Ty* __std_find_last_trivial(_Ty* const _First, _Ty* const _Last, const _TVal _V
         static_assert(_Always_false<_Ty>, "Unexpected size");
     }
 }
+
+template <class _Ty>
+_Ty* __std_find_first_of_trivial(
+    _Ty* const _First1, _Ty* const _Last1, _Ty* const _First2, _Ty* const _Last2) noexcept {
```
StephanTLavavej, line 171: This breaks existing code with varying constness, which also indicates missing test coverage:

```
```
```cpp
#include <algorithm>
#include <cassert>
#include <vector>
using namespace std;

int main() {
    const vector<char> haystack{'m', 'e', 'o', 'w', 'C', 'A', 'T', 'S'};
    vector<char> needle{'R', 'S', 'T'};
    const auto result = find_first_of(haystack.begin(), haystack.end(), needle.begin(), needle.end());
    assert(result == haystack.begin() + 6);
}
```
```
D:\GitHub\STL\out\x64>cl /EHsc /nologo /W4 /std:c++latest /MTd /Od /D_USE_STD_VECTOR_ALGORITHMS=0 meow.cpp && meow
meow.cpp

meow.cpp
D:\GitHub\STL\out\x64\out\inc\algorithm(3367): error C2672: 'std::__std_find_first_of_trivial': no matching overloaded function found
D:\GitHub\STL\out\x64\out\inc\algorithm(170): note: could be '_Ty *std::__std_find_first_of_trivial(_Ty *const ,_Ty *const ,_Ty *const ,_Ty *const ) noexcept'
D:\GitHub\STL\out\x64\out\inc\algorithm(3367): note: '_Ty *std::__std_find_first_of_trivial(_Ty *const ,_Ty *const ,_Ty *const ,_Ty *const ) noexcept': template parameter '_Ty' is ambiguous
D:\GitHub\STL\out\x64\out\inc\algorithm(3367): note: could be '_Ty'
        with
        [
            _Ty=char
        ]
D:\GitHub\STL\out\x64\out\inc\algorithm(3367): note: or       '_Ty'
        with
        [
            _Ty=const char
        ]
D:\GitHub\STL\out\x64\out\inc\algorithm(3367): note: '_Ty *std::__std_find_first_of_trivial(_Ty *const ,_Ty *const ,_Ty *const ,_Ty *const ) noexcept': could not deduce template argument for '_Ty *const ' from '_Ty *'
        with
        [
            _Ty=char
        ]
D:\GitHub\STL\out\x64\out\inc\algorithm(3367): note: the template instantiation context (the oldest one first) is
meow.cpp(9): note: see reference to function template instantiation '_FwdIt1 std::find_first_of<std::_Vector_const_iterator<std::_Vector_val<std::_Simple_types<_Ty>>>,std::_Vector_iterator<std::_Vector_val<std::_Simple_types<_Ty>>>>(const _FwdIt1,const _FwdIt1,const _FwdIt2,const _FwdIt2)' being compiled
        with
        [
            _FwdIt1=std::_Vector_const_iterator<std::_Vector_val<std::_Simple_types<char>>>,
            _Ty=char,
            _FwdIt2=std::_Vector_iterator<std::_Vector_val<std::_Simple_types<char>>>
        ]
D:\GitHub\STL\out\x64\out\inc\algorithm(3397): note: see reference to function template instantiation '_FwdIt1 std::find_first_of<_FwdIt1,_FwdIt2,std::equal_to<void>>(_FwdIt1,const _FwdIt1,const _FwdIt2,const _FwdIt2,_Pr)' being compiled
        with
        [
            _FwdIt1=std::_Vector_const_iterator<std::_Vector_val<std::_Simple_types<char>>>,
            _FwdIt2=std::_Vector_iterator<std::_Vector_val<std::_Simple_types<char>>>,
            _Pr=std::equal_to<void>
        ]
D:\GitHub\STL\out\x64\out\inc\algorithm(3367): error C2737: '_Result': const object must be initialized
D:\GitHub\STL\out\x64\out\inc\algorithm(3371): error C3536: '_Result': cannot be used before it is initialized
D:\GitHub\STL\out\x64\out\inc\algorithm(3371): error C2440: '=': cannot convert from 'int' to 'const char *'
D:\GitHub\STL\out\x64\out\inc\algorithm(3371): note: Conversion from integral type to pointer type requires reinterpret_cast, C-style cast or parenthesized function-style cast
```

stl/src/vector_algorithms.cpp
```diff
@@ -2075,6 +2075,75 @@ namespace {
         }
         return _Result;
     }
+
+    template <class _Ty>
+    const void* __stdcall __std_find_first_of_trivial_impl(
+        const void* _First1, const void* const _Last1, const void* const _First2, const void* const _Last2) noexcept {
+#ifndef _M_ARM64EC
+        constexpr bool _Bytes = sizeof(_Ty) == 1;
```
StephanTLavavej, line 2083: Suggestion: `_Bytes` sounds like a number of bytes. We recently used `_Single_byte_elements` which would be consistently descriptive here.

stl/src/vector_algorithms.cpp
```diff
@@ -2075,6 +2075,75 @@ namespace {
...
     }
+
+    template <class _Ty>
+    const void* __stdcall __std_find_first_of_trivial_impl(
+        const void* _First1, const void* const _Last1, const void* const _First2, const void* const _Last2) noexcept {
+#ifndef _M_ARM64EC
+        constexpr bool _Bytes = sizeof(_Ty) == 1;
+        constexpr int _Op =
+            (_Bytes ? _SIDD_UBYTE_OPS : _SIDD_UWORD_OPS) | _SIDD_CMP_EQUAL_ANY | _SIDD_LEAST_SIGNIFICANT;
+        constexpr int _Part_size_el = _Bytes ? 16 : 8;
```
StephanTLavavej, line 2086: While these are lower risk because they're `constexpr` (modifiable state would be higher risk), it would still be better to scope these within the conditional statement where they're used.

stl/src/vector_algorithms.cpp
```diff
@@ -2075,6 +2075,75 @@ namespace {
...
+
+        const size_t _Needle_length = _Byte_length(_First2, _Last2);
+
+        if (_Use_sse42() && _Needle_length <= 16) {
+            const int _Needle_length_el = static_cast<int>(_Byte_length(_First2, _Last2) / sizeof(_Ty));
```
StephanTLavavej, line 2091: This should reuse `_Needle_length`.

stl/inc/algorithm
```diff
@@ -3321,6 +3358,26 @@ _NODISCARD _CONSTEXPR20 _FwdIt1 find_first_of(
     const auto _ULast1  = _STD _Get_unwrapped(_Last1);
     const auto _UFirst2 = _STD _Get_unwrapped(_First2);
     const auto _ULast2  = _STD _Get_unwrapped(_Last2);
+#if _USE_STD_VECTOR_ALGORITHMS
+    constexpr ptrdiff_t _Threshold = 16; // vectorization is likely to be a win after this size
```
StephanTLavavej, line 3362: Question: Is this intentionally an element threshold (versus bytes)?

StephanTLavavej (changes requested):

stl/inc/algorithm
AlexGuteniev: I observe that for vectored implementation the constant cost part is big, and per loop iteration cost is small. Which is kind of expected. We compare that with scalar where per-element cost is big. So it is elements.

stl/inc/algorithm
```diff
@@ -3321,6 +3358,26 @@ _NODISCARD _CONSTEXPR20 _FwdIt1 find_first_of(
     const auto _ULast1  = _STD _Get_unwrapped(_Last1);
     const auto _UFirst2 = _STD _Get_unwrapped(_First2);
     const auto _ULast2  = _STD _Get_unwrapped(_Last2);
+#if _USE_STD_VECTOR_ALGORITHMS
+    constexpr ptrdiff_t _Threshold = 16; // vectorization is likely to be a win after this size
+
```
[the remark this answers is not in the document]
AlexGuteniev: Will update here.
I'm was just afraid you wouldn't even consider this, so went with minimum version on the first take 😄

stl/inc/algorithm
AlexGuteniev: This vector algorithm has unusually high constant cost. I suspect, based on what I observe in profiler, that `memcpy` is expensive here, which makes sense, because it should either branch on sizes, or have per-byre loop. Since we don't do reading beyond the range tricks (anymore), I don't see any faster `memcpy` alternatives.

The loop iteration is only moderately expensive, it has one unusually slow instruction (which is still very fast given how many comparisons it replaces), but the loop path consists of only that instruction (repeated intrinsic with other suffix letter queries different output of the already executed instruction) and pointer increment/end checking.

StephanTLavavej: Thanks, good to know! :bulb:

stl/src/vector_algorithms.cpp
AlexGuteniev: It is only used twice in other `constexpr` below, so we can skip naming it

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for vectorizing more algorithms! :rocket: :fast_forward: :tada:

PR #4458, opened 2024-03-08
Delete `<cvt/meow>`

StephanTLavavej:
Fixes #282.

`<cvt/meow>` was deprecated by #3924 in VS 2022 17.8, so we'll have shipped two long-term-support releases (17.8 and 17.10) with that deprecation. @MahmoudGSaleh has confirmed that we can go ahead and remove these headers in 17.11. These headers were never documented.

In VS, this will delete 78 files, 4.8 MB, which is the major benefit that users will experience. For us, it reduces our test costs and maintenance surface area.

* Delete `stl/inc/cvt` (78 files, 5,061,936 bytes).
* `stl/CMakeLists.txt`: Drop all `${CMAKE_CURRENT_LIST_DIR}/inc/cvt/` lines.
* `yvals_core.h`: `_DEPRECATE_STDEXT_CVT` is now unused; replace it with a tombstone for STL4044.
* Delete `tests/tr1/tests/cvt` (76 files, 32,051 bytes).
* Delete `tests/tr1/include/cvt_xtest.h` (1 file, 4,990 bytes).
* `tests/tr1/test.lst`: Drop all `tests\cvt\` lines.
* `tests/std/tests`: Delete `Dev11_0453373_codecvt_compiles` and `VSO_0000000_instantiate_cvt` (4 files, 7,866 bytes).
* `tests/std/test.lst`: Drop `Dev11_0453373_codecvt_compiles` and `VSO_0000000_instantiate_cvt`.
* `P0218R1_filesystem`: Drop/replace test coverage using `stdext::cvt` facets.
  + There are no replacements for `stdext::cvt::codecvt_cp1251` and `stdext::cvt::codecvt_sjis`.
  + We can replace `stdext::cvt::codecvt_utf8_utf16` with the deprecated `std::codecvt_utf8_utf16`.
  + This was the last use of `_SILENCE_STDEXT_CVT_DEPRECATION_WARNING`.

:warning: Note to self: this will require a slightly larger-than-usual amount of MSVC-internal work to remove the `cvt` subdirectory from the build and setup authoring.

PR #4470, opened 2024-03-12
Delete `std::experimental::erase/erase_if`

StephanTLavavej:
My `std::experimental::erase/erase_if` were deprecated by #236 implementing `std::erase/erase_if`, merged on 2019-11-01. This [shipped](https://learn.microsoft.com/en-us/visualstudio/releases/2019/release-notes-v16.5) on 2020-03-16 as part of VS 2019 16.5. Users have had *4 years* of deprecation notice, and this machinery was always marked as experimental. It's time to follow through on our promise. This slightly reduces our surface area (making our repo and VS installations slightly smaller, avoiding the slight navigation headache of having both `vector` and `experimental/vector`), but my real motivation is modules - only `std::erase/erase_if` are available through the Standard Library Modules, so we should encourage users to write Standard code. (While `std::erase/erase_if` are C++20, any users who eagerly reached for experimental TS machinery should be expected to have eagerly upgraded their Standard mode. Users can also write simple fallbacks if upgrading their Standard mode is somehow impractical, described below.)

Aside: The equally MS-deprecated `<experimental/filesystem>`, `<hash_meow>`, and `stdext::checked_array_iterator` were much more widely used (unlike `<cvt/meow>`, which was never documented and which I believe was virtually unused outside of our test suite, hence #4458 removing it), and they involve *types*, not just functions like `erase/erase_if`. (Accordingly, I don't think it's feasible to remove those types until vNext.) While the experimental Uniform Container Erasure functions probably got a bit of use, removing these functions is a non-event for ABI, and migration is very simple to deal with. (Either upgrade to C++20 and use the Standard functions, or fall back to classic `erase-remove[_if]` for `vector`-like containers, member `remove[_if]` for `list`-like containers, or a node removal loop for associative containers.)

Structured commits:

* Delete the experimental headers for `erase/erase_if` (9 files, 11,814 bytes).
* Remove all mentions of the headers.
* Replace the deprecation macro with a tombstone.
* Remove all mentions of the silencing macro.
* Remove the feature-test macro and its test.
* Remove compile-only test coverage of `std::experimental::fundamentals_v2::erase/erase_if`.
  + We're still testing `std::erase/erase_if`.
  + Move `auto pr1` and `vector<bool> vb` within `#if _HAS_CXX20` to avoid any potential unused-variable warnings. (We surely instantiate `vector<bool>` elsewhere, so we're not giving up any test coverage here.)
* Remove test coverage for `std::experimental::erase/erase_if`.
  + This was copy-pasted for `// P1209R0 erase_if(), erase()` below, further enhanced to inspect the return values that were added to the Standard, so we aren't losing any test coverage.
  + The test coverage for `P0458R2 contains()` used the final contents of the ordered/unordered associative containers, so I'm defining them with copy-list-initialization to match the `const` containers below.
* Followup: Fuse `_HAS_CXX20` regions.
* Avoid Clang `-Wunused-parameter` warnings.

:warning: As usual, this will require MSVC-internal changes to remove files.


CaseyCarter (approved):

StephanTLavavej:

PR #4471, opened 2024-03-12
`<generator>`: Various fixes and cleanups

StephanTLavavej:
My apologies for the grab-bag PR (slightly excusable for a feature branch).

This starts with trivial followups to #4464, then addresses pre-existing issues that I found:

* `std::` => `_STD`
* Include `<cstddef>` for `std::nullptr_t`.
* Extract a verbose immediately-invoked lambda for clarity.
* Bugfix: Use `__stdcall` to make `<generator>` immune to default calling conventions.
  + We must always do this when mentioning function pointer types in headers.
* Scalar `constexpr size_t _Align` doesn't need to be `static`.
* Fix unintentional move from `const _Alloc&`.
  + This is now consistent with the other occurrence of `auto& _Stored_al`.
* Consistently use `_RANGES`.
* `_Promise` => `_CoroPromise` to avoid intolerable shadowing of `<future>`.
  + I clearly missed this in the shipping `<coroutine>`. We conventionally never allow template parameters to shadow other machinery like this: https://github.com/microsoft/STL/blob/d6efe9416e4ad7d6e245ae9e96023d413794d1eb/stl/inc/future#L1001-L1002
* `throw;` => `_RERAISE;`
  + We never directly `try/catch/throw` in headers. Using the macros respects `_HAS_EXCEPTIONS=0`.
* Overhaul `generator`'s `static_assert`s.
  + Add Standard citations.
  + "generator's second argument" was bogus; use "generator's selected reference type" to describe italic-monospace-reference. Then say "an actual reference type" for clarity.
  + "an iterator with \[...\] cannot model indirectly_readable" was confusing because it was describing a bad situation, but the other messages are describing how situations must be good. Rephrase this for clarity.
* Mark `_Dealloc_fn` as `_NOEXCEPT_FNPTR`.
  + This is important because the compiler can't see through function pointers.
* Use `static` member functions instead of lambdas to make Clang happy.
  + Clang `-fno-ms-compatibility` refuses to convert stateless lambdas to `__stdcall` function pointers, which I suppose is a defensible choice.

### DIFF
--- a/stl/inc/generator
+++ b/stl/inc/generator
@@ -258,7 +258,7 @@ public:
         if (_Info) {
             _Info->_Except = _STD current_exception();
         } else {
-            throw;
+            _RERAISE;
         }
     }
 
--- a/stl/inc/generator
+++ b/stl/inc/generator
@@ -270,10 +270,10 @@ private:
             return false;
         }
 
-        template <class _Promise>
-        constexpr void await_suspend(coroutine_handle<_Promise> _Handle) noexcept {
+        template <class _CoroPromise>
+        constexpr void await_suspend(coroutine_handle<_CoroPromise> _Handle) noexcept {
 #ifdef __cpp_lib_is_pointer_interconvertible // TRANSITION, LLVM-48860
-            _STL_INTERNAL_STATIC_ASSERT(is_pointer_interconvertible_base_of_v<_Base, _Promise>);
+            _STL_INTERNAL_STATIC_ASSERT(is_pointer_interconvertible_base_of_v<_Base, _CoroPromise>);
 #endif // ^^^ no workaround ^^^
 
             _Base& _Current = _Handle.promise();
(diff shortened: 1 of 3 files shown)


stl/inc/generator
```diff
@@ -409,25 +412,27 @@ private:
 };
 
 _EXPORT_STD template <class _Rty, class _Vty, class _Alloc>
-class generator : public ranges::view_interface<generator<_Rty, _Vty, _Alloc>> {
+class generator : public _RANGES view_interface<generator<_Rty, _Vty, _Alloc>> {
```
AlexGuteniev, line 415: Why use `_RANGES` macro on types? Are they susceptible to ADL?

stl/inc/generator
```diff
@@ -231,20 +231,23 @@ public:
...
 
     template <_RANGES input_range _Rng, class _Alloc>
         requires convertible_to<_RANGES range_reference_t<_Rng>, _Yielded>
     _NODISCARD auto yield_value(_RANGES elements_of<_Rng, _Alloc> _Elem) {
         using _Vty              = _RANGES range_value_t<_Rng>;
         using _Nested_awaitable = _Nested_awaitable_provider<_Yielded, _Vty, _Alloc>::_Awaitable;
-        return _Nested_awaitable{[](allocator_arg_t, _Alloc, _RANGES iterator_t<_Rng> _It,
-                                     const _RANGES sentinel_t<_Rng> _Se) -> generator<_Yielded, _Vty, _Alloc> {
+
+        auto _Lambda = [](allocator_arg_t, _Alloc, _RANGES iterator_t<_Rng> _It,
```
AlexGuteniev, line 243: Maybe make this a function? This lambda does not capture.
(Remember that lambdas contribute to code bloat when not inline, as they have unique real type per TU)

stl/inc/generator
StephanTLavavej: We're in `std`, not `std::ranges` here, so we need to qualify. Qualifying with `_RANGES` consistently makes searching easier.

stl/inc/generator
StephanTLavavej: Yep, I have to do this for Clang (which doesn't permit lambdas to be converted to `__stdcall` function pointers under `-fno-ms-compatibility`).

CaseyCarter (approved):

PR #4392, opened 2024-02-15
Toolset update: VS 2022 17.10 Preview 1

StephanTLavavej:
# Overview
- Code cleanups:
  * Removed compiler bug workarounds.
- Infrastructure improvements:
  * Updated dependencies.
    + Updated build compiler to VS 2022 17.10 Preview 1 (now required).
    + Updated CMake to 3.28 (now required).
    + Updated Python to 3.12.2.

# Commits
* Use PowerShell 7.4.1.
* Python 3.12.2.
* New pool.
* VS 2022 17.10 Preview 1.
* CMake 3.28.
* Require VS 2022 17.10 (aka MSVC 19.40).
* Require CMake 3.28.
* Update citations: DevCom-10543093 (not entirely fixed) => VSO-1956799
  + Reported VSO-1956799 "C1XX `constexpr` ICE when invoking a base PMD on a `reference_wrapper<Derived>`".
* Update citations: LLVM-45398 llvm/llvm-project#44743 (resolved as duplicate) => LLVM-35450 llvm/llvm-project#34798
* Update citations: VSO-1854243 (duplicate of upstream) => google/sanitizers#295
* Update citations: URL => `google/sanitizers#295`
  + To be more searchable.
* Update citations: OS-11107628 (archived) => ArchivedOS-12440914
* Update citations: OS-17090155 (archived) => ArchivedOS-17501595
* Remove UCRT workaround for ArchivedOS-17501595.
  + On 2018-04-20, @amyw-msft noted: "We actually fixed this in RS1"
* Remove Clang workarounds for LLVM-48173 llvm/llvm-project#47517.
* Remove Clang workaround for LLVM-48305 llvm/llvm-project#47649.
* Remove Clang workaround in `test_header_units_and_modules.hpp`.
  + We aren't testing Clang with header units and modules yet, but we require Clang 17, and this aligns with how we changed `<source_location>` in #4259.
* Remove Clang workarounds in `P0009R18_mdspan_mdspan`.
* Remove Clang workaround in `P2374R4_views_cartesian_product_recommended_practices`.
* Remove MSVC FAILs for tests that are passing now.
* Remove `_MSVC_INTERNAL_TESTING` workaround in `P1208R6_source_location`.
* Remove `_MSVC_INTERNAL_TESTING` workarounds in `P2465R3_standard_library_modules`.
  + This activates `#include`-then-`import` coverage.
* Remove MSVC workaround for DevCom-1516290.
* Remove MSVC workaround for DevCom-10095944.
[...]

PR #2791, opened 2022-06-15
Toolset update: VS 2022 17.3 Preview 2, CUDA 11.6.0

StephanTLavavej:
This updates the toolset to:

* VS 2022 17.3 Preview 2.
  + Required in the README, but not in `<yvals_core.h>` because the internal build still uses 19.32.
* CMake 3.23.
  + Required in the README and `CMakeLists.txt`.
* Python 3.10.5.
  + We still require Python 3.9 because the VS Installer still has 3.9.7.
* CUDA 11.6.0.
  + `azure-devops/provision-image.ps1` now installs all CUDA features, which is much easier than enumerating almost all of them.
  + Required in `<yvals_core.h>`.
    - CUDA 11.6.0 was the first release that officially supported VS 2022, so this aligns our support.
    - Approved by our boss and double boss, who are awesome! :smile_cat:

Then we can remove workarounds:

* `tests/std/tests/GH_000639_nvcc_include_all/env.lst`
  + Drop `--allow-unsupported-compiler`, now that we're using a CUDA-supported scenario.
* `tests/std/tests/P1502R1_standard_library_header_units/test.cpp`
  + Remove workarounds for VSO-1496084 "Standard Library Header Units: Splitting `<chrono>` emits fatal error C1116: unrecoverable error importing header unit `'<chrono>'`".
* Remove workarounds for VSO-1269037 "Compiler ICE on reference to pointer in `constexpr`".
  + `<vector>` 2051-2056: This is clearer with the workaround as `_Temp` is used repeatedly, but we can extract it.
  + `<vector>` 3282: This is clearer with the workaround as `_Temp` is used repeatedly.
  + `<xmemory>` 1212: Remove the workaround as `_Temp` is used only once.
  + `<xmemory>` 1233-1236: The original code used pointers-to-pointers, which were difficult to read. The workaround was significantly clearer, but still had subtle behavior - it was writing to `auto& _Pnext` repeatedly, ultimately nulling it out (so the separate assignment was redundant). As @CaseyCarter observed, we can avoid the subtlety by using `exchange()` to null out `_Myproxy->_Myfirstiter` immediately, and then the rest of the loop can use a non-reference variable.
* `<format>`
  + Remove workaround for VSO-1444086 "`consteval` + `/presetPadding` error".
* Remove workarounds for internal test machines running older OSes.
[...]

PR #2714, opened 2022-05-11
Toolset update: VS 2022 17.3 Preview 1

StephanTLavavej:
* This updates the toolset to VS 2022 17.3 Preview 1 (and the May 2022 Patch Tuesday). Effectively the only changes are compiler bugfixes since #2651 updated to 17.2 Preview 3.
  + Also use PowerShell 7.2.3 to prepare the VMSS.
* In `stl/CMakeLists.txt`, we don't need to include `crt/src/concrt` anymore.
  + This became unnecessary after #1194 removed our ConcRT dependency and #1325 eliminated our inclusion of `<concrtinternal.h>` etc.
  + Will be paired with MSVC-internal changes to `src/vctools/crt/crt_build.settings.targets`.
  + This is technically unrelated to the toolset update, even in theory, but I'm already changing the build system below.
* Mention the Windows 11 SDK in the README.
  + The MSVC-internal build already uses the 22000 SDK.
  + Taking this dependency (for building/testing the STL only; consumers of the STL are unaffected as we don't drag in `<Windows.h>` from headers) allows us to remove a couple of long-standing workarounds.
* In `filesystem.cpp`, we can directly use `FILE_DISPOSITION_INFO_EX` now.
  + I verified with `static_assert` that these `struct`s are layout-compatible, and that the constants have identical values.
  + In `CMakeLists.txt`, we need to increase `NTDDI_VERSION` in order to get the declaration of `FileDispositionInfoEx`. Instead of increasing it to the exact value we need, I am going further and increasing it to the maximum supported value right now, `NTDDI_WIN10_CO`, so that we can access other declarations if necessary (see the comment above).
  + Technically we don't need to increase `_WIN32_WINNT` (due to an apparent mistake in the WinSDK headers, the `FILE_DISPOSITION_FLAG_DELETE` etc. macros are effectively unguarded), but I am increasing it to its maximum value of Win10 for consistency.
  + Will be paired with MSVC-internal changes to `src/vctools/crt/crt-common.settings.targets`, setting `Win32WinNt` and `NtddiVersion`.
* Fix GH-854 by removing the winsdk matrix workaround. This simply deletes all of the winsdk matrices (which were burdensome to maintain, and had accumulated a small amount of unintentional divergence) and changes the affected tests to use the corresponding plain matrices.
  + :warning: This is inherently a stealth build break for PRs in flight - we will need to watch out for any that are adding tests with the winsdk matrix.
* Remove `/Za` test coverage. (No product changes.)
[...]

PR #4475, opened 2024-03-13
Toolset update: VS 2022 17.10 Preview 2, WinSDK 22621, CUDA 12.4.0

StephanTLavavej:
# :scroll: Changelog

- Code cleanups:
  * Updated dependencies.
    + Updated build compiler to VS 2022 17.10 Preview 2.
    + Updated to Windows 11 SDK 22621. This is now required for building and testing the STL, but not for using it.
    + Updated CUDA to 12.4.0 (now required).

# :gear: Details

## WinSDK

WinSDK 22621 is now used by the MSVC-internal build (this was MSVC-PR-407828 which flowed into `prod/fe` via MSVC-PR-528863 on 2024-02-15).

In `CMakeLists.txt`, this defines `NTDDI_VERSION=NTDDI_WIN10_NI` to make the latest machinery available, although we don't need any of it yet. #2714 was when we last increased this to the maximum supported value (getting `FILE_DISPOSITION_INFO_EX` along the way). Note: Just because the latest machinery is declared, doesn't mean that we'll start requiring the latest OS - we always have runtime OS checks when attempting to actually use new machinery.

## CUDA

CUDA 12.4.0 is the first version that allows VS 2022 17.10 (`_MSC_VER` `1940`), allowing us to remove `--allow-unsupported-compiler` (recently added by #4392). The last time we increased our requirement was #2791 when CUDA 11.6.0 began supporting VS 2022.

Note that our escape hatch `_ALLOW_COMPILER_AND_STL_VERSION_MISMATCH` remains available if a user can't upgrade their CUDA at the same time as their VS for some reason.

## AV exclusions

I'm dropping the AV exclusions in `provision-image.ps1`. Nobody has hissed at us yet, but I saw MSVC-PR-533609 making similar changes in MSVC-internal scripts, so I think we should do this now (even though our VMs are, of course, completely untrusted, unprivileged, and ephemeral).

## Compiler bug workaround removals

* VSO-1888157 "\[EDG\] failure to compare past-the-end pointer with `nullptr` in `constexpr`", fixed 2023-10-30.
* DevCom-10112408 VSO-1588505 "Implicit conversion to function parameter of a non-movable type is still broken", closed as fixed on 2023-11-06.
* VSO-1538698 "Better handling for non-exported friend function declarations", fixed in 17.10 Preview 2.

# :warning: Note to self:

NON-GitHub change required to define `NTDDI_VERSION` to `NTDDI_WIN10_NI` in `src/vctools/crt/crt-common.settings.targets`.


CaseyCarter (approved):

StephanTLavavej:

PR #4476, opened 2024-03-13
Update GitHub workflows

StephanTLavavej:
* Delete GitHub workflows for the old Code Reviews project.
  + "Assign new PRs" is built-in functionality for new-style projects - except for the draft distinction, which we can live without.
  + The `/pr wip` and `/pr review` commands were rarely used by contributors, and now that the backlog is much smaller, they aren't worth replacing/maintaining. (I believe they broke when MS started requiring permissions and I didn't grant them in #4223.)
* `update-status-chart.yml`: `.ts` => `.mts`
  + I'll need to merge #4460 on the same day as this.
* Update node-version to 21.7.1.
  + Updating this version is always "nice to have" - the infrastructure usually retrieves the latest (and is already retrieving 21.7.1). I just don't want to let it get too outdated, in case some stale version starts getting selected from a cache, significantly diverging from what I'm locally using.

:warning: Note to self:
===
After merging, we should update the repo's settings for Actions Permissions.


CaseyCarter (approved):

StephanTLavavej:

PR #4478, opened 2024-03-14
Fix `atomic_ref<16 bytes>::is_always_lock_free`

### LINKED ISSUE #4472: `<atomic>`: With `_STD_ATOMIC_ALWAYS_USE_CMPXCHG16B` defined to `1`, `atomic_ref<16 bytes>` does not report `is_lock_free` and `is_always_lock_free` correctly
With `_STD_ATOMIC_ALWAYS_USE_CMPXCHG16B` defined to `1` on x64, `atomic_ref<16 bytes>` reports `is_lock_free` and `is_always_lock_free` as both `false`. It should be both `true`.

AlexGuteniev:
Fixes  #4472

* Invent always-lock-free compile-time check independently from `atomic<T>`
  *  Sharing the checks implementation is a bomb for vNext. In vNext I expect `atomic<3 bytes>` to be always-lock-free, but not `atomic_ref<3 bytes>`
  * Having the template constant in global scope is unnecessary, `atomic_ref<T>` doesn't have `volatile` things
* Expand test matrix for that
  * Requires DCAS machine, but it is Win8+ requirement
  * `clang-cl` wants `cx16` feature enabled for that


tests/std/tests/P0019R8_atomic_ref/env.lst
```diff
@@ -2,3 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 RUNALL_INCLUDE ..\usual_20_matrix.lst
+RUNALL_CROSSLIST
+PM_CL=""
+PM_CL="/D_STD_ATOMIC_ALWAYS_USE_CMPXCHG16B=1"
```
StephanTLavavej, line 7: I believe this is missing star-tab to respect ASAN.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:radioactive: :hammer_and_wrench: :heart_eyes_cat:
===

PR #4490, opened 2024-03-19
`vector_algorithms.cpp`: clearer control flow in `bitset::to_string`

AlexGuteniev:
Use `else` instead of early return

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -2237,10 +2237,7 @@ __declspec(noalias) void __stdcall __std_bitset_to_string_1(
         }
 
         _mm256_zeroupper(); // TRANSITION, DevCom-10331414
-        return;
-    }
-
-    if (_Use_sse2()) {
+    } else if (_Use_sse2()) {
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
Is this clearer? I need to think about whether the finality of an early `return` is better than chaining.

AlexGuteniev:
It now stands out so that it is clear that it is only a fallbaxk, and not tail handling

jovibor:
Honestly, it's more of a personal preference than anything.
Also, this construct:
```cpp
    } else
#endif // !defined(_M_ARM64EC)
```
looks clunky. (`else` immediately following by `endif`)

StephanTLavavej:
@AlexGuteniev Hmm, I suppose it's at least equally clear.

@jovibor It looks weird, but we've increasingly used this pattern when we need to interleave preprocessor conditions with `if constexpr` or runtime `if` control flow. I have to imagine both preprocessor situations in sequence when reasoning about such code, but as long as the interleaving isn't too tangled, it ends up being reasonably readable.

StephanTLavavej:

StephanTLavavej:
:broom: :eyeglasses: :sparkles:
===

PR #4492, opened 2024-03-20
libcxx `stol.pass.cpp` is now passing for ASAN

StephanTLavavej:
Followup to the VS 2022 17.10 Preview 2 toolset update #4475.

This toolset fixed VSO-1875597 "\[ASAN\]\[STL\] Interception breaks strtol: TEST 'libc++ std/strings/string.conversions/stol.pass.cpp:1' FAILED". @CaseyCarter explained the bug:

> The implementation of `std::stol` expects `::strtol` to set `errno` to `ERANGE` when given an unrepresentable input, which it does normally, but not with ASAN

We can now un-skip the affected test.

(We missed a few opportunities to discover this during the toolset update; I should have run ASAN, etc. I'll try to improve our development practices in the future.)


StephanTLavavej:

zacklj89 (approved):

AraHaan:
This looks good.

PR #4494, opened 2024-03-20
Mirror internal MSBuild changes again

StephanTLavavej:
This mirrors @d-winsor's internal MSVC-PR-533225 "Update VS Solution file" as of Iteration 19. Like the previous PR #4428, this is limited to `stl/msbuild`, so we can tolerate the temporary divergence of these changes merging into internal `prod/be` (instead of our usual home `prod/fe`).


zacklj89 (approved):
I didn't take a close look at this one since it's already in prod/be, and I personally haven't seen any problems yet 😛

PR #4495, opened 2024-03-21
`mismatch` vectorization

AlexGuteniev:
Not only `mismatch`, also useful in future `lexicographical_compare`

## Tail masking

A novel for `vector_algorithms.cpp` thing here is also using AVX2 mask for tail and not yielding to SSE2.  This allows
 * Not entering SSE2 code path, avoiding its startup cost
 * Reducing maximum tail size, making the algorithms more efficient

Unfortunately, SSE2 doesn't have usable mask instructions, and AVX2 has 4-byte granularity as its finest mask, so still have at most 3 byte tails with AVX2 and up to 15 byte tails with SSE2.

I think other vector algorithms can benefit from this approach too.

If ASan hisses at this, this would be ASan bug, as AVX2 masked loads/stores are well documented and empirically proved to skip reading/writing fairly: no data races created, no exceptions on inaccessible memory, etc.

## Integer indexing

Other vector algorithms use pointer increments instead of index. This allows instructions with memory operand avoid complex indexing, whereas keeping everything else in loops the same. The compiler may perform this optimization itself, but may not, so pointer increment is useful.

With two arrays algorithms, where both arrays are processed in the same direction, the situation is ambiguous. Integer indexing requires one increment, whereas pointer indexing takes two increments, so integer indexing saves an instruction. The compiler may perform an optimization to make one pointer relative to the other, and use that pointer instead of using index, to make some of accesses simple-indexed. Surprising only MSVC does this optimization, clang and gcc do not. This optimization cannot be done manually, as it is UB in C++, and is very squirrelly, so only the compiler can do this.

As the situation on index vs pointer is ambiguous here, and the difference is expected to be small, I just used the simpler approach, that is using index.

## Benchmark

Before:
```
bm<uint8_t>/8/3              4.05 ns         4.08 ns    172307692
bm<uint8_t>/24/22            11.8 ns         12.0 ns     64000000
bm<uint8_t>/105/-1           43.7 ns         43.9 ns     16000000
bm<uint8_t>/4021/3056         989 ns          984 ns       746667
bm<uint16_t>/8/3             4.77 ns         4.76 ns    144516129
bm<uint16_t>/24/22           8.91 ns         9.07 ns     89600000
bm<uint16_t>/105/-1          41.6 ns         41.7 ns     17230769
bm<uint16_t>/4021/3056        983 ns          977 ns       640000
bm<uint32_t>/8/3             4.59 ns         4.55 ns    154482759
bm<uint32_t>/24/22           18.7 ns         18.8 ns     37333333
bm<uint32_t>/105/-1          88.9 ns         87.9 ns      7466667
bm<uint32_t>/4021/3056       2353 ns         2354 ns       298667
bm<uint64_t>/8/3             3.94 ns         4.01 ns    179200000
bm<uint64_t>/24/22           12.0 ns         12.2 ns     64000000
bm<uint64_t>/105/-1          44.5 ns         44.5 ns     15448276
bm<uint64_t>/4021/3056       1020 ns         1025 ns       746667
```

After:
```
bm<uint8_t>/8/3              5.57 ns         5.58 ns    112000000
bm<uint8_t>/24/22            5.56 ns         5.58 ns    112000000
bm<uint8_t>/105/-1           9.80 ns         9.77 ns     64000000
bm<uint8_t>/4021/3056         109 ns          110 ns      6400000
bm<uint16_t>/8/3             4.80 ns         4.76 ns    144516129
bm<uint16_t>/24/22           5.84 ns         5.72 ns    112000000
bm<uint16_t>/105/-1          9.10 ns         9.21 ns     74666667
bm<uint16_t>/4021/3056        126 ns          126 ns      5600000
bm<uint32_t>/8/3             4.55 ns         4.55 ns    154482759
bm<uint32_t>/24/22           5.71 ns         5.78 ns    100000000
bm<uint32_t>/105/-1          17.3 ns         17.3 ns     40727273
bm<uint32_t>/4021/3056        247 ns          246 ns      2800000
bm<uint64_t>/8/3             4.56 ns         4.55 ns    154482759
bm<uint64_t>/24/22           7.35 ns         7.32 ns     89600000
bm<uint64_t>/105/-1          20.1 ns         19.9 ns     34461538
bm<uint64_t>/4021/3056        507 ns          502 ns      1120000

```


AlexGuteniev:
Thanks @frederick-vs-ja for #4138, especially for the coverage

stl/src/vector_algorithms.cpp
```diff
@@ -2009,6 +2027,79 @@ namespace {
...
+
+        if (_Use_avx2()) {
+            const size_t _Count_bytes          = _Count * sizeof(_Ty);
+            const size_t _Count_bytes_avx_full = _Count_bytes & ~size_t{0x1F};
+
+            for (; _Result != _Count_bytes_avx_full; _Result += 0x20) {
+                const __m256i _Elem1 = _mm256_loadu_si256(reinterpret_cast<const __m256i*>(_First1_ch + _Result));
+                const __m256i _Elem2 = _mm256_loadu_si256(reinterpret_cast<const __m256i*>(_First2_ch + _Result));
+                const auto _Bingo = ~static_cast<unsigned int>(_mm256_movemask_epi8(_Traits::_Cmp_avx(_Elem1, _Elem2)));
+                if (_Bingo != 0) {
```
Alcaro, line 2047: If you're expecting this loop to run long, perhaps you should swap this for if (!_mm256_testc_si256(traits::cmp_avx(), _mm256_set1_epi32(-1))), then put the bingo stuff inside the braces.

(If you expect it to run short, that would be a pessimization; the movemask is needed for the last iteration. Make sure to test it.)

StephanTLavavej:
I resolved adjacent-add conflicts in `vector_algorithms.cpp`, and renamed the impl function for consistency.

StephanTLavavej:
Regardless of what the STL-ASan-CI says, after the unsized `ranges::find` problem, I think I'd really like to avoid any read overruns here, even if they "should be" okay, and even if this gives up some performance.

AlexGuteniev:
> azp run STL-ASan-CI
When and how do we know the results

AlexGuteniev:
Also please then hold on with this, if this is of so much concern. I'll add dedicated coverage thet attempts to break ASan and run locally

AlexGuteniev:
> Regardless of what the STL-ASan-CI says, after the unsized `ranges::find` problem, I think I'd really like to avoid any read overruns here, even if they "should be" okay, and even if this gives up some performance.

Here we don't have a read overrun,
I'm just thinking that as ASan is currently far from perfect, it might misinterpret correctly masked read as an overrun.

StephanTLavavej:
> When and how do we know the results?

The way I look them up is to go to the README, click on the STL-ASan-CI status badge, go back one level in the "breadcrumbs" at the top to the STL-ASan-CI component of "vclibs / STL / Pipelines / STL-ASan-CI", which is https://dev.azure.com/vclibs/STL/_build?definitionId=5&_a=summary (so you can bookmark that if you want to be smarter and more efficient than me :joy_cat:), then click on the PR of interest.

x64 passed and x86 looks like it's passing (I had to rerun one timed-out shard just now) so things are looking good so far.

Basically my concern is that if the user gives us pointers from address 0x1729 to 0x2097 or whatever, that we don't read any bytes outside of that range.

stl/src/vector_algorithms.cpp
AlexGuteniev: I'll measure.
Ideally I don't want to introduce vectorization threshold and favor smaller ranges too.

AlexGuteniev: There's no unambiguous winner. The below strange results are consistent.
I suspect some thin effect, like the exact loop alignment.
I'll keep the current version, it is just simpler.

current
```
C:\Project\STL>out\bench\benchmark-mismatch.exe --benchmark_min_time=2s
2024-03-23T07:58:04+02:00
Running out\bench\benchmark-mismatch.exe
Run on (12 X 2208 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 256 KiB (x6)
  L3 Unified 9216 KiB (x1)
bm<uint8_t>/8/3              5.66 ns         5.65 ns    497777778
bm<uint8_t>/24/22            5.62 ns         5.61 ns    437073171
bm<uint8_t>/105/-1           10.6 ns         10.5 ns    259710145
bm<uint8_t>/4021/3056         111 ns          111 ns     25239437
bm<uint16_t>/8/3             4.81 ns         4.81 ns    597333333
bm<uint16_t>/24/22           6.34 ns         6.36 ns    484324324
bm<uint16_t>/105/-1          9.52 ns         9.52 ns    303728814
bm<uint16_t>/4021/3056        129 ns          129 ns     22123457
bm<uint32_t>/8/3             4.73 ns         4.71 ns    597333333
bm<uint32_t>/24/22           5.62 ns         5.62 ns    497777778
bm<uint32_t>/105/-1          13.5 ns         13.6 ns    154482759
bm<uint32_t>/4021/3056        250 ns          251 ns     10541176
bm<uint64_t>/8/3             4.96 ns         4.96 ns    617931034
bm<uint64_t>/24/22           7.83 ns         7.83 ns    381276596
bm<uint64_t>/105/-1          20.2 ns         20.2 ns    137846154
bm<uint64_t>/4021/3056        516 ns          515 ns      5430303
```

vptest
```
C:\Project\STL>out\bench\benchmark-mismatch.exe --benchmark_min_time=2s
2024-03-23T07:54:44+02:00
Running out\bench\benchmark-mismatch.exe
Run on (12 X 2208 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 256 KiB (x6)
  L3 Unified 9216 KiB (x1)
bm<uint8_t>/8/3              5.74 ns         5.73 ns    471578947
bm<uint8_t>/24/22            5.75 ns         5.77 ns    484324324
bm<uint8_t>/105/-1           9.13 ns         9.10 ns    298666667
bm<uint8_t>/4021/3056        75.7 ns         75.8 ns     37333333
bm<uint16_t>/8/3             5.47 ns         5.46 ns    497777778
bm<uint16_t>/24/22           6.73 ns         6.72 ns    381276596
bm<uint16_t>/105/-1          9.93 ns         9.94 ns    284444444
bm<uint16_t>/4021/3056        134 ns          134 ns     21590361
bm<uint32_t>/8/3             4.58 ns         4.60 ns    617931034
bm<uint32_t>/24/22           5.84 ns         5.84 ns    484324324
bm<uint32_t>/105/-1          12.7 ns         12.7 ns    208372093
bm<uint32_t>/4021/3056        242 ns          242 ns     11946667
bm<uint64_t>/8/3             4.83 ns         4.84 ns    578064516
bm<uint64_t>/24/22           7.61 ns         7.60 ns    365714286
bm<uint64_t>/105/-1          20.1 ns         20.0 ns    128000000
bm<uint64_t>/4021/3056        552 ns          553 ns      4715789
```

like current, but with subtraction
```
C:\Project\STL>out\bench\benchmark-mismatch.exe --benchmark_min_time=2s
2024-03-23T08:11:11+02:00
Running out\bench\benchmark-mismatch.exe
Run on (12 X 2208 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 256 KiB (x6)
  L3 Unified 9216 KiB (x1)
bm<uint8_t>/8/3              5.03 ns         5.04 ns    543030303
bm<uint8_t>/24/22            5.01 ns         4.99 ns    560000000
bm<uint8_t>/105/-1           9.33 ns         9.31 ns    303728814
bm<uint8_t>/4021/3056        91.1 ns         91.0 ns     31438596
bm<uint16_t>/8/3             5.20 ns         5.22 ns    560000000
bm<uint16_t>/24/22           6.16 ns         6.15 ns    426666667
bm<uint16_t>/105/-1          9.79 ns         9.75 ns    275692308
bm<uint16_t>/4021/3056        126 ns          127 ns     21590361
bm<uint32_t>/8/3             4.88 ns         4.88 ns    560000000
bm<uint32_t>/24/22           6.08 ns         6.09 ns    459487179
bm<uint32_t>/105/-1          15.4 ns         15.5 ns    182857143
bm<uint32_t>/4021/3056        308 ns          309 ns      8960000
bm<uint64_t>/8/3             5.04 ns         5.05 ns    578064516
bm<uint64_t>/24/22           9.53 ns         9.56 ns    308965517
bm<uint64_t>/105/-1          31.2 ns         31.3 ns     85333333
bm<uint64_t>/4021/3056        818 ns          818 ns      3381132
```

Alcaro: Looks roughly as expected for me. vptest is mostly faster for long-running loops, but for short ones, reusing more instructions between main loop and the found-it branch reduces the constant factor. Also plenty of noise, which is also expected; modern cpu performance is more random than std::mt19937.

Yet another thing you could try is skip the cmpeq. xor elem1/elem2 instead, then _mm256_testz_si256 the result with itself. That saves having to fill a register with -1 in the prologue (or in the actual loop, if the compiler doesn't hoist it).

But that's at the cost of putting the cmpeq in the found-it branch, meaning one more instruction if the target is found early, so it'll probably just amplify the wins and losses you've already seen.

Probably not worth it; your choice if it's even worth checking. The bottleneck is probably memory bandwidth and not the computation, anyways.

AlexGuteniev: `vpxor`+`vptest`
```
bm<uint8_t>/8/3              6.10 ns         6.10 ns    448000000
bm<uint8_t>/24/22            6.22 ns         6.23 ns    426666667
bm<uint8_t>/105/-1           9.04 ns         9.04 ns    293770492
bm<uint8_t>/4021/3056        85.4 ns         85.3 ns     29866667
bm<uint16_t>/8/3             5.43 ns         5.42 ns    484324324
bm<uint16_t>/24/22           6.87 ns         6.86 ns    389565217
bm<uint16_t>/105/-1          9.70 ns         9.68 ns    289032258
bm<uint16_t>/4021/3056        139 ns          137 ns     20363636
bm<uint32_t>/8/3             5.13 ns         5.14 ns    578064516
bm<uint32_t>/24/22           5.92 ns         5.95 ns    459487179
bm<uint32_t>/105/-1          13.7 ns         13.7 ns    205977011
bm<uint32_t>/4021/3056        267 ns          265 ns      9955556
bm<uint64_t>/8/3             4.80 ns         4.80 ns    560000000
bm<uint64_t>/24/22           8.20 ns         8.20 ns    325818182
bm<uint64_t>/105/-1          23.4 ns         23.4 ns    119466667
bm<uint64_t>/4021/3056        665 ns          664 ns      4167442
```
Looks like a worse version of `vpcmpeqb`+`vptest`

AlexGuteniev:
> Also please then hold on with this, if this is of so much concern. I'll add dedicated coverage that attempts to break ASan and run locally

The extra coverage is added, though I didn't run ASan locally.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -376,6 +376,165 @@ void test_min_max_element_special_cases() {
...
+            test_case_mismatch(input_b, input_a);
+            input_b.pop_back();
+        }
+
+        // actual mismatch (or maybe not, depending on random)
+        if (!input_b.empty()) {
+            uniform_int_distribution<size_t> mismatch_dis(0, input_a.size() - 1);
+
+            for (size_t attempts = 0; attempts < mismatchCount; ++attempts) {
+                size_t possible_mismatch_pos   = mismatch_dis(gen);
```
StephanTLavavej, line 429: Nitpick: `possible_mismatch_pos` could be `const`.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -376,6 +376,165 @@ void test_min_max_element_special_cases() {
...
+
+    template <class T, size_t Size, size_t PadSize>
+    char stack_array_various_alignments_impl() {
+        with_pad<T, Size + 1, PadSize + 1> a = {};
+        with_pad<T, Size + 1, PadSize + 1> b = {};
+        assert(mismatch(begin(a.v), end(a.v), begin(b.v), end(b.v)) == make_pair(end(a.v), end(b.v)));
+        return 0;
+    }
+
+    template <class T, size_t Size, size_t... PadsSizes>
```
StephanTLavavej, line 465: Should just be `PadSizes`.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -376,6 +376,165 @@ void test_min_max_element_special_cases() {
...
+            free(a);
+            free(b);
+        }
+
+        // subarray from stack array. We would have wrong result if run out of the range.
+        T a[range + 1] = {};
+        T b[range + 1] = {};
+        for (size_t i = 0; i != range; ++i) {
+            a[i + 1] = 1;
+            // whole range mistmatch finds mismatch after past-the-end
```
StephanTLavavej, line 512: I'll fix comment typos. Also, this comment is a bit confusing because it's talking about "after past-the-end" of the subarray, not the whole array.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -376,6 +376,165 @@ void test_min_max_element_special_cases() {
...
+            assert(mismatch(a, a + range + 1, b, b + range + 1) == make_pair(a + i + 1, b + i + 1));
+            // limited range mismatch gets to past-the-end
+            assert(mismatch(a, a + i, b, b + i) == make_pair(a + i, b + i));
+            a[i + 1] = 0;
+        }
+    }
+} // namespace test_mismatch_sizes_and_alignments
+
+template <class C1, class C2>
+void test_mismatch_containers() {
```
StephanTLavavej, line 522: Nitpick: This isn't being defined in the order that it's being called. Following the same order makes it easier to ensure that everything is actually being called.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for improving the performance of this useful algorithm! :rocket: :runner: :leopard:

PR #4538, opened 2024-03-28
`vector_algorithms.cpp`: Fix ARM64EC build break

StephanTLavavej:
Followup to #4495.

The usage of `_Avx2_tail_mask_32()` is always guarded by `#ifndef _M_ARM64EC`, and we need to do the same for the definition. (It's still within an unnamed namespace.)

I'm adding a newline to make the preprocessor structure of the file clearer.

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -10,6 +10,7 @@
 #include <cstdint>
 #include <cstring>
 #include <xtr1common>
+
 #ifndef _M_ARM64EC
 #include <intrin.h>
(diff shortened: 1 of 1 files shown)


stl/src/vector_algorithms.cpp
```diff
@@ -10,6 +10,7 @@
 #include <cstdint>
 #include <cstring>
 #include <xtr1common>
+
```
zacklj89, line 13: This seems like a good convention nonetheless 👍

zacklj89 (approved):

PR #4550, opened 2024-04-01
Manually vectorize for at least SSE4.2

### LINKED ISSUE #4536: `vector_algorithms.cpp`: Remove the distinction between SSE2 and SSE4.2
During code review, I've prevented two bugs where usage of post-SSE2 instructions was being incorrectly guarded with `_Use_sse2()` - see https://github.com/microsoft/STL/pull/4384#discussion_r1485643686 and https://github.com/microsoft/STL/pull/4495#discussion_r1541843401. This is extremely hazardous, and the correctness of the STL shouldn't depend on whether I've had 270 mg of caffeine every single time I've reviewed a vectorization PR.

At this time, we still need to support the tiny fraction (~0.7%, I've heard) of processors that have SSE2 but not SSE4.2. However, we don't need to extend novel optimizations to them - they were perfectly happy running classic STL algorithms up to 2019.

We should prevent this class of mistakes by removing the distinction between SSE2 and SSE4.2 in `vector_algorithms.cpp`. That is, we should test for the presence of SSE4.2 only, before attempting to use anything up to and including SSE4.2. (This will supersede the error-prone `_Traits::_Sse_available()`.)

We'll still need a distinction between "SSE4.2 is available" and "AVX2 is available", but I consider this to be much less dangerous, because AVX/AVX2 intrinsics and types are very distinctive.

AlexGuteniev:
Resolves #4536

Also checked and improved where can take advantage of new assumptions:
 * `_mm_set1_epi8` replaced with `_mm_shuffle_epi8` with zero input, slight improvement for 1-byte elements `find` and `count`
 * Simplified and improved `__std_bitset_to_string_*`
   * Used `_mm_shuffle_epi8` to populate a `__m128i` variable instead of the sequence
   * Used `_mm_blendv_epi8` to avoid `xor`

Didn't measure the results. If measured on a modern machine, that would be artificial to alter `__isa_enabled`, and still not relevant.

Also removed SSE level of instructions from comments. Kept `__popcnt` comment though.

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -30,14 +30,6 @@ namespace {
         return __isa_enabled & (1 << __ISA_AVAILABLE_SSE42);
     }
 
-    bool _Use_sse2() noexcept {
-#ifdef _M_IX86
-        return __isa_enabled & (1 << __ISA_AVAILABLE_SSE2);
-#else
-        return true;
-#endif
-    }
-
     struct [[nodiscard]] _Zeroupper_on_exit { // TRANSITION, DevCom-10331414
         _Zeroupper_on_exit() = default;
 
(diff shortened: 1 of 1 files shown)


StephanTLavavej (approved):

StephanTLavavej:
Perfect! :heart_eyes_cat:

StephanTLavavej:

StephanTLavavej:
Thanks for dramatically reducing our risk here, and slightly improving perf for modern processors! :heart_eyes_cat: :tada: :rocket:

OlafvdSpek:
> Thanks for dramatically reducing our risk here, and slightly improving perf for modern processors! 😻 🎉 🚀

What risk does this reduce?

PR #4587, opened 2024-04-13
Vectorize `find_first_of` for 4 and 8 byte elements

AlexGuteniev:
The specialized approach shuffles the needle so that every element is on every place and compares it with the haystack part. Then `or`s all comparison results to find the lowest index match, then if there's a match returns it.

The generic approach finds a haystack element in the needle, it is like `find` with reversed haystack/needle, but instead of `vpmovmskb` and checking that index and computing offset, we have `vptest`, as we don't need the index.

Looks like we will have way better results if specialized approach is generalized,, instead of having different and simpler generic approach. But the complexity would grow.

If only the generic approach is used, the complexity would be reduced, but the performance for small needles would be worse, the generic approach for small needles and 64-bit elements even loses to the scalar implementation.

## Benchmark results

Before:
```
---------------------------------------------------------------
Benchmark                     Time             CPU   Iterations
---------------------------------------------------------------
bm<uint32_t>/2/3           2.84 ns         2.19 ns    527058824
bm<uint32_t>/7/4           13.5 ns         11.4 ns    112000000
bm<uint32_t>/9/3           13.3 ns         8.06 ns    128000000
bm<uint32_t>/22/5          48.9 ns         39.1 ns     33185185
bm<uint32_t>/58/2          58.8 ns         41.4 ns     24888889
bm<uint32_t>/102/4          135 ns          105 ns     10000000
bm<uint32_t>/325/1         15.4 ns         12.7 ns    112000000
bm<uint32_t>/1011/11       3720 ns         3125 ns       640000
bm<uint32_t>/3056/7        6076 ns         4688 ns       280000
bm<uint64_t>/2/3           2.78 ns         2.15 ns    560000000
bm<uint64_t>/7/4           13.5 ns         10.8 ns    100000000
bm<uint64_t>/9/3           13.6 ns         10.3 ns    100000000
bm<uint64_t>/22/5          49.4 ns         31.2 ns     32000000
bm<uint64_t>/58/2          59.7 ns         45.8 ns     35840000
bm<uint64_t>/102/4          132 ns         89.1 ns     12800000
bm<uint64_t>/325/1         42.5 ns         32.2 ns     40727273
bm<uint64_t>/1011/11       3611 ns         2846 ns       560000
bm<uint64_t>/3056/7        6175 ns         2084 ns       757262
```

After:
```
---------------------------------------------------------------
Benchmark                     Time             CPU   Iterations
---------------------------------------------------------------
bm<uint32_t>/2/3           2.99 ns         1.90 ns    560000000
bm<uint32_t>/7/4           14.1 ns         6.58 ns    182857143
bm<uint32_t>/9/3           4.34 ns         1.67 ns   1000000000
bm<uint32_t>/22/5          7.78 ns         3.92 ns    358400000
bm<uint32_t>/58/2          4.80 ns         2.49 ns    670690059
bm<uint32_t>/102/4         12.0 ns         5.01 ns    321126400
bm<uint32_t>/325/1         15.7 ns         7.08 ns    218536585
bm<uint32_t>/1011/11       1002 ns          507 ns      2560000
bm<uint32_t>/3056/7         522 ns          455 ns      2986667
bm<uint64_t>/2/3           2.84 ns         1.48 ns   1000000000
bm<uint64_t>/7/4           13.8 ns         6.44 ns    344615385
bm<uint64_t>/9/3           4.93 ns         2.86 ns    995555556
bm<uint64_t>/22/5          26.5 ns         13.7 ns    136533334
bm<uint64_t>/58/2          10.1 ns         4.94 ns    344615385
bm<uint64_t>/102/4         20.8 ns         10.8 ns    100000000
bm<uint64_t>/325/1         38.0 ns         17.4 ns     74666667
bm<uint64_t>/1011/11       1313 ns         1094 ns      1000000
bm<uint64_t>/3056/7        3186 ns         1496 ns       814545
```

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -42,9 +42,10 @@ namespace {
     };
 
     __m256i _Avx2_tail_mask_32(const size_t _Count_in_dwords) noexcept {
-        // _Count_in_dwords must be within [1, 7].
-        static constexpr unsigned int _Tail_masks[14] = {~0u, ~0u, ~0u, ~0u, ~0u, ~0u, ~0u, 0, 0, 0, 0, 0, 0, 0};
-        return _mm256_loadu_si256(reinterpret_cast<const __m256i*>(_Tail_masks + (7 - _Count_in_dwords)));
+        // _Count_in_dwords must be within [0, 8].
+        static constexpr unsigned int _Tail_masks[16] = {
+            ~0u, ~0u, ~0u, ~0u, ~0u, ~0u, ~0u, ~0u, 0, 0, 0, 0, 0, 0, 0, 0};
+        return _mm256_loadu_si256(reinterpret_cast<const __m256i*>(_Tail_masks + (8 - _Count_in_dwords)));
     }
 } // namespace
 #endif // !defined(_M_ARM64EC)
@@ -2038,7 +2039,26 @@ namespace {
     }
 
     template <class _Ty>
-    const void* __stdcall __std_find_first_of_trivial_impl(
+    const void* __stdcall __std_find_first_of_trivial_fallback(
+        const void* _First1, const void* const _Last1, const void* const _First2, const void* const _Last2) {
+        auto _Ptr_haystack           = static_cast<const _Ty*>(_First1);
+        const auto _Ptr_haystack_end = static_cast<const _Ty*>(_Last1);
+        const auto _Ptr_needle       = static_cast<const _Ty*>(_First2);
+        const auto _Ptr_needle_end   = static_cast<const _Ty*>(_Last2);
+
+        for (; _Ptr_haystack != _Ptr_haystack_end; ++_Ptr_haystack) {
+            for (auto _Ptr = _Ptr_needle; _Ptr != _Ptr_needle_end; ++_Ptr) {
+                if (*_Ptr_haystack == *_Ptr) {
+                    return _Ptr_haystack;
+                }
+            }
+        }
+
+        return _Ptr_haystack;
+    }
+
+    template <class _Ty>
+    const void* __stdcall __std_find_first_of_trivial_pcmpestri_impl(
         const void* _First1, const void* const _Last1, const void* const _First2, const void* const _Last2) noexcept {
 #ifndef _M_ARM64EC
         if (_Use_sse42()) {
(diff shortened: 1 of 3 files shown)


AlexGuteniev:
> Looks like we will have way better results if specialized approach is generated, instead of having different and simpler generic approach. But the complexity would grow.

On the other hand, despite having more code, it would be an uniform approach.

stl/src/vector_algorithms.cpp
```diff
@@ -2037,161 +2038,375 @@ namespace {
...
+
+                const void* _Stop2 = _First2;
+                _Advance_bytes(_Stop2, _Needle_length & ~size_t{0x1F});
 
+                for (auto _Ptr1 = static_cast<const _Ty*>(_First1); _Ptr1 != _Last1; ++_Ptr1) {
+                    const auto _Data1 = _Traits::_Set_avx(*_Ptr1);
+                    for (auto _Ptr2 = _First2; _Ptr2 != _Stop2; _Advance_bytes(_Ptr2, 32)) {
+                        const __m256i _Data2 = _mm256_loadu_si256(static_cast<const __m256i*>(_Ptr2));
+                        const __m256i _Eq    = _Traits::_Cmp_avx(_Data1, _Data2);
+                        if (!_mm256_testz_si256(_Eq, _Eq)) {
```
StephanTLavavej, line 2390: `_mm256_testz_si256` returns `int`, so this is directly negating an integer, which we conventionally avoid. Occurs below.

StephanTLavavej (approved):

stl/src/vector_algorithms.cpp
Alcaro: That function is documented to only return 0 and 1, so personally I'd say it should be treated as a boolean. It's probably typed as int because there is no bool type in the C89 language.

But it's your repo, your rules.

StephanTLavavej: After looking at this again, I agree with you - the direct testing is clearer to understand (avoids getting confused with layers of negation). I'll revert this commit. Thanks!

StephanTLavavej (approved):

stl/src/vector_algorithms.cpp
```diff
@@ -2037,161 +2038,375 @@ namespace {
...
+                        return _Ptr_haystack;
+                    }
+                }
+            }
+
+            return _Ptr_haystack;
+        }
+
+        template <class _Ty>
+        const void* __stdcall __pcmpestri_impl(const void* _First1, const void* const _Last1, const void* const _First2,
```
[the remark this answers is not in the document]
AlexGuteniev: I'm wondering about namespace name then. Should it be `_Ugly` too?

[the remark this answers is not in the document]
StephanTLavavej: I'm neutral on that. It contains functions that want to exactly match the `extern "C"` interface like the fallback, so there's an argument to share the same prefix.

StephanTLavavej:

StephanTLavavej:
Thanks for *finding* so many ways to improve performance! :detective: :bulb: :rocket:

PR #4605, opened 2024-04-19
Emit 'MS extension' deprecation warnings in all Standard modes

StephanTLavavej:
There are three Microsoft extensions that we've deprecated, with the intent of eventually removing them:

* VS 2022 17.8: #3818
  + Patched by #3924
* VS 2022 17.9: #4006
* VS 2022 17.10: #4284

Currently, these deprecations are activated by C++17 mode, but there's no real justification for that. It was *practically* convenient, because leaving C++14 mode unaffected limited the source-breaking impact on legacy codebases that we immediately had to deal with. However, if we really want to remove this machinery, we need to warn C++14 users. Fortunately, some ecosystem cleanup has already occurred (e.g. Qt stopped using `stdext::checked_array_iterator` and backported that to their various release branches) since usage of C++17 and later is somewhat popular. Deprecating and removing things always requires effort, but by "fractionating the dose" we can spread out the impact over releases. This is the next step.

This PR changes these MS deprecations to be unconditionally emitted, but still with their fine-grained and coarse-grained `_SILENCE_ALL_MS_EXT_DEPRECATION_WARNINGS` escape hatches.

### DIFF
--- a/stl/inc/yvals_core.h
+++ b/stl/inc/yvals_core.h
@@ -1470,8 +1470,7 @@ _EMIT_STL_ERROR(STL1004, "C++98 unexpected() is incompatible with C++23 unexpect
 #define _CXX23_DEPRECATE_DENORM
 #endif // ^^^ warning disabled ^^^
 
-#if _HAS_CXX17 && !defined(_SILENCE_STDEXT_ARR_ITERS_DEPRECATION_WARNING) \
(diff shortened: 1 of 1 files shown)


CaseyCarter (approved):

StephanTLavavej:

PR #4607, opened 2024-04-19
Remove `lower_bound()`/`upper_bound()` from `hash_meow`/`unordered_meow`

StephanTLavavej:
These member functions were unconditionally deprecated by MSVC-PR-181971 on 2019-05-23.

It's been *almost 5 years*, so it's time to remove this machinery now. Note that there is no loss of functionality - as the deprecation warnings advised, any calls can be exactly converted into `find()` or `equal_range()`.

We can remove the tr1 test coverage because we already have exactly equivalent coverage of `find()` and `equal_range()`.

### DIFF
--- a/stl/inc/xhash
+++ b/stl/inc/xhash
@@ -1247,36 +1247,6 @@ public:
         }
     }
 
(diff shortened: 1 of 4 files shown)


CaseyCarter (approved):

StephanTLavavej:

PR #4608, opened 2024-04-19
`<ranges>`: Refactor `ranges::to`

### LINKED ISSUE #4562: `<ranges>`: Poor error message when `ranges::to` cannot find a way to create the result
Currently, if `ranges::to` cannot create the result, the error message just says "the program is ill-formed per N4950 [range.utility.conv.to]/2.3".
https://github.com/microsoft/STL/blob/be81252ed1f5e5fc6d77faca0b5dbbbdae8143a2/stl/inc/ranges#L10377
This doesn't help an average user to understand the problem.

It should be possible to give a more detailed message. What I have in mind:
```c++
if constexpr (_Ref_converts<_Rng, _Container>) {
    static_assert(false, "cannot find a way to create the result: "
                         "neither constructor call nor insertion at the end works");
} else {
    static_assert(
        false, "elements of the destination container must be "
               "implicitly convertible from elements of the source range");
}
```

cpplearner:
- Extract the uses of `_Ref_converts` from the helper concepts.
  - This matches the structure of [\[range.utility.conv.to\]/2](http://eel.is/c++draft/range.utility.conv.to#2).
- Provide more detailed error messages.

Closes #4562

### DIFF
--- a/stl/inc/ranges
+++ b/stl/inc/ranges
@@ -10297,17 +10297,8 @@ namespace ranges {
         !input_range<_Container> || convertible_to<range_reference_t<_Rng>, range_value_t<_Container>>;
 
     template <class _Rng, class _Container, class... _Types>
-    concept _Converts_direct_constructible =
-        _Ref_converts<_Rng, _Container> && constructible_from<_Container, _Rng, _Types...>;
-
-    template <class _Rng, class _Container, class... _Types>
-    concept _Converts_tag_constructible = _Ref_converts<_Rng, _Container>
-                                       // per LWG issue unnumbered as of 2022-08-08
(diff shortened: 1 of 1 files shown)


StephanTLavavej (approved):

StephanTLavavej:
This is awesome, thank you! :heart_eyes_cat: I had earlier noticed how a significant mental transformation was needed to match the previous implementation to the Standardese, and your restructuring is much easier to follow. Your error messages are also excellent, they describe the Standardese in an understandable manner.

StephanTLavavej:

StephanTLavavej:
Thanks again for this `ranges::to` fast, `ranges::to` furious improvement! :car: :joy_cat: :tada:

PR #4618, opened 2024-04-22
`<random>`: Skip reset in `normal_distribution::operator()` with custom params

horenmar:
`normal_distribution` works by computing a pair of coefficients that are multiplied with the stddev to compute how far (and in which direction) the result is from the mean. These coefficients are not reliant on the actual stddev (or mean), so there is no reason to avoid reusing one computed during last invocation, even if it was with different params.

AFAIK usage of this overload is rare (personally I think this overload should've never been standardized), the old code still needlessly penalized it, in the worst case halving the performance.


tests/std/tests/GH_004618_mixed_operator_usage_keeps_statistical_properties/test.cpp
```diff
@@ -0,0 +1,57 @@
...
+static const size_t N = 1000000;
+
+void check_results(std::vector<double> const& generated, double expected_mean, double expected_variance) {
+    const double actual_mean = std::accumulate(generated.begin(), generated.end(), 0.0) / generated.size();
+    assert(std::abs((actual_mean - expected_mean) / expected_mean) < 0.01);
+
+    double actual_variance = 0;
+    double actual_skew     = 0;
+    for (auto value : generated) {
+        auto deviation = value - actual_mean;
```
StephanTLavavej, line 21: Nitpicks: We should iterate in-place (`const auto&`) and we can mark `deviation` as `const` too.

tests/std/tests/GH_004618_mixed_operator_usage_keeps_statistical_properties/test.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <cmath>
+#include <cstddef>
+#include <numeric>
+#include <random>
+#include <vector>
+
+
+static const size_t N = 1000000;
+
+void check_results(std::vector<double> const& generated, double expected_mean, double expected_variance) {
```
StephanTLavavej, line 14: `expected_mean` and `expected_variance` could be `const`. We don't always rigorously follow this practice (especially in test code), but it's preferable especially in longer functions because it prevents mistakes and draws attention to parameters that actually are modified.

tests/std/tests/GH_004618_mixed_operator_usage_keeps_statistical_properties/test.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+int main() {
+    // The basic idea is to generate values from two different distributions,
+    // one defined through the type's constructor, one through the params
+    // overload of operator(). The generated values are then checked for
+    // their expected statistical properties.
+    std::mt19937 rng;
+    std::normal_distribution<> dist(5., 4.);
+    using dist_params = std::normal_distribution<>::param_type;
+    dist_params params(50., 0.5);
```
StephanTLavavej, line 40: `params` could also be `const`.

StephanTLavavej (approved):

StephanTLavavej:
Thanks again for improving performance and simplifying code! :rocket: :sparkles: :heart_eyes_cat:

PR #4619, opened 2024-04-23
`<generator>`: An attempt to merge `_Top` and `_Info`

JMazurkiewicz:
Merge `promise_type::_Top` and `promise_type::_Info` into single `promise_type::_Data` member of type `uintptr_t`.

### DIFF
--- a/stl/inc/generator
+++ b/stl/inc/generator
@@ -311,14 +311,13 @@ private:
             _STL_INTERNAL_STATIC_ASSERT(is_pointer_interconvertible_base_of_v<_Base, _CoroPromise>);
 #endif // ^^^ no workaround ^^^
 
-            _Nest_info* _Info = _Handle.promise()._Try_get_nest_info();
-            if (!_Info) {
-                return _STD noop_coroutine();
+            if (_Nest_info* _Info = _Handle.promise()._Try_get_nest_info()) {
+                coroutine_handle<_Base> _Cont = _Info->_Parent;
+                _Info->_Root.promise()._Set_top(_Cont);
+                return _Cont;
             }
 
-            coroutine_handle<_Base> _Cont = _Info->_Parent;
-            _Info->_Root.promise()._Set_top(_Cont);
-            return _Cont;
+            return _STD noop_coroutine();
         }
 
         void await_resume() noexcept {}
--- a/stl/inc/generator
+++ b/stl/inc/generator
@@ -388,8 +388,8 @@ private:
     friend struct _Gen_iter_provider;
 
     // Least significant bit of `_Data` indicates stored information:
-    // 0 - `_Data` is a top coroutine handle,
-    // 1 - `_Data ^ 1U` is a pointer to an object of type `_Nest_info`.
+    // LSB 0: `_Data` is a top coroutine handle,
+    // LSB 1: `_Data ^ 1U` is a pointer to an object of type `_Nest_info`.
     uintptr_t _Data              = reinterpret_cast<uintptr_t>(coroutine_handle<_Base>::from_promise(*this).address());
     add_pointer_t<_Yielded> _Ptr = nullptr;
 };
--- a/stl/inc/generator
+++ b/stl/inc/generator
@@ -387,7 +387,7 @@ private:
     template <class, class>
     friend struct _Gen_iter_provider;
 
-    // First bit of `_Data` indicates stored information:
+    // Least significant bit of `_Data` indicates stored information:
     // 0 - `_Data` is a top coroutine handle,
     // 1 - `_Data ^ 1U` is a pointer to an object of type `_Nest_info`.
     uintptr_t _Data              = reinterpret_cast<uintptr_t>(coroutine_handle<_Base>::from_promise(*this).address());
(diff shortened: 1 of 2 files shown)


JMazurkiewicz:
> For my own education, I'd like to confirm - the size of `promise_type` is important enough to be worth this additional complexity?

It's debatable, that's why I called this PR `An attempt`.

My opinion is yes - it is important. Since promise object is dynamically allocated, smaller size can be useful when users supply their own allocator and there is not a lot of memory to give. At first it seems that saving 4/8 bytes is not important, but when you think about recursive case, then we can gain much more. Also, the additional complexity is negligible in non-recursive case - for example, a call to `_Get_top` should be optimized away in `/O1` build (I'm going to verify it).

Maybe we should mark this PR as `decision needed`?

StephanTLavavej:
Being dynamically allocated is compelling enough to me, thanks!

stl/inc/generator
```diff
@@ -365,13 +363,35 @@ private:
...
+    void _Set_top(coroutine_handle<_Base> _Top) noexcept {
+        _Data = reinterpret_cast<uintptr_t>(_Top.address());
+    }
+
     template <class, class>
     friend struct _Gen_iter_provider;
 
-    // _Top and _Info are mutually exclusive, and could potentially be merged.
-    coroutine_handle<_Base> _Top = coroutine_handle<_Base>::from_promise(*this);
+    // First bit of `_Data` indicates stored information:
```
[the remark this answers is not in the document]
JMazurkiewicz: I agree, this should be changed later. For now, it's not worth resetting CI.

stl/inc/generator
```diff
@@ -365,13 +363,35 @@ private:
...
+    // 0 - `_Data` is a top coroutine handle,
+    // 1 - `_Data ^ 1U` is a pointer to an object of type `_Nest_info`.
```
StephanTLavavej, line 392: Really nitpicky, but these dashes kind of look like minus operators during a quick read.

stl/inc/generator
```diff
@@ -311,14 +311,13 @@ private:
             _STL_INTERNAL_STATIC_ASSERT(is_pointer_interconvertible_base_of_v<_Base, _CoroPromise>);
 #endif // ^^^ no workaround ^^^
 
-            _Base& _Current = _Handle.promise();
-            if (!_Current._Info) {
+            _Nest_info* _Info = _Handle.promise()._Try_get_nest_info();
+            if (!_Info) {
                 return _STD noop_coroutine();
             }
```
StephanTLavavej, line 317: Style suggestion: This is a minimal transformation from the existing code (a good thing, considered in isolation), but I think a slightly larger change would be even better. The other uses are scoping `_Info` to an `if`-statement, and that would seem to be natural here too.

stl/inc/generator
```diff
@@ -266,7 +266,7 @@ public:
     void return_void() const noexcept {}
 
     void unhandled_exception() {
-        if (_Info) {
+        if (auto _Info = _Try_get_nest_info()) {
```
StephanTLavavej, line 269: We could make this `const auto`, and do that consistently below.

stl/inc/generator
```diff
@@ -311,14 +311,13 @@ private:
...
-            _Base& _Current = _Handle.promise();
-            if (!_Current._Info) {
+            _Nest_info* _Info = _Handle.promise()._Try_get_nest_info();
+            if (!_Info) {
                 return _STD noop_coroutine();
             }
 
-            coroutine_handle<_Base> _Cont        = _Current._Info->_Parent;
-            _Current._Info->_Root.promise()._Top = _Cont;
-            _Current._Info                       = nullptr;
```
StephanTLavavej, line 321: I see that `_Current._Info = nullptr;` (i.e. `_Handle.promise()._Info = nullptr;`) is being dropped, but I don't completely understand why.

The new `_Info->_Root.promise()._Set_top(_Cont);` (i.e. `_Handle.promise()._Try_get_nest_info()->_Root.promise()._Set_top(_Cont);`) seems to be modifying a different promise - are they actually the same?

(Note that I am very uninformed about the data structure relationships here.)

StephanTLavavej (approved):

stl/inc/generator
JMazurkiewicz: > I see that `_Current._Info = nullptr;` (i.e. `_Handle.promise()._Info = nullptr;`) is being dropped, but I don't completely understand why.

We are in final awaiter, so current coroutine is about to be destroyed. Setting `_Current._Info` to `nullptr` is unnecessary.

> The new `_Info->_Root.promise()._Set_top(_Cont);` (i.e. `_Handle.promise()._Try_get_nest_info()->_Root.promise()._Set_top(_Cont);`) seems to be modifying a different promise - are they actually the same?

`_Info->_Root.promise()._Set_top(_Cont);` (used to be `_Current._Info->_Root.promise()._Top = _Cont;`) modifies root promise which in this situation cannot be the same as `_Handle.promise()`.

If `_Handle.promise()` was the same as root, then `_Handle.promise._Try_get_nest_info()` would be equal to `nullptr` (there is no nested generator), and in such case we return `noop_coroutine`.

StephanTLavavej: Got it, thanks! :heart_eyes_cat:

PR #4621, opened 2024-04-23
`<generator>`: Don't use `operator new[]` and `operator delete[]`

JMazurkiewicz:
`_Promise_allocator<void>`'s `operator new(size_t)` is specified to allocate memory with `allocator<_Aligned_block>` (see [[coro.generator.promise]/17](http://eel.is/c++draft/coro.generator#promise-17)), which internally uses `::operator new`, not `::operator new[]` (see [[allocator.members]/5](http://eel.is/c++draft/default.allocator#allocator.members-5)).

### DIFF
--- a/tests/std/tests/P2502R2_generator_promise/test.cpp
+++ b/tests/std/tests/P2502R2_generator_promise/test.cpp
@@ -5,6 +5,7 @@
 #include <coroutine>
 #include <cstddef>
 #include <cstdint>
+#include <cstdlib>
 #include <forward_list>
 #include <generator>
 #include <list>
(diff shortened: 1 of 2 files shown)


tests/std/tests/P2502R2_generator_promise/test.cpp
```diff
@@ -22,6 +22,16 @@
 
 using namespace std;
 
+#pragma warning(disable : 28251) // Inconsistent annotation for 'new[]': this instance has no annotations.
+
+void* operator new[](size_t) {
+    unreachable();
```
AlexGuteniev, line 28: Better `abort()` to crash on reaching rather than UB on reaching. Same below.
(Include `<cstdlib>`)

StephanTLavavej:
The STL-CI pipeline was (again mysteriously) disabled; I've re-enabled it. It sounds like another commit will be pushed, so I haven't manually queued a run here.

StephanTLavavej (approved):

StephanTLavavej:
Great catch! :smiley_cat: I adjusted the PR title due to a mega-nitpick - `operator new[]` is the function being avoided here, whereas the `new` operator is when `new T` appears in source code. (`new T[n]` isn't really using a `new[]` operator, although it can generally be thought of that way.)

PR #4631, opened 2024-04-26
Measure display width in tuple formatter

### LINKED ISSUE #4479: `<format>`: Misbehavior for huge strings
Noticed while reviewing #4438, but affects `main`. I don't see any Standardese allowing us to assume that sizes and widths can fit in an `int`.

```
```
```cpp
#include <cassert>
#include <chrono>
#include <cstddef>
#include <format>
#include <print>
#include <string>
#include <string_view>
#include <tuple>
using namespace std;

static_assert(sizeof(void*) == 8, "This test requires 64-bit.");

void test_string(const size_t n) {
    const auto start = chrono::steady_clock::now();

    const string str(n, 'X');

    {
        const string result = format("{}", str);
        assert(result.size() == n);
        assert(result == str);
    }

    {
        const string result = format("{:o^{}}", str, n + 4);
        assert(result.size() == n + 4);
        assert(result.starts_with("oo"));
        assert(string_view{result}.substr(2, n) == str);
        assert(result.ends_with("oo"));
    }

    const auto finish  = chrono::steady_clock::now();
    const auto elapsed = chrono::duration_cast<chrono::seconds>(finish - start);

    println("test_string({}) passed, {} elapsed.", n, elapsed);
}

#ifdef TEST_TUPLE
void test_tuple(const size_t n) {
    const auto start = chrono::steady_clock::now();

    tuple<string> t;
    get<0>(t).resize(n, 'X');

    {
        const string result = format("{}", t);
        assert(result.size() == n + 4);
        assert(result.starts_with("(\""));
        assert(string_view{result}.substr(2, n) == get<0>(t));
        assert(result.ends_with("\")"));
    }

    {
        const string result = format("{:o^{}}", t, n + 8);
        assert(result.size() == n + 8);
        assert(result.starts_with("oo(\""));
        assert(string_view{result}.substr(4, n) == get<0>(t));
        assert(result.ends_with("\")oo"));
    }

    const auto finish  = chrono::steady_clock::now();
    const auto elapsed = chrono::duration_cast<chrono::seconds>(finish - start);

    println("test_tuple({}) passed, {} elapsed.", n, elapsed);
}
#endif // TEST_TUPLE

int main() {
    test_string(5);
    test_string(50'000'000);
    test_string(500'000'000);
    test_string(2'000'000'000);
    test_string(3'000'000'000);

#ifdef TEST_TUPLE
    test_tuple(5);
    test_tuple(50'000'000);
    test_tuple(500'000'000);
    test_tuple(2'000'000'000);
    test_tuple(3'000'000'000);
#endif // TEST_TUPLE
}
```
```
D:\GitHub\STL\out\x64>cl /EHsc /nologo /W4 /std:c++latest /MT /O2 meow.cpp && meow
meow.cpp
test_string(5) passed, 0s elapsed.
test_string(50000000) passed, 1s elapsed.
test_string(500000000) passed, 14s elapsed.
test_string(2000000000) passed, 56s elapsed.
[***SILENT CRASH***]
```

### LINKED ISSUE #4635: `P2286R8_text_formatting_tuple` fails on non-English locale
```
FAIL: std :: tests/P2286R8_text_formatting_tuple:22 (1448 of 1578)
******************** TEST 'std :: tests/P2286R8_text_formatting_tuple:22' FAILED ********************
Build setup steps:
Build steps:
Build step failed unexpectedly.
Command: "C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Tools\MSVC\14.40.33807\bin\HostX64\x64\cl.EXE" "C:\Project\STL\tests\std\tests\P2286R8_text_formatting_tuple\test.cpp" "-IC:\Project\STL\out\x64\out\inc" "-IC:\Project\STL\llvm-project\libcxx\test\support" "-IC:\Project\STL\tests\std\include" "/nologo" "/Od" "/W4" "/w14061" "/w14242" "/w14265" "/w14582" "/w14583" "/w14587" "/w14588" "/w14749" "/w14841" "/w14842" "/w15038" "/w15214" "/w15215" "/w15216" "/w15217" "/w15262" "/sdl" "/WX" "/D_ENABLE_STL_INTERNAL_CHECK" "/bigobj" "/FIforce_include.hpp" "/w14365" "/w14668" "/w15267" "/D_ENFORCE_FACET_SPECIALIZATIONS=1" "/D_STL_CALL_ABORT_INSTEAD_OF_INVALID_PARAMETER" "/w14640" "/Zc:threadSafeInit-" "/EHsc" "/std:c++latest" "/MTd" "/permissive" "-fsanitize=address" "/Zi" "-FeC:\Project\STL\out\x64\tests\std\tests\P2286R8_text_formatting_tuple\Output\22\P2286R8_text_formatting_tuple.exe" "-link" "-LIBPATH:C:\Project\STL\out\x64\out\lib\amd64" "-LIBPATH:C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Tools\MSVC\14.40.33807\lib\x64" "/MANIFEST:EMBED" "/debug"
Exit Code: 2 (0x2)
Standard Output:
--
test.cpp
C:\Project\STL\tests\std\tests\P2286R8_text_formatting_tuple\test.cpp(161): error C2220: the following warning is treated as an error
C:\Project\STL\tests\std\tests\P2286R8_text_formatting_tuple\test.cpp(161): warning C4566: character represented by universal-character-name '\u00D6' cannot be represented in the current code page (1251)
C:\Project\STL\tests\std\tests\P2286R8_text_formatting_tuple\test.cpp(162): warning C4566: character represented by universal-character-name '\u00D6' cannot be represented in the current code page (1251)
--
```

cpplearner:
Currently, the tuple formatter uses `static_cast<int>(_Tmp_buf.size())` to compute the width of output (in order to determine the number of fill characters to insert). This gives the wrong width if

1. The output contains multibyte Unicode characters, or
1. `_Tmp_buf.size()` is larger than `INT_MAX` (see #4479).

Both problems can be avoided by using [`_Measure_string_prefix`](https://github.com/microsoft/STL/blob/0515a05b394596de92d08cb0f352614479a2a883/stl/inc/format#L3343-L3383), which computes the display width and clamps the result to `INT_MAX` (which means no fill character will be inserted).

Closes #4479

While working on this, I noticed that the tuple formatter test uses `\u00d6` which cannot be represented in some legacy text encodings. This PR adjusted the test to make it encoding-independent.

Closes #4635 

Future work: if no *width* is specified in the format string, there should be no need to use a temporary buffer. The use of `_Tmp_buf` should be skipped in this case.

### DIFF
--- a/stl/inc/format
+++ b/stl/inc/format
@@ -3961,6 +3961,13 @@ _NODISCARD size_t formatted_size(const locale& _Loc, const wformat_string<_Types
 _FMT_P2286_END
 
 #if _HAS_CXX23
+template <class _CharT>
+_NODISCARD int _Measure_display_width(const basic_string_view<_CharT> _Value) {
+    int _Width = -1;
+    (void) _Measure_string_prefix(_Value, _Width);
+    return _Width;
+}
+
 enum class _Fmt_tuple_type : uint8_t { _None, _Key_value, _No_brackets };
 
 template <class _CharT>
@@ -4127,8 +4134,9 @@ protected:
         }(index_sequence_for<_ArgTypes...>{});
         _STD _Copy_unchecked(_Closing_bracket._Unchecked_begin(), _Closing_bracket._Unchecked_end(), _Tmp_ctx.out());
 
-        return _STD _Write_aligned(_Fmt_ctx.out(), static_cast<int>(_Tmp_buf.size()), _Format_specs, _Fmt_align::_Left,
-            [&](typename _FormatContext::iterator _Out) {
+        const int _Width = _Measure_display_width<_CharT>(_Tmp_buf);
+        return _STD _Write_aligned(
+            _Fmt_ctx.out(), _Width, _Format_specs, _Fmt_align::_Left, [&](typename _FormatContext::iterator _Out) {
                 return _STD _Fmt_write(_STD move(_Out), basic_string_view<_CharT>{_Tmp_buf});
             });
     }
--- a/tests/std/tests/P2286R8_text_formatting_escaping_legacy_text_encoding/test.cpp
+++ b/tests/std/tests/P2286R8_text_formatting_escaping_legacy_text_encoding/test.cpp
@@ -17,6 +17,15 @@ void test_escaped_string() {
     assert(format("{:?}", "\x81\x40\x40\x81") == "\"\\u{3000}\x40\\x{81}\"");
 }
 
+template <class TupleOrPair>
+void test_tuple_or_pair_escaping(TupleOrPair&& input) {
+    get<1>(input) = "hell\uff2f"; // U+FF2F FULLWIDTH LATIN CAPITAL LETTER O
+    assert(format("{}", input) == "('*', \"hell\uff2f\")");
+    assert(format("{:#^16}", input) == "('*', \"hell\uff2f\")#");
+}
+
 int main() {
     test_escaped_string();
+    test_tuple_or_pair_escaping(make_pair('*', ""));
+    test_tuple_or_pair_escaping(make_tuple('*', ""));
 }
--- a/tests/std/tests/P2286R8_text_formatting_escaping_utf8/test.cpp
+++ b/tests/std/tests/P2286R8_text_formatting_escaping_utf8/test.cpp
@@ -21,15 +21,32 @@ void test_escaped_string() {
     assert(format("{:?}", "a\u0300") == "\"a\u0300\"");
 }
 
-int main() {
+template <class TupleOrPair>
+void test_tuple_or_pair_escaping(TupleOrPair&& input) {
+    get<1>(input) = "hell\u00d6"; // U+00D6 LATIN CAPITAL LETTER O WITH DIAERESIS
+    assert(format("{}", input) == "('*', \"hell\u00d6\")");
+    assert(format("{:#^16}", input) == "#('*', \"hell\u00d6\")#");
+
+    get<1>(input) = "hell\uff2f"; // U+FF2F FULLWIDTH LATIN CAPITAL LETTER O
+    assert(format("{}", input) == "('*', \"hell\uff2f\")");
+    assert(format("{:#^16}", input) == "('*', \"hell\uff2f\")#");
+}
+
+void run_test() {
     test_escaped_string();
+    test_tuple_or_pair_escaping(make_pair('*', ""));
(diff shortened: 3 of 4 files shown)


fsb4000 (commented):
Your comment is not enough for auto-linking: https://github.com/microsoft/STL/issues/4635#issuecomment-2080603635

Could you add `Closes #4631` to your first message?

stl/inc/format
```diff
@@ -3961,6 +3961,13 @@ _NODISCARD size_t formatted_size(const locale& _Loc, const wformat_string<_Types
 _FMT_P2286_END
 
 #if _HAS_CXX23
+template <class _CharT>
+_NODISCARD int _Measure_display_width(const basic_string_view<_CharT> _Value) {
+    int _Width = -1;
+    (void) _Measure_string_prefix(_Value, _Width);
+    return _Width;
+}
+
```
barcharcraz, line 3970: I'd prefer if this was inlined, it's really a simple wrapper and it only saves one line in each caller, in exchange for the 6-line function definition (not that code-golf is something we optimize for). I just think it doesn't increase clarity to redirect through this vs just calling _Measure_string_prefix directly.

barcharcraz: oh, it's only used once too. Definitely inline this.

barcharcraz (changes requested):

stl/inc/format
cpplearner: I added this wrapper because:

1. The same functionality is needed by `range_formatter` (#4642) and `formatter<std::filesystem::path>`, and might be needed by new formatters in the future.
2. The name `_Measure_string_prefix` is IMO misleading here: it's measuring the entire string, not a (proper) prefix.
3. I'd also prefer to avoid uses of out parameters.

StephanTLavavej: That rationale seems compelling to me (particularly adapting the out-param to a normal return type).

StephanTLavavej (approved):

StephanTLavavej:

stl/inc/format
barcharcraz: I can live with it. I still prefer the function be inlined but if it's used more than once it's more stylistic.

barcharcraz (approved):

StephanTLavavej:
Thanks for fixing this runtime correctness bug in a hopefully-popular new feature! :heart_eyes_cat: :lady_beetle: :hammer_and_wrench:

PR #4646, opened 2024-05-01
`<xutility>`: unify min/max limit

AlexGuteniev:


StephanTLavavej (approved):

StephanTLavavej:
Thanks! I verified that `_Meow_limit` works for all integral types (including `bool`, even though `bool` is handled before the new usage occurs):

```
```
```cpp
#include <print>
#include <string_view>
#include <utility>
using namespace std;

template <typename Type, typename As>
void print_limits_as(const string_view name) {
    println("{}: min {}, max {}", name, static_cast<As>(_Min_limit<Type>()), static_cast<As>(_Max_limit<Type>()));
}

int main() {
    print_limits_as<char, int>("char");
    print_limits_as<wchar_t, int>("wchar_t");
    print_limits_as<char8_t, int>("char8_t");
    print_limits_as<char16_t, int>("char16_t");
    print_limits_as<char32_t, unsigned int>("char32_t");
    print_limits_as<bool, int>("bool");
}
```
```
meow.cpp
char: min -128, max 127
wchar_t: min 0, max 65535
char8_t: min 0, max 255
char16_t: min 0, max 65535
char32_t: min 0, max 4294967295
bool: min 0, max 1
```

PR #4652, opened 2024-05-03
Avoid triggering ASan bug in `Dev09_056375_locale_cleanup`

CaseyCarter:
Avoid triggering VSO-2046190 in `Dev09_056375_locale_cleanup` by leaving the DLL loaded in ASan instrumented statically-linked configs.

This test is failing with 17.10p5 and current dev compilers, I suspect due to a change in the ASan runtime's interception of `memcpy`. Recall that `memcpy` is part of VCRuntime, so statically-linked configs of this test have a copy of `memcpy` linked to the program and another copy of `memcpy` linked into the DLL. When the program unloads the DLL, I believe the ASan runtime's "pointer to actual memcpy" is left pointing into unallocated space where the DLL's copy used to be.

I'm going to disable unloading the DLL for this case to avoid the problem until we can implement and ship a fix.

### DIFF
--- a/tests/std/tests/Dev09_056375_locale_cleanup/test.cpp
+++ b/tests/std/tests/Dev09_056375_locale_cleanup/test.cpp
@@ -65,7 +65,9 @@ void test_dll() {
     TheFuncProc pFunc = reinterpret_cast<TheFuncProc>(GetProcAddress(hLibrary, "DllTest"));
     assert(pFunc != nullptr);
     pFunc();
+#if defined(_DLL) || !defined(__SANITIZE_ADDRESS__) // TRANSITION, VSO-2046190
     FreeLibrary(hLibrary);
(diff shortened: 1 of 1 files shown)


CaseyCarter:
[STL-ASan-CI](https://dev.azure.com/vclibs/STL/_build/results?buildId=16786&view=results) is in progress.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: I updated the PR title to be more meaningful when scanning git history.

zacklj89 (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for scouting ahead and fixing this test for ASan before the next toolset update encountered it! :tada: :heart_eyes_cat: :magic_wand:

PR #4656, opened 2024-05-05
Reduce `mismatch` test run time

AlexGuteniev:
Before this change the mismatch and lex compare test consumed more time than all other tests in vector algorithms test combined.

Done the following:
 * Reused lgk mismatch result in lgk lex compare
 * Reduced the number of random mismatches


StephanTLavavej (approved):

StephanTLavavej:
Thanks! On my local machine, this results in a test speedup of 1.61x, very nice! :heart_eyes_cat:

StephanTLavavej:

StephanTLavavej:
Thanks for noticing and fixing how this test was wasting time! :stopwatch: :hourglass_flowing_sand: :heart_eyes_cat:

Issue #3922, opened 2023-08-04
Should we require SSE2?


StephanTLavavej:
We still support targeting Win7, but before it stopped receiving security updates, it was patched to require SSE2. This was [KB4103718](https://support.microsoft.com/en-us/topic/may-8-2018-kb4103718-monthly-rollup-c4c01989-faca-af5f-46f4-2bdc2d0171fd) in May 2018. It's now **5 years later**. I think it would be safe for the STL to begin assuming unconditional support for SSE2, because this would cause problems only if:

* An end user has a still-functioning machine with a truly ancient non-SSE2 processor,
* They're running Win7 without any security updates past May 2018,
* They receive x86 software from a user-programmer using a fresh version of VS 2022 (whether an updated program itself, or a VCRedist)

This seems like it's not worth worrying about anymore. The changes on the STL's side would be minimal, but would be an improvement for all other x86-targeting user-programmers - we would remove `/arch:IA32` from the STL's build system (and test coverage), and the vectorized algorithms could begin assuming unconditional support for SSE2.

AlexGuteniev:
What are gains? For separately compiled vector algorithms they are minimal

StephanTLavavej:
Slightly fewer configurations to test.

AlexGuteniev:
Personally I don't care much -- the project I'm working on simply requires x64.

Yet, I would see such a decision as some lack of discipline. If the compiler and the CRT supports a mode, why would SLT silently break? I'd expect a consistent decision across the toolset, with another decision, like should the last version of toolset for a prrvious mode be cannedfor some extended use, like the XP toolset, or not.

AlexGuteniev:
> this would cause problems only if:
> 
> * An end user has a still-functioning machine with a truly ancient non-SSE2 processor,
> * They're running Win7 without any security updates past May 2018,
> * They receive x86 software from a user-programmer using a fresh version of VS 2022 (whether an updated program itself, or a VCRedist)

My experience also suggests that the amount of users for which this may be a problem may be very small and insignificant for you to care of, but it is not likely to be zero.

Like you have million developer users with billion end users, and you expect the set for your criteria is still empty? Some users may be servicing old production machines or kiosks or something, getting updates to their IDE, possibly not even realizing they are updating the STL with that, and they will be broken by such a change.

I'm not really defending keeping a legacy mode forever, but I still think that the removal should be graceful, with blocking the legacy mode across the toolset and a sympathetic apology message.

[...]

PR #2821, opened 2022-06-24
Fix vectorized `min/max/minmax_element` for 64-bit types on x86

StephanTLavavej:
This fixes a regression caused by #2447 that was specific to 64-bit types on x86, reported by internal VSO-1558536 "[RWC][prod/fe][Regression][x86] LLVM one test 'Profile/c-counter-overflows.c' failed".

Thanks to @AlexGuteniev for finding the root cause and providing the fix. The problem is that `_mm_extract_epi32` and `_mm_cvtsi128_si32` both return `int`. By directly casting to `_Unsigned_t`, i.e. `uint64_t` in `_Minmax_traits_8`, high-bit values were being sign-extended, but the desired behavior was zero-extension. Adding initial casts of `static_cast<uint32_t>` fixes this.

Then, I'm enhancing the test coverage:

* Adding a few specific test cases that failed.
* Truly randomizing the test, so that every run provides unique coverage.
* Using wider distributions of values for the min/max testing, since the limited `[1, 20]` range concealed this bug. Now, integers use the full range, while floating-point values use a large range centered around 0. (Unfortunately, `uniform_real_distribution` requires `b - a <= max`, so we can't generate a full range without extra work. As floating-point is not yet vectorized, I felt that this was sufficient for now.)

Fortunately, this regression has not yet shipped in a Preview.

PR #4659, opened 2024-05-07
Use AVX2 in `minmax_element` vectorization

AlexGuteniev:
Resolves #2803

This is not final optimization. At least, we should use AVX masks here too.
But this one is complex enough already, so the rest would be follow-up PR(s).

I also notice that `Both_val` 8-bit and 16-bit cases are slow. 
The vectorization for them is not engaged, it is a separate issue from the AVX.

<details>
<summary>Benchmark results</summary>
Before:

```
---------------------------------------------------------------------------
Benchmark                                 Time             CPU   Iterations
---------------------------------------------------------------------------
bm<uint8_t, 8021, Op::Min>              248 ns         96.3 ns      7466667
bm<uint8_t, 8021, Op::Max>              238 ns         96.9 ns      6291661
bm<uint8_t, 8021, Op::Both>             371 ns          124 ns      8432941
bm<uint8_t, 8021, Op::Min_val>          131 ns         56.6 ns     16000000
bm<uint8_t, 8021, Op::Max_val>          128 ns         67.0 ns      7466667
bm<uint8_t, 8021, Op::Both_val>        4197 ns         1611 ns       407273
bm<uint16_t, 8021, Op::Min>             459 ns          226 ns      4072727
bm<uint16_t, 8021, Op::Max>             445 ns          200 ns      3521123
bm<uint16_t, 8021, Op::Both>            685 ns          331 ns      2357895
bm<uint16_t, 8021, Op::Min_val>         247 ns          122 ns      4480000
bm<uint16_t, 8021, Op::Max_val>         252 ns          105 ns      7466667
bm<uint16_t, 8021, Op::Both_val>       4239 ns         3299 ns       203636
bm<uint32_t, 8021, Op::Min>             979 ns          364 ns      1544828
bm<uint32_t, 8021, Op::Max>             932 ns          586 ns      1120000
bm<uint32_t, 8021, Op::Both>           1439 ns          921 ns       746667
bm<uint32_t, 8021, Op::Min_val>         501 ns          321 ns      1947826
bm<uint32_t, 8021, Op::Max_val>         494 ns          414 ns      2036364
bm<uint32_t, 8021, Op::Both_val>        673 ns          500 ns      1000000
bm<uint64_t, 8021, Op::Min>            4252 ns         3128 ns       344615
bm<uint64_t, 8021, Op::Max>            4360 ns         2567 ns       280000
bm<uint64_t, 8021, Op::Both>           4397 ns         2783 ns       213333
[...]

PR #4661, opened 2024-05-07
`vector_algorithms.cpp`: `minmax` for 64-bit elements: replace ugly x86 workaround with a nice one

AlexGuteniev:
This: piece
https://github.com/microsoft/STL/blob/8dc4faadafb52e3e0a627e046b41258032d9bc6a/stl/src/vector_algorithms.cpp#L1116-L1123
works around the oddity of not having `_mm_cvtsi128_si64` on 32-bit x86

It has been problematic:
 * An internal bug was reported; fixed by #2821 
 * A compiler bug was caught, it blocked #3922

I have discovered a nicer workaround!

If we spill the reg into the stack, the spill will optimize away. 
On 32-bit with at least `/arch:SSE2` it even produces better code than the existing workaround.
Demo: https://godbolt.org/z/ErGWz8GYT

It still does the actual spill on `/arch:IA32`. But given that this path is executed only once per function call (there are no intermediate reductions for 64-bit elements), and there's a plan to lift to `/arch:SSE2`, I think that's fine.

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -1110,7 +1110,7 @@ namespace {
         }
 
         static uint64_t _Get_any_u(const __m128i _Cur) noexcept {
-            return _Minmax_traits_8::_Get_any(_Cur);
+            return _Minmax_traits_8::_Get_v_pos(_Cur, 0);
         }
 
         static uint64_t _Get_v_pos(const __m128i _Idx, const unsigned long _H_pos) noexcept {
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -1114,9 +1114,7 @@ namespace {
         }
 
         static uint64_t _Get_v_pos(const __m128i _Idx, const unsigned long _H_pos) noexcept {
-            uint64_t _Array[2];
-            _mm_storeu_si128(reinterpret_cast<__m128i*>(&_Array), _Idx);
-            return _Array[_H_pos >> 3];
+            return _Minmax_traits_8::_Get_v_pos(_Idx, _H_pos);
         }
 
         static __m128d _Cmp_eq(const __m128d _First, const __m128d _Second) noexcept {
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -901,13 +901,11 @@ namespace {
         }
 
         static _Signed_t _Get_any(const __m128i _Cur) noexcept {
-#ifdef _M_IX86
-            return static_cast<_Signed_t>(
-                (static_cast<_Unsigned_t>(static_cast<uint32_t>(_mm_extract_epi32(_Cur, 1))) << 32)
-                | static_cast<_Unsigned_t>(static_cast<uint32_t>(_mm_cvtsi128_si32(_Cur))));
-#else // ^^^ x86 / x64 vvv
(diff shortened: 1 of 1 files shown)


stl/src/vector_algorithms.cpp
```diff
@@ -1114,12 +1112,7 @@ namespace {
...
-#ifdef _M_IX86
-            return (static_cast<uint64_t>(static_cast<uint32_t>(_mm_extract_epi32(_Cur, 1))) << 32)
-                 | static_cast<uint64_t>(static_cast<uint32_t>(_mm_cvtsi128_si32(_Cur)));
-#else // ^^^ x86 / x64 vvv
-            return static_cast<uint64_t>(_mm_cvtsi128_si64(_Cur));
-#endif // ^^^ x64 ^^^
+            return _Minmax_traits_8::_Get_any(_Cur);
         }
 
         static uint64_t _Get_v_pos(const __m128i _Idx, const unsigned long _H_pos) noexcept {
```
StephanTLavavej, line 1118: Pre-existing: `_Minmax_traits_8` and `_Minmax_traits_d` have exactly identical `_Get_v_pos` (modulo typedef usage), which makes sense since `_Get_any_u` and `_Get_v_pos` are the parts that still behave like integers (whereas the other member functions were `double`-ified). We should just call `_Minmax_traits_8` here.

stl/src/vector_algorithms.cpp
```diff
@@ -1114,12 +1112,7 @@ namespace {
         }
 
         static uint64_t _Get_any_u(const __m128i _Cur) noexcept {
-#ifdef _M_IX86
-            return (static_cast<uint64_t>(static_cast<uint32_t>(_mm_extract_epi32(_Cur, 1))) << 32)
-                 | static_cast<uint64_t>(static_cast<uint32_t>(_mm_cvtsi128_si32(_Cur)));
-#else // ^^^ x86 / x64 vvv
-            return static_cast<uint64_t>(_mm_cvtsi128_si64(_Cur));
-#endif // ^^^ x64 ^^^
+            return _Minmax_traits_8::_Get_any(_Cur);
```
StephanTLavavej, line 1115: `_Minmax_traits_8::_Get_any` returns `_Minmax_traits_8::_Signed_t` so this is a value-modifying conversion to `uint64_t`.

However, if we "punch through" and call `_Minmax_traits_8::_Get_v_pos`, that returns `_Unsigned_t` so no `static_cast` would be needed.

StephanTLavavej (approved):

StephanTLavavej:
Thanks, this is great! :heart_eyes_cat: I pushed further changes to centralize the logic, please meow if you have concerns.

AlexGuteniev:
This cantralization is already done in  #4659 , there would just be more conflicts, after doing again here

AlexGuteniev:
Oh, you also did the implementation of `_Get_any` via `_Get_v_pos`. Didn't think about it, bu seems it will be no worse, as 0 is expected to constant propagate.

StephanTLavavej:
I'm drowning in PRs right now, so I think I'm going to have to clear out the backlog in multiple batches (in between investigating a non-STL bug that I can't weasel out of investigating forever). I'd like to land this PR first, then resolve conflicts in #4659.

StephanTLavavej:

StephanTLavavej:
Thanks for noticing how to make this code way more elegant! :magic_wand: :heart_eyes_cat: :green_heart:

PR #3337, opened 2023-01-09
P2093R14: Formatted Output

tylerbrawl:
This PR contains the complete implementation of [P2093R14](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2093r14.html), along with the changes proposed in [P2539R3](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2539r3.html). If merged, it should also close #2916.

# Added Files
Some files were added to implement this PR. These may require internal changes to the MSVC and/or Visual Studio which are outside of my control. The added files are as follows:

* `<print>`: This is the public header described in [P2093R14](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2093r14.html) containing the `FILE*` overloads of `std::print()` et al..

* `<xfilebuf.h>`: This is a private, non-core header which contains the definition for `std::basic_filebuf`, which was initially located in `<fstream>`. This was necessary in order to define the `std::ostream&` overloads of `std::print()` et al. within `<ostream>`; otherwise, there would be a header circular dependency (`<ostream>` -> `<fstream>` -> `<istream>` -> `<ostream>` -> ...).

* `<xprint.h>`: This is a private, core header which declares the C API exposing the functionality for printing to Unicode consoles as described in [P2093R14](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2093r14.html).

* `src\msvcp_print.src`: This is a template export definition file for a new print satellite DLL, much like `src\msvcp_atomic_wait.src`.

* `src\print.cpp`: This implements the C API declared in `<xprint.h>`.

# Implementation Notes
* I'm not particularly proud of the current method for checking if `std::ostream` instances refer to Unicode consoles. The proposed implementation uses a `dynamic_cast` to try to convert the `std::streambuf` of the `std::ostream` instance to a `std::filebuf`, and then extracts its `FILE*` if successful. I'm open to suggestions as to how this could be improved. The main difficulty is coming up with a solution which doesn't break ABI. (This is also the method taken by the [fmt](https://github.com/fmtlib/fmt) library.)

* As a result of the aforementioned implementation detail, the `std::ostream&` overloads of `std::print()` et al. are currently deleted if RTTI is disabled. (The versions which take a `FILE*` instead are still available in that case.)

* Some optimizations were done to improve code generation in common scenarios:
[...]

PR #2210, opened 2021-09-18
`tests/tr1`: Fix sporadic failures caused by `tmpnam`/`_tempnam`

StephanTLavavej:
While investigating failures that @mnatsuhara encountered in the internal test harness, @CaseyCarter encountered sporadic failures in the `tr1` legacy test suite:

<details><summary>Click to expand test failure output from tests/tr1/tests/fstream1 and fstream2.</summary>

```
RUN.PM line 750:  Executing: "fstream1.exe".
FAIL test 061 at line 247 in .\test.cpp: nfs.is_open()
FAIL test 062 at line 248 in .\test.cpp: nfs.good()
GOT "" != "this is a test"
FAIL test 063 at line 253 in .\test.cpp: buf == "this is a test"
GOT "" != "this is only a test"
FAIL test 064 at line 255 in .\test.cpp: buf == "this is only a test"
GOT "" != "this is still just a test"
FAIL test 065 at line 257 in .\test.cpp: buf == "this is still just a test"
GOT "" != "one last test"
FAIL test 066 at line 259 in .\test.cpp: buf == "one last test"
GOT "" != "one more last test"
FAIL test 067 at line 261 in .\test.cpp: buf == "one more last test"
FINISHED testing <fstream>, part 1
```

```
RUN.PM line 750:  Executing: "fstream2.exe". 
GOT L"this is still just a test" != L"one more last test" 
FAIL test 063 at line 232 in .\test.cpp: buf == L"one more last test"
```
</details>

The problem appears to be that `tr1` uses [`tmpnam`](https://en.cppreference.com/w/cpp/io/c/tmpnam) and [`_tempnam`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/tempnam-wtempnam-tmpnam-wtmpnam?view=msvc-160) to generate temporary file names. By default, these functions generate temporary file names in the user-global temp directory, and they're either short enough to be collision-prone, or sequential. :scream_cat: This is a problem when tests are run in parallel:

<details><summary>Click to expand example output from tmpnam and _tempnam, plus the new function being added in this PR.</summary>

```
C:\Temp>type meow.cpp
```
```cpp
#include "temp_file_name.h"
#include <filesystem>
#include <stdio.h>

using namespace std;
using namespace std::filesystem;

int main() {
    printf("             tmpnam: %s\n", tmpnam(nullptr));
    printf("           _tempnam: %s\n", _tempnam(".", ""));
[...]

PR #4665, opened 2024-05-10
Improve `P0218R1_filesystem` test reliability and temp filename generation

StephanTLavavej:
Making an exception to our usual rule of "we don't spend a lot of effort on cleaning up tests that are well-behaved, because it's low reward and risks disrupting what they were trying to test". I had to investigate why `P0218R1_filesystem` was behaving slightly flaky, and I ended up making a bunch of targeted improvements.

* Centralize the definition of `temp_file_name()`.
  + In #3337, I pushed a commit to copy `tests/tr1/include/temp_file_name.h` to `tests/std/include/temp_file_name.hpp`, because `tests/std` can't use `tests/tr1/include`. What I forgot is that `tests/tr1` *can* use `tests/std/include`, in both the GitHub and MSVC-internal test harnesses. Let's deduplicate these files now.
  + Include `<temp_file_name.hpp>` with angle brackets because it's outside the current directory (this is our usual convention, with only a few exceptions that I'll clean up later).
* Improve `temp_file_name()`: `"msvc_stl_"` prefix, 128 bits.
  + The `".tmp"` extension is already clear, so let's change the prefix to make the source of these files obvious.
  + In #2210, I chose 256 bits of entropy without careful consideration, and it was way too much. 128 bits is plenty; see Wikipedia's birthday problem article for a [probability table](https://en.wikipedia.org/wiki/Birthday_problem#Probability_table).
* Conversely, increase `get_test_directory_subname()` from 16 hexits / 64 bits to 32 hexits / 128 bits.
* Rename part 1: `get_test_directory` => `get_experimental_test_directory`
* Rename part 2: `get_new_test_directory` => `get_test_directory`
  + Standard `<filesystem>` should be the default assumption.
* Simplify `get_test_directory_subname()` by templating it.
  + We no longer need `<cstring>` for `strlen()`.
* Make `test_temp_directory` resistant to misuse.
  + Mark it as a `[[nodiscard]]` guard type and make it noncopyable.
* `test_temp_directory`'s ctor/dtor can use local `error_code ec;`.
  + Nobody needed to access this as a public data member after construction.
* Consistently name `test_temp_directory`, part 1: `recursiveTests` => `tempDir`
* Consistently name `test_temp_directory`, part 2: `followSymlinkTests` => `tempDir`
* Use consistent names for test directory bases.
  + `"recursive_directory_iterator-specific"`: Separate with a dash, instead of an accursed space.
  + `"recursive_directory_iterator-VSO-649431"`: Separate with a dash, instead of an underscore.
  + `"status"`: We use the name of the filesystem function we're testing, not our outer test function.
  + `"create_directories-and-remove_all"`: Separate with dashes, don't abbreviate.
* Directly pass `path`s instead of `.native()`.
  + These calls to `create_directories()`, `create_directory()`, and `exists()` were totally inconsistent for no reason. We don't need to construct temporary identical paths.
* Improve testing of "Standard rename where target doesn't exist".
  + We should call the `ec` form (since we're not catching exceptions), verify that `ec` is good, and verify that the source file no longer exists.
* Skip affected `rename()` tests for `_MSVC_INTERNAL_TESTING`.


StephanTLavavej:

barcharcraz (approved):

PR #4687, opened 2024-05-22
Toolset update: VS 2022 17.11 Preview 1

### LINKED ISSUE #4632: Submodule checkout failure should short-circuit checks
Rarely, submodule checkout fails due to network issues. When this happens, the Checkout Submodule step doesn't report a failure to Azure Pipelines, so it continues to configure/build the STL and then run tests, which will fail horribly. This makes the cause of the failure harder to understand (especially for newer contributors who won't recognize this pattern), and wastes machine time.

<details><summary>Click to expand sample logs:</summary>

```
Starting: Checkout llvm-project Submodule
==============================================================================
Task         : PowerShell
Description  : Run a PowerShell script on Linux, macOS, or Windows
Version      : 2.237.5
Author       : Microsoft Corporation
Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/powershell
==============================================================================
Generating script.
========================== Starting Command Output ===========================
"C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe" -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command ". 'D:\a\_work\_temp\bc3f730b-fdc7-4723-9176-e19c9d80cc0c.ps1'"
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
Initialized empty Git repository in D:/a/_work/1/s/llvm-project/.git/
hint: 	git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
hint: 	git branch -m <name>
From https://github.com/llvm/llvm-project
 * branch            2e2b6b53f5f63179b52168ee156df7c76b90bc71 -> FETCH_HEAD
error: RPC failed; curl 56 Failure when receiving data from the peer
error: 2334 bytes of body are still expected
fetch-pack: unexpected disconnect while reading sideband packet
fatal: early EOF
fatal: fetch-pack: invalid index-pack output
fatal: could not fetch 0757c610be56f8aeb01980d95b3bb20f5532ccc2 from promisor remote
Finishing: Checkout llvm-project Submodule
```
</details>

Filing this as a reminder-to-self; I'll investigate when I get a chance. If I wanted to be very fancy, I could look into making failure perform a directory cleanup and automatic retry. (Rerunning flaky *tests* is an abomination, but retrying a network failure is morally reasonable.)

StephanTLavavej:
# :scroll: Changelog
  * Overhauled our Azure Pipelines machinery:
    + Improved the reliability of the submodule checkout task.
    + Improved how clang-format diffs are uploaded and logged.
  * Updated dependencies.
    + Updated build compiler to VS 2022 17.11 Preview 1.

# :gear: Azure Pipelines Commits
Followup to #4594. I manually tested all of the success and failure scenarios for these commits.

* Detect provision-image.ps1 failure.
  + This fixes a major headache in the toolset update process. Previously, if provision-image.ps1 failed while downloading and installing components on the VM, it would stop, but create-1es-hosted-pool.ps1 would keep going on the maintainer's machine, creating a totally doomed pool. When failure occurs, `Invoke-AzVMRunCommand` doesn't indicate that, so I have a simple strategy - have the VM script print a special message for success, then have the maintainer script search for that message.
  + Use case-sensitive matching for `'PROVISION_IMAGE_SUCCEEDED'`.
* Permanently install PowerShell 7.4.2.
  + We'll need this for the following commits.
  + We need to preserve the original filename `'PowerShell-7.4.2-win-x64.msi'` instead of generating `'RANDOM.exe'`, otherwise we won't be able to run it.
  + This means that provision-image.ps1 no longer attempts an "in-place upgrade" of the PowerShell version that's running it. This significantly simplifies the script, and we don't need any PowerShell 7.x features here.
* Download installers to `D:\installerTemp`.
  + CUDA is 3 GB, so this may be helpful to make the script run faster.
  + On the VM, `D:` is fast temporary storage.
* Guard against locally running provision-image.ps1.
  + I've never made this mistake, but let's make sure it never happens for any maintainers or contributors. (Obviously, checking for the computer being named `'PROTOTYPE'` isn't absolutely perfect, but it's the simplest way to recognize the VM.)
  + This uses a case-sensitive comparison; the default is case-insensitive.
* Skip a blank line emitted by `cmd /c ver`.
* Set `pwsh: true`.
  + Needed for the following commit. I'm changing the 'Create Diff' task for consistency.
* Fix #4632.
  + PowerShell 7.4 added `$PSNativeCommandUseErrorActionPreference`: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.4#psnativecommanduseerroractionpreference
* Retry submodule checkout up to 4 times.
  + See: https://learn.microsoft.com/en-us/azure/devops/release-notes/2021/pipelines/sprint-195-update#automatic-retries-for-a-task
  + This should essentially fix one of the biggest remaining causes of sporadic PR/CI failure.
  + The release notes said "The failing task is retried immediately" but I now observe quadratic backoff, waiting 1, 4, 9 seconds (which is better than immediate).
* checkout-submodule.yml: Use case-sensitive `-creplace` and `-ceq`.
  + `git submodule status` emits lowercase hexits, and we pass lowercase `'llvm-project'`.
  + Although PowerShell defaults to case-insensitive, choosing case-sensitive for string operations should align with C++ expectations and avoid surprises.
* Style: Wrap `Restart-AzVM` arguments.
* Move `Get-AzVM` up so we can use `$VM.ID` earlier.
  + `$VM.ID` and `$VM.StorageProfile.OsDisk.Name` are unchanged between these locations.
  + Calling `Get-AzVM` is necessary; the return value of `New-AzVm` isn't useful.
* Inline away the `Wait-Shutdown` function.
  + The progress bar message makes the `Write-Host` messages unnecessary.
  + `-ResourceId $VM.ID` is much shorter than `-ResourceGroupName $ResourceGroupName -Name $Name`, allowing us to directly say `'PowerState/stopped'` without wrapping.
  + The original implementation in #633 (named `Start-WaitForShutdown`) was verbose, justifying a separate function, but now a direct loop is clearer.
* Consistently use `$VM.ID`.
* Run sysprep with `-ScriptString`.
  + `$ErrorActionPreference` wasn't useful because sysprep is a native command.
  + `Write-Host` wasn't useful because we aren't printing the message returned by `Invoke-AzVMRunCommand`.
  + The PowerShell call operator `&` wasn't useful because we aren't expanding any variables here.
  + In addition to being slightly simpler, this avoids having a script file that someone might try to run.
* Style: For `New-AzVMConfig`, use `-VMName` instead of the alias `-Name`.
  + See: https://learn.microsoft.com/en-us/powershell/module/az.compute/new-azvmconfig?view=azps-12.0.0#-vmname
* `New-AzVm` defaults to Trusted Launch as of Azure PowerShell 11.0.0 (November 2023).
  + See: https://techcommunity.microsoft.com/t5/azure-compute-blog/breaking-change-for-vm-amp-vmss-powershell-cli-customers/ba-p/3937037
  + See: https://learn.microsoft.com/en-us/powershell/azure/release-notes-azureps?view=azps-12.0.0#azcompute-700
* Give `validationBuildOutputLocation` a dedicated name, and clean it.
* Inline create-prdiff.ps1, part 1.
  + No changes except for:
    - Drop redundant banner.
    - `$DiffFile` was never passed as an argument, so set it as an ordinary variable.
    - Indentation.
* Inline create-prdiff.ps1, part 2.
  + Simplify by using the PowerShell redirection operator `>` instead of `Start-Process`.
  + https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_redirection?view=powershell-7.4#powershell-redirection-operators
  + Verified that the diff can be applied cleanly.
* Inline create-prdiff.ps1, part 3.
  + Overhaul messaging.
* Inline create-prdiff.ps1, part 4.
  + Actually name the file format.diff before uploading, as this appears in the zip.
* Inline create-prdiff.ps1, part 5.
  + Name the zip differently from the file within.
* Inline create-prdiff.ps1, part 6.
  + Style: Avoid Yoda condition when checking `git diff` output.
* Improve provision-image.ps1's synopsis/description.
  + They were talking about our old scale set system. Now they talk about hosted pools, and explain the overall process in more detail.

# :toolbox: Toolset Update Commits
* Require Node.js 22.2.0.
* New pool.
* VS 2022 17.11 Preview 1.
  + Not yet required - the internal toolset is still 19.40.
* Remove compiler warning suppressions.
  + MSVC no longer emits warning C4521 'multiple copy constructors specified', since at least MSVC-PR-368907 on 2021-12-09.
* Work around VSO-2064546 "EDG ICE when deriving from `std::num_get`".

STL-ASan-CI passed.


barcharcraz (approved):

StephanTLavavej:

PR #4695, opened 2024-05-25
Clean up ARM64 `popcount`

### LINKED ISSUE #2129: <intrin0.h>: needs some stuff for `_M_ARM64`
For https://github.com/microsoft/STL/pull/2127 I need

1) `__uint64ToN64_v`
2) `neon_cnt`
3) `neon_addv8`
4) `__n64`

from `<arm64_neon.h>`

https://godbolt.org/z/MbEqfshGG

### LINKED ISSUE #4683: ARM64 `__popcnt` intrinsics
In VS 2022 17.11 Preview 1, MSVC-PR-530436 updated `intrin0.inl.h` to extend the previously x86/x64 `__popcnt` intrinsic family to ARM64:

```cpp
__MACHINEX86_X64_ARM64(unsigned int __popcnt(unsigned int))
__MACHINEX86_X64_ARM64(unsigned short __popcnt16(unsigned short))
__MACHINEARM64_X64(unsigned __int64 __popcnt64(unsigned __int64))
```

It's unclear to me whether it would be simpler and/or faster to replace our existing ARM64 codepath:

https://github.com/microsoft/STL/blob/63354c3fa9c1fb2ab1fccb58c47d23c6af1c290f/stl/inc/__msvc_bit_utils.hpp#L351-L356

AlexGuteniev:
It is both way simpler (don't need to use another intrinsics set) and better for codegen (as the previous version is only for 64-bits and does not take advantage of smaller parameter size).

I have not tested on actual ARM64 and ARM64EC.

Resolves #4683
Resolves #2129
Resolves llvm/llvm-project#50830


AlexGuteniev:
Honestly, what I'm doing for ARM64EC here is uneducated guess. 
This needs to be checked by an ARM64EC expert.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for greatly simplifying this code! :heart_eyes_cat: :sparkles: :tada:

PR #4702, opened 2024-05-31
llvm-project Meso-Update

StephanTLavavej:
* This updates our llvm-project submodule for the first time in 4 months. Previously:
  + #4263
  + #4348
* To make this possible, I had to fix the following things upstream:
  + llvm/llvm-project#93255
  + llvm/llvm-project#93257
  + llvm/llvm-project#93259
* Tons of things have been fixed upstream, allowing us to remove many expected failures and update comments:
  + llvm/llvm-project#77948
  + llvm/llvm-project#79978
  + llvm/llvm-project#79856
  + llvm/llvm-project#83734
  + llvm/llvm-project#74534
  + llvm/llvm-project#88669
  + llvm/llvm-project#90981
  + llvm/llvm-project#88998
  + llvm/llvm-project#74630
  + llvm/llvm-project#79793
* libcxx replaced old Features with 'has-1024-bit-atomics' (and unconditional Features should be defined in features.py instead of tests.py):
  + llvm/llvm-project#75553
  + llvm/llvm-project#83841
* We 'has-no-zdump' (a unix program).
* libcxx now uses 'c++23' instead of 'c++2b'. I believe this was:
  + https://github.com/llvm/llvm-project/commit/71400505ca04
  + llvm/llvm-project#69035
* Remove C++26 features from 'MISSING STL FEATURES'.
  + Now that we're identifying ourselves as being a C++23 implementation.
* Update magic comments.
  + These are used by the internal test harness to skip tests.
  + The two removed lines no longer appear in libcxx.
* range_adaptor_closure.pass.cpp needs WG21-P1169R4 `static operator()`.
* Add missing newline.
* Mark vector.modifiers/destroy_elements.pass.cpp as failing due to constexpr step limits.
* Categorize failing tests.
* format.range.formatter/parse.pass.cpp is now passing after #4642.
* Recategorize format.range.formatter/format.pass.cpp.
* Skip ranges.ends_with.pass.cpp which sporadically fails with "fatal error C1060: compiler is out of heap space".
* Finally, we need to update `tests/std/tests/P0088R3_variant/test.cpp` because it uses LLVM machinery. Both `VariantAllowsNarrowingConversions` and `variant<T&>` coverage were removed:
  + llvm/llvm-project#83928
  + llvm/llvm-project#84222

I'm working on a full regeneration of `P0088R3_variant`, but that's going to be a *much* larger change and I don't want to mix that into this PR.

### DIFF
--- a/tests/libcxx/expected_results.txt
+++ b/tests/libcxx/expected_results.txt
@@ -22,10 +22,6 @@ std/time/time.syn/formatter.year_month_day_last.pass.cpp:1 FAIL
 std/time/time.syn/formatter.year_month_weekday.pass.cpp:0 FAIL
 std/time/time.syn/formatter.year_month_weekday.pass.cpp:1 FAIL
 
-# LLVM-74221: [libc++][test] nasty_char_traits::move is incompatible with constexpr
-std/strings/basic.string/string.modifiers/string_append/initializer_list.pass.cpp:2 FAIL
-std/strings/basic.string/string.modifiers/string_assign/string.pass.cpp:2 FAIL
-
 # LLVM-74756: [libc++][test] overload_compare_iterator doesn't support its claimed iterator_category
 std/utilities/memory/specialized.algorithms/uninitialized.copy/uninitialized_copy.pass.cpp FAIL
 std/utilities/memory/specialized.algorithms/uninitialized.move/uninitialized_move.pass.cpp FAIL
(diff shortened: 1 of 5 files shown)


cpplearner:
The `P0088R3_variant` test is failing because libc++ removed the definition of `TEST_VARIANT_HAS_NO_REFERENCES` (along with the guarded test code in the libc++ test suite). See LLVM-84222.

StephanTLavavej:
Yeah, the `<variant>` tests have accumulated a bunch of changes. I'm performing a full manual regeneration.

MahmoudGSaleh (approved):
LGTM

PR #4712, opened 2024-06-04
`<__msvc_int128.hpp>`: Fix warning C26437 about slicing

StephanTLavavej:
In internal MSVC-PR-555396, @MahmoudGSaleh updated our checked-in toolset to VS 2022 17.11 Preview 2 and encountered a newly emitted [warning C26437](https://learn.microsoft.com/en-us/cpp/code-quality/c26437?view=msvc-170): "Do not slice (es.63)." One affected call:

https://github.com/microsoft/STL/blob/e36ee6c2b9bc6f5b1f70776c18cf5d3a93a69798/stl/inc/__msvc_int128.hpp#L902-L903

This can be avoided, for all callsites, by changing `_Base128`'s `_Divide()` and `_Modulo()` to copy their parameters locally. (This is still slicing, of course, just in a way that the compiler won't complain about. :smirk_cat:)

I'm also updating our test coverage to prevent regressions. Note that 17.11 Preview 2 newly warns about this code; 17.11 Preview 1 didn't warn.

### DIFF
--- a/stl/inc/__msvc_int128.hpp
+++ b/stl/inc/__msvc_int128.hpp
@@ -444,8 +444,8 @@ struct
         return _Result;
     }
     _NODISCARD static constexpr _Base128 _Divide(const _Base128& _Num_arg, const _Base128& _Den_arg) noexcept {
-        _Base128 _Num{_Num_arg};
-        _Base128 _Den{_Den_arg};
+        _Base128 _Num{_Num_arg}; // Avoid warning C26437 by creating local (sliced) copies
+        _Base128 _Den{_Den_arg}; // instead of taking parameters by value.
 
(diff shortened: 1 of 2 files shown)


AlexGuteniev:
Maybe comment like "slice explicitly to avoid C26437" ?

StephanTLavavej:
The static analysis team has confirmed that this is actually a compiler bug, VSO-2080466 "Code Analysis: False C26437 warnings due to badly constructed AST for assignment operation from base class object for trivially copiable base class". While `_Divide` was technically slicing here, the compiler was designed to not warn when the derived class adds no additional data members. The fact that it's emitting a warning is a regression. We're going to disable the warning in the internal build, instead of modifying the STL's sources, until the compiler's warning logic is properly fixed.

PR #4715, opened 2024-06-08
Fix warning C26818 about `switch` statements needing `default` cases

StephanTLavavej:
Fixes DevCom-10670613 VSO-2081258 / AB#2081258.

In general, we haven't attempted to be clean with respect to Core Guidelines warnings, as many of them are noisy and/or inappropriate for Standard Library code. However, warning C26818 "Switch statement does not cover all cases. Consider adding a 'default' label ([es.79][])." aligns with our conventions. We always like to have a `default` as a reminder to think about what happens when none of the `case`s are selected (this is one of the relatively few scenarios where we like to write code that isn't necessary). The only exception to this convention is when we're switching on an enum and we've handled all of the enumerators (it can still be okay to have a "can't happen" `default`, but sometimes we don't bother), which fortunately this warning doesn't complain about.

[es.79]: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-default

I added test coverage for all of these fixes, i.e. removing any fix will cause `GH_002094_cpp_core_guidelines` to fail.

All of the empty `default: break;` cases being added are used, i.e. making them `abort()` will cause other tests to fail.

In `<format>`, we have a couple of "can't happen" cases, where I've added `_STL_UNREACHABLE` with comments explaining why.

In `<xlocmon>`, we have `default` cases that can only be activated by users giving us a bogus `money_base::pattern`, so I've added `_STL_ASSERT(false, "message")` citing the same Standardese each time.

While auditing all of our `switch` statements (which is how I found half of these), I found a few cases in old code where we weren't obeying our syntax convention: a `switch`'s last `case` or `default` must always `break`. Falling off of the end is an unacceptable maintenance hazard.

Finally, I removed an unnecessary, inconsistent comment in `<xlocmon>`.

### DIFF
--- a/stl/inc/format
+++ b/stl/inc/format
@@ -276,6 +276,8 @@ _NODISCARD constexpr _Decode_result<char> _Decode_utf(const char* _First, const 
             return {_First + 1, false};
         }
         break;
+    default:
+        break;
     }
 
     // mask out the "value bits" in the leading byte,
@@ -1260,6 +1262,8 @@ _NODISCARD constexpr const _CharT* _Parse_align(
             break;
         case '^':
             _Parsed_align = _Fmt_align::_Center;
+            break;
+        default:
             break;
         }
 
@@ -2851,6 +2855,8 @@ _NODISCARD _OutputIt _Write_integral(
         break;
     case 'o':
         _Base = 8;
+        break;
+    default:
         break;
     }
 
--- a/stl/inc/format
+++ b/stl/inc/format
@@ -292,6 +292,9 @@ _NODISCARD constexpr _Decode_result<char> _Decode_utf(const char* _First, const 
     case 4:
         _Val &= 0b111u;
         break;
+    default:
+        _STL_INTERNAL_CHECK(false); // can't happen, see how the value of _Num_bytes is determined
+        break;
     }
 
     for (int _Idx = 1; _Idx < _Num_bytes; ++_Idx) {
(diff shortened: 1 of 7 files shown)


stl/inc/format
```diff
@@ -4534,6 +4545,10 @@ public:
             }
 
             break;
+
+        default:
+            _STL_INTERNAL_CHECK(false); // can't happen, we've handled all possible outputs from _Parse_range_specs()
```
AlexGuteniev, line 4550: `_STL_UNREACHABLE` ?

stl/inc/format
```diff
@@ -292,6 +294,9 @@ _NODISCARD constexpr _Decode_result<char> _Decode_utf(const char* _First, const
     case 4:
         _Val &= 0b111u;
         break;
+    default:
+        _STL_INTERNAL_CHECK(false); // can't happen, see how the value of _Num_bytes is determined
```
AlexGuteniev, line 298: `_STL_UNREACHABLE` ?

AlexGuteniev (commented):
We should use `std:;unreachable` for "can't happen" default handlers to aid optimization

StephanTLavavej:
Sure, thanks @AlexGuteniev! :heart_eyes_cat: Pushed a commit.

StephanTLavavej:

barcharcraz (approved):

PR #4724, opened 2024-06-11
`<ranges>`: Remove redundant `_HAS_CXX23` guards

StephanTLavavej:
This removes redundant `_HAS_CXX23` guards around `_Compile_time_max_size` specializations, for views that are already guarded for C++23.

Note that these changes are very low risk because the primary template for `_Compile_time_max_size` is guarded for C++23. If we attempted to specialize it in C++20 mode, we'd get instant errors:

https://github.com/microsoft/STL/blob/e36ee6c2b9bc6f5b1f70776c18cf5d3a93a69798/stl/inc/ranges#L44-L55

This also adds a couple of `#endif` comments; they aren't strictly necessary but they helped me ripgrep for the redundant guards.

Noticed while reviewing #4707.

### DIFF
--- a/stl/inc/ranges
+++ b/stl/inc/ranges
@@ -2135,13 +2135,11 @@ namespace ranges {
     template <class _Rng>
     constexpr bool enable_borrowed_range<as_rvalue_view<_Rng>> = enable_borrowed_range<_Rng>;
 
-#if _HAS_CXX23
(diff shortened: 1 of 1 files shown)


StephanTLavavej:

barcharcraz (approved):

PR #4725, opened 2024-06-12
Toolset update: VS 2022 17.11 Preview 2

StephanTLavavej:
# :scroll: Changelog
- Code cleanups:
  * Updated dependencies.
    + Updated build compiler to VS 2022 17.11 Preview 2.
    + Updated Python to 3.12.4.

# :bulb: Notes
The internal build compiler still needs to be updated (@MahmoudGSaleh is working on it, but it's a time-consuming process), so we still can't require `_MSC_VER` `1941`.

Similarly, a new WinSDK is available, but the internal build isn't using it yet, so I'm not updating the image or the readme to specify it. (This is updated at a slower pace, so it might not happen for a while.)

STL-ASan-CI passed.

# :gear: Workarounds removed
* VSO-1913897 DevCom-10508937 "Startup issue when building the static debug STL with ASan enabled"
* VSO-1985428 "`<limits.h>` emits warning C4668"
* VSO-2016422 "EDG hasn't (completely?) implemented CWG-2518 making `static_assert(false)` usable"
  + Product code still needs this workaround for CUDA 12.4, so I've updated the comment in `<type_traits>`.
* VSO-2064546 "EDG ICE when deriving from `std::num_get`"


AlexGuteniev:
How can you remove workarounds without updating internal build?

StephanTLavavej:
That's a good question! (And one that I didn't fully think through before submitting this PR :joy_cat:)

In this case I think I'll get away with all of these workaround removals. EDG is updated independently from the checked-in MSVC toolset, and is actually usually a newer version than what's shipping publicly, so the EDG workaround removals should be fine. The `<limits.h>` workaround removal depends on updated VCRuntime sources that I fixed, which are present in the internal repo (they aren't consumed from the checked-in toolset). So that leaves the ASAN issue, which I *think* will be okay - the checked-in toolset is used only for the initial build, whereas the tests run with the freshly built compiler. Even though we're building the compiler with ASAN these days (or will shortly, I forget), this bug was specific to the static debug STL, but the compiler doesn't consume the STL statically.

StephanTLavavej:

barcharcraz (approved):

PR #4729, opened 2024-06-16
Fix `atomic_ref::is_lock_free()` on x64

### LINKED ISSUE #4728: `<atomic>`: On x64, `atomic_ref::is_lock_free()` incorrectly returns `true` when it shouldn't
Repros with VS 2022 17.11 Preview 2 x64:

```
C:\Temp>type meow.cpp
```
```cpp
#include <atomic>
#include <print>
using namespace std;

struct Large {
    char str[100]{};
};

int main() {
    Large lg{};
    atomic_ref<Large> atom_ref{lg};
    println("atomic_ref<Large> is_lock_free(): {}", atom_ref.is_lock_free());
    atom_ref.store(Large{"cute fluffy kittens"});
    println("lg.str: {}", lg.str);
}
```
```
C:\Temp>cl /EHsc /nologo /W4 /std:c++latest /MTd /Od /Zi /Fdmeow.pdb meow.cpp
meow.cpp

C:\Temp>meow
atomic_ref<Large> is_lock_free(): true
lg.str: cute fluffy kittens
```

`atomic_ref<Large>::store()` definitely takes a lock:

https://github.com/microsoft/STL/blob/e36ee6c2b9bc6f5b1f70776c18cf5d3a93a69798/stl/inc/atomic#L630-L635

`atomic_ref::is_lock_free()` is bogus because:

https://github.com/microsoft/STL/blob/e36ee6c2b9bc6f5b1f70776c18cf5d3a93a69798/stl/inc/atomic#L2367-L2376

This *meant* to say: "If we're always lock-free then `true`, otherwise we have to be *potentially* lock-free *and* actually have cmpxchg16b."

I have a fix.

StephanTLavavej:
Fixes #4728.

For `!_ATOMIC_HAS_DCAS`:

* If we're `is_always_lock_free` (power of two, at most 1 pointer), return true.
* Otherwise, we have to be `_Is_potentially_lock_free` (power of two, at most 2 pointers) and have cmpxchg16b.

### DIFF
--- a/stl/inc/atomic
+++ b/stl/inc/atomic
@@ -2370,8 +2370,10 @@ public:
 #else // ^^^ _ATOMIC_HAS_DCAS / !_ATOMIC_HAS_DCAS vvv
         if constexpr (is_always_lock_free) {
             return true;
+        } else if constexpr (_Is_potentially_lock_free) {
+            return __std_atomic_has_cmpxchg16b() != 0;
         } else {
-            return _Is_potentially_lock_free && __std_atomic_has_cmpxchg16b() != 0;
+            return false;
         }
 #endif // ^^^ !_ATOMIC_HAS_DCAS ^^^
     }
--- a/stl/inc/atomic
+++ b/stl/inc/atomic
@@ -2371,7 +2371,7 @@ public:
         if constexpr (is_always_lock_free) {
             return true;
         } else {
-            return __std_atomic_has_cmpxchg16b() != 0;
+            return _Is_potentially_lock_free && __std_atomic_has_cmpxchg16b() != 0;
         }
 #endif // ^^^ !_ATOMIC_HAS_DCAS ^^^
     }
--- a/tests/std/tests/P0019R8_atomic_ref/test.cpp
+++ b/tests/std/tests/P0019R8_atomic_ref/test.cpp
@@ -447,6 +447,22 @@ void test_gh_4472() {
     assert(ar.is_lock_free());
 }
 
+// GH-4728 "<atomic>: On x64, atomic_ref::is_lock_free() incorrectly returns true when it shouldn't"
+void test_gh_4728() {
+    struct Large {
+        char str[100]{};
+    };
+
+    alignas(std::atomic_ref<Large>::required_alignment) Large lg{};
+
+    static_assert(std::atomic_ref<Large>::required_alignment == alignof(Large));
+
+    static_assert(!std::atomic_ref<Large>::is_always_lock_free);
+
+    std::atomic_ref<Large> ar{lg};
+    assert(!ar.is_lock_free());
+}
+
 int main() {
     test_ops<false, char>();
     test_ops<false, signed char>();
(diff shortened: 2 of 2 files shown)


stl/inc/atomic
```diff
@@ -2371,7 +2371,7 @@ public:
         if constexpr (is_always_lock_free) {
             return true;
         } else {
-            return __std_atomic_has_cmpxchg16b() != 0;
+            return _Is_potentially_lock_free && __std_atomic_has_cmpxchg16b() != 0;
```
AlexGuteniev, line 2374: I think it should be `if constexpr` and `else return false`

StephanTLavavej: No, because 16-byte `atomic_ref` is lock-free on non-potato chips. (I discovered this because I have changes for a future PR to remove the support for potato chips.) See the earlier test:

https://github.com/microsoft/STL/blob/e36ee6c2b9bc6f5b1f70776c18cf5d3a93a69798/tests/std/tests/P0019R8_atomic_ref/test.cpp#L427-L448

Debug into this:

```
C:\Temp>type meow.cpp
```
```cpp
#include <atomic>
#include <print>
using namespace std;

struct TwoPointers {
    int* left  = nullptr;
    int* right = nullptr;
};

using AtomicRef = atomic_ref<TwoPointers>;

int main() {
    int x = 17;
    int y = 29;
    alignas(AtomicRef::required_alignment) TwoPointers two_pointers{};
    AtomicRef atom_ref{two_pointers};

    println("AtomicRef::is_always_lock_free: {}", AtomicRef::is_always_lock_free);
    println("       atom_ref.is_lock_free(): {}", atom_ref.is_lock_free());
    println("Debug into atom_ref.store():");

    atom_ref.store(TwoPointers{&x, &y});

    println("{} {}", *two_pointers.left, *two_pointers.right);
}
```
```
C:\Temp>cl /EHsc /nologo /W4 /std:c++latest /MTd /Od /Zi /Fdmeow.pdb meow.cpp
meow.cpp

C:\Temp>meow
AtomicRef::is_always_lock_free: false
       atom_ref.is_lock_free(): true
Debug into atom_ref.store():
17 29

C:\Temp>devenv meow.exe meow.cpp
```

It goes through:

https://github.com/microsoft/STL/blob/e36ee6c2b9bc6f5b1f70776c18cf5d3a93a69798/stl/inc/atomic#L1299-L1300

AlexGuteniev: I still don't get how this prevent the following interim solution:
```c++
        if constexpr (is_always_lock_free) {
            return true;
        } else if constexpr (_Is_potentially_lock_free)
            return __std_atomic_has_cmpxchg16b() != 0;
        } else {
            return false;
        }
```

StephanTLavavej: Oh, that's strictly superior, thanks!

StephanTLavavej:

barcharcraz (approved):

PR #4753, opened 2024-06-29
Inline `_Memcmp_ranges` that is used only once

AlexGuteniev:
It is used only once after #4654


StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for the attention to detail in keeping the codebase clean! :broom: :eagle: :eye:

PR #4778, opened 2024-07-02
Fix insert for flat_map and flat_multimap

duanqn:
`_Flat_map_base::insert` overloads were hidden because of the additional `insert` overloads defined in the children classes.
Use `using _Mybase::insert` to make them visible.
Also add test cases.

### DIFF
--- a/tests/std/tests/P0429R9_flat_map/test.cpp
+++ b/tests/std/tests/P0429R9_flat_map/test.cpp
@@ -418,8 +418,16 @@ void test_insert() {
     assert(res3.first->second == 'w');
     assert(res3.second);
 
-    assert(check_key_content(fm, {10, 90}));
-    assert(check_value_content(fm, {'m', 'w'}));
+    flat_map<int, char>::iterator res4 = fm.insert(fm.cbegin(), std::pair<int, char>{30, 'c'});
+    assert(res4->first == 30);
+    assert(res4->second == 'c');
+
+    flat_map<int, char>::iterator res5 = fm.insert(fm.cbegin(), std::pair<char, char>{static_cast<char>(40), 'd'});
+    assert(res5->first == 40);
+    assert(res5->second == 'd');
+
+    assert(check_key_content(fm, {10, 30, 40, 90}));
+    assert(check_value_content(fm, {'m', 'c', 'd', 'w'}));
 
     flat_multimap<int, char> fmm;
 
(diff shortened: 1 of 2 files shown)


tests/std/tests/P0429R9_flat_map/test.cpp
```diff
@@ -431,13 +439,21 @@ void test_insert() {
...
+    assert(it4->second == 'w');
+
+    flat_multimap<int, char>::iterator it5 = fmm.insert(fmm.cend(), std::pair<char, char>{static_cast<char>(70), 'q'});
+    assert(it5->first == 70);
+    assert(it5->second == 'q');
 
-    assert(check_key_content(fmm, {10, 10, 90}));
-    assert(check_value_content(fmm, {'n', 'm', 'w'}));
+    assert(check_key_content(fmm, {10, 10, 70, 70, 90}));
+    assert(check_value_content(fmm, {'n', 'm', 'q', 'p', 'w'}));
```
AlexGuteniev, line 456: Is the particular order of q and p guaranteed by the standard? If not, should use the other overload of check_value_content.

duanqn: I think not, but I see that the order of 'n' and 'm' was checked here. I will check the standard.

AlexGuteniev: Yes, I think the intention of "as close as possible" is exactly that

AlexGuteniev: You may add `// FIXME` comment here, if you don't want to fix it right away in this PR

duanqn: Added.

duanqn:
@StephanTLavavej Could you please take a look at this? Thanks!

tests/std/tests/P0429R9_flat_map/test.cpp
```diff
@@ -418,8 +418,16 @@ void test_insert() {
     assert(res3.first->second == 'w');
     assert(res3.second);
 
-    assert(check_key_content(fm, {10, 90}));
-    assert(check_value_content(fm, {'m', 'w'}));
+    flat_map<int, char>::iterator res4 = fm.insert(fm.cbegin(), std::pair<int, char>{30, 'c'});
+    assert(res4->first == 30);
+    assert(res4->second == 'c');
+
+    flat_map<int, char>::iterator res5 = fm.insert(fm.cbegin(), std::pair<char, char>{static_cast<char>(40), 'd'});
```
StephanTLavavej, line 425: The previous 3 insertions said `const auto` to store the returned iterators, and it would be reasonable to be consistent here. Ditto below.

(It's fine to explicitly say `std::pair<int, char>` and `std::pair<char, char>` to emphasize the types being inserted, even though `std::pair` CTAD would give the same result.)

tests/std/tests/P0429R9_flat_map/test.cpp
```diff
@@ -431,13 +439,24 @@ void test_insert() {
...
+    assert(it4->first == 90);
+    assert(it4->second == 'w');
+
+    flat_multimap<int, char>::iterator it5 = fmm.insert(fmm.cend(), std::pair<char, char>{static_cast<char>(70), 'q'});
+    assert(it5->first == 70);
+    assert(it5->second == 'q');
+
+    assert(check_key_content(fmm, {10, 10, 70, 70, 90}));
+    // FIXME: The expected order of the last 3 value element is 'p', 'q', 'w' (no permutation allowed), according to
+    // [associative.reqmts.general]/72. The current implementation gives 'q', 'p', 'w', which is wrong.
```
StephanTLavavej, line 457: Should say "last 3 value elements" plural. Also, we should cite the WP number when citing paragraphs, because paragraphs are frequently renumbered; this is valid for N4981. (The WP number is optional when citing a stable name by itself.)

StephanTLavavej (approved):

StephanTLavavej:
@duanqn Looks great, thanks! I pushed small cosmetic changes and I'll merge this after the tests pass. Apologies for not getting to this sooner.

PR #4782, opened 2024-07-03
Remove `_NODISCARD_FRIEND` workaround

### LINKED ISSUE #4104: Does CUDA 12.4.0 still need `_NODISCARD_FRIEND`?
* Shortly before 2018-05: The bug EDGcpfe/19388, corresponding to VSO-568006 "REPORTED: Attributes are not accepted on friend definitions inside class templates", was fixed upstream by EDG.
* 2018-05: The EDG fix was picked up by IntelliSense.
* 2019-08: CUDA 10.1 Update 2 was released (it did not contain the EDG fix, to the best of my knowledge).
* 2021-12: #2208 introduced `_NODISCARD_FRIEND` to centralize our workaround.
* 2022-01: CUDA 11.6.0 was released.
* 2022-06: #2791 increased our required version from CUDA 10.1 Update 2 to CUDA 11.6.0.
* 2024-03: #4475 increased our required version to CUDA 12.4.0.

From this timeline, I suspect, but haven't confirmed, that we no longer need this workaround. If someone has time to check, that would be great. (I don't think our minimal CUDA test coverage is detailed enough to have confidence that simply removing the workaround and looking for test failures would be sufficient validation.)

fsb4000:
Fixes #4104 

I believe the workaround is not needed since CUDA 11.3.0: https://gcc.godbolt.org/z/bPh1av5f9

```
C:\Dev\STL\playground>type main.cpp
template<int>
struct test {
    int i;
     [[nodiscard]] friend bool operator==(const test& lhs, const test& rhs) {
        return lhs.i == rhs.i;
    }
};

int main() {
    test<1> t1{1};
    test<1> t2{1};
    t1 == t2;
}

C:\Dev\STL\playground>nvcc main.cpp
main.cpp
main.cpp(12): warning C4834: discarding return value of function with [[nodiscard]] attribute

C:\Dev\STL\playground>nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2024 NVIDIA Corporation
Built on Tue_Feb_27_16:28:36_Pacific_Standard_Time_2024
Cuda compilation tools, release 12.4, V12.4.99
Build cuda_12.4.r12.4/compiler.33961263_0
```

изображение

I used this version Cuda 12.4.0: https://developer.download.nvidia.com/compute/cuda/12.4.0/network_installers/cuda_12.4.0_windows_network.exe

### DIFF
--- a/stl/inc/functional
+++ b/stl/inc/functional
@@ -2050,8 +2050,8 @@ struct _Select_fixer<_Cv_TiD, false, false, _Jx> { // placeholder fixer
 
 template <class _Cv_TiD, class _Untuple>
 constexpr auto _Fix_arg(_Cv_TiD& _Tid, _Untuple&& _Ut) noexcept(
-    noexcept(_Select_fixer<_Cv_TiD>::_Fix(_Tid, _STD move(_Ut)))) -> decltype(_Select_fixer<_Cv_TiD>::_Fix(_Tid,
-    _STD move(_Ut))) { // translate an argument for bind
+    noexcept(_Select_fixer<_Cv_TiD>::_Fix(_Tid, _STD move(_Ut))))
+    -> decltype(_Select_fixer<_Cv_TiD>::_Fix(_Tid, _STD move(_Ut))) { // translate an argument for bind
     return _Select_fixer<_Cv_TiD>::_Fix(_Tid, _STD move(_Ut));
 }
 
--- a/stl/inc/functional
+++ b/stl/inc/functional
@@ -1946,7 +1946,7 @@ public:
         _Fn1._Swap(_Fn2);
     }
 
-    _NODISCARD_FRIEND bool operator==(const move_only_function& _This, nullptr_t) noexcept {
+    _NODISCARD friend bool operator==(const move_only_function& _This, nullptr_t) noexcept {
         return _This._Is_null();
     }
 };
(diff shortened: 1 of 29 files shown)


stl/inc/functional
```diff
@@ -2050,8 +2050,8 @@ struct _Select_fixer<_Cv_TiD, false, false, _Jx> { // placeholder fixer
 
 template <class _Cv_TiD, class _Untuple>
 constexpr auto _Fix_arg(_Cv_TiD& _Tid, _Untuple&& _Ut) noexcept(
-    noexcept(_Select_fixer<_Cv_TiD>::_Fix(_Tid, _STD move(_Ut))))
-    -> decltype(_Select_fixer<_Cv_TiD>::_Fix(_Tid, _STD move(_Ut))) { // translate an argument for bind
+    noexcept(_Select_fixer<_Cv_TiD>::_Fix(_Tid, _STD move(_Ut)))) -> decltype(_Select_fixer<_Cv_TiD>::_Fix(_Tid,
+    _STD move(_Ut))) { // translate an argument for bind
```
StephanTLavavej, line 2054: Unrelated reflow is happening here.

StephanTLavavej (approved):

StephanTLavavej:
Thanks for doing the archaeology to figure out exactly when this workaround became unnecessary! :heart_eyes_cat: This change is super safe because we began requiring CUDA 11.6.0 with #2791 in VS 2022 17.4, two years ago.

miscco (approved):
This mirrors what we are doing in cccl

miscco (approved):

stl/inc/functional
fsb4000: Yes, I did `clang-format -i filename` for every changed file. Probably clang-format changed that for some reason.

StephanTLavavej:

StephanTLavavej:
Thanks for cleaning up this old workaround! :broom: :magic_wand: :heart_eyes_cat:

PR #4817, opened 2024-07-07
Fix `<bitset>` benchmark correctness

AlexGuteniev:
I see more things to do with `bitset` regarding performance, but first let's fix the current benchmark.

Fortunately, the discovered issues didn't have impact on results, and on performance decisions.

## Add `DoNotOptimize` to input

Should use this to prevent compiler making the benchmarked function or its part constant.
Currently the compiler doesn't do this, but it might improve in the future

## Make 512 case as large

Small cases were initialized with `uint64_t` which was not enough to initialize 512 bit bitset.

The 512 rows in the benchmark improved, but it is gooseberries vs watermelons: small cases have inner loop whereas large doesn't. 

(The loop in small cases is to defeat branch prediction for the cases of branchy implementation activation. It is not needed for larger bitsets, as these would be vectorized, and if they wouldn't, the amount of bits should be enough to overflow the branch prediction)

### DIFF
--- a/benchmarks/src/bitset_to_string.cpp
+++ b/benchmarks/src/bitset_to_string.cpp
@@ -13,18 +13,18 @@
 using namespace std;
 
 namespace {
-    template <size_t N>
+    template <size_t Elems>
     auto random_bits_init() {
         mt19937_64 rnd{};
-        array<uint64_t, N> arr;
+        array<uint64_t, Elems> arr;
         for (auto& d : arr) {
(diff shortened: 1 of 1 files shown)


StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for improving how we benchmark all of these bits! :rocket: :computer: :stopwatch:

PR #4818, opened 2024-07-07
Fewer allocations to put smaller bitset to stream

AlexGuteniev:
The value of 128 bytes is arbitrary, could be greater if we are brave.

I kept variable `_Len` as opposer to `_Bits` constant, as constant propagation apparently engages a very harmful optimization.

Didn't benchmark precisely output to stream, I think if falls under "definitely good" category; if not, I can add, but then will need to wait for #4817

### DIFF
--- a/stl/inc/bitset
+++ b/stl/inc/bitset
@@ -18,9 +18,9 @@ _STL_DISABLE_CLANG_WARNINGS
 #pragma push_macro("new")
 #undef new
 
-#ifndef _STD_BITSET_TO_STRREAM_STACK_RESERVATION
-#define _STD_BITSET_TO_STRREAM_STACK_RESERVATION 128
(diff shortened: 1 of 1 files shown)


AlexGuteniev:
Can go furthr and convert to digit array by parts, and stream those parts, this way can avoid sllocations altogether.

But this would be more complex change

StephanTLavavej (approved):

AlexGuteniev:
> anyone touching iostreams isn't going to have great perf in the first place

Well, `stringstream` is not as that, but generally agree.

StephanTLavavej:

StephanTLavavej:
Thanks for letting the allocator take some vacation days! :beach_umbrella: :partying_face: :joy_cat:

PR #4835, opened 2024-07-12
Remove `std::ranges::view_interface` visualizer.

### LINKED ISSUE #4830: `STL.natvis`: `ranges::view_interface` doesn't always have `size()`
https://github.com/microsoft/STL/blob/ef1d621d51263285aff8e560a214f5477d63d687/stl/inc/xutility#L3766-L3779
https://github.com/microsoft/STL/blob/ef1d621d51263285aff8e560a214f5477d63d687/stl/debugger/STL.natvis#L2004-L2012

Reported by @xiangfan-ms:

> The 'size' member function doesn't always exist in the specialization. When this happens, the entry will be ignored by the debugger.

Present since this visualizer was added by #2191 on 2022-07-11 in VS 2022 17.4.

fsb4000:
Fixes #4830 

Actually I tested it and I think `std::ranges::view_interface` natvis may be useless.

My test code:
```
#include <iostream>
#include <ranges>
#include <vector>
#include <sstream>

template<class T, class A>
class VectorView : public std::ranges::view_interface<VectorView<T, A>>
{
public:
    VectorView() = default;

    VectorView(const std::vector<T, A>& vec) :
        m_begin(vec.cbegin()), m_end(vec.cend())
    {}

    auto begin() const { return m_begin; }

    auto end() const { return m_end; }

private:
    typename std::vector<T, A>::const_iterator m_begin{}, m_end{};
};

int main()
{
    std::vector<int> v = { 1, 4, 9, 16 };

    VectorView view1{ v };
    view1.size();
    view1.front();
    view1.empty();
    view1.back();
    view1.data();

    auto words = std::istringstream{ "today is yesterday’s tomorrow" };
    auto istream_view = std::views::istream<std::string>(words);;
    std::ranges::view_interface<std::ranges::basic_istream_view<std::basic_string<char, std::char_traits<char>, std::allocator<char> >, char, std::char_traits<char> > > view2 = istream_view;

    return 0;
}
```

For `std::ranges::view_interface` without `size()` I used `istream_view`.

For `std::ranges::view_interface` with`size()` I used `VectorView` from https://en.cppreference.com/w/cpp/ranges/view_interface 
I also called `size()`, `front()`, `empty()`, `back()`,  `data()` because the compiler didn't generate these methods by default and Natvis showed `unknown size`.

My results:
изображение

As I can see Natvis never calls methods. It works great only for memory variables and intrinsics.
iota_view example: 
```
<Intrinsic Optional="true" Name="size" Expression="(size_t)(_Bound - _Value)" />
```

And because `std::ranges::view_interface` is dependent from `_Derived`, we can't create an intrinsic for its size.
Am I correct and we always will receive "???" or "This expression has side effects and will not be evaluate"?

### DIFF
--- a/stl/debugger/STL.natvis
+++ b/stl/debugger/STL.natvis
@@ -2001,19 +2001,6 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
     </Expand>
   </Type>
 
-  <Type Name="std::ranges::view_interface&lt;*&gt;">
-    <!-- the view_interface has size() -->
-    <DisplayString Optional="true">{{ size={size()} }}</DisplayString>
-    <!-- the view_interface doesn't have size() -->
-    <DisplayString Optional="true">unknown size</DisplayString>
-    <Expand>
-      <Item Optional="true" Name="empty">empty()</Item>
-      <Item Optional="true" Name="front">front()</Item>
-      <Item Optional="true" Name="back">back()</Item>
-      <Item Optional="true" Name="data">data()</Item>
-    </Expand>
-  </Type>
-
   <Type Name="std::ranges::dangling">
     <DisplayString>dangling</DisplayString>
   </Type>
--- a/stl/debugger/STL.natvis
+++ b/stl/debugger/STL.natvis
@@ -2002,7 +2002,10 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
   </Type>
 
   <Type Name="std::ranges::view_interface&lt;*&gt;">
-    <DisplayString>{{ size={size()} }}</DisplayString>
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
Hmm, this does seem like an inherently not-useful/not-implementable visualizer. I think we should simply remove it, and then if someone identifies a scenario in which it can be usefully implemented, we can add a working version later.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for fixing this! We still need to triple-mirror this into VS so it'll take effect, but this gets us one step closer. :eyes: :magic_wand: :smile_cat:

PR #4836, opened 2024-07-12
`move_iterator` visualizer should use `_Current`

### LINKED ISSUE #4829: `STL.natvis`: `move_iterator` visualizer should use `_Current`, not `current`
https://github.com/microsoft/STL/blob/ef1d621d51263285aff8e560a214f5477d63d687/stl/inc/xutility#L4149-L4152
https://github.com/microsoft/STL/blob/ef1d621d51263285aff8e560a214f5477d63d687/stl/debugger/STL.natvis#L1522-L1527

And traditionally, we've used `Name="[current]"` to indicate that we're displaying a pretty name for an ugly data member.

Regression introduced by #1080 on 2020-07-27 in VS 2019 16.8.

Reported by @xiangfan-ms.

fsb4000:
Fixes #4829

Test code:
```c++
#include <iterator>
#include <string>
#include <vector>

int main()
{
    std::vector<std::string> v{ "This", "_", "is", "_", "an", "_", "example" };
    auto begin = std::make_move_iterator(v.begin());
    return 0;
}
```

Result:
изображение

### DIFF
--- a/stl/debugger/STL.natvis
+++ b/stl/debugger/STL.natvis
@@ -1520,9 +1520,9 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 
   <Type Name="std::move_iterator&lt;*&gt;">
-      <DisplayString>move_iterator {current}</DisplayString>
+      <DisplayString>move_iterator {_Current}</DisplayString>
       <Expand>
-          <Item Name="current">current</Item>
+          <Item Name="[current]">_Current</Item>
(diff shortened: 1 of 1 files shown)


StephanTLavavej (approved):

StephanTLavavej:
Oops, I thanked you in the commit message itself, instead of a comment. First time I've ever made that mistake :zany_face: :no_entry: :brain:

PR #4486, opened 2024-03-16
Remove attempt on unsized find vectorization...

AlexGuteniev:
... except for 1-byte element, where we can still yield to `memchr` 

Fixes #4485

Drive-by: signed coverage doesn't make difference here

PR #4839, opened 2024-07-13
Vectorize `bitset` from string construction

AlexGuteniev:
# 📜 The approach

We need to:
*  Compare against '0' and '1' (or other characters) and report if there's neither of them
*  Reverse the order of characters
*  Convert byte mask or `wchar_t` to bit mask
*  Handle bitset and string odd length tails
*  Handle too short and too long string properly

It is implemented as follows:
* We compare against '1' vector to get byte or `wchar_t` mask. We xor the input with '0' vector, and check if anything not masked in the resulting byte mask is zero. This way we both validate, and have the mask.
* If we have longer string than bitset, we don't do further steps than byte or `wchar_t` mask, as we only need to validate,
* To both reverse the order, and convert `wchar_t` mask to byte mask we use `(v)pshufb` (`..._shuffle_epi8`). The AVX2 version is not cross-lane, and additional cross-lane instruction is somewhat expensive, so we fix up that with bits, rather than now.
* `(v)pmovmskb` is the instruction that does byte mask to bit mask conversion, it is emitted by `..._movemask_epi8` intrinsic.
* Now we do rotations to fix up the missing AVX2 cross-lane reordering
* Then put that to bitset. We assume padding, even if bitset length is smaller.
* Repeat until string ends
* Write zeros to remaining portion of bitset, if there's something left

# ⚠️ Padding assumption

To avoid unnecessary complication, and have a bit better performance, an assumption was made that bitset has at least a multiple of 2 padding on SSE4.2 code path, and at lest a multiple of 4 padding on AVX2 code path.

For large bitset we have units of `uint64_t`, and it is not reasonable to have unit less than native integer size, and AVX2 code path has the threshold of 256, since it is not efficient on smaller sizes.

For small bitset we have units of `uint32_t`, but there's a vNext comment to consider smaller units. But we have a threshold on the whole vectorization as 16 bits, so 8 or less bits bitset that could potentially be less that 2 bytes, would not vectorize.

# ⚖️ Threshold selection

According to my benchmarks, the vectorized approach benefits for 16 bits, but doesn't for 8 bits. I tried 12, and it seems in favor of non-vectorization. Let it be round up to 16 then.

# 📏 String length

There are overloads with implicit and explicit length. 

For implicit length the non-vectorized implementation calls `strlen`/`wcslen` (via character traits).

Hypothetically, vectorized implementation could determine length during conversion. We don't do that here, because:
 * This cannot be done ASAN-friendly, see #4486
 * Even if it was possible, this adds complication to already not very simple algorithm

Unfortunately, this causes implicit length overload to benefit much less from vectorization.

# ⏱️ Benchmark results

Benchmark                                                     |     main  |     this
--------------------------------------------------------------|-----------|-----------
**explicit length, 1-byte chars**
BM_bitset_from_string<length_type::char_count, 15, char>      |  1813 ns  |  1778 ns
BM_bitset_from_string<length_type::char_count, 16, char>      |  1794 ns  |   921 ns
BM_bitset_from_string<length_type::char_count, 36, char>      |  2230 ns  |   651 ns
BM_bitset_from_string<length_type::char_count, 64, char>      |  2075 ns  |   274 ns
BM_bitset_from_string<length_type::char_count, 512, char>     |  2026 ns  |  78.1 ns
BM_bitset_from_string<length_type::char_count, 2048, char>    |  2072 ns  |  65.7 ns
**explicit length, 2-byte chars**
BM_bitset_from_string<length_type::char_count, 15, wchar_t>   |  1777 ns  |  1774 ns
BM_bitset_from_string<length_type::char_count, 16, wchar_t>   |  1786 ns  |   979 ns
BM_bitset_from_string<length_type::char_count, 36, wchar_t>   |  2230 ns  |   741 ns
BM_bitset_from_string<length_type::char_count, 64, wchar_t>   |  1988 ns  |   349 ns
BM_bitset_from_string<length_type::char_count, 512, wchar_t>  |  1970 ns  |   189 ns
BM_bitset_from_string<length_type::char_count, 2048, wchar_t> |  1912 ns  |   143 ns
**implicit length, 1-byte chars**
BM_bitset_from_string<length_type::null_term, 15, char>       |  2825 ns  |  2777 ns
BM_bitset_from_string<length_type::null_term, 16, char>       |  2118 ns  |  1439 ns
BM_bitset_from_string<length_type::null_term, 36, char>       |  2571 ns  |  1173 ns
BM_bitset_from_string<length_type::null_term, 64, char>       |  2710 ns  |  1018 ns
BM_bitset_from_string<length_type::null_term, 512, char>      |  2482 ns  |   606 ns
BM_bitset_from_string<length_type::null_term, 2048, char>     |  2407 ns  |   566 ns
**implicit length, 2-byte chars**
BM_bitset_from_string<length_type::null_term, 15, wchar_t>    |  2098 ns  |  2102 ns
BM_bitset_from_string<length_type::null_term, 16, wchar_t>    |  2515 ns  |  1613 ns
BM_bitset_from_string<length_type::null_term, 36, wchar_t>    |  2128 ns  |  1246 ns
BM_bitset_from_string<length_type::null_term, 64, wchar_t>    |  2739 ns  |  1055 ns
BM_bitset_from_string<length_type::null_term, 512, wchar_t>   |  2491 ns  |   719 ns
BM_bitset_from_string<length_type::null_term, 2048, wchar_t>  |  2407 ns  |   639 ns


stl/src/vector_algorithms.cpp
```diff
@@ -3664,5 +3656,263 @@ __declspec(noalias) void __stdcall __std_bitset_to_string_2(
...
+                if (!_Traits::_Check(_Val, _Ex1, _Dx0)) {
+                    return false;
+                }
+
+                _Out(_Ex1);
+            }
+        }
+
+        template <class _Traits, class _Elem>
+        static bool _Impl(void* _Dest, const _Elem* _Src, size_t _Size_bytes, size_t _Size_bits, size_t _Size_chars,
```
StephanTLavavej, line 3813: This is within an unnamed namespace, so it shouldn't be `static`.

StephanTLavavej:
I finished reviewing the benchmark but still need to review the product and test code. I pushed changes since I've got to work on the toolset update next and I'll forget my context.

benchmarks/src/bitset_from_string.cpp
```diff
@@ -0,0 +1,93 @@
...
+        std::array<charT, actual_size> result;
+
+        for (auto dest = result.begin(); dest != result.end();) {
+            for (size_t i = 0; i != N; ++i, ++dest) {
+                *dest = static_cast<charT>(dis(rnd));
+            }
+
+            *dest = charT{'\0'};
+            ++dest;
+        }
```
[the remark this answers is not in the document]
AlexGuteniev: What about Fizz Buzz pattern then?

```c++
for (size_t i = 0; i < actual_size; ++i) {
    if ((i % (N + 1)) == N) {
        result[i] = charT{'\0'};
    } else {
        result[i] = static_cast<charT>(dis(rnd));
    }
}
```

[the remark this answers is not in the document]
StephanTLavavej: Sure, that would be an improvement! (I'd drop the parens around the modulo - equality has well-known lower precedence than the multiplicative operators.)

StephanTLavavej:
I've pushed a conflict-free merge with main, no clang-format regen needed, and the FizzBuzz change.

stl/inc/bitset
```diff
@@ -115,6 +123,30 @@ public:
 private:
     template <class _Traits, class _Elem>
     _CONSTEXPR23 void _Construct(const _Elem* const _Ptr, size_t _Count, const _Elem _Elem0, const _Elem _Elem1) {
+#if _USE_STD_VECTOR_ALGORITHMS
+        constexpr size_t _Bitset_from_string_vector_threshold = 16;
+        if constexpr (_Bits >= _Bitset_from_string_vector_threshold
+                      && _Is_specialization_v<_Traits, char_traits> && sizeof(_Elem) <= 2) {
```
AlexGuteniev, line 129: ## ⚠️ 🦄🦄🦄  ⚠️
Maybe it would be best to wait for #4951  and replace `_Is_specialization_v<_Traits, char_traits>` with `_Is_implementation_handled_char_traits<_Traits>`

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -1038,6 +1047,33 @@ void test_bitset(mt19937_64& gen) {
...
            == U"000000000001111111011011100101110101001100001110110010101000011001000010000"); // not vectorized
 
+    assert(bitset<0>("").to_ullong() == 0);
+    assert(bitset<0>("1").to_ullong() == 0);
+    assert_throws_inv([] { (void) bitset<0>("x"); });
+
+    assert(bitset<45>("101110000000111010001011100101001111111111111").to_ullong() == 0x1701D1729FFFULL);
+    assert(bitset<45>("110101001100001110110010101000011001000010000").to_ullong() == 0x1A9876543210ULL);
+    assert(bitset<45>("111").to_ullong() == 0x7);
+    assert_throws_inv([] { (void) bitset<45>("11x11"); });
```
StephanTLavavej, line 1057: We need test coverage that the vectorized implementation rejects bogus characters outside of the bitset's size. (`bitset<0>("x")` falls below the vectorization threshold.)

StephanTLavavej (approved):

AlexGuteniev:
> and a single 0.91 scenario is nothing to worry about. 

This is one of the four 15 cases which are just below vectorization threshold. Just one of them has exactly 1.00 as expected

StephanTLavavej:

StephanTLavavej:
Thanks ~~a bit~~ a lot!

:joy_cat: :zero: :one:
===

PR #4857, opened 2024-07-24
Temporarily restore support for Win7 / Server 2008 R2

StephanTLavavej:
# :world_map: Overview
My ultrabosses have decided that dropping support for targeting Win7 / Server 2008 R2 in VS 2022 17.12 is too aggressive, so I'm being required to restore it. However, dropping support has been *approved* for the 18.0 Preview 1 release (whenever that will happen, and whatever its major version branding will be), so this restoration is *temporary*.

**Windows 7 and Server 2008 R2 are insecure operating systems, which haven't been receiving security updates for many months. Targeting them is making the world a worse place.**

# :scroll: Commits
* Revert #4742.
  + Resolve conflicts with #4751:
  + In `yvals_core.h`, we dropped `_STL_WIN32_WINNT_WIN7`, then we dropped `_STL_WIN32_WINNT_WINBLUE`.
  + In `atomic_wait.cpp`, we dropped `_ATOMIC_WAIT_ON_ADDRESS_STATICALLY_AVAILABLE` machinery, then we dropped `__std_atomic_compare_exchange_128_fallback`.
* \[cherry-pick\] Cauterize usage of Win8 `GetCurrentPackageId`.
  + As previously explained:
    > This is technically a bonus cleanup - instead of removing machinery dealing with the conditional absence of Win8 APIs, it's removing machinery dealing with the conditional *presence* of Win8 APIs.
  + We should preserve this improvement, because it doesn't affect Win7 targeting.
  + Conflict resolutions: We want to remove `DEFINEFUNCTIONPOINTER`/`STOREFUNCTIONPOINTER` for `GetCurrentPackageId` next to where the machinery for `GetSystemTimePreciseAsFileTime` has been re-added.
* Restore the `system_clock` comment change.
  + We don't need to mention the `GetSystemTimeAsFileTime` fallback for Win7 here.
* Restore the `P1135R6_atomic_wait` test changes.
  + This made it self-contained (instead of centralized in `test_atomic_wait.hpp`).
  + This also made it not bother to verify the API level. `__std_atomic_set_api_level` exists for test purposes only, and our dev/test machines are all modern.
* Rename `P1135R6_atomic_wait_vista` to `P1135R6_atomic_wait_win7`.
  + The directory name was confusing, since we've totally dropped Vista support.
* Fuse `test_atomic_wait.hpp` into `P1135R6_atomic_wait_win7`.
  + No changes other than dropping the duplicate banner and `#pragma once`.
  + Note divergence: #4221 (merged 2024-01-17, before we removed Win7 on 2024-06-24) added compile-only test coverage for incomplete types to `P1135R6_atomic_wait` only. There's no need to replicate that coverage.

Making `P1135R6_atomic_wait` and `P1135R6_atomic_wait_win7` independent obviously duplicates test code, but will make final removal simpler (in 18.0p1).

# :warning: Note to self:
I will need to revert some of MSVC-PR-559637's internal changes, specifically "NON-GitHub: Link `src/vctools/PDB` binaries against `synchronization.lib`."


MahmoudGSaleh (approved):
LGTM. Thank you!

jovibor:
A PR that was fraught with pain and sorrow😶‍🌫️...

darkalexis1990:
Windows Server 2008 R2 is still supported ->[Grandfathered](https://en.wikipedia.org/wiki/Grandfather_clause)[[7]](https://en.wikipedia.org/wiki/Windows_Server_2008_R2#cite_note-PA2-7) Premium Assurance security update support until January 13, 2026.[[8]](https://en.wikipedia.org/wiki/Windows_Server_2008_R2#cite_note-PA-8)[[9]](https://en.wikipedia.org/wiki/Windows_Server_2008_R2#cite_note-PA3-9)

PR #4859, opened 2024-07-24
`README.md`: Add instructions for building ARM and ARM64.

### LINKED ISSUE #1211: README.md: Consider explaining how to build ARM and ARM64
We merged #873, removing an incomplete explanation of how to build the STL for ARM and ARM64. We should consider adding a complete explanation, which would involve selecting ARM and ARM64 support in the VS installer. #3730 added CMake Presets, centralizing everything else.

fsb4000:
Fixes #1211 
ARM
ARM64
arm64ec

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,8 +143,10 @@ Just try to follow these rules, so we can spend more time fixing bugs and implem
 
 1. Install Visual Studio 2022 17.11 Preview 3 or later.
     * Select "Windows 11 SDK (10.0.22621.0)" in the VS Installer.
-    * Select "MSVC v143 - VS 2022 C++ ARM64/ARM64EC build tools (Latest)" in the VS Installer if you would like to build ARM64/ARM64EC library.
-    * Select "MSVC v143 - VS 2022 C++ ARM build tools (Latest)" in the VS Installer if you would like to build ARM library.
+    * Select "MSVC v143 - VS 2022 C++ ARM64/ARM64EC build tools (Latest)" in the VS Installer
+    if you would like to build the ARM64/ARM64EC target.
+    * Select "MSVC v143 - VS 2022 C++ ARM build tools (Latest)" in the VS Installer
+    if you would like to build the ARM target.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -180,6 +188,30 @@ To build the x64 target (recommended):
 3. `cmake --preset x64`
 4. `cmake --build --preset x64`
 
+To build the ARM target:
+
+1. Open an "x64 Native Tools Command Prompt for VS 2022 Preview".
+2. Change directories to the previously cloned `STL` directory.
+3. `"C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Auxiliary\Build\vcvarsall.bat" amd64_arm`
```
cpplearner, line 195: This assumes that VS 2022 is installed under `C:\Program Files`, which is not always the case.

fsb4000: изображение

README.md
```diff
@@ -180,6 +188,30 @@ To build the x64 target (recommended):
 3. `cmake --preset x64`
 4. `cmake --build --preset x64`
 
+To build the ARM target:
+
+1. Open an "x64 Native Tools Command Prompt for VS 2022 Preview".
+2. Change directories to the previously cloned `STL` directory.
+3. `"%VSINSTALLDIR%VC\Auxiliary\Build\vcvarsall.bat" amd64_arm`
```
StephanTLavavej, line 195: This results in a horrible mess for the INCLUDE, PATH, etc. environment variables. It works, but I think it's easier to depict the default path and tell people to adjust accordingly if they've customized the location.

StephanTLavavej, line 195: vcvarsall.bat understands `x64_MEOW` as a synonym, and we try to prefer the vendor-neutral and less-confusing x64 whenever possible.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:scroll: :muscle: :mechanical_arm:
===

PR #4864, opened 2024-07-26
Simplify the comparison of sizes of different ranges

hewillk:
Explicit casting of the result of `ranges::size` is too verbose; simply comparing the values ​​of `ranges::distance` is sufficient as they are guaranteed to be signed integers.
This aligns with the implementation in `ranges::is_permutation`:
https://github.com/microsoft/STL/blob/8657d15b3ddfacb121e88e307f9a18b3fa379185/stl/inc/algorithm#L1084-L1088


StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for this nice simplification! :magic_wand: :smile_cat: :black_cat:

PR #4874, opened 2024-08-02
Allow spectre build of debug libraries

amyw-msft:
Historically, /Qspectre was only supported in optimized code so spectre-mitigated debug builds were disabled. This is now supported, so this PR enables the build path. Spectre-mitigated debug libraries are useful primarily to reduce friction for customers that would like to release debug versions of their binaries, but have policies preventing production or publication of binaries that are not spectre-mitigated.

Mirrored as MSVC-PR-569343.

### DIFF
--- a/stl/msbuild/stl_2/xmd/dirs.proj
+++ b/stl/msbuild/stl_2/xmd/dirs.proj
@@ -6,9 +6,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
     <Import Project="$(MSBuildThisFileDirectory)..\..\..\..\..\crt-common.settings.targets" />
 
-    <ItemGroup Condition="('$(SpectreBuildMode)' == '' or
-                           '$(Configuration)' == 'chk' or
-                           '$(Configuration)' == 'dbg')">
+    <ItemGroup>
         <ProjectFile Include="msvcp_2_xmd_app.vcxproj" />
         <ProjectFile Include="msvcp_2_xmd_kernel32.vcxproj" />
         <ProjectFile Include="msvcp_2_xmd_onecore.vcxproj" />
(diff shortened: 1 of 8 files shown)


stl/msbuild/stl_2/xmd/dirs.proj
```diff
@@ -6,9 +6,7 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
     <Import Project="$(MSBuildThisFileDirectory)..\..\..\..\..\crt-common.settings.targets" />
 
-    <ItemGroup Condition="('$(SpectreBuildMode)' == '' or
-                           '$(Configuration)' == 'chk' or
-                           '$(Configuration)' == 'dbg')">
+    <ItemGroup>
         <ProjectFile Include="msvcp_2_xmd_app.vcxproj" />
         <ProjectFile Include="msvcp_2_xmd_kernel32.vcxproj" />
         <ProjectFile Include="msvcp_2_xmd_onecore.vcxproj" />
```
StephanTLavavej, line 12: (Also commented in MSVC-PR-569343, expand context immediately below.)

I observe that 6 files are still guarding netfx projects for spectre. Is this intentional?

StephanTLavavej (approved):

stl/msbuild/stl_2/xmd/dirs.proj
amyw-msft: Yep, this is only to update the binaries we ship in VS. CLR doesn't need spectre-mitigated debug libraries.

StephanTLavavej:
Marking as blocked until the MSVC-PR has a clean setup build.

CaseyCarter:
Verified the changes still match MSVC-PR-569343 as of iteration 4.

StephanTLavavej:
:ghost: :lady_beetle: :ghost:
===

PR #4878, opened 2024-08-05
Enforce [alg.three.way]/2 mandate

### LINKED ISSUE #4877: `<algorithm>`: `lexicographical_compare_three_way` missing _Mandates_ check
There is a _Mandates_ in [[alg.three.way]](https://eel.is/c++draft/alg.three.way), which seems to be missing in the MSVC-STL implementation:

> _Mandates_: `decltype(comp(*b1, *b2))` is a comparison category type.

```cpp
#include <algorithm>

struct Hack {
  Hack(auto);
  bool operator==(int) const;
};

struct Cmp {
  Hack operator()(const auto&, const auto&) const;
};

int main() {
  int x[] = {42};
  auto r = std::lexicographical_compare_three_way(x, x+1, x, x+1, Cmp{}); // only ok in MSVC-STL
}
```
https://godbolt.org/z/P6jPedes9

CaseyCarter:
Fixes #4877


StephanTLavavej (approved):

hewillk (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for requiring users to be good kitties! :cat2: :smirk_cat: :smile_cat:

Issue #1900, opened 2021-05-04
@statementreply [observes](https://github.com/microsoft/STL/pull/1870#discussion_r617528729) that the current `[time.format]` implementation calls `put_time` once for each formatting field - we should investigate improving performance here (perhaps by leveraging internal machinery).


mnatsuhara:
[Observed](https://github.com/microsoft/STL/pull/1870#discussion_r617528729) by @statementreply

The current `[time.format]` implementation calls `put_time` once for each formatting field - we should investigate improving performance here (perhaps by leveraging internal machinery).

PR #4824, opened 2024-07-09
Toolset update: VS 2022 17.11 Preview 3

StephanTLavavej:
# :scroll: Changelog
- Infrastructure improvements:
  * Updated dependencies. [#4824](https://github.com/microsoft/STL/pull/4824)
    + Updated build compiler to VS 2022 17.11 Preview 3 (now required).
    + Updated CMake to 3.29 (now required).

# :gear: Commits
* PowerShell 7.4.3.
* Move to eastus2.
  + This is required for 1ES Hosted Pools quota.
* VS 2022 17.11 Preview 3, now required.
* CMake 3.29, now required.
* Remove workaround for the MSVC-internal toolset.
* New pool.

PR #4883, opened 2024-08-08
Fix put_time() crash on invalid struct tm data

### LINKED ISSUE #924: <iomanip>: put_time crashes, allocating a huge memory block
Environment:
Version 16.6.2
Open x64 Native command prompt for VS 2019.
Microsoft (R) C/C++ Optimizing Compiler Version 19.26.28806 for x64

Problem:
std::put_time crashed if all "tm" structure fields are equal -1.
If calling localtime_s with time_t > _MAX__TIME64_T then all "tm" fields filled by -1.
In real application put_time allocate tons of memory and application hang for some time.

```
#include <iomanip>
#include <sstream>
#include <iostream>
int main(int, char**)
{
    std::time_t t = 335303598060; // t > _MAX__TIME64_T
    tm t_tm;
    auto err = localtime_s(&t_tm, &t); // error, all fileds in t_tm = -1
    std::cout << "errno = " << err << "\n"; // errno = 22
    std::stringstream ss;
    ss << std::put_time(&t_tm, "%Y-%m-%d %H:%M:%S"); // crash, allocate tons on memory
    std::cout << ss.str() << "\n";
    return 0;
}

c:\~repro\cl /EHsc repro.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.26.28806 for x64

repro.cpp
Microsoft (R) Incremental Linker Version 14.26.28806.0

/out:repro.exe
repro.obj

c:\~repro>repro.exe
errno = 22
crash
```

Expected behavior:
put_time not crashed.

GCC/Clang (checked in Wandbox):
replace
```
    auto err = localtime_s(&t_tm, &t); // error, all fields in t_tm = -1
    std::cout << "errno = " << err << "\n"; // errno = 22
```
by
```
    localtime_r(&t, &t_tm);
    std::cout << "errno = " << errno << "\n";
```
Result:
```
errno = 0
12595-05-08 01:21:00
```

Why MS localtime_s has limitations for time_t maximum value?
As result we have an unsafe code, it is enough to pass a date >3000 year and the application crashes.

### LINKED ISSUE #4882: <iomanip>: std::put_time should not crash on invalid/out-of-range tm struct values
# Describe the bug
If an invalid or out-of-range tm struct is passed to `std::put_time()` it will lead to crash.

# Command-line test case
```
C:\Temp>type repro.cpp
#include <iostream>
#include <iomanip>
#include <sstream>
#include <ctime>

int main() {
    using namespace std;

    time_t t = time(nullptr);
    tm currentTime;
    localtime_s(&currentTime, &t);
    currentTime.tm_hour = 25; // set invalid hour

    stringstream ss;
    ss << put_time(&currentTime, "%Y-%m-%d-%H-%M");
}

C:\Temp>cl /EHsc /W4 /WX /std:c++20 .\repro.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.36.32546 for x64

repro.cpp
Microsoft (R) Incremental Linker Version 19.36.32546.0

/out:repro.exe
repro.obj

C:\Temp>.\repro.exe
!!!App Crashed!!!
```

# Expected behavior
The program should NOT crash.

# Additional context
The crash is happening because of `strftime` which validates the data and if something is wrong it leads to crash if no handler is set.

TheStormN:
Fixes #4882
Fixes #924

My next PR about fixing `std::put_time` crashes.

If all goes well I will try to also fix the performance issues mentioned in #3575 and #1900 before VS 17.12 cut.

UPDATE: The result now outputs single ? for every invalid tm struct data based on specifier. For example if tm struct contains invalid year and the format string is "%Y-%m-%d-%H-%M", the output will be like `?-08-10-00-45`.


tests/std/tests/Dev11_0836436_get_time/test.cpp
```diff
@@ -799,13 +801,13 @@ void test_invalid_argument() {
     {
         wstringstream wss;
         wss << put_time(&currentTime, L"%Y-%m-%d-%H-%M");
-        assert(wss.rdstate() == ios_base::badbit);
+        assert(wss.rdstate() == ios_base::failbit);
```
StephanTLavavej, line 804: I'm concerned about this behavior change and can't find justification for it in the Standard. Please see my comment at https://github.com/microsoft/STL/issues/4882#issuecomment-2276788199 .

Note that this section of the test is calling `_set_invalid_parameter_handler(test_invalid_parameter_handler)` may have been incorrect to start with.

stl/inc/xloctime
StephanTLavavej, line 1: There are test failures.

* Please inspect the Azure Pipelines logs to find all of the failing tests.
  + You can click in the UI to see each individual failure and its emitted errors, or you can download the XML logs and decode them with `tools/scripts/print_failures.py`. Feel free to ask for assistance on Discord.
* After attempting to fix the test failures, run the affected tests locally and ensure that they're passing before pushing changes.
  + This is both much, much faster for you to iterate on your changes, and helps avoid consuming CI system resources (our budget for test runs is large but finite - don't worry about consuming it when pushing changes, we simply want to avoid people using the CI as a substitute for local test runs which is wildly inefficient).
  + Running a subset of tests is extremely fast, just pass the relevant directories to `stl-lit.py`.

My `print_failures.py` can be used to decode local test logs too (they're emitted in JSON format). Here's my usual incantations, where `%STL%` is my repo root `D:\GitHub\STL`. As written, this will run *all tests* (which takes an hour-ish even on my 16C/32T machine); replace `%STL%\tests\std %STL%\tests\tr1 %STL%\llvm-project\libcxx\test` with `%STL%\tests\std\tests\DIRECTORY_ONE %STL%\tests\std\tests\DIRECTORY_TWO` etc. to run a couple of directories:

```
python tests\utils\stl-lit\stl-lit.py -o testing_x64.log -Dnotags=ASAN --order=random %STL%\tests\std %STL%\tests\tr1 %STL%\llvm-project\libcxx\test

python %STL%\tools\scripts\print_failures.py testing_x64.log
```

stl/inc/xloctime
```diff
@@ -666,6 +666,57 @@ _NODISCARD constexpr bool _Is_valid_strftime_specifier(const char _Specifier) {
...
+        return false;
+    }
+
+    // days since January 1 - [0, 365]
+    if (_Pt->tm_yday < 0 || _Pt->tm_yday > 365) {
+        return false;
+    }
+
+    // daylight savings time flag
+    // _Pt->tm_isdst - no need from validation as UCRT treats it as boolean
```
StephanTLavavej, line 715: Comment typo, presumably "for validation" was intended.

stl/inc/xloctime
```diff
@@ -666,6 +666,57 @@ _NODISCARD constexpr bool _Is_valid_strftime_specifier(const char _Specifier) {
...
+
+    return true;
```
StephanTLavavej, line 717: I observe that this logic could be considerably condensed into a single chain:

```cpp
return _Pt
    && ONE_LO <= _Pt->ONE && _Pt->ONE <= ONE_HI // DATA MEMBER ONE - [ONE_LO, ONE_HI]
    && TWO_LO <= _Pt->TWO && _Pt->TWO <= TWO_HI // for example:
    && 0 <= _Pt->tm_min && _Pt->tm_min <= 59 // minutes after the hour - [0, 59]
    // etc...
```

tests/std/tests/Dev11_0836436_get_time/test.cpp
```diff
@@ -944,3 +946,265 @@ void test_gh_4820() {
...
+
+    // Check tm_sec - START
+    localtime_s(&currentTime, &t);
+    currentTime.tm_sec = -1; // set invalid lower limit
+
+    {
+        wstringstream wss;
+        wss << put_time(&currentTime, L"%Y-%m-%d-%H-%M");
+        assert(wss.rdstate() == ios_base::failbit);
+    }
```
StephanTLavavej, line 963: The following code is highly repetitive. Could we extract this into a helper function, and tell it which data member to damage to what value? That could be done with either a series of lambdas, or a constant table of pointers-to-data-members and ints.

Usually having somewhat repetitive test code is fine (makes it easier to figure out what the test is doing, and debug into it, compared to having to dig through layers of test abstractions), but I think the volume here is worth reducing.

stl/inc/xloctime
```diff
@@ -666,6 +666,57 @@ _NODISCARD constexpr bool _Is_valid_strftime_specifier(const char _Specifier) {
     return false;
 }
 
+_NODISCARD constexpr bool _Is_valid_strftime_tm_data(const tm* _Pt) {
```
StephanTLavavej, line 669: `_Pt` could be top-level `const`. We're not totally disciplined about this but we try to use `const` as much as possible in new code.

StephanTLavavej (changes requested):

TheStormN:
The CI failed and it would seem that this validation is expected to be done ad-hoc based on the formatting string being passed. I.e. if there are invalid values into the tm struct, they can be ignored in case the formatting string does not require them.

I guess I underestimated the issue at hand. Doing per specifier validation should not be too hard, Will try to alter it tomorrow.

tests/std/tests/Dev11_0836436_get_time/test.cpp
TheStormN: Yep, I've also replied on the issue how I interpret the things.

Also you are right, `_set_invalid_parameter_handler` is no longer needed here as I validate the struct before letting `strftime` to crash or call the handler if set(as is the case here).

stl/inc/xloctime
TheStormN: Thanks for the info. Yep when I happen to have failures I try to run the debug the tests locally by following the main repo readme file. Thanks for the additional tips about `print_failures.py`. This part seems to be missing in the main docs.

P.S. In general I find debugging(using IDE debugger) tests in this repo for extremely inconvenient given the instructions on the main page. It would be super nice if I can just select and run specific test from the IDE. :)

stl/inc/xloctime
TheStormN: I'm not sure I understand the comment. It is already `const`, so I'm not sure how to interpret "top-level" `const`. Perhaps you mean to also make the pointer const, not just the data, like `const tm* const _Pt`?

StephanTLavavej:
That makes sense - regardless of my open question about what we should do in the case of invalid values, I definitely didn't see any wording that suggested that invalid values that aren't needed by the formatting string are problematic in any way, so per-specifier validation does appear to be necessary. Good catch!

stl/inc/xloctime
StephanTLavavej: Yep, I have a personal todo about explaining `print_failures.py` (which I'll try to get to after working through the review backlog).

To debug an individual test, I compile it on the command line, adding options to emit debug info, and then I invoke the debugger. Extracting the command line from the logs is simple (this is the most important part of any log, i.e. figure out what it did, and our logs are good about this). The additional incantations look like this:

```
C:\Temp>type meow.cpp
```
```cpp
#include <print>
using namespace std;

int main() {
    for (int i = 0; i < 5; ++i) {
        println("Debug into this! {}", i);
    }
}
```
```
meow.cpp
Debug into this! 0
Debug into this! 1
Debug into this! 2
Debug into this! 3
Debug into this! 4

C:\Temp>cl /EHsc /nologo /W4 /std:c++latest /MTd /Od /Zi /Fdmeow.pdb meow.cpp
meow.cpp

C:\Temp>devenv meow.exe meow.cpp
```

All that's necessary is to add `/Zi` for debug info (I additionally use `/Fd` to control the PDB name), then invoke the IDE (`devenv`) to launch the EXE (first argument) and open the source file (second argument). This allows the program to be restarted at will with F5.

I don't know how to make the test suite more friendly to IDE users since I don't use the IDE beyond this technique. I am aware of VSCode's JSON files to integrate with its test tab or whatever, but I've never figured out how to configure that (I spent a bit of time on it but bounced off due to complexity; maybe I should try again).

stl/inc/xloctime
StephanTLavavej: Yes, that's what top-level means - `const` that applies to the pointer itself, instead of what it points to.

stl/inc/xloctime
TheStormN: Yes, this technique is explained in the main readme and this is what I meant by "extremely inconvenient". :)

Perhaps CTest would be an option here for better integration into the IDE, but I will look into it another time.

tests/std/tests/Dev11_0836436_get_time/test.cpp
StephanTLavavej: Also, if "store unspecified characters and succeed" is determined to be the right approach, then instead of clamping values to zero (or maximum), writing the "expected" number of characters but using a fallback like `'*'` might be ideal, since it'll indicate that something was wrong with the input (e.g. `"2024-08-**"`).

tests/std/tests/Dev11_0836436_get_time/test.cpp
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+#include <algorithm>
```
TheStormN, line 4: CI format checks made me put this at the top. I guess it is just alphabetical order.

TheStormN:
A single test on `x86 x86 8` failed:
```
1: Failed Tests (1):
1:   libc++ :: std/thread/thread.condition/thread.condition.condvarany/notify_one.pass.cpp:2
```

However, this seems like a flaky test, unrelated to my PR. @StephanTLavavej could you manually rerun this specific step, as it seems user/pass are required and I can't do it.

CaseyCarter:
> A single test on `x86 x86 8` failed:
> 
> ```
> 1: Failed Tests (1):
> 1:   libc++ :: std/thread/thread.condition/thread.condition.condvarany/notify_one.pass.cpp:2
> ```
> 
> However, this seems like a flaky test, unrelated to my PR. @StephanTLavavej could you manually rerun this specific step, as it seems user/pass are required and I can't do it.

I've submitted #4885 to disable this test tempermanently, and rerun the failed job from your CI run.

tests/std/tests/Dev11_0836436_get_time/test.cpp
CaseyCarter: Yes, our clang-format setup sorts headers.

tests/std/tests/Dev11_0836436_get_time/test.cpp
```diff
@@ -944,3 +909,68 @@ void test_gh_4820() {
         assert(wss.str() == L"\x043a%\x043e%\x0448%E\x043a%O\x0430");
     }
 }
+
+void test_gh_4882() {
+    // GH-4882 <iomanip>: std::put_time should set failbit on invalid/out-of-range tm struct, instead of crash
```
StephanTLavavej, line 914: To avoid confusion, the title of this issue should be updated, and then this comment should mirror it. It would be fine to just have the issue say that we shouldn't crash for out-of-range values, no need to say what we should do.

tests/std/tests/Dev11_0836436_get_time/test.cpp
```diff
@@ -944,3 +909,68 @@ void test_gh_4820() {
...
+        localtime_s(&currentTime, &t);
+
+        (currentTime.*field) = value;
+
+        stringstream ss;
+        ss << put_time(&currentTime, format.c_str());
+        assert(ss.rdstate() == ios_base::goodbit);
+        const auto result = ss.str();
+        assert(result.length() == format.length() / 2);
+        assert(std::all_of(result.cbegin(), result.cend(), [](const char c) { return c == '?'; }));
```
StephanTLavavej, line 927: This test is already `using namespace std;` so there should be no `std::` qualification. (Occurs below.)

tests/std/tests/Dev11_0836436_get_time/test.cpp
```diff
@@ -944,3 +909,68 @@ void test_gh_4820() {
...
+        tm currentTime;
+        localtime_s(&currentTime, &t);
+
+        (currentTime.*field) = value;
+
+        stringstream ss;
+        ss << put_time(&currentTime, format.c_str());
+        assert(ss.rdstate() == ios_base::goodbit);
+        const auto result = ss.str();
+        assert(result.length() == format.length() / 2);
```
StephanTLavavej, line 926: Style nitpick: `size()` and `length()` are synonyms for `string` but we always say `size()`. We might be less disciplined about this in test code but that's the overwhelming majority of usage.

tests/std/tests/Dev11_0836436_get_time/test.cpp
```diff
@@ -944,3 +909,68 @@ void test_gh_4820() {
...
+        assert(std::all_of(result.cbegin(), result.cend(), [](const char c) { return c == '?'; }));
+
+        // Narrow conversion is good enough for our ASCII only format strings
+        wstring wformat(format.length(), L' ');
+        transform(format.cbegin(), format.cend(), wformat.begin(), [](const char c) { return wchar_t(c); });
```
StephanTLavavej, line 931: We conventionally avoid functional-style C-semantics casts like `wchar_t(c)`. (They're willing to reinterpret! Try `size_t(ptr)`.) This should be a `static_cast`.

StephanTLavavej (commented):
Thanks, this is looking really good! I'm about to fall asleep so I want to take another look at this, but here are the comments I noticed.

TheStormN:
> Thanks, this is looking really good! I'm about to fall asleep so I want to take another look at this, but here are the comments I noticed.

All comments addressed. Although there seems to be an issue with the format target. Seems that sometimes the IDE is missing some points and the CI failure says I should execute `cmake --build --preset x64 --target format` to fix any issues. However, cmake reports that such target does not exists.

TheStormN:
I've discovered I'm missing some cases. Will update the PR again soon.

stl/inc/xloctime
```diff
@@ -666,6 +666,110 @@ _NODISCARD constexpr bool _Is_valid_strftime_specifier(const char _Specifier) {
...
 
+_NODISCARD constexpr bool _Is_valid_strftime_tm_sec(const tm* const _Pt) noexcept {
+    // seconds after the minute - [0, 60] including leap second
+    return _Pt->tm_sec >= 0 && _Pt->tm_sec <= 60;
+}
+
+_NODISCARD constexpr bool _Is_valid_strftime_tm_min(const tm* const _Pt) noexcept {
+    // minutes after the hour - [0, 59]
+    return _Pt->tm_min >= 0 && _Pt->tm_min <= 59;
+}
```
TheStormN, line 677: I wanted to do these as internal lambdas, but on older standards this is not allowed and because this header is included as part of the public includes I have to make it compatible with all flavors.

TheStormN:
Ok, now I think I've got all the cases covered (and perhaps brain damage). :)

Btw, the tests are also compiling in C++14 mode. Is there any specific reason for that? It would be really nice if we can use some neat C++20 tricks there.

StephanTLavavej:
> Sometimes the IDE is missing some points

I'm not sure what that means - you should configure your editor to clang-format on save, using the same clang-format that ships in VS (ensuring that the exact same version is being used, as its behavior can change with the version). I can provide guidance for VSCode; I think full VS can do it too but I don't know how.

> the CI failure says I should execute `cmake --build --preset x64 --target format` to fix any issues. However, cmake reports that such target does not exists.

What exactly are you seeing? This works for me:

```
D:\GitHub\STL>"C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Auxiliary\Build\vcvarsall.bat" x64
** Visual Studio 2022 Developer Command Prompt v17.11.0-pre.7.0
** Copyright (c) 2022 Microsoft Corporation
[vcvarsall.bat] Environment initialized for: 'x64'

D:\GitHub\STL>cmake --preset x64
-- The CXX compiler identification is MSVC 19.41.34119.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Preview/VC/Tools/MSVC/14.41.34117.0/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Performing Test WINDOWS_SDK_VERSION_CHECK
-- Performing Test WINDOWS_SDK_VERSION_CHECK - Success
-- The ASM_MASM compiler identification is MSVC
-- Found assembler: C:/Program Files/Microsoft Visual Studio/2022/Preview/VC/Tools/MSVC/14.41.34117.0/bin/Hostx64/x64/ml64.exe
-- Searching for VS clang-format
-- Searching for VS clang-format - found
-- Found Python: C:/Users/stl/AppData/Local/Programs/Python/Python312/python.exe (found suitable version "3.12.4", minimum required is "3.12") found components: Interpreter
-- Boost.Math: standalone mode ON
-- Configuring done (3.2s)
-- Generating done (0.2s)
-- Build files have been written to: D:/GitHub/STL/out/x64

D:\GitHub\STL>cmake --build --preset x64 --target format
[1328/1328] C:\WINDOWS\system32\cmd.exe /C "cd /D D:\GitHu...GitHub/STL/tools/format/../../tools/validate/validate.cpp"
```

> Btw, the tests are also compiling in C++14 mode. Is there any specific reason for that? It would be really nice if we can use some neat C++20 tricks there.

Yes - the reason is that we need to ensure that C++14 library features (like `put_time()`) use only C++14 core language features in their implementation, so that users can actually use them under `/std:c++14`. If we tested old library features under newer language modes *only*, then it would be too easy to unintentionally use new compiler features in the product code. (We do test old library features under a mix of old and new modes, because we need to ensure that users can use all modes.)

The exception is that we can use a limited number of core language features in older Standard modes, when compilers allow this and emit suppressible warnings that we're using Future Technology. This is how we can use C++17 `if constexpr` and C++20 `explicit(bool)` everywhere in product code. (Test code is still subject to the warnings unless we suppress them.)

Test code can conditionally activate new logic under `#if _HAS_CXX20` but it should generally do so only when testing C++20-only library features, or testing interaction of old library features with new core language features (e.g. spaceship logic, or whatever). Otherwise we do need to sometimes write more primitive test code than we would like to in an ideal world.

BTW - it isn't a huge deal, but we generally recommend that after code review has begun, you should avoid force-pushing unless it's really important. GitHub makes it more difficult to see what's changed when force pushes happen, which can interfere with incremental reviewing. (It's not the end of the world though, and sometimes force pushes can help reviewing, e.g. by reorganizing messy commit history into logical chunks, or replaying commits to get a complicated merge out of the history.)

stl/inc/xloctime
StephanTLavavej: Does this stuff actually need to be `constexpr`? It's in iostreams, so the top-level is non-`constexpr`, unless I am confused.

TheStormN: Well you did so for the previous PR where I check for invalid specifies, instead of using static local variable, so I thought we are seeking the same for this part. Otherwise I can move them to lambdas, no problem.

TheStormN:
I use VS and on save it didn't reordered the headers. Might be something on my end. Will check again. 

About force pushing. I did so exactly for readability reasons and it was for new commits after the last review. 

When I get home I will show the exact errors I get for the format targets.

TheStormN:
```
** Visual Studio 2022 Developer Command Prompt v17.10.5
** Copyright (c) 2022 Microsoft Corporation
[vcvarsall.bat] Environment initialized for: 'x64'

C:\Program Files\Microsoft Visual Studio\2022\Community>cd D:\Projects\Microsoft\STL

C:\Program Files\Microsoft Visual Studio\2022\Community>D:

D:\Projects\Microsoft\STL>cmake --build --preset x64
ninja: no work to do.

D:\Projects\Microsoft\STL>cmake --build --preset x64 --target format
ninja: error: unknown target 'format'

D:\Projects\Microsoft\STL>
```

Nope, does not work. :)

StephanTLavavej:
Hmm, I see that you're using VS 2022 17.10.5, but we require the latest Preview (as mentioned in the README). I'm not sure if that's the root cause, though. In fact, after I merged #4824 on 2024-07-11, the STL's build will emit an error for anything older than VS 2022 17.11, so I'm not sure how you've successfully built.

Can you:

* Install VS Preview
* Show your CMake and Ninja versions
* Pull `main` (remember to `git submodule update` if necessary), clean your repo, and try again?

Here's what I have:

```
C:\Temp>"C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Auxiliary\Build\vcvarsall.bat" x64
** Visual Studio 2022 Developer Command Prompt v17.11.0-pre.7.0
** Copyright (c) 2022 Microsoft Corporation
[vcvarsall.bat] Environment initialized for: 'x64'

C:\Temp>where cmake
C:\Program Files\Microsoft Visual Studio\2022\Preview\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe

C:\Temp>cmake --version
cmake version 3.29.5-msvc4

CMake suite maintained and supported by Kitware (kitware.com/cmake).

C:\Temp>where ninja
C:\Program Files\Microsoft Visual Studio\2022\Preview\Common7\IDE\CommonExtensions\Microsoft\CMake\Ninja\ninja.exe

C:\Temp>ninja --version
1.11.0
```

stl/inc/xloctime
StephanTLavavej: Oh, that was basically a case of "`constexpr` doesn't hurt here, so I have no objections". Unlike applying `const` to variables and parameters whenever possible, we have no consistent convention about marking helper functions as `constexpr` when possible.

fsb4000:
I just want to show how `cmake --build --preset x64 --target format` works for us:
изображение

fsb4000:
> Hmm, I see that you're using VS 2022 17.10.5, but we require the latest Preview (as mentioned in the README). I'm not sure if that's the root cause, though.

Probably. In file `tools/format/CMakeLists.txt` we have a hardcoded path for `clang-format`:
```
find_program(CLANG_FORMAT
    NAMES clang-format
    PATHS "C:/Program Files/Microsoft Visual Studio/2022/Preview/VC/Tools/Llvm/bin"
    DOC "The clang-format program to use"
    NO_DEFAULT_PATH
    NO_CMAKE_PATH
    NO_CMAKE_ENVIRONMENT_PATH
    NO_SYSTEM_ENVIRONMENT_PATH
    NO_CMAKE_SYSTEM_PATH
)
```
and because his `clang-format` has a different path, we don't add the custom target:
```
add_custom_target(run-format)
```

TheStormN:
I see, yes. It's strange to hardcode paths for tools.

TheStormN:
While I can somehow agree with your statements, still, if I had installed VS Preview on a non-standard location, it would still not work with hardcoded path.

StephanTLavavej:
Yeah, you've definitely found a real problem there. I'm looking into that CMakeLists.txt now, thanks! :smiley_cat:

TheStormN:
> Yeah, you've definitely found a real problem there. I'm looking into that CMakeLists.txt now, thanks! 😺

Never should've mentioned it... Got your attention away from the PR. :D

StephanTLavavej:
Created #4888. Now back to actual work :joy_cat:

stl/inc/xloctime
```diff
@@ -666,6 +666,110 @@ _NODISCARD constexpr bool _Is_valid_strftime_specifier(const char _Specifier) {
...
+    case 'R':
+        return _Is_valid_strftime_tm_hour(_Pt) && _Is_valid_strftime_tm_min(_Pt);
+    case 'D':
+    case 'x':
+    case 'F':
+        return _Is_valid_strftime_tm_year(_Pt) && _Is_valid_strftime_tm_mon(_Pt) && _Is_valid_strftime_tm_mday(_Pt);
+    case 'g':
+    case 'G':
+    case 'V':
+        return _Is_valid_strftime_tm_year(_Pt) && _Is_valid_strftime_tm_wday(_Pt) && _Is_valid_strftime_tm_yday(_Pt);
```
MattStephanson, line 754: At least in SDK version 10.0.22000.0, there is no year validation for `'V'`. See ucrt\time\wcsftime.cpp, line 942--947.

If I have time tonight I'll check others, but that's what I've noticed for now.

TheStormN: In this case, you are correct, but bare in mind that some of the validations are not directly within the `case` label. Once you are done with all the checks I will do any needed modifications.

MattStephanson: Everything here appears to match the validation done in wcsftime.cpp, except for `V` and the local-dependent flags `c`, `x`, and `X`.

In both the POSIX and C specs, the local-dependent flags specify that all the fields must be valid, though that seems like overkill for `X`, which only prints the time.

For `V`, the CRT doesn't validate explicitly, but the leap year check will overflow for values greater than `INT_MAX - 1900`. POSIX and C don't place any limitation on this field, so I would suggest being permissive here, either allowing any year or only guarding against `> INT_MAX - 1900`.

stl/inc/xloctime
```diff
@@ -666,6 +666,110 @@ _NODISCARD constexpr bool _Is_valid_strftime_specifier(const char _Specifier) {
...
+    return _Pt->tm_mday >= 1 && _Pt->tm_mday <= 31;
+}
+
+_NODISCARD constexpr bool _Is_valid_strftime_tm_mon(const tm* const _Pt) noexcept {
+    // months since January - [0, 11]
+    return _Pt->tm_mon >= 0 && _Pt->tm_mon <= 11;
+}
+
+_NODISCARD constexpr bool _Is_valid_strftime_tm_year(const tm* const _Pt) noexcept {
+    // years since 1900 - UCRT max range is up until 8099
```
MattStephanson, line 695: Since this is a limitation imposed by the UCRT rather than the `strftime` spec, would it make sense to be more permissive for Clang?

TheStormN:
@StephanTLavavej I think we are all good here. All comments resolved. :)

tests/std/tests/Dev11_0836436_get_time/test.cpp
```diff
@@ -944,3 +909,72 @@ void test_gh_4820() {
...
+        {&tm::tm_year, 8100, "%C%y%Y%D%x%F%g%G%c"},
+
+        {&tm::tm_yday, -1, "%j%U%W"},
+        {&tm::tm_yday, 366, "%j%U%W"},
+
+        {&tm::tm_wday, -1, "%a%A%u%w%c"},
+        {&tm::tm_wday, 7, "%a%A%u%w%c"},
+
+        {&tm::tm_wday, -1, "%U%W"},
+        {&tm::tm_wday, 7, "%U%W"},
```
StephanTLavavej, line 974: `%g%G` also cares about `tm_wday` and `tm_yday`:

```cpp
    case 'g':
    case 'G':
        return _Is_valid_strftime_tm_year(_Pt) && _Is_valid_strftime_tm_wday(_Pt) && _Is_valid_strftime_tm_yday(_Pt);
```

tests/std/tests/Dev11_0836436_get_time/test.cpp
```diff
@@ -944,3 +909,72 @@ void test_gh_4820() {
...
+
+    struct FormatTestData {
+        int tm::*field;
+        int value;
+        string format;
+    };
+
+    const FormatTestData testDataList[] = {
+        {&tm::tm_sec, -1, "%S%r%X%T%c"},
+        {&tm::tm_sec, 61, "%S%r%X%T%c"},
```
StephanTLavavej, line 950: Suggestion: We could avoid a lot of repetition by having this table store `lo` and `hi` values.

StephanTLavavej (approved):

TheStormN:
Man, if we could just push a bit the UCRT team to replicate the behavior of other C libraries regarding `strftime`, we would save so much pain. :)

StephanTLavavej:

StephanTLavavej:
Thanks for fixing this long-standing bug and making this code so much more robust! :tada: :smile_cat: :rocket:

PR #4885, opened 2024-08-09
Skip libcxx tests with timing assumptions

CaseyCarter:
These tests are all annotated with `ALLOW_RETRIES: n` comments, indicating that sporadic failures due to timing assumptions are expected. We don't tolerate sporadic failures, so let's disable these tests locally.

### DIFF
--- a/tests/libcxx/expected_results.txt
+++ b/tests/libcxx/expected_results.txt
@@ -518,6 +518,7 @@ std/thread/futures/futures.shared_future/wait_until.pass.cpp SKIPPED
 std/thread/futures/futures.unique_future/get.pass.cpp SKIPPED
 std/thread/futures/futures.unique_future/wait.pass.cpp SKIPPED
 std/thread/futures/futures.unique_future/wait_for.pass.cpp SKIPPED
+std/thread/thread.condition/thread.condition.condvarany/notify_one.pass.cpp SKIPPED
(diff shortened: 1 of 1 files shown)


tests/libcxx/expected_results.txt
```diff
@@ -509,25 +509,17 @@ std/utilities/smartptr/unique.ptr/unique.ptr.class/unique.ptr.asgn/move_convert.
 std/utilities/smartptr/unique.ptr/unique.ptr.class/unique.ptr.asgn/move_convert.single.pass.cpp FAIL
 std/utilities/smartptr/unique.ptr/unique.ptr.class/unique.ptr.asgn/move.pass.cpp FAIL
 
-# Not analyzed, likely bogus tests. Appears to be timing assumptions.
+# `ALLOW_RETRIES` comments indicate tests with timing assumptions
```
CaseyCarter, line 512: The change markup isn't great. There are actually no removals here, I simply pulled the skips with `ALLOW_RETRIES` comments out into a new section, and added two more on green lines 515 and 522. The other red lines here are exactly the other green lines below.

StephanTLavavej: Yep, double-checked the partition - thanks! :heart_eyes_cat:

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for noticing and fixing these intolerable sporadic failures! :slot_machine: :game_die: :hammer_and_wrench:

PR #4887, opened 2024-08-10
Add visualizer for `std::flat_map`

Nerixyz:
This PR adds visualizers for `std::flat_map` with any combination of `std::vector` and `std::deque` as the backing containers.

I'm not sure if there's a way of making this smaller. The only difference between the types is the implementation of `size`, `key_at`, and `value_at` - `<Expand>` is always the same. Ideally, there would be a way of having a condition based on the relevant type in from the template (`$T4` for `size` and `key_at`; `$T5` for `value_at`). Alternatively, it would be fine keeping just the type where both containers are vectors, since that's the default, and have the fallback that shows keys and values.

~~Initially, I had a `<DisplayString>{{ size={size()} }}</DisplayString>`, but I wanted to have at least some preview of the items contained in the map in the unexpanded view. Unfortunately, the `[comparator]` still shows up in the display string, so only one value is shown.~~ See https://github.com/microsoft/STL/pull/4887#issuecomment-2435773058

Code_2024-08-10_16-56-49

<details><summary>Code</summary>

```cpp
#include <deque>
#include <flat_map>
#include <string>

template <typename KeyContainer, typename MappedContainer> auto make() {
  std::flat_map<int, std::string, std::less<int>, KeyContainer, MappedContainer>
      map;
  map.emplace(42, "hello world");
  map.emplace(1, "meow");
  map.emplace(2, "hey");
  return map;
}

int main() {
  auto vectorVector = make<std::vector<int>, std::vector<std::string>>();
  auto vectorDeque = make<std::vector<int>, std::deque<std::string>>();
  auto dequeVector = make<std::deque<int>, std::vector<std::string>>();
  auto dequeDeque = make<std::deque<int>, std::deque<std::string>>();
  return 0;
}
```
</details>


stl/debugger/STL.natvis
```diff
@@ -2322,4 +2322,126 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
     </Expand>
   </Type>
 
+  <!-- KeyContainer: vector, MappedContainer: vector -->
+  <Type Name="std::flat_map&lt;*,*,*,std::vector&lt;*&gt;,std::vector&lt;*&gt;&gt;">
```
CaseyCarter, line 2326: I note that these visualizers are not going to work with `vector<bool>`. I don't think we should block the PR because of that, but I can imagine `vector<bool>` being used with `flat_map` so we should probably file an issue to correct the lack of support.

CaseyCarter (changes requested):

stl/debugger/STL.natvis
```diff
@@ -2322,4 +2322,126 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
     </Expand>
   </Type>
 
+  <!-- KeyContainer: vector, MappedContainer: vector -->
+  <Type Name="std::flat_map&lt;*,*,*,std::vector&lt;*&gt;,std::vector&lt;*&gt;&gt;">
+    <Intrinsic
+      Name="size"
+      Expression="(size_t)(_Data.keys._Mypair._Myval2._Mylast - _Data.keys._Mypair._Myval2._Myfirst)"
+    />
+    <Intrinsic Name="key_at" Expression="($T1*)(&amp;_Data.keys._Mypair._Myval2._Myfirst[index])">
```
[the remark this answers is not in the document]
Nerixyz: Not sure why I did this. But I removed the casts.

Nerixyz:
I added a `<DisplayString>` to be more in-line with the visualizers for `std::map` and `std::unordered_map`.

<details><summary>Preview</summary>

devenv_2024-10-24_18-43-25

</details> 

<details><summary>Code</summary>

```cpp
#include <deque>
#include <flat_map>
#include <map>
#include <string>
#include <unordered_map>

static void fillMap(auto &map) {
  map.emplace(42, "hello world");
  map.emplace(1, "meow");
  map.emplace(2, "hey");
}

template <typename T> T make() {
  T map{};
  fillMap(map);
  return map;
}

template <typename KeyContainer, typename MappedContainer> auto make() {
  return make<std::flat_map<int, std::string, std::less<int>, KeyContainer,
                            MappedContainer>>();
}

int main() {
  auto vectorVector = make<std::vector<int>, std::vector<std::string>>();
  auto vectorDeque = make<std::vector<int>, std::deque<std::string>>();
  auto dequeVector = make<std::deque<int>, std::vector<std::string>>();
  auto dequeDeque = make<std::deque<int>, std::deque<std::string>>();

  auto stdMap = make<std::map<int, std::string>>();
  auto stdUnorderedMap = make<std::unordered_map<int, std::string>>();
  return 0;
}
```
</details>

CaseyCarter (approved):

CaseyCarter:
Thanks for yet another great piece of visualizer work! 👁️🕶️👓🥽👀

stl/debugger/STL.natvis
CaseyCarter: That issue is #5037.

PR #4894, opened 2024-08-15
Use `wmemchr` in optimizations

AlexGuteniev:
Follow up on #4873 

I didn't measure the effect. I expect it to be negative currently, until @mcfi changes are picked up.

I changed the compatibility part too, as it was meant to optimize originally, why shouldn't save it, if possible.

### DIFF
--- a/stl/inc/xutility
+++ b/stl/inc/xutility
@@ -6100,7 +6100,7 @@ namespace ranges {
                 _Ptr_t _Result;
 #if _USE_STD_VECTOR_ALGORITHMS
                 if constexpr (_Is_sized) {
-                    // When _Is_sized && _Single_byte_elements, prefer this over memchr() for performance
+                    // When _Is_sized && _Elements_are_1_or_2_bytes, prefer this over memchr()/wmemchr() for performance
                     const auto _Last_ptr = _First_ptr + (_Last - _First);
                     _Result              = _STD _Find_vectorized(_First_ptr, _Last_ptr, _Val);
                 } else
--- a/stl/inc/xutility
+++ b/stl/inc/xutility
@@ -6021,12 +6021,8 @@ _NODISCARD _CONSTEXPR20 _InIt _Find_unchecked(_InIt _First, const _InIt _Last, c
                     _Result = static_cast<_Ptr_t>(_CSTD memchr(_First_ptr, static_cast<unsigned char>(_Val), _Count));
                 } else {
                     _STL_INTERNAL_STATIC_ASSERT(sizeof(_Iter_value_t<_InIt>) == 2);
-
-                    using _Src_type =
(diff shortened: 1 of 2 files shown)


stl/inc/xutility
```diff
@@ -6093,15 +6106,26 @@ namespace ranges {
...
+                            static_cast<_Ptr_t>(_CSTD memchr(_First_ptr, static_cast<unsigned char>(_Val), _Count));
+                    } else {
+                        _STL_INTERNAL_STATIC_ASSERT(sizeof(_Iter_value_t<_It>) == 2);
+
+                        using _Src_type =
+                            conditional_t<is_const_v<remove_reference_t<_Iter_ref_t<_It>>>, const wchar_t*, wchar_t*>;
+
+                        _Result = reinterpret_cast<_Ptr_t>(
+                            _CSTD wmemchr(reinterpret_cast<_Src_type>(_First_ptr), static_cast<wchar_t>(_Val), _Count));
+                    }
```
AlexGuteniev, line 6128: This is an exact duplicate of the above sequence, but I don't have an idea for function name to extract this into

stl/inc/xutility
```diff
@@ -6009,17 +6010,31 @@ _NODISCARD _CONSTEXPR20 _InIt _Find_unchecked(_InIt _First, const _InIt _Last, c
...
+                if constexpr (sizeof(_Iter_value_t<_InIt>) == 1) {
+                    _Result = static_cast<_Ptr_t>(_CSTD memchr(_First_ptr, static_cast<unsigned char>(_Val), _Count));
+                } else {
+                    _STL_INTERNAL_STATIC_ASSERT(sizeof(_Iter_value_t<_InIt>) == 2);
+
+                    using _Src_type =
+                        conditional_t<is_const_v<remove_reference_t<_Iter_ref_t<_InIt>>>, const wchar_t*, wchar_t*>;
+
+                    _Result = reinterpret_cast<_Ptr_t>(
+                        _CSTD wmemchr(reinterpret_cast<_Src_type>(_First_ptr), static_cast<wchar_t>(_Val), _Count));
```
StephanTLavavej, line 6029: `reinterpret_cast` can add constness, even though it can't remove constness. It would be simpler to always reinterpret to `const wchar_t*` for the input, then `const_cast` the output.

stl/inc/xutility
StephanTLavavej: I think this is fine as-is. With only 2 occurrences and not a lot of code duplication (especially after simplifying away `_Src_type`), I don't see a need to extract this. 3 occurrences is usually my recommended threshold.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for making the codebase consistently use `wmemchr`! :100: :tada: :cat2:

PR #4895, opened 2024-08-17
Workaround Intellisense bug in some `ranges::to` tests

CaseyCarter:
There's a regression in the latest Intellisense compiler, we'll avoid the problem until it's fixed. Tracked as VSO-2208356.

This mirrors the STL portion of MSVC-PR-571275.


StephanTLavavej (approved):

StephanTLavavej:
Thanks for ensuring that our compiler bug workarounds are in no danger of going extinct! :t-rex: :sauropod: :joy_cat:

PR #4900, opened 2024-08-20
Casey's accumulated miscellaneous changes

CaseyCarter:
A grab bag of minor changes broken down for quick-and-easy commit-wise review:

* `<ranges>`: simplify `zip_view::_Size_closure` by making it a `static constexpr` data member instead of a function
* Use conventional LLVM issue linking style
* Simplify `_Tuple_like` metaprogramming: implements `_Tuple_like_impl` as a variable template with four partial specializations instead of as a disjunction of four other variable template specializations. Consequently, each evaluation of `_Tuple_like` causes the compiler to memoize one constexpr variable instead of five.
* Don't return a decayed *boolean-testable* since the exposition-only concept doesn't allow for construction of an object of decayed type from a model of *boolean-testable*. Note that the changes in `erase_if(forward_list, pred)` and `erase_if(list, pred)` are the proposed resolution of LWG-4135, which I feel is a no-brainer and can be implemented without comments. I audited for other occurrences after noticing we do the same thing in `forward_list::remove` and `list::remove`.
  The fold expressions in `<ranges>` are a weird case: they are only problematic when the pack has one element. For zero or more than one elements, the fold expression has type `bool`, but with exactly one element it's a *boolean-testable* that the lambda implicitly decays on return. I think it's a bit more elegant to fix these by adding `|| false` to the fold expressions to make them binary than to explicitly specify the return type of the lambda.
* Don't `return` in `cartesian_product_view::_Iterator::iter_swap`. We _can_ `return` an expression with `void` type in a function with return type `void`, but we typically don't do so when the types are concrete.

I could be convinced that the *boolean-testable* change deserves to be in a separate PR with some test coverage.

### DIFF
--- a/stl/inc/__msvc_iter_core.hpp
+++ b/stl/inc/__msvc_iter_core.hpp
@@ -456,8 +456,10 @@ _EXPORT_STD using ranges::get;
 template <class _It, class _Se, ranges::subrange_kind _Ki>
 constexpr bool _Is_subrange_v<ranges::subrange<_It, _Se, _Ki>> = true;
 
+#if _HAS_CXX23
 template <class _It, class _Se, ranges::subrange_kind _Ki>
 constexpr bool _Tuple_like_impl<ranges::subrange<_It, _Se, _Ki>> = true;
+#endif // _HAS_CXX23
 
 template <class _It, class _Se, ranges::subrange_kind _Ki>
 struct tuple_size<ranges::subrange<_It, _Se, _Ki>> : integral_constant<size_t, 2> {};
--- a/stl/inc/__msvc_iter_core.hpp
+++ b/stl/inc/__msvc_iter_core.hpp
@@ -457,6 +457,9 @@ template <class _It, class _Se, ranges::subrange_kind _Ki>
 constexpr bool _Is_subrange_v<ranges::subrange<_It, _Se, _Ki>> = true;
 
 template <class _It, class _Se, ranges::subrange_kind _Ki>
+constexpr bool _Tuple_like_impl<ranges::subrange<_It, _Se, _Ki>> = true;
+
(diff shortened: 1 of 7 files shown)


stl/inc/__msvc_iter_core.hpp
```diff
@@ -456,6 +456,9 @@ _EXPORT_STD using ranges::get;
 template <class _It, class _Se, ranges::subrange_kind _Ki>
 constexpr bool _Is_subrange_v<ranges::subrange<_It, _Se, _Ki>> = true;
 
+template <class _It, class _Se, ranges::subrange_kind _Ki>
+constexpr bool _Tuple_like_impl<ranges::subrange<_It, _Se, _Ki>> = true;
```
StephanTLavavej, line 460:
```
D:\build\out\inc\__msvc_iter_core.hpp(460,16): error: no variable template matches partial specialization
  460 | constexpr bool _Tuple_like_impl<ranges::subrange<_It, _Se, _Ki>> = true;
      |                ^
```

It appears that the primary template is C++23, but this partial specialization is guarded for C++20:

https://github.com/microsoft/STL/blob/62205ab155d093e71dd9588a78f02c5396c3c14b/stl/inc/utility#L212-L215

StephanTLavavej (changes requested):

stl/inc/__msvc_iter_core.hpp
CaseyCarter: "Surely I must have tested these changes from months ago." :roll_eyes:

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Yay various cleanups! :broom: :tada: :black_cat:

PR #4904, opened 2024-08-22
Optimize equality comparison of two empty `string_view`s

### LINKED ISSUE #3310: Optimize `string_view::operator==` when comparing 2 empty `string_view`s
I discovered room for optimization when comparing 2 empty std::string_view s in GCC. However, Clang already seems to be optimized for this use case. See this quick bench run: https://quick-bench.com/q/FZc2abFbpI5Q_lEHXF7sNGC_Jn4 . Looking at this implementation of the STL, it seems like there is room for improvement as well: if both std:string_view s have the same size, something like memcmp is called. I couldn't verify this because quick bench doesn't support MSVC (yet).

I originally discovered this optimization here: https://github.com/microsoft/vcpkg-tool/pull/843

YexuanXiao:
Fixes #3310.


benchmarks/src/sv_equal.cpp
```diff
@@ -0,0 +1,74 @@
...
+    std::array<std::string_view, count> result{};
+
+    if constexpr (length == 0) {
+        return result;
+    }
+
+    using namespace std::views;
+
+    std::ranges::copy(haystack | chunk(length) | transform([](auto&& t) { return std::string_view(t); }) | take(count),
+        result.begin());
```
CaseyCarter, line 47: If we put everything after the `if constexpr` into an `else` branch, it will save us the time to instantiate all those templates only to then optimize them out of existence. Of course we'll then have two branches ending with `return result`, which is weird, so we may as well have only the `else` branch:
```suggestion
    if constexpr (length != 0) {
        using namespace std::views;

        std::ranges::copy(haystack | chunk(length) | transform([](auto&& t) { return std::string_view(t); }) | take(count),
            result.begin());
    }
```

benchmarks/CMakeLists.txt
```diff
@@ -122,6 +122,7 @@ add_benchmark(replace src/replace.cpp)
 add_benchmark(search src/search.cpp)
 add_benchmark(std_copy src/std_copy.cpp)
 add_benchmark(swap_ranges src/swap_ranges.cpp)
+add_benchmark(sv_equal src/sv_equal.cpp)
```
CaseyCarter, line 125: These are ordered alphabetically; `sv_equal` should precede `swap_ranges`:
```suggestion
add_benchmark(sv_equal src/sv_equal.cpp)
add_benchmark(swap_ranges src/swap_ranges.cpp)
```

benchmarks/src/sv_equal.cpp
```diff
@@ -0,0 +1,74 @@
...
+}
+
+BENCHMARK(equal<0>);
+BENCHMARK(equal<8>);
+BENCHMARK(equal<16>);
+BENCHMARK(equal<32>);
+BENCHMARK(equal<64>);
+BENCHMARK(equal<128>);
+BENCHMARK(equal<256>);
+
```
CaseyCarter, line 73: We should `static_assert` here that `haystack.size() >= count * 256` to demonstrate that the haystack is large enough for the biggest test case. This won't prevent someone breaking the invariant, but it will hopefully remind them an invariant exists.
```suggestion

static_assert(haystack.size() >= count * 256, "haystack is too small");

```

CaseyCarter (approved):
Some minor nits that I'll apply.

benchmarks/src/sv_equal.cpp
```diff
@@ -0,0 +1,75 @@
...
+        std::ranges::copy(
+            haystack | chunk(length) | transform([](auto&& t) { return std::string_view(t); }) | take(count),
+            result.begin());
+    }
+
+    return result;
+}
+
+template <std::size_t length>
+void equal(benchmark::State& state) {
```
StephanTLavavej, line 52: Naming nitpicks: Although this benchmark isn't `using namespace std;` so there's no actual shadowing, it's visually confusing to include `<algorithm>` and then have a constant named `count` and a function template named `equal`. It would be clearer to name them distinctly, e.g. `Count` and `sv_equal`. (Also, we usually capitalize template parameters in test code, so `Length` would look more like a compile-time constant.)

StephanTLavavej (approved):

StephanTLavavej:
Thanks, this is great! :heart_eyes_cat: I pushed a commit for trivial naming nitpicks, and verified that there's a significant speedup for empty comparisons (I observed 1.7x, from 37.4 ns down to 22.0 ns).

We merge PRs to our GitHub and MSVC-internal repos simultaneously in a semi-manual process, batched up to save time. Your PR will be part of the next batch. I'll post comments here as I prepare this for merging!

StephanTLavavej:

StephanTLavavej:
Thanks for this significant performance improvement - it's a small change but it affects extremely common operations (both `string` and `string_view`). Amazing for a first-time contribution! :heart_eyes_cat: :rocket: :tada:

PR #4908, opened 2024-08-24
`<ranges>`: rename `views::iota` and `views::repeat` parameters

SuperWig:
The Visual Studio IDE has an "inline parameter hint" feature which will show the parameter names inline.  The STL generally has nice descriptive names which can make this feature helpful.
When I went to use `views::iota` I forgot if the second arg is a count or not, I was hoping the parameter names would save me a trip to cppreference, but alas, I was greeted with useless names.

`views::repeat` has the same issue (which is my fault; I just copied and pasted iota)

image

I also have a [feature request](https://developercommunity.visualstudio.com/t/De-uglify-STL-identifiers-in-inline-hint/10697009) to de-uglify them (might make another to not show the type information for these views, or I guess more generally, long nested templated types)

### DIFF
--- a/stl/inc/ranges
+++ b/stl/inc/ranges
@@ -1051,11 +1051,12 @@ namespace ranges {
             }
 
             template <class _Ty1, class _Ty2>
-            _NODISCARD _STATIC_CALL_OPERATOR constexpr auto operator()(_Ty1&& _Val1, _Ty2&& _Val2) _CONST_CALL_OPERATOR
-                noexcept(noexcept(iota_view(static_cast<_Ty1&&>(_Val1), static_cast<_Ty2&&>(_Val2))))
-                requires requires { iota_view(static_cast<_Ty1&&>(_Val1), static_cast<_Ty2&&>(_Val2)); }
+            _NODISCARD _STATIC_CALL_OPERATOR constexpr auto operator()(
+                _Ty1&& _Start, _Ty2&& _Bound) _CONST_CALL_OPERATOR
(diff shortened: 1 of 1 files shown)


StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: This is a strict improvement over the status quo's currently terrible names, so I'm happy here.

I observe that this isn't perfectly consistent with either the underlying views' parameters, or the Standard-depicted parameters, but I don't care enough to nitpick it further. (The views are the preferred interface for users, and the names you've chosen are a bit more descriptive than "value, bound", which is enough rationale for me.)

I'll defer to Casey if he has other thoughts.

SuperWig:
> I observe that this isn't perfectly consistent with either the underlying views' parameters, or the Standard-depicted parameters, but I don't care enough to nitpick it further. (The views are the preferred interface for users, and the names you've chosen are a bit more descriptive than "value, bound", which is enough rationale for me.)

After I made the commit I realised the same thing... I then also came to pretty much the same conclusion so opted to just create the PR anyway 😁.

CaseyCarter (approved):

CaseyCarter:
> `views::repeat` has the same issue (which is my fault; I just copied and pasted iota)

Hey, at least we didn't name them `(_Val2, _Val1)`.

StephanTLavavej:

StephanTLavavej:
`_Thanks1, _Thanks2` :joy_cat: :zany_face: :tada:

PR #4914, opened 2024-08-26
Adjust tests relying on CWG-1351 noexcept behavior

joemmett:
Update some noexcept tests in variant and invoke to remove old permissive behavior. This is guarded in a check for an as-yet unreleased MSVC compiler.

This mirrors the STL portion of MSVC-PR-566073.


frederick-vs-ja (approved):

CaseyCarter:
😻 Thanks for ensuring these tests are _no exception_ to our suite's quality and correctness!

PR #4921, opened 2024-08-27
Avoid `rand()` in the legacy `tr1` test suite

StephanTLavavej:
Our bosses and boss-like entities heard that [`rand()` is considered harmful](https://learn.microsoft.com/en-us/shows/goingnative-2013/rand-considered-harmful) and are starting a push to avoid unnecessary use.

Almost all of our mentions of `rand` are Standard-mandated (or Standard stable name citations), but I found a few unnecessary uses in our legacy `tr1` test suite. We avoid cleaning up its code unless there's a specific reason to mess with it, and this easily qualifies.

* `tr1/tests/algorithm`
  + Give `rand_gen` an `mt19937` data member. (Default construction is fine; we weren't attempting to seed `rand()` before.) This is purely masking bits with `& 0xfffff`, so there's no need for a `uniform_int_distribution`.
  + Replace the `frand()` wrapper with an `rng_func` lambda. To avoid `/clr:pure` problems with init-captures, I'm capturing `mt` by reference (so the lambda doesn't need to be `mutable` and doesn't need to be passed via `ref()`). We need to adapt the half-open interface to `uniform_int_distribution`'s fully-closed range.
* `tr1/tests/memory2`
  + Construct a local `mt19937 mt;` within the `static void ctors()` worker function. (It's important to not use a global engine with multiple threads!) Again, we weren't attempting to seed `rand()`, so default-constructing `mt19937` is fine. And because we're just looking at a bit, we don't need a `uniform_int_distribution`.

### DIFF
--- a/tests/tr1/tests/memory2/test.cpp
+++ b/tests/tr1/tests/memory2/test.cpp
@@ -7,7 +7,7 @@
 #include "tdefs.h"
 #include <memory>
 #include <mutex>
-#include <stdlib.h>
+#include <random>
 #include <thread>
 #include <vector>
 
@@ -15,6 +15,7 @@ using STD shared_ptr;
 using STD weak_ptr;
 
 using STD lock_guard;
+using STD mt19937;
 using STD mutex;
 using STD thread;
 using STD vector;
@@ -27,6 +28,7 @@ static shared_ptr<int> sp;
 static mutex start_mtx;
 
 static void ctors() { // construct a gazillion shared and weak pointers
+    mt19937 mt;
 
     { // wait for access
         lock_guard<mutex> lock(start_mtx);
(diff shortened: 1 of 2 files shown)


tests/tr1/tests/memory2/test.cpp
```diff
@@ -27,14 +28,15 @@ static shared_ptr<int> sp;
...
+    mt19937 mt;
 
     { // wait for access
         lock_guard<mutex> lock(start_mtx);
     }
 
     for (int i = 0; i < NSETS; ++i) {
         for (int j = 0; j < NREPS; ++j) {
-            if (rand() % 2 != 0) {
+            if (mt() % 2 != 0) {
```
CaseyCarter, line 39: That is one very expensive bit of entropy. (No change requested.)

CaseyCarter (approved):

StephanTLavavej:

StephanTLavavej:
I had to push an additional commit to avoid `/clr:pure` problems with init-captures.

CaseyCarter (approved):

PR #4922, opened 2024-08-30
Fix constraint on `unique_ptr<T, D>(auto_ptr<U>&&)`

CaseyCarter:
Per suggestion of Howard Hinnant on the LWG reflector. We initialize `unique_ptr`'s stored `pointer` with the result of calling `release()` on the `auto_ptr`. `release()` has type `U*`, so the correct constraint is to require `U*` to be convertible to `pointer`.

STL will probably laugh himself to tears before rejecting this PR, but it's probably the only chance I'll ever get to fix an `auto_ptr`-related bug so here it is =)

### DIFF
--- a/stl/inc/memory
+++ b/stl/inc/memory
@@ -3380,7 +3380,7 @@ public:
 
 #if _HAS_AUTO_PTR_ETC
     template <class _Ty2,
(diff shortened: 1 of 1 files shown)


StephanTLavavej (approved):

StephanTLavavej:
Approved with uncontrollable vomiting :vomiting_face: :joy_cat:

StephanTLavavej:

StephanTLavavej:
:test_tube: :biohazard: :skull_and_crossbones:
===

PR #4937, opened 2024-09-05
Avoid testing infinities in `/fp:fast`

AlexGuteniev:
Towards #4931

There remaining warnings do not originate from the test itself.

### DIFF
--- a/tests/std/tests/VSO_0000000_vector_algorithms_floats/test.cpp
+++ b/tests/std/tests/VSO_0000000_vector_algorithms_floats/test.cpp
@@ -30,11 +30,19 @@ void test_min_max_element_floating_any(mt19937_64& gen) {
 
     constexpr auto input_of_input_size = dataCount / 2;
     vector<T> input_of_input(input_of_input_size);
-    input_of_input[0] = -numeric_limits<T>::infinity();
-    input_of_input[1] = +numeric_limits<T>::infinity();
-    input_of_input[2] = -0.0;
-    input_of_input[3] = +0.0;
-    for (size_t i = 4; i < input_of_input_size; ++i) {
+
+    input_of_input[0] = -0.0;
+    input_of_input[1] = +0.0;
+#ifndef _M_FP_FAST
+    constexpr size_t special_values_count = 2;
+#else // ^^^ defined(_M_FP_FAST) / !defined(_M_FP_FAST) vvv
+    input_of_input[2] = -numeric_limits<T>::infinity();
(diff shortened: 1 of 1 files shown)


tests/std/tests/VSO_0000000_vector_algorithms_floats/test.cpp
```diff
@@ -30,11 +30,19 @@ void test_min_max_element_floating_any(mt19937_64& gen) {
...
-    input_of_input[1] = +numeric_limits<T>::infinity();
-    input_of_input[2] = -0.0;
-    input_of_input[3] = +0.0;
-    for (size_t i = 4; i < input_of_input_size; ++i) {
+
+    input_of_input[0] = -0.0;
+    input_of_input[1] = +0.0;
+#ifndef _M_FP_FAST
+    constexpr size_t special_values_count = 2;
+#else // ^^^ defined(_M_FP_FAST) / !defined(_M_FP_FAST) vvv
```
StephanTLavavej, line 38: The logic and comments disagree; in this case I believe the logic is reversed, so the PR is doing the opposite of what it claims.

tests/std/tests/VSO_0000000_vector_algorithms_floats/test.cpp
```diff
@@ -30,11 +30,19 @@ void test_min_max_element_floating_any(mt19937_64& gen) {
...
+    input_of_input[2] = -numeric_limits<T>::infinity();
+    input_of_input[3] = +numeric_limits<T>::infinity();
+
+    constexpr size_t special_values_count = 4;
+#endif // ^^^ !defined(_M_FP_FAST) ^^^
+
+    for (size_t i = special_values_count; i < input_of_input_size; ++i) {
```
StephanTLavavej, line 45: Suggestion: There's nothing wrong with this `special_values_count` technique, but it seems increasingly elaborate for something that could be simpler. Could we simply use a range-for to fill the vector with `dis(gen)`, then unconditionally *overwrite* elements 0 and 1 with negative and positive zero, and then conditionally overwrite elements 2 and 3 with negative and positive infinity? It's not going to cost us significant test execution time to throw away a few distribution/engine invocations.

StephanTLavavej (changes requested):

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:infinity: :white_check_mark: :tada:
===

PR #4939, opened 2024-09-05
`<source_location>`: Remove EDG workaround for DevCom-10199227

fsb4000:
`Fixed In: Visual Studio 2022 version 17.11.2` DevCom-10199227


CaseyCarter (approved):

CaseyCarter:
I see test failures in both `tests\P1208R6_source_location` and `tests\P2465R3_standard_library_modules`.

fsb4000:
Should we add `/D_USE_DETAILED_FUNCTION_NAME_IN_SOURCE_LOCATION=0` to one of `PM_CL` in `usual_20_matrix.lst` to test basic function name? After my current changes we test only detailed names.

StephanTLavavej:
We have so little test coverage of `<source_location>`, I think we should *not* modify the centralized matrices.

fsb4000:
By the way,  how do you find out C1XX values in compile time?
I mean this:
```
#if !defined(__clang__) && !defined(__EDG__) // TRANSITION, VSO-1285783
    if (is_constant_evaluated()) {
        assert(s.x.loc.function_name() == "main"sv);
    } else
#endif // ^^^ workaround ^^^
```
It's different from `_EDG_` and C1XX runtime, and you can't print them in compile time...

For `_EDG_` I created a mini app and hover my mouse to variables to find out:
изображение
изображение

tests/std/tests/P1208R6_source_location/env.lst
```diff
@@ -2,3 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 RUNALL_INCLUDE ..\usual_20_matrix.lst
+RUNALL_CROSSLIST
+*	PM_CL="" # Test default setting
+*	PM_CL="/D_USE_DETAILED_FUNCTION_NAME_IN_SOURCE_LOCATION=0" # Test escape hatch, see GH-4055
```
CaseyCarter: Comments at the end of lines with configuration mean something to the internal test runner, let's not do this. I personally don't think we need the comment: we can search for `_USE_DETAILED_FUNCTION_NAME_IN_SOURCE_LOCATION` and figure out what's going on here very quickly. If disagree enough to restore the comments, please put them on their own lines above the configurations they describe.

tests/std/tests/P1208R6_source_location/test.cpp
```diff
@@ -102,10 +110,19 @@ constexpr void sloc_constructor_test() {
...
+#endif // ^^^ Other ^^^
     }
 #else // ^^^ detailed / basic vvv
-    assert(x.loc.function_name() == "sloc_constructor_test"sv);
+#if !defined(__clang__) && !defined(__EDG__) // TRANSITION, VSO-1285783
+    if (is_constant_evaluated()) {
+        assert(x.loc.function_name() == "main"sv);
+    } else
+#endif // ^^^ workaround ^^^
+        assert(x.loc.function_name() == "sloc_constructor_test"sv);
```
CaseyCarter, line 126: This sometimes-`else` branch should be fully braced:
```suggestion
    {
        assert(x.loc.function_name() == "sloc_constructor_test"sv);
    }
```

tests/std/tests/P1208R6_source_location/test.cpp
```diff
@@ -143,10 +164,19 @@ constexpr void sub_member_test() {
...
+#endif // ^^^ Other ^^^
     }
 #else // ^^^ detailed / basic vvv
-    assert(s.x.loc.function_name() == "sub_member_test"sv);
+#if !defined(__clang__) && !defined(__EDG__) // TRANSITION, VSO-1285783
+    if (is_constant_evaluated()) {
+        assert(s.x.loc.function_name() == "main"sv);
+    } else
+#endif // ^^^ workaround ^^^
+        assert(s.x.loc.function_name() == "sub_member_test"sv);
```
CaseyCarter, line 182: Ditto "fully braced":
```suggestion
    {
        assert(s.x.loc.function_name() == "sub_member_test"sv);
    }
```

tests/std/tests/P1208R6_source_location/test.cpp
```diff
@@ -185,19 +219,29 @@ constexpr void lambda_test() {
...
     assert(x1.function_name() == "void __cdecl lambda_test(void)"sv);
+#endif // ^^^ Other ^^^
 #else // ^^^ detailed / basic vvv
     assert(x1.function_name() == "lambda_test"sv);
 #endif // ^^^ basic ^^^
     const string_view fun2{x2.function_name()};
 #if !_USE_DETAILED_FUNCTION_NAME_IN_SOURCE_LOCATION
+#if !defined(__clang__) && !defined(__EDG__)
+    assert(fun2 == "operator ()"sv);
+#else // ^^ C1XX / Other vv
```
CaseyCarter:
```suggestion
#else // ^^^ C1XX / Other vvv
```

tests/std/tests/P1208R6_source_location/test.cpp
```diff
@@ -185,19 +219,29 @@ constexpr void lambda_test() {
...
     assert(fun2 == "operator()"sv);
+#endif // ^^ Other ^^
```
CaseyCarter:
```suggestion
#endif // ^^^ Other ^^^
```

CaseyCarter (approved):

tests/std/tests/P1208R6_source_location/env.lst
AlexGuteniev: Guess it gas precedent here
https://github.com/microsoft/STL/blob/c7c5ca7d6adde7c1554b1e52c0b97cb2a7f7ca60/tests/std/tests/VSO_0000000_vector_algorithms/env.lst#L5-L7

CaseyCarter:
> By the way, how do you find out C1XX values in compile time?

You preserve the compile time value until runtime where you can examine it. In this case, the function names all have static storage duration like string literals. You can simply return the value from the function to initialize a `constinit const char*` that you print in main: https://www.godbolt.org/z/v3jPe1MrW.

tests/std/tests/P1208R6_source_location/env.lst
CaseyCarter: Thanks, @AlexGuteniev - I'll clean that up in a separate change.

StephanTLavavej (approved):

StephanTLavavej:
Relabeling as this is a workaround removal in product code, purr! :heart_eyes_cat:

StephanTLavavej:

StephanTLavavej:
Thanks for the extensive workaround removal! :broom: :magic_wand: :smiley_cat:

PR #4940, opened 2024-09-06
Explicitly spell the noexcept-specifier of the move constructor and move assignment operator of packaged_task

heckerpowered:
https://en.cppreference.com/w/cpp/thread/packaged_task/packaged_task
https://en.cppreference.com/w/cpp/thread/packaged_task/operator%3D

Cppref states that the move constructor and move assignment operator of packaged_task are noexcept, so I think there should be a noexcept specifier here.

### DIFF
--- a/stl/inc/future
+++ b/stl/inc/future
@@ -1268,9 +1268,9 @@ public:
     template <class _Fty2, enable_if_t<!is_same_v<_Remove_cvref_t<_Fty2>, packaged_task>, int> = 0>
     explicit packaged_task(_Fty2&& _Fnarg) : _MyPromise(new _MyStateType(_STD forward<_Fty2>(_Fnarg))) {}
 
-    packaged_task(packaged_task&&) = default;
+    packaged_task(packaged_task&&) noexcept = default;
 
-    packaged_task& operator=(packaged_task&&) = default;
+    packaged_task& operator=(packaged_task&&) noexcept = default;
 
 #if _HAS_FUNCTION_ALLOCATOR_SUPPORT
     template <class _Fty2, class _Alloc, enable_if_t<!is_same_v<_Remove_cvref_t<_Fty2>, packaged_task>, int> = 0>


frederick-vs-ja:
This PR seems to be an improvement as it makes `noexcept` specified by the Standard explicitly spelt.

However, the touched move functions are already implicitly `noexcept`. E.g. the following code snippet is accepted with MSVC STL ([Godbolt link](https://godbolt.org/z/aP9ab15bY)).

```C++
#include <future>
#include <type_traits>

static_assert(std::is_nothrow_move_constructible_v<std::packaged_task<void()>>);
static_assert(std::is_nothrow_move_constructible_v<std::packaged_task<int(long)>>);

static_assert(std::is_nothrow_move_assignable_v<std::packaged_task<void()>>);
static_assert(std::is_nothrow_move_assignable_v<std::packaged_task<int(long)>>);
```

I think the title of this PR should be changed to
> Explicitly spell the `noexcept`-specifier of the move constructor and move assignment operator of `packaged_task`

CaseyCarter (approved):

frederick-vs-ja (approved):
Oops, the original explicit `noexcept`-specifiers were dropped by #3315... by me.

I used to think that implicit exception specification should be sufficient and would slightly speed up compilation, but I ignored the drawback that it would be harder for users to confirm and rely on the fact that these move functions are always `noexcept`.

heckerpowered:

CaseyCarter:
In the past I think we've chosen to not state properties of 

> Oops, the original explicit `noexcept`-specifiers were dropped by #3315... by me.

It's not all on you: in the past we've avoided explicit `noexcept`s - and even some `constexpr`s - on defaulted member functions. I think it was just to have less code to read, both for us and (as you suggest) the compiler.

I agree that we should be explicit given the enormous number of readers the code gets with a diversity of experiences. It would be nice not to have to read the entire header simply to determine that `packaged_task`'s move constructor is `noexcept`, or example.

CaseyCarter (approved):

CaseyCarter:
image

@heckerpowered, for future reference we prefer that authors don't force push PR branches after review starts because it tends to break incremental code review. (GitHub's "show changes since your last review" feature gets confused. Or at least it did, I haven't checked for a while.) It's obviously a non-issue for this two-line change, but for bigger things we will cry and gnash our teeth with frustration. And nobody likes gnashing.

StephanTLavavej (approved):

StephanTLavavej:
I remembered to check for `R&` and `void` specializations (as seen in other `<future>` technology) but `packaged_task` doesn't have them, so this change is correct and complete. :smile_cat:

StephanTLavavej:

StephanTLavavej:
Thanks for this clarity improvement and congratulations on your first microsoft/STL commit! :tada: :smile_cat: :rocket:

PR #4942, opened 2024-09-07
Silence CodeQL false positive warnings

StephanTLavavej:
* "Unclear validation of array index" silenced by `// lgtm [cpp/unclear-array-index-validation]`
  + `__msvc_string_view.hpp` 668: We have `bool _Matches[256]` and the index is cast to `unsigned char`.
  + `__msvc_string_view.hpp` 1334: `basic_string_view::operator[]` has optional validation controlled by `_CONTAINER_DEBUG_LEVEL` and intentionally doesn't validate by default.
  + `charconv`: `_Digit_from_byte` has 256 elements (enforced by `_STL_INTERNAL_STATIC_ASSERT`) and the index is cast to `unsigned char`.
* "Cast from `char*` to `wchar_t*`" silenced by `// lgtm [cpp/incorrect-string-type-conversion]`
  + `xutility`: `_Copy_memmove()` is correctly bypassing pointer arithmetic for performance.
  + `StlLCMapStringA.cpp`: This function has [`LCMapStringA`][]'s narrow interface, but implements `LCMAP_SORTKEY` with wide [`LCMapStringEx`][]. "The sort key is stored in the buffer and treated as an opaque array of bytes" so there is no correctness issue.

[`LCMapStringA`]: https://learn.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-lcmapstringa
[`LCMapStringEx`]: https://learn.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-lcmapstringex

### DIFF
--- a/stl/inc/__msvc_string_view.hpp
+++ b/stl/inc/__msvc_string_view.hpp
@@ -665,7 +665,7 @@ public:
     }
 
     constexpr bool _Match(const _Elem _Ch) const noexcept { // test if _Ch is in the bitmap
-        return _Matches[static_cast<unsigned char>(_Ch)];
+        return _Matches[static_cast<unsigned char>(_Ch)]; // lgtm [cpp/unclear-array-index-validation]
     }
 
 private:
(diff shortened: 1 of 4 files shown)


AlexGuteniev:
> * `__msvc_string_view.hpp` 668: We have `bool _Matches[256]` and the index is cast to `unsigned char`.

This looks like clear false positive that might be reported

StephanTLavavej:
We're trying to get in touch with the maintainers to report severe usability issues - if we get traction on those, then we can report specific false positives like these.

StephanTLavavej:

PR #4944, opened 2024-09-10
Begin removing EDG workaround from `ranges::to` tests

CaseyCarter:
VSO-2208356 has been fixed internally. Disable the workaround internally, and mark the workaround for cleanup when VS 17.12p3 releases.

This mirrors MSVC-PR-577053.

### DIFF
--- a/tests/std/tests/P1206R7_ranges_to_mappish/test.cpp
+++ b/tests/std/tests/P1206R7_ranges_to_mappish/test.cpp
@@ -156,13 +156,13 @@ struct mappish_instantiator {
             assert(c5.get_allocator().state == 13);
             assert(ranges::is_permutation(c5, expected, any_pair_eq));
         }
-#ifndef __EDG__ // TRANSITION, VSO-2208356
+#if defined(_MSVC_INTERNAL_TESTING) || !defined(__EDG__) // TRANSITION, VS 17.12p3
         {
             std::same_as<T> auto c6 = R{some_pairs} | ranges::to<T>(Alloc{13});
             assert(c6.get_allocator().state == 13);
             assert(ranges::is_permutation(c6, expected, any_pair_eq));
         }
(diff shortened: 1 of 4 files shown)


tests/std/tests/P1206R7_ranges_to_mappish/test.cpp
```diff
@@ -156,13 +156,13 @@ struct mappish_instantiator {
...
         }
-#ifndef __EDG__ // TRANSITION, VSO-2208356
+#if defined(_MSVC_INTERNAL_TESTING) || !defined(__EDG__) // TRANSITION, VS 17.12p3
         {
             std::same_as<T> auto c6 = R{some_pairs} | ranges::to<T>(Alloc{13});
             assert(c6.get_allocator().state == 13);
             assert(ranges::is_permutation(c6, expected, any_pair_eq));
         }
-#endif // ^^^ no workaround ^^^
+#endif
```
CaseyCarter, line 165: We should restore these "no workaround" comments so we know to remove the conditional, and not its contents, then VS 17.12p3 ships. I'll do so in a followup, this needs to match the internal PR which has already merged.

StephanTLavavej (approved):

StephanTLavavej:
Citing the MSVC-PR being mirrored in the PR description is great (yay autolink references), but it makes for an exceptionally opaque PR title / commit subject. When scanning history, it's much more useful to have brief descriptions of what changed. I've changed the title and added a bit more clarity than before (mentioning that this affects `ranges::to` tests).

(Yes I am a very picky kitty :cat2:)

tests/std/tests/P1206R7_ranges_to_mappish/test.cpp
CaseyCarter: Done in #4945.

PR #4952, opened 2024-09-12
Polish `std::generator`

CaseyCarter:
This should complete the `generator` feature branch and get it ready to merge. There are very few product code changes here, it's mostly expanded test coverage with some cleanup and reorganization. There's also some documentation of the "stack of coroutines" and the memory layout of allocations that will make it easier for folks to get up to speed on how the code works under the covers.

It's broken down nicely by commit for ease of review; I'll replicate the commit messages here as an overview. Comments without specific context refer to `P2502R2_generator` where most changes were made. The `P2502R2_generator_iterator` and `P2502R2_generator_promise` tests were already very nicely complete.

* Update `generator` template argument mandates
  * Update citations to WG21-N4988
  * Reorder checks to specification order
  * Make `_RRef` more obviously reflect the wording for `RRef` in WG21-N4988 [coro.generator.class]/1.4.
* Allocator testing updates
  * Simplify and correct `StatelessAlloc`
    * Deriving publicly from `std::allocator` not a great idea, as witnessed by the recent addition of `allocate_at_least`.
    * `deallocate` should be `noexcept` for a `Cpp17Allocator`.
    * We don't need to reimplement `std::allocator` when we can simply use it.
    * The domain of equality for `Cpp17Allocators` is an entire `rebind` family, i.e., `StatelessAlloc<T>` and `allocator_traits<StatelessAlloc<T>>::rebind_alloc<U>` must be comparable.
  * PascalCase `stateful_alloc` for consistency and move into the header with `StatelessAlloc`.
* Promise test tweaks
  * "whose member `await_ready` returns `false`" in WG21-N4988 [coro.generator.promise]/11 implies the return type is exactly `bool`, not "convertible to `bool`". Update `test_yield_elements_of_range` accordingly.
  * Move non-portable size check from `P2502R2_generator` into `P2502R2_generator_promise`
* Expand and complete `static_checks`: Implement a complete `generator` traits in the header to use in both `P2502R2_generator_iterator` and `P2502R2_generator`'s `static_checks`.
* Create generic `test_one` template which takes a generator, a description of its static properties, and the expected result of piping the generator through a provided range adaptor. `test_one` validates the static properties with `static_checks`, and confirms the output is as expected.
* Extract test cases for weird reference types (mutable lvalue and rvalue references) from `main` into new function `test_weird_reference_types`.
* Several small tweaks:
  * It's no longer significant that `co_upto` wasn't an example in the proposal; strike the comment.
  * Enforce `co_upto`'s precondition so it's nicely documented.
  * Consistently prefer `same_as` to `is_same_v`.
  * Regroup calls to test cases in `main` topically and title each category.
* Reorganize test code to agree with call order in `main`. [This is the largest individual commit; it is a pure reordering.]
* All product code assertions now depend on `_CONTAINER_DEBUG_LEVEL`. These are all simple O(1) checks, there's no reason not to promote them from `_DEBUG` to `_CONTAINER_DEBUG_LEVEL`.
* Clarify allocation mechanisms, including some fancy memory layout diagrams.
  * Rename `_Promise_allocator` to `_Coro_promise_allocator` to avoid any confusion with `_Promise`. Rename its template parameter `_Allocator` to `_Proto_allocator` to avoid confusion with the rebound allocator type `_Alloc`.
  * Expand `static_assert` message in `operator new`. If and when `_Coro_promise_allocator` is reused by other coroutine types we can worry about making the message more generic.
  * Extract block size computations.
* `Element_awaiter` must direct-non-list-initialize its stored object, per WG21-N4988 [coro.generator.promise]/7.
* Document the "stack of coroutine handles" in a code comment with another work of art.

### DIFF
--- a/stl/inc/generator
+++ b/stl/inc/generator
@@ -409,12 +409,16 @@ public:
 
     _NODISCARD _Ref operator*() const
         noexcept(noexcept(static_cast<_Ref>(*_Coro.promise()._Get_top().promise()._Ptr))) {
-        _STL_ASSERT(!_Coro.done(), "Can't dereference generator end iterator");
+#if _CONTAINER_DEBUG_LEVEL > 0
+        _STL_VERIFY(!_Coro.done(), "Can't dereference generator end iterator");
+#endif // _CONTAINER_DEBUG_LEVEL > 0
         return static_cast<_Ref>(*_Coro.promise()._Get_top().promise()._Ptr);
     }
 
     _Iterator& operator++() {
-        _STL_ASSERT(!_Coro.done(), "Can't increment generator end iterator");
+#if _CONTAINER_DEBUG_LEVEL > 0
+        _STL_VERIFY(!_Coro.done(), "Can't increment generator end iterator");
+#endif // _CONTAINER_DEBUG_LEVEL > 0
         _Coro.promise()._Get_top().resume();
         return *this;
     }
@@ -491,7 +495,9 @@ public:
 
     _NODISCARD _Gen_iter_provider<_Value, _Ref>::_Iterator begin() {
         // Pre: _Coro is suspended at its initial suspend point
-        _STL_ASSERT(_Coro, "Can't call begin on moved-from generator");
+#if _CONTAINER_DEBUG_LEVEL > 0
+        _STL_VERIFY(_Coro, "Can't call begin on moved-from generator");
+#endif // _CONTAINER_DEBUG_LEVEL > 0
         _Coro.resume();
         return typename _Gen_iter_provider<_Value, _Ref>::_Iterator{
             _Gen_secret_tag{}, coroutine_handle<_Gen_promise_base<yielded>>::from_address(_Coro.address())};
--- a/stl/inc/generator
+++ b/stl/inc/generator
@@ -40,22 +40,38 @@ struct alignas(__STDCPP_DEFAULT_NEW_ALIGNMENT__) _Aligned_block {
 template <class _Alloc>
 concept _Has_real_pointers = same_as<_Alloc, void> || is_pointer_v<typename allocator_traits<_Alloc>::pointer>;
 
-template <class _Allocator>
-class _Promise_allocator { // statically specified allocator type
+template <class _Proto_allocator>
+class _Coro_promise_allocator { // statically specified allocator type
 private:
-    using _Alloc           = _Rebind_alloc_t<_Allocator, _Aligned_block>;
+    using _Alloc           = _Rebind_alloc_t<_Proto_allocator, _Aligned_block>;
     using _Alloc_size_type = allocator_traits<_Alloc>::size_type;
 
+    static size_t _Allocation_block_size(const size_t _Size) noexcept {
+        // Compute the number of _Aligned_blocks needed to store a coroutine
+        // frame of _Size bytes and, if necessary, an _Alloc.
+        if constexpr (default_initializable<_Alloc> && allocator_traits<_Alloc>::is_always_equal::value) {
+            // allocator is stateless, we need no additional space
+            return (_Size + sizeof(_Aligned_block) - 1) / sizeof(_Aligned_block);
+        } else {
+            // allocator is stateful, we need space for it
+            constexpr size_t _Align = (_STD max)(alignof(_Alloc), sizeof(_Aligned_block));
+            return (_Size + sizeof(_Alloc) + _Align - 1) / sizeof(_Aligned_block);
+        }
+    }
+
     static void* _Allocate(_Alloc _Al, const size_t _Size) {
+        // memory layout:
+        // |--- coroutine frame --|- alignment padding -|---- _Alloc -----|
+        // |--- always present ---|----------- stateful only -------------|
+        // all rounded up to a multiple of sizeof(_Aligned_block)
+
+        const size_t _Size_in_blocks = _Allocation_block_size(_Size);
         if constexpr (default_initializable<_Alloc> && allocator_traits<_Alloc>::is_always_equal::value) {
             // do not store stateless allocator
-            const size_t _Count = (_Size + sizeof(_Aligned_block) - 1) / sizeof(_Aligned_block);
-            return _Al.allocate(_Convert_size<_Alloc_size_type>(_Count));
+            return _Al.allocate(_Convert_size<_Alloc_size_type>(_Size_in_blocks));
         } else {
             // store stateful allocator
-            constexpr size_t _Align = (_STD max)(alignof(_Alloc), sizeof(_Aligned_block));
-            const size_t _Count     = (_Size + sizeof(_Alloc) + _Align - 1) / sizeof(_Aligned_block);
-            void* const _Ptr        = _Al.allocate(_Convert_size<_Alloc_size_type>(_Count));
+            void* const _Ptr = _Al.allocate(_Convert_size<_Alloc_size_type>(_Size_in_blocks));
             const auto _Al_address =
                 (reinterpret_cast<uintptr_t>(_Ptr) + _Size + alignof(_Alloc) - 1) & ~(alignof(_Alloc) - 1);
             ::new (reinterpret_cast<void*>(_Al_address)) _Alloc(_STD move(_Al));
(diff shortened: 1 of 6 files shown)


stl/inc/generator
```diff
@@ -503,6 +564,9 @@ public:
 
 private:
     coroutine_handle<promise_type> _Coro = nullptr;
+    // The stack of coroutine handles depicted in the Standard is realized by a linked structure of promises and
+    // awaitable objects so all necessary storage is allocated in coroutine frames. See the description in the body of
```
CaseyCarter, line 568:
```suggestion
    // awaitable objects, so all necessary storage is allocated in coroutine frames. See the description in the body of
```

tests/std/include/test_generator_support.hpp
```diff
@@ -25,37 +57,52 @@ class StatelessAlloc : public std::allocator<T> {
...
+    }
 };
 
 static_assert(std::default_initializable<StatelessAlloc<int>>);
 
+template <class T>
+struct StatefulAlloc {
+    using value_type = T;
+
+    int domain;
```
CaseyCarter, line 83: Could be `private`, in which case this should be a `class`, not a `struct`.

stl/inc/generator
```diff
@@ -40,22 +40,38 @@ struct alignas(__STDCPP_DEFAULT_NEW_ALIGNMENT__) _Aligned_block {
...
             // do not store stateless allocator
-            const size_t _Count = (_Size + sizeof(_Aligned_block) - 1) / sizeof(_Aligned_block);
-            return _Al.allocate(_Convert_size<_Alloc_size_type>(_Count));
+            return _Al.allocate(_Convert_size<_Alloc_size_type>(_Size_in_blocks));
         } else {
             // store stateful allocator
-            constexpr size_t _Align = (_STD max)(alignof(_Alloc), sizeof(_Aligned_block));
-            const size_t _Count     = (_Size + sizeof(_Alloc) + _Align - 1) / sizeof(_Aligned_block);
-            void* const _Ptr        = _Al.allocate(_Convert_size<_Alloc_size_type>(_Count));
+            void* const _Ptr = _Al.allocate(_Convert_size<_Alloc_size_type>(_Size_in_blocks));
```
StephanTLavavej, line 74: Now that `_Allocation_block_size()` has been extracted (which is an awesome clarity improvement :heart_eyes_cat:), I observe that `_Al.allocate(_Convert_size<_Alloc_size_type>(_Size_in_blocks))` is identical on both branches here. Can/should we lift out `_Ptr`? We could also push down `return _Ptr;` and invert the `if constexpr` condition so that we only do work when we need to "store stateful allocator" although I don't think that's as important for clarity.

stl/inc/generator
```diff
@@ -71,23 +87,23 @@ public:
...
     }
 
     static void operator delete(void* const _Ptr, const size_t _Size) noexcept {
         if constexpr (default_initializable<_Alloc> && allocator_traits<_Alloc>::is_always_equal::value) {
-            // make stateless allocator
+            // recreate stateless allocator
             _Alloc _Al{};
-            const size_t _Count = (_Size + sizeof(_Aligned_block) - 1) / sizeof(_Aligned_block);
-            _Al.deallocate(static_cast<_Aligned_block*>(_Ptr), static_cast<_Alloc_size_type>(_Count));
+            const size_t _Size_in_blocks = _Allocation_block_size(_Size);
```
StephanTLavavej, line 105: Similarly, should we extract `_Size_in_blocks` here? (There's no opportunity for further unification though, since the nature of how we create `_Al` varies.)

stl/inc/generator
```diff
@@ -126,38 +154,40 @@ private:
         _Alloc _Al{_STD move(_Stored_al)};
         _Stored_al.~_Alloc();
 
-        const size_t _Count = (_Size + sizeof(_Al) + _Align - 1) / sizeof(_Aligned_block);
-        _Al.deallocate(static_cast<_Aligned_block*>(_Ptr), static_cast<allocator_traits<_Alloc>::size_type>(_Count));
+        const size_t _Size_in_blocks = (_Size + sizeof(_Alloc) + _Align - 1) / sizeof(_Aligned_block);
```
StephanTLavavej, line 157: Should this be using `_Allocation_block_size()` now?

stl/inc/generator
```diff
@@ -275,6 +307,10 @@ private:
     struct _Element_awaiter {
         remove_cvref_t<_Yielded> _Val;
 
+        _Element_awaiter(const remove_reference_t<_Yielded>& _Val_)
```
StephanTLavavej, line 310: Can/should this be `explicit`?

StephanTLavavej (changes requested):

tests/std/tests/P2502R2_generator/test.cpp
```diff
@@ -77,133 +115,66 @@ generator<tuple<ranges::range_reference_t<Rng1>, ranges::range_reference_t<Rng2>
     }
 }
 
-// Not from the proposal:
+void zip_example() {
+    using V = tuple<int, int>;
+    using R = tuple<int&, int&>;
+
+    auto g0 = co_zip(array{1, 2, 3}, vector{10, 20, 30, 40, 50});
+    test_one<gen_traits<R, V>, V, R, R>(move(g0), std::identity{}, array{tuple{1, 10}, tuple{2, 20}, tuple{3, 30}});
```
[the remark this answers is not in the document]
CaseyCarter: It's too weird that all but one callsite uses `identity`; I should just add an overload of `test_one`.

tests/std/tests/P2502R2_generator/test.cpp
```diff
@@ -280,64 +253,173 @@ void adl_proof_test() {
...
+}
 
+void operator delete(void* const p, size_t, align_val_t) noexcept {
+    ::_aligned_free(p);
+}
+
+class malloc_resource final : public pmr::memory_resource {
+    void* do_allocate(size_t bytes, size_t align) override {
+        assert(align <= __STDCPP_DEFAULT_NEW_ALIGNMENT__);
+        if (!bytes) {
```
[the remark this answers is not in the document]
CaseyCarter: While we're here, `return nullptr` seems not to conform to [mem.res.private]/2: "A derived class shall implement this function to return a pointer to **allocated storage** [emphasis added] with a size of at least bytes, aligned to the specified alignment."

stl/inc/generator
CaseyCarter: I could put the allocator recreation/retrieval in an IILE to eliminate the tail duplication. That normally feels a little too "clever" to me, but it actually seems readable here. I'll push it and see what reviewers think.

stl/inc/generator
CaseyCarter: Also, why are we repeating `default_initializable<_Alloc> && allocator_traits<_Alloc>::is_always_equal::value` 783249 times and not giving the condition a short semantically meaningful name?

tests/std/include/test_generator_support.hpp
```diff
@@ -6,13 +6,45 @@
...
+#include <generator>
 #include <memory>
 #include <memory_resource>
-#include <new>
 #include <type_traits>
 
+template <class R, class V = void, class A = void>
+struct gen_traits : gen_traits<R, V> {
+    using generator = std::generator<R, V, A>;
+};
```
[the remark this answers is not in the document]
CaseyCarter: I figured it wasn't quite worth an extra declaration to get the more normal ordering, but I'll yield to your more objective opinion.

tests/std/tests/P2502R2_generator/test.cpp
```diff
@@ -95,6 +95,11 @@ void test_one(generator<R, V, A> g0, invocable<generator<R, V, A>> auto adaptor,
     assert(i == ranges::cend(expected));
 }
 
+template <class Traits, class ValueType, class ReferenceType, class RvalueReferenceType, class R, class V, class A>
+void test_one(generator<R, V, A> g0, ranges::input_range auto&& expected) {
+    return test_one<Traits, ValueType, ReferenceType, RvalueReferenceType>(move(g0), identity{}, expected);
```
StephanTLavavej, line 100: This is passing `expected` as an lvalue instead of forwarding, intentional?

StephanTLavavej (approved):

tests/std/tests/P2502R2_generator/test.cpp
StephanTLavavej: (If this is an issue, OK to merge as-is and fix in the final `<generator>` PR into `main`)

CaseyCarter: Yes, intentional. General ranges are typically accepted by forwarding reference and handled as lvalues. The only operations are `begin` and `end`, and it doesn't make sense to call either without calling the other, so preserving rvalue-ness is mostly a waste of time. I suppose we could theoretically accept a general range and pass it to something with specialized behavior for a specific subset of types - "I take all `input_range`s but do something special if they happen to be `vector` rvalues" - and I'd `forward` in that case.

Yeah, I suppose it makes sense to `forward` in mid-layer code like this (1) to avoid needing to change mid layers when the ends change, and (2) to avoid the "you didn't `forward` this forwarding reference!" reflex.

I'll leave this open and address it in the final merge.

CaseyCarter: Addressed in #4953.

PR #4953, opened 2024-09-13
P2502R2 `<generator>`: Synchronous Coroutine Generator For Ranges

### LINKED ISSUE #2936: P2502R2 `<generator>`: Synchronous Coroutine Generator For Ranges

CaseyCarter:
This is the 44 commits from `feature/generator` plus one more to add a `forward` in a `test_one` overload in `P2502R2_generator`. 
I've verified that we implement the resolutions of the three LWG issues mentioned in #2936, and recently validated that we address all the "add a new header requirements".

Closes #2936, fixes https://github.com/microsoft/STL/pull/4952#discussion_r1757906602.

I stuffed all the implementation details in `<generator>` into `namespace _Gen_detail`, and simplified names that no longer need disambiguation. The delta looks huge, but the largest commit (b61b7ba55dc7a05ac2e498478823d8221ce6fd4d) is mostly just additional indentation, and the remainder are a single rename each.


stl/inc/generator
```diff
@@ -0,0 +1,592 @@
...
+#undef empty_bases
+
+_STD_BEGIN
+
+struct alignas(__STDCPP_DEFAULT_NEW_ALIGNMENT__) _Aligned_block {
+    unsigned char _Pad[__STDCPP_DEFAULT_NEW_ALIGNMENT__];
+};
+
+template <class _Alloc>
+concept _Has_real_pointers = same_as<_Alloc, void> || is_pointer_v<typename allocator_traits<_Alloc>::pointer>;
```
frederick-vs-ja, line 41: It it better to name it as `_Allocator_uses_raw_pointers`?

CaseyCarter: The `same_as<_Alloc, void> ||` means this isn't really reusable, so I'm not too concerned about making the name discoverable. If anything, the `ispointer_v` term should be `_Allocator_uses_raw_pointers`, and this should be stuffed into a `generator`-specific detail namespace (`_Gen`?) along with some other entities that are not intended to be global but are shared by multiple parts of the `generator` impl.

StephanTLavavej: I wouldn't mind a detail namespace, but `_Gen` would be very shadowy.

CaseyCarter: I chose `_Gen_detail`. There's a full description of the change in the OP.

stl/inc/generator
```diff
@@ -0,0 +1,600 @@
...
+        friend struct _Iter_provider;
+
+        // Least significant bit of `_Data` indicates stored information:
+        // LSB 0: `_Data` is a top coroutine handle,
+        // LSB 1: `_Data ^ 1U` is a pointer to an object of type `_Nest_info`.
+        uintptr_t _Data = reinterpret_cast<uintptr_t>(coroutine_handle<_Promise_base>::from_promise(*this).address());
+        add_pointer_t<_Yielded> _Ptr = nullptr;
+    };
+
+    struct _Secret_tag {};
```
StephanTLavavej, line 455: We should have an explicit default ctor to make this moderately difficult to misuse, right?

tests/std/include/test_generator_support.hpp
```diff
@@ -0,0 +1,150 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#pragma once
+
+#include <concepts>
+#include <cstddef>
+#include <cstdlib>
+#include <generator>
+#include <memory>
+#include <memory_resource>
```
StephanTLavavej, line 11: `<cstdlib>` and `<memory_resource>` became unused after #4952.

StephanTLavavej:
Thanks! :heart_eyes_cat: :zap:

* Pushed some changes, including the significant extraction of `_Get_allocator_address()` - please double-check.
* Had a couple of outstanding questions (see un-resolved comments).
* Reviewed everything except for the 4 test `.cpp` files before my brain ran out of caffeine - will resume tomorrow.

stl/inc/generator
```diff
@@ -0,0 +1,601 @@
...
+
+        _Iterator(_Iterator&& _That) noexcept : _Coro{_STD exchange(_That._Coro, {})} {}
+
+        _Iterator& operator=(_Iterator&& _That) noexcept {
+            _Coro = _STD exchange(_That._Coro, {});
+            return *this;
+        }
+
+        _NODISCARD _Ref operator*() const
+            noexcept(noexcept(static_cast<_Ref>(*_Coro.promise()._Get_top().promise()._Ptr))) {
```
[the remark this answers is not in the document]
cpplearner: See LWG-4057.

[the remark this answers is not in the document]
CaseyCarter:
```suggestion
        _NODISCARD _Ref operator*() const
            noexcept(noexcept(static_cast<_Ref>(*_Coro.promise()._Get_top().promise()._Ptr))) /* See LWG-4057 */ {
```

stl/inc/generator
```diff
@@ -0,0 +1,592 @@
...
+
+    struct _Nest_info {
+        exception_ptr _Except;
+        coroutine_handle<_Gen_promise_base> _Parent;
+        coroutine_handle<_Gen_promise_base> _Root;
+    };
+
+    struct _Final_awaiter {
+        _NODISCARD bool await_ready() noexcept {
+            return false;
```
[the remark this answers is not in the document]
CaseyCarter: No reason at all. I suspect there was something interesting here once, but obviously not now.

tests/std/tests/P2502R2_generator/test.cpp
```diff
@@ -0,0 +1,429 @@
...
+        auto pos = r.begin();
+        for (int i = 0; i < n / 2; ++i, ++*pos, ++pos) {
+            assert(pos != r.end());
+            assert(*pos == 2 * i);
+        }
+        assert(pos == r.end());
+    }
+
+#if !(defined(__clang__) && defined(_M_IX86)) // TRANSITION, LLVM-56507
+    { // Test with mutable xvalue reference type
```
StephanTLavavej, line 156: "xvalue reference type" is not actually a thing.

tests/std/tests/P2502R2_generator/test.cpp
```diff
@@ -0,0 +1,429 @@
...
+        auto woof = [](size_t size, size_t count) -> generator<vector<int>&&> {
```
StephanTLavavej, line 157: As much as I love animal noises for test data or simple test functions whose purpose is obvious, this is a fairly large lambda doing something fairly complicated. Would it be possible to give the lambda a more descriptive name?

StephanTLavavej, line 157: `size` could be `const`.

tests/std/tests/P2502R2_generator/test.cpp
```diff
@@ -0,0 +1,429 @@
...
+            random_device rd{};
+            uniform_int_distribution dist{0, 99};
```
StephanTLavavej, line 159: This is producing truly random values, but not logging them. (We have a system for randomized tests where we truly-random seed a Mersenne Twister and log the seed, so we can reproduce failures if necessary.) But then I observe that this test doesn't appear to actually care about any of the values being generated into the vector - it looks like they could all be constant. The use of `dist(rd)` therefore appears to be a distraction.

Could we just fill the vector with a constant? (Or if variation is needed for some reason I'm missing, `static_cast<int>(count)`?) Additionally, it looks like this might be the only use of `<random>` in the test, if so the header should be dropped.

tests/std/tests/P2502R2_generator/test.cpp
```diff
@@ -0,0 +1,429 @@
...
+    }
+}
+
+void recursive_test() {
+    static constexpr auto might_throw = []() -> generator<int> {
+        co_yield 0;
+        throw runtime_error{"error"};
+    };
+
+    static constexpr auto nested_ints = []() -> generator<int> {
```
StephanTLavavej, line 202: As scalars, these don't need to be `static`, they can be plain `constexpr`.

tests/std/tests/P2502R2_generator/test.cpp
```diff
@@ -0,0 +1,429 @@
...
+        try {
+            co_yield ranges::elements_of(might_throw());
+        } catch (const runtime_error& e) {
+            assert(e.what() == "error"sv);
```
StephanTLavavej, line 206: Need to include `<string_view>`.

tests/std/tests/P2502R2_generator/test.cpp
```diff
@@ -0,0 +1,429 @@
...
+
+class malloc_resource final : public pmr::memory_resource {
+private:
+    void* do_allocate(size_t bytes, size_t align) override {
+        assert(align <= __STDCPP_DEFAULT_NEW_ALIGNMENT__);
+        if (bytes == 0) {
+            bytes = 1;
+        }
+
+        if (void* result = malloc(bytes)) {
```
StephanTLavavej, line 369: Hyper nitpick: Previous lines said `void* const result`.

tests/std/tests/P2502R2_generator_iterator/test.cpp
```diff
@@ -0,0 +1,171 @@
...
+template <class Ref, class V = void, class Alloc = void>
+void test_one() {
+    using Gen  = generator<Ref, V, Alloc>;
+    using Iter = ranges::iterator_t<Gen>;
+    static_assert(input_iterator<Iter>);
+    static_assert(sizeof(Iter) == sizeof(void*)); // NB: implementation defined
+
+    // Test member types
+    static_assert(same_as<typename Iter::value_type, gen_value_t<Ref, V>>);
+    static_assert(same_as<typename Iter::difference_type, ptrdiff_t>);
```
StephanTLavavej, line 64: Should include `<cstddef>` for `ptrdiff_t`.

tests/std/tests/P2502R2_generator_iterator/test.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+    // Test copying functions
+    static_assert(!is_copy_constructible_v<Iter>);
+    static_assert(!is_copy_assignable_v<Iter>);
+
+    { // Test move constructor
+        Gen g  = generate_zero<Ref, V, Alloc>();
+        Iter i = g.begin();
+        Iter j = move(i);
```
StephanTLavavej, line 73: `<utility>` for `move()`.

tests/std/tests/P2502R2_generator_iterator/test.cpp
```diff
@@ -0,0 +1,171 @@
...
+    test_with_allocator<Proxy<T>&, T>();
+    test_with_allocator<const Proxy<T>&, T>();
+    test_with_allocator<Proxy<T>&&, T>();
+    test_with_allocator<const Proxy<T>&&, T>();
+}
+
+int main() {
+    test_with_type<int>();
+    test_with_type<float>();
+    test_with_type<string>();
```
StephanTLavavej, line 169: `<string>` for `string`. Skulls for the Skull Throne! :skull:

tests/std/tests/P2502R2_generator_promise/test.cpp
```diff
@@ -0,0 +1,242 @@
...
+    using Promise = Gen::promise_type;
+    using Alloc   = generator_allocator<Gen>::type;
+
+    // Test 'operator new(size_t)'
+    constexpr bool has_op_new1 = HasOperatorNew<Promise, size_t>;
+    static_assert(has_op_new1 == (same_as<Alloc, void> || default_initializable<Alloc>) );
+    if constexpr (has_op_new1) {
+        const size_t size = __STDCPP_DEFAULT_NEW_ALIGNMENT__;
+        void* const mem   = p.operator new(size);
+        assert(reinterpret_cast<intptr_t>(mem) % __STDCPP_DEFAULT_NEW_ALIGNMENT__ == 0);
```
StephanTLavavej, line 76: Suggestion: I would prefer to see `uintptr_t` here. (The less I have to think about negative values, the better.) Occurs below.

tests/std/tests/P2502R2_generator_promise/test.cpp
```diff
@@ -0,0 +1,242 @@
...
+    if constexpr (is_rvalue_reference_v<Yielded>
+                  && constructible_from<remove_cvref_t<Yielded>, const remove_reference_t<Yielded>&>) {
+        using Lval      = const remove_reference_t<Yielded>&;
+        using Awaitable = decltype(p.yield_value(declval<Lval>()));
+
+        static_assert(convertible_to<decltype(declval<Awaitable&>().await_ready()), bool>);
+        static_assert(is_void_v<decltype(declval<Awaitable&>().await_resume())>);
+        static_assert(noexcept(p.yield_value(declval<Lval>()))
+                      == is_nothrow_constructible_v<remove_cvref_t<Yielded>,
+                          const remove_reference_t<Yielded>&>); // strengthened
```
StephanTLavavej, line 145: Should use `Lval` here for brevity.

tests/std/tests/P2502R2_generator_promise/test.cpp
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    using ValTy = conditional_t<is_void_v<V>, remove_cvref_t<Ref>, V>;
+    if constexpr (!TestingIncomplete && convertible_to<ValTy&, Yielded>) {
+        // Test 'yield_value(ranges::elements_of<range>)'
+        test_yield_elements_of_range<Gen, vector<ValTy>>(p);
+        test_yield_elements_of_range<Gen, list<ValTy>>(p);
+        test_yield_elements_of_range<Gen, forward_list<ValTy>>(p);
+        test_yield_elements_of_range<Gen,
+            test::range<input_iterator_tag, ValTy, test::Sized::no, test::CanDifference::no, test::Common::no,
```
StephanTLavavej, line 167: `range_algorithm_support.hpp` drags it in, but we should probably include `<iterator>` for this direct mention of `input_iterator_tag`.

StephanTLavavej (changes requested):

StephanTLavavej:
:heart_eyes_cat: Looks great, thanks! I've finished reviewing both the product and test code and have one last round of nitpicks plus a significant question about the use of `<random>` (which is why I didn't just push changes).

tests/std/tests/P2502R2_generator/test.cpp
CaseyCarter: So, meow?

tests/std/tests/P2502R2_generator/test.cpp
CaseyCarter: My brain cannon has a taxonomy of iterators mostly based on the value category of `*i`: lvalue iterators, xvalue iterators, pure prvalue iterators (reference type and value type are the same), and proxy iterators (value category of `*i` is a prvalue, but not `value_type`). These comments are abbreviations of "test with reference type chosen so the iterator produced is a meow iterator" which is great and meaningful if you happen to be me. Apparently not so good for not me.

tests/std/tests/P2502R2_generator/test.cpp
CaseyCarter: There is just _way_ too much extraneous cruft here. Let's distill this down to just what it's actually testing.

tests/std/tests/P2502R2_generator_promise/test.cpp
CaseyCarter: Time for my regular "It's Standard that `<ranges>` includes `<iterator>` includes `<concepts>`" reminder. On a related note, I can't wait to start writing `import std;` in tests.

tests/std/tests/P2502R2_generator/test.cpp
CaseyCarter: I think I burned away enough cruft that this is now a proper short-and-sweet test case so the naming comment no longer applies. Shout if you disagree.

tests/std/tests/P2502R2_generator/test.cpp
StephanTLavavej: Ah, the second has to capture the first by value. That's a bigger change than I thought (the lambda is no longer empty) but I think making the dependency more explicit is a win.

tests/std/tests/P2502R2_generator/test.cpp
StephanTLavavej: Yes, this is much much easier to understand, thank you!

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
I had to push an additional commit to work around VSO-2254804 "EDG ICE in `cpfe.dll!make_coroutine_result_expression` with C++23 `<generator>` test" (an internal assertion, which is why we didn't see it with shipping EDG). I checked how much code I could enable without ICEing, and it turned out that the regions we need to workaround are exactly the same as for Clang x86.

CaseyCarter:
> ... it turned out that the regions we need to workaround are exactly the same as for Clang x86.

I suspect this is related to the confusion about coroutines and x86 MSABI discussed in LLVM-59382.

StephanTLavavej:
:zap: :rocket: :heart_eyes_cat:
===

Thank you @CaseyCarter, @AlexGuteniev, @cpplearner, @frederick-vs-ja, and @JMazurkiewicz for working together to land this major C++23 feature!

PR #4963, opened 2024-09-17
Don't deduce the return type of `emplace`{,`_back`,`_front`}

CaseyCarter:
The instantiation used to deduce the return types of template functions with placeholders in their declared return type happens outside the immediate context of any other template instantiation. Notably, that means the body of such an instantiation isn't an unevaluated context. 

Switching to concrete return types will:
1. Avoid return type deduction, correcting the problem described in DevCom-10745303, and
2. Correct the ODR problem described in #4962.

Fixes DevCom-10745303 / VSO-2252142 / AB#2252142

### DIFF
--- a/stl/inc/xmemory
+++ b/stl/inc/xmemory
@@ -2621,6 +2621,12 @@ _NODISCARD _Elem* _UIntegral_to_buff(_Elem* _RNext, _UTy _UVal) { // used by bot
 }
 _STD_END
 
+#if _HAS_CXX17
+#define _CONTAINER_EMPLACE_RETURN reference
+#else // ^^^ _HAS_CXX17 ^^^ / vvv !_HAS_CXX17 vvv
+#define _CONTAINER_EMPLACE_RETURN void
+#endif // ^^^ _HAS_CXX17 ^^^
+
 #pragma pop_macro("new")
 _STL_RESTORE_CLANG_WARNINGS
 #pragma warning(pop)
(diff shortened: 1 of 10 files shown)


stl/inc/xmemory
```diff
@@ -2621,6 +2621,12 @@ _NODISCARD _Elem* _UIntegral_to_buff(_Elem* _RNext, _UTy _UVal) { // used by bot
 }
 _STD_END
 
+#if _HAS_CXX17
```
AlexGuteniev, line 2624: Maybe add a comment like:
```c++
// Don't deduce `emplace_back` return type to guard against omniconvertible types
// and prevent ODR violation with C++14//C++17 mixing; see GH-4963
```

StephanTLavavej: We talked about this at the weekly maintainer meeting and we believe that the usage and intent of this macro is sufficiently clear that it doesn't need to be commented. Our git history and test coverage should be sufficient to remember why we did this.

Importantly, this area isn't being actively modified, and future containers and container-like entities will not have varying return types, so there won't be increasing usage of this macro over time.

stl/inc/xmemory
```diff
@@ -2621,6 +2621,12 @@ _NODISCARD _Elem* _UIntegral_to_buff(_Elem* _RNext, _UTy _UVal) { // used by bot
 }
 _STD_END
 
+#if _HAS_CXX17
+#define _CONTAINER_EMPLACE_RETURN reference
+#else // ^^^ _HAS_CXX17 ^^^ / vvv !_HAS_CXX17 vvv
+#define _CONTAINER_EMPLACE_RETURN void
+#endif // ^^^ _HAS_CXX17 ^^^
```
StephanTLavavej, line 2628: Typo!

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for emplacing this fix into the STL! :magic_wand: :gear: :smile_cat:

PR #4970, opened 2024-09-20
Use `char_traits` in streaming `operator>>` of `bitset`

### LINKED ISSUE #4956: `<bitset>`: streaming `operator >>` does not use character traits
`_Char` is directly compared against `_Elem0` and `_Elem1`
https://github.com/microsoft/STL/blob/73b5791e5c9eff1ece3ce593571fb30c31bf08d9/stl/inc/bitset#L589-L600

AlexGuteniev:
Fixes #4956

### DIFF
--- a/stl/inc/bitset
+++ b/stl/inc/bitset
@@ -589,15 +589,20 @@ basic_istream<_Elem, _Tr>& operator>>(basic_istream<_Elem, _Tr>& _Istr, bitset<_
             if (_Tr::eq_int_type(_Tr::eof(), _Meta)) { // end of file, quit
                 _State |= _Istr_t::eofbit;
                 break;
-            } else if (!_Tr::eq((_Char = _Tr::to_char_type(_Meta)), _Elem0) && !_Tr::eq(_Char, _Elem1)) {
+            }
+
+            if (!_Tr::eq((_Char = _Tr::to_char_type(_Meta)), _Elem0) && !_Tr::eq(_Char, _Elem1)) {
                 break; // invalid element
-            } else if (_Str.max_size() <= _Str.size()) { // no room in string, give up (unlikely)
+            }
+
+            if (_Str.max_size() <= _Str.size()) { // no room in string, give up (unlikely)
                 _State |= _Istr_t::failbit;
                 break;
-            } else { // valid, append '0' or '1'
-                _Str.push_back('0' + _Tr::eq(_Char, _Elem1));
-                _Changed = true;
             }
+
+            // valid, append '0' or '1'
+            _Str.push_back('0' + _Tr::eq(_Char, _Elem1));
+            _Changed = true;
         }
         _CATCH_IO_(_Istr_t, _Istr)
     }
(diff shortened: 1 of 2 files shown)


stl/inc/bitset
```diff
@@ -589,13 +589,13 @@ basic_istream<_Elem, _Tr>& operator>>(basic_istream<_Elem, _Tr>& _Istr, bitset<_
             if (_Tr::eq_int_type(_Tr::eof(), _Meta)) { // end of file, quit
                 _State |= _Istr_t::eofbit;
                 break;
-            } else if ((_Char = _Tr::to_char_type(_Meta)) != _Elem0 && _Char != _Elem1) {
+            } else if (!_Tr::eq((_Char = _Tr::to_char_type(_Meta)), _Elem0) && !_Tr::eq(_Char, _Elem1)) {
```
CaseyCarter, line 592: This line was a mess before, and adding to it hasn't improved readability. Could we drop the `else` (note that the `if` branch ends with `break`, so the `else` is extraneous) and move the assignment to its own line? (I wouldn't mind removing the other occurrences of `else`-after-`break`, but I won't demand it either.)

tests/std/tests/GH_004930_char_traits_user_specialization/test.cpp
```diff
@@ -280,6 +285,16 @@ CONSTEXPR20 bool test_gh_4930() {
 static_assert(test_gh_4930());
 #endif // _HAS_CXX20
 
+void test_gh_4956() {
+    basic_string<char, odd_char_traits<char>> s("QQPPQ", 5);
```
CaseyCarter, line 289: It's nice to have a short description of the issue the regression test is testing for:
```suggestion
    // bitset's stream extraction operator was using `!=` to compare characters instead of `traits::eq`
    basic_string<char, odd_char_traits<char>> s("QQPPQ", 5);
```

CaseyCarter (approved):

stl/inc/bitset
StephanTLavavej: Agreed, I'll validate and push changes. In new code we almost always avoid mixing modifications and tests. I think this is worth mixing a cleanup into a correctness change (but messing with the other else-breaks, not so much).

tests/std/tests/GH_004930_char_traits_user_specialization/test.cpp
```diff
@@ -280,6 +285,16 @@ CONSTEXPR20 bool test_gh_4930() {
 static_assert(test_gh_4930());
 #endif // _HAS_CXX20
 
+void test_gh_4956() {
+    basic_string<char, odd_char_traits<char>> s("QQPPQ", 5);
+
+    bitset<7> bs;
+    std::basic_stringstream<char, odd_char_traits<char>>(s) >> bs;
```
StephanTLavavej, line 292: This is constructing a temporary with parentheses. I'd prefer to split the construction and streaming to improve clarity.

Also, this can be a `basic_istringstream` - we aren't performing output here.

Also, `std::`.

tests/std/tests/GH_004930_char_traits_user_specialization/test.cpp
StephanTLavavej: I also captured both bug titles exactly.

StephanTLavavej (approved):

CaseyCarter (approved):

StephanTLavavej:

StephanTLavavej:
:zero: :one: :lady_beetle:
===

PR #4973, opened 2024-09-21
Use `steady_clock` in timed mutex test

### LINKED ISSUE #1472: VSO_0226079_mutex is sporadically failing
We encountered a sporadic test failure:

```
Assertion failed: time_execution([this] { unique_lock<Mutex> ul(mtx, 50ms); assert(!ul.owns_lock()); }) >= 50ms,
file C:\a\1\s\tests\std\tests\VSO_0226079_mutex\test.cpp, line 327
```

https://github.com/microsoft/STL/blob/ea156e730cf3370d0551a0532669b7de90c78915/tests/std/tests/VSO_0226079_mutex/test.cpp#L324-L327

It appears that `system_clock` behavior may be making this test unreliable.

AlexGuteniev:
_Can I tell you only one solution to this problem? Ditch the second clock!_

Fixes #1472

With the series of PRs ended with #4755 we should be fine with this change.

### DIFF
--- a/tests/std/tests/VSO_0226079_mutex/test.cpp
+++ b/tests/std/tests/VSO_0226079_mutex/test.cpp
@@ -148,12 +148,9 @@ public:
 
 template <typename Func>
 nanoseconds time_execution(Func&& f) {
-    // system_clock currently powers mutex waits, so we're using system_clock
-    // rather than high_resolution_clock here for consistency. This should be
-    // fixed with VSO-133414, VSO-166543, VSO-189735.
-    const auto startTime = system_clock::now();
+    const auto startTime = steady_clock::now();
     forward<Func>(f)();
-    return duration_cast<nanoseconds>(system_clock::now() - startTime);
+    return duration_cast<nanoseconds>(steady_clock::now() - startTime);
 }
 
 template <typename Mutex>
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
Re-labeling as "test" because this isn't a product code bug.

tests/std/tests/VSO_0226079_mutex/test.cpp
```diff
@@ -148,12 +148,9 @@ class other_mutex_thread {
 
 template <typename Func>
 nanoseconds time_execution(Func&& f) {
-    // system_clock currently powers mutex waits, so we're using system_clock
-    // rather than high_resolution_clock here for consistency. This should be
-    // fixed with VSO-133414, VSO-166543, VSO-189735.
```
StephanTLavavej, line 153: No change requested: For the record, these were

* VSO-133414 `<condition_variable>`: `wait_until()` is implemented with relative time
* VSO-166543 `<condition_variable>`: `wait_until` and `sleep_until` functions don't behave correctly for user-defined clocks
* VSO-189735 `<thread>`: `std::this_thread::sleep_for` hangs when system time is moved backwards

which indeed should all be fixed now.

StephanTLavavej (approved):

StephanTLavavej:
Awesome - thanks for digging this up and fixing it! :heart_eyes_cat:

StephanTLavavej:

StephanTLavavej:
:stopwatch: :lock: :hammer_and_wrench:
===

PR #4980, opened 2024-09-24
Deduce the return type of `queue::emplace` and `stack::emplace`

StephanTLavavej:
Followup to #4963. Fixes VSO-2261359 / AB#2261359 "\[VCPKG\] rocksdb installation failed with error C2440":

```
stack(125): error C2440: 'return': cannot convert from 'void' to 'rocksdb::TransactionBaseImpl::SavePoint &'
```

As the container adaptors are required to use `decltype(auto)` in C++17, the "warning C5046: Symbol involving type with internal linkage not defined" in the `convertible_to_any` scenario and the ODR violation when mixing C++14 and C++17 TUs are unavoidable.

In `<xmemory>`, I'm lexicographically sorting `_ADAPTOR_EMPLACE_RETURN` before `_CONTAINER_EMPLACE_RETURN`, which will avoid an adjacent-add conflict with #4975 adding `_LIST_REMOVE_RETURN` (also in lex order).

### DIFF
--- a/stl/inc/queue
+++ b/stl/inc/queue
@@ -135,7 +135,7 @@ public:
 #endif // _HAS_CXX23
 
     template <class... _Valty>
-    _CONTAINER_EMPLACE_RETURN emplace(_Valty&&... _Val) {
+    _ADAPTOR_EMPLACE_RETURN emplace(_Valty&&... _Val) {
 #if _HAS_CXX17
         return c.emplace_back(_STD forward<_Valty>(_Val)...);
 #else // ^^^ _HAS_CXX17 / !_HAS_CXX17 vvv
--- a/stl/inc/stack
+++ b/stl/inc/stack
@@ -120,7 +120,7 @@ public:
 #endif // _HAS_CXX23
 
     template <class... _Valty>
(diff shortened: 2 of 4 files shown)


CaseyCarter (approved):

CaseyCarter:
Woops 😊.

stl/inc/queue
```diff
@@ -135,7 +135,7 @@ public:
 #endif // _HAS_CXX23
 
     template <class... _Valty>
-    _CONTAINER_EMPLACE_RETURN emplace(_Valty&&... _Val) {
+    _ADAPTOR_EMPLACE_RETURN emplace(_Valty&&... _Val) {
 #if _HAS_CXX17
         return c.emplace_back(_STD forward<_Valty>(_Val)...);
 #else // ^^^ _HAS_CXX17 / !_HAS_CXX17 vvv
```
AlexGuteniev, line 141: I understand you avoid the void return for no reason, but here it could save a preprocessor branch, why not do that?

AlexGuteniev: Oh, actually the void return would enforce the void return type in C++4 mode, which is certainly not desired.

StephanTLavavej:

PR #4987, opened 2024-09-28
`remove` vectorization

AlexGuteniev:
# 📜   The algorithm

In-place `remove` algorithm that uses bit mask of vector comparisons as shuffle index to remove certain elements. 
The destination is advanced by a size taken from a lookup table too, although `popcnt` could have been used.

The details vary on depending on element size: 
 * The tables are selected to be of up to 256 entries, to process up to 8 elements at once. Bigger tables don't fit top level caches well. Some tables are smaller, when less than 8 elements can fit the vector.
 * 8 and 16 bit variants use SSE only, as 8 elements of 8 or 16 bits fit SSE register, and more elements will take bigger table, Also there's no cross-lane shuffle within SSE4.2 / AVX2 that works with elements of such sizes. They use the famous `pshufb` / `_mm_shuffle_epi8` to remove elements.
 * 32 and 64 bit variants use AVX2, and 64 bit variant uses only a table of 16 entries, as there are up to 4 elements only. They use `vpermd` / `_mm256_permutevar8x32_epi32` to remove elements, which is cross lane. SEE fallbacks are used with smaller tables, still surprisingly more efficient than scalar.
 * Need contiguous mask of bits, single bit per element, not few bits for the same comparison. 8-bit uses `pmovmskb`. 32 and 64 bit use `vmovmskps` / `vmovmskpd`, though they are for floating types, they fit well, and avoid the need of cross-lane swizzling to compress the mask. For 16-bit, `packsswb` is used, although `pshufb` could have been used as well.

# 🔍 Find first!

Before even starting, find is performed to find the first mismatch element. This is done for the correctness, and also there are   performance reasons why it is good:
* **Correctness**. [[algorithms.requirements]/3](https://eel.is/c++draft/algorithms.requirements#3) states: _For purposes of determining the existence of data races, algorithms shall not modify objects referenced through an iterator argument unless the specification requires such modification_. Whereas [[alg.remove]](https://eel.is/c++draft/alg.remove) is vague on how the algorithm should work, I think we should only write to elements that has to be written to
* **Vectorization**, We can have full AVX2 vector size as the step always, not only for 32 and 64 bit elements
* **Memory bandwidth**. The vectorized algorithm might be memory bound, saving writes may make it faster
* **Number of operations**. Fewer ops to just test the content

The existing `find` implementation is called. Hypothetically I could implement it inline and save some instructions in some cases, but such optimization has too negligible effect on performance,  while increasing complexity noticeably. Though this might be revised for future `remove_copy` if that and this would share the implementation.

# ⚠️ Correctness doubt - superfluous writes

The algorithm removes elements from the source vector (of 8 or less elements) by a shuffle operation, so that non-removed elements are placed contiguously in that vector. Then it writes the whole vector to the destination, and advances the destination pointer to the size of non-removed elements.

As a result:
* In the remaining range some elements are overwritten with some values, before they are overwritten with expected values
* In the removed range, some amount of elements (up to 8 of them) are overwritten with values of other elements, and never restored.

I have no doubts that overwriting elements in the resulting range to to some intermediate values before setting them to the expected values is correct. The write and the data race (in abstract machine terms) exist anyway, so extra write is not observable.

I have concerns regarding damaging the removed range. Changing these values is observable.

I'd appeal to that elements in removed range stay in valid-but-uspecified state, although I understand that the purpose of standard saying that is to enable moving of non-trivially-copyables, but not to do what I did.

Note that:
* It is possible to avoid to do any of superfluous write, but it will have some cost
* The cost of avoiding superfluous writes is small for 32 and 64 bit elements, and larger for 8 and 16 bit elements
* When//if vectorizing `remove_copy` in a similar way have to avoid superfluous writes anyway

# 🗄️ Memory usage

Unlike most other vectorization algorithms, this one uses large lookup tables. 8 and 32 bit variants use 2 KiB table, 16 bit variant uses 4 KiB table.

This has different performance characteristics, compared to pure-computational optimizations. In particular, it tends to behave worse in some programs that don't fit cache well on their critical path. This doesn't apply to benchmarks, but unfortunately often applies to realistic programs, especially the ones that are not written with having performance in mind.

I believe that the optimization is still good or at least not bad most of the time where it is needed.

# ⏱️ Benchmark results

Benchmark                           |     main  |    this     
------------------------------------|-----------|-------------
r<alg_type::std_fn, std::uint8_t>   |   944 ns  |  294 ns     
r<alg_type::std_fn, std::uint16_t>  |  1470 ns  |  297 ns     
r<alg_type::std_fn, std::uint32_t>  |  1059 ns  |  403 ns     
r<alg_type::std_fn, std::uint64_t>  |  1498 ns  |  884 ns     
r<alg_type::rng, std::uint8_t>      |  1208 ns  |  307 ns     
r<alg_type::rng, std::uint16_t>     |  1386 ns  |  288 ns     
r<alg_type::rng, std::uint32_t>     |  1218 ns  |  397 ns     
r<alg_type::rng, std::uint64_t>     |  1411 ns  |  842 ns


CaseyCarter:
Irony: a PR that adds vectorization entitled "`remove` vectorization".

AlexGuteniev:
> * **Correctness**. [[algorithms.requirements]/3](https://eel.is/c++draft/algorithms.requirements#3) states: _For purposes of determining the existence of data races, algorithms shall not modify objects referenced through an iterator argument unless the specification requires such modification_. Whereas [[alg.remove]](https://eel.is/c++draft/alg.remove) is vague on how the algorithm should work, I think we should only write to elements that has to be written to

There's a caveat: only the elements before the first removed one are not written. Elements after the resulting range (up to 8 of them) may be overwritten with values of some other elements. That's _valid but unspecified state_ as in [alg.remove]/7

If that's not fine, have to use AVX2 masks, which would be slower, and resort to having only 32-bit and 64-bit versions.

AlexGuteniev:
`remove_copy` if it succeeds going to be very similar, but would always use AVX2 mask, and so available only for 32 and 64 bit elements. I can try this within the same PR, although it seems big already.

AlexGuteniev:
Modern AMD data would be interesting. 
To avoid tricky swizzling, I added mixing integers and floats in 39974d1. The overall results are better for me, but this mixing seems to have more penalty on AMD than on Intel

stl/src/vector_algorithms.cpp
```diff
@@ -3626,6 +3626,175 @@ __declspec(noalias) void __stdcall __std_replace_8(
...
+
+        return _Dest;
+    }
+
+
+    template <size_t _Size_v, size_t _Size_h>
+    constexpr auto _Make_remove_patterns(const unsigned _Mul, const unsigned _Ew) {
+        struct {
+            uint8_t _Data[_Size_v][_Size_h];
+            uint8_t _Count[_Size_v];
```
AlexGuteniev, line 3652: I want to elaborate on the table element bit with decision, though it probably doesn't deserve a code comment.
`_Data` is explicitly widened for 32-bit and 64-bit cases using `_mm256_cvtepi8_epi32(_mm_loadu_si64(...));`
`_Count` is implicitly widened for all cases. 

This widening causes some extra instructions. 

Still prefer that over having larger tables. Larger tables are consuming more cache, which is hard to notice in a synthetic benchmark, but likely to be noticeable in a realistic program.

Currently the `_Data` table, which is the largest of the two, is 128 bytes for 64-bit elements, 1 KiB for 8 and 32 bit elements, 2 KiB for 16 bit elements. Typical L1 cache size is 64 KiB per core.

stl/src/vector_algorithms.cpp
```diff
@@ -3626,6 +3626,175 @@ __declspec(noalias) void __stdcall __std_replace_8(
...
+
+        return _Dest;
+    }
+
+
+    template <size_t _Size_v, size_t _Size_h>
+    constexpr auto _Make_remove_patterns(const unsigned _Mul, const unsigned _Ew) {
+        struct {
+            uint8_t _Shuf[_Size_v][_Size_h];
+            uint8_t _Size[_Size_v];
```
AlexGuteniev, line 3652: I want to elaborate on the table element bit with decision, though it probably doesn't deserve a code comment.
`_Shuf` is explicitly widened for 32-bit and 64-bit cases using `_mm256_cvtepu8_epi32(_mm_loadu_si64(...));`
`_Size` is implicitly widened for all cases. 

This widening causes some extra instructions for `_Shuf`. For `_Size` apparently it is free (`movzx` instead of `mov`)

I prefer having to widen over having larger tables. Larger tables are consuming more cache, which is hard to notice in a synthetic benchmark, but likely to be noticeable in a realistic program.

Currently the `_Shuf` table, which is the largest of the two, is 128 bytes for 64-bit elements, 1 KiB for 8 and 32 bit elements, 2 KiB for 16 bit elements. Typical L1 cache size is 64 KiB per core.

AlexGuteniev:
> `remove_copy` if it succeeds going to be very similar, but would always use AVX2 mask, and so available only for 32 and 64 bit elements. I can try this within the same PR, although it seems big already.

I'm now seeing multiple solutions how to do `remove_copy` for 8 and 16 bit elements

stl/src/vector_algorithms.cpp
```diff
@@ -3845,6 +3845,208 @@ __declspec(noalias) void __stdcall __std_replace_8(
...
+            _Advance_bytes(_First, 16);
+        } while (_First != _Stop);
+    }
+
+    return _Remove_fallback(_First, _Last, _Out, _Val);
+}
+
+} // extern "C"
+
 #ifndef _M_ARM64EC
```
StephanTLavavej, line 4050: There's a trivial merge conflict with `main` here.

stl/src/vector_algorithms.cpp
```diff
@@ -3845,6 +3845,208 @@ __declspec(noalias) void __stdcall __std_replace_8(
...
+
+            ++_Src;
+        }
+
+        return _Dest;
+    }
+
+
+    template <size_t _Size_v, size_t _Size_h>
+    constexpr auto _Make_remove_tables(const unsigned _Mul, const unsigned _Ew) {
```
StephanTLavavej, line 3868: This new code is missing `_M_ARM64EC` guards. Both direct use of intrinsics, and support functions like `_Make_remove_tables`, should be guarded.

StephanTLavavej, line 3868: We conventionally say `unsigned int`, never `unsigned`. Occurs below.

stl/src/vector_algorithms.cpp
```diff
@@ -3845,6 +3845,208 @@ __declspec(noalias) void __stdcall __std_replace_8(
...
+        } _Result;
+
+        for (unsigned _Vx = 0; _Vx != _Size_v; ++_Vx) {
+            unsigned _Nx = 0;
+
+            // Make shuffle mask for pshufb / vpermd corresponding to _Vx bit value.
+            // Every bit set corresponds to element skipped.
+            for (unsigned _Hx = 0; _Hx != _Size_h / _Ew; ++_Hx) {
+                if ((_Vx & (1 << _Hx)) == 0) {
+                    // Inner loop needed for cases where shuffle mask operate on element pars rather than whole elements
```
StephanTLavavej, line 3881: pairs?

stl/inc/algorithm
```diff
@@ -4589,6 +4589,30 @@ namespace ranges {
             _STL_INTERNAL_STATIC_ASSERT(sentinel_for<_Se, _It>);
             _STL_INTERNAL_STATIC_ASSERT(indirect_binary_predicate<ranges::equal_to, projected<_It, _Pj>, const _Ty*>);
 
+#if _USE_STD_VECTOR_ALGORITHMS
+#if _HAS_CXX20
+            if constexpr (_Vector_alg_in_find_is_safe<_It, _Ty> && sized_sentinel_for<_Se, _It>
+                          && is_same_v<_Pj, identity>) {
```
StephanTLavavej, line 4595: This appears to be hyper-bogus. The "eligible for optimization" check shouldn't be guarded. Additionally, this is ranges code, so C++20 is guaranteed.

stl/inc/xmemory
```diff
@@ -2198,6 +2231,31 @@ _NODISCARD_REMOVE_ALG _CONSTEXPR20 _FwdIt remove(_FwdIt _First, const _FwdIt _La
     _UFirst           = _STD _Find_unchecked(_UFirst, _ULast, _Val);
     auto _UNext       = _UFirst;
     if (_UFirst != _ULast) {
+#if _USE_STD_VECTOR_ALGORITHMS
+        if constexpr (_Vector_alg_in_find_is_safe<decltype(_UFirst), _Ty>) {
+#if _HAS_CXX20
+            if (!_STD is_constant_evaluated())
+#endif // _HAS_CXX20
```
StephanTLavavej, line 2238: Can use `_Is_constant_evaluated`.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -776,6 +776,62 @@ void test_swap_ranges(mt19937_64& gen) {
...
+void test_remove(mt19937_64& gen) {
+    using TD = conditional_t<sizeof(T) == 1, int, T>;
+    binomial_distribution<TD> dis(10);
+
+    vector<T> source;
+    vector<T> in_out_expected;
+    vector<T> in_out_actual;
+    vector<T> in_out_actual_r;
+
+    for (auto v : {&source, &in_out_expected, &in_out_actual, &in_out_actual_r}) {
```
StephanTLavavej, line 819: I know they're pointers but I'd still prefer to see `const auto&` here. Ditto below.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -922,6 +978,16 @@ void test_vector_algorithms(mt19937_64& gen) {
...
+    test_remove<signed char>(gen);
+    test_remove<unsigned char>(gen);
+    test_remove<short>(gen);
+    test_remove<unsigned short>(gen);
+    test_remove<int>(gen);
+    test_remove<unsigned int>(gen);
+    test_remove<long long>(gen);
+    test_remove<unsigned long long>(gen);
+
     test_swap_ranges<char>(gen);
```
StephanTLavavej, line 991: It looks like the definitions aren't being added in the same order as their calls. When possible, it's good to follow the same order, making it easier to verify that all test code is actually being called.

StephanTLavavej (changes requested):

stl/src/vector_algorithms.cpp
AlexGuteniev: parts

stl/inc/xmemory
AlexGuteniev: This construct is used everywhere else for `_CONSTEXPR20`

stl/src/vector_algorithms.cpp
AlexGuteniev: Let it be `unit32_t` then to avoid too much verbosity. 

Intrinsics code isn't portable, so saying `unsigned int` versus `uint32_t` does not actually convey any more information.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
AlexGuteniev: Swapped the definitions

stl/inc/xmemory
StephanTLavavej: Not everywhere:

https://github.com/microsoft/STL/blob/2c9e3e7f10366dbd93a16da73264464c8478cdcb/stl/inc/algorithm#L2121-L2136

StephanTLavavej:
Thanks for the detailed writeup! I have no correctness concerns here - the Standard has a note that the garbage values are valid-but-unspecified, so we're fully within our rights to leave totally random values in there, even if the range originally contained (for example) only `10` and `20`.

"If you want `partition`, you know where to find it."

stl/inc/xmemory
AlexGuteniev: Ok, maybe it indeed worth doing hereforth. The cost of debug codegen/throughput is tiny, and only for pre-C++20

stl/src/vector_algorithms.cpp
```diff
@@ -3859,6 +3859,218 @@ __declspec(noalias) void __stdcall __std_replace_8(
...
+    return _Remove_fallback(_First, _Last, _Out, _Val);
+}
+
+void* __stdcall __std_remove_4(void* _First, void* const _Last, const uint32_t _Val) noexcept {
+    _First     = const_cast<void*>(__std_find_trivial_4(_First, _Last, _Val));
+    void* _Out = _First;
+
+#ifndef _M_ARM64EC
+    const size_t _Size_bytes = _Byte_length(_First, _Last);
+    if (_Use_avx2() && _Size_bytes >= 32) {
```
StephanTLavavej, line 3991: This can be `if`-scoped, as it is in the other algorithms. It's still visible to the `else`, like in `__std_remove_8`.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: Pushed changes as usual, good results on my 5950X:

Benchmark                            |  Before |   After | Speedup |
-------------------------------------|--------:|--------:|--------:|
`r<alg_type::std_fn, std::uint8_t>`  | 1291 ns |  360 ns | 3.59
`r<alg_type::std_fn, std::uint16_t>` | 1291 ns |  338 ns | 3.82
`r<alg_type::std_fn, std::uint32_t>` | 1285 ns |  491 ns | 2.62
`r<alg_type::std_fn, std::uint64_t>` | 1504 ns | 1151 ns | 1.31
`r<alg_type::rng, std::uint8_t>`     | 1317 ns |  355 ns | 3.71
`r<alg_type::rng, std::uint16_t>`    | 1250 ns |  330 ns | 3.79
`r<alg_type::rng, std::uint32_t>`    | 1330 ns |  489 ns | 2.72
`r<alg_type::rng, std::uint64_t>`    | 2090 ns | 1082 ns | 1.93

StephanTLavavej:

StephanTLavavej (approved):

StephanTLavavej:
I've pushed a merge with `main` to resolve merge conflicts in `search.cpp` (structured as one commit to replicate the stupid thing I did in MSVC, followed by a proper fix, so I can mirror the latter).

Now, this directly constructs `std::string` from the `std::string_view`s `src_haystack` and `src_needle`, and constructs `std::vector` from their `.begin()` and `.end()`. These are stylistic improvements to reduce verbosity.

StephanTLavavej:
Thanks for removing time when users call `remove`! :joy_cat: :zany_face: :stopwatch:

Issue #314, opened 2019-11-20
STL: Visualize more types


StephanTLavavej:
It's been a while since we've done a "visualizer pass", looking at all of the types that have been added to the STL in recent years, and determining whether their representations are comprehensible in the debugger, or whether we need to improve how they're displayed.

`std::filesystem::path` and `std::shared_ptr<T[]>` probably need to be visualized, but the list is likely longer.

**Visualizer note:** The Visual Studio debugger "visualizes" STL data structures using a file named `stl.natvis`. (Example location: `C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\Common7\Packages\Debugger\Visualizers\stl.natvis`) This file is updated by Microsoft's STL maintainers because it's deeply connected to the representations of STL data structures, but at this time it isn't part of this GitHub repo.

CaseyCarter:
Per the Debugger team, the debugger now understands `char8_t`; we should ensure that `std::u8string` and `std::u8string_view` visualize properly.

mwinterb:
`std::u8string_view` works great since the visualizer for `std::basic_string_view` doesn't do anything overly special. The `std::basic_string` visualizers, however, are specialized for each known character type, `char`, `wchar_t`, `unsigned short`, `char16_t`, and `char32_t`. It looks like adding an `AlternativeType` entry for `char8_t` under the `char` specialization is enough.
![image](https://user-images.githubusercontent.com/1081839/97369059-64667100-1869-11eb-8c5a-a509d73b5592.png)
First block is for `std::basic_string<char8_t>`, second block is for `std::basic_string<char8_t>::iterator`
```
Comparing files woo.stl.natvis and STL.NATVIS
***** woo.stl.natvis
  <Type Name="std::basic_string&lt;char,*&gt;">
    <Intrinsic Name="size" Expression="_Mypair._Myval2._Mysize" />
***** STL.NATVIS
  <Type Name="std::basic_string&lt;char,*&gt;">
    <AlternativeType Name="std::basic_string&lt;char8_t,*&gt;" />
    <Intrinsic Name="size" Expression="_Mypair._Myval2._Mysize" />
*****

***** woo.stl.natvis
      <AlternativeType Name="std::_String_const_iterator&lt;std::_String_val&lt;std::_Simple_types&lt;char&gt; &gt; &gt;" />
      <SmartPointer Usage="Indexable">_Ptr,na</SmartPointer>
***** STL.NATVIS
[...]

PR #5005, opened 2024-10-08
Add visualizer for `std::chrono::system_clock::time_point`

Nerixyz:
Adds a visualizer for `std::chrono::system_clock::time_point`. To compute the date representation, it uses the same algorithm as [`_Civil_from_days`](https://github.com/microsoft/STL/blob/faccf0084ed9b8b58df103358174537233b178c7/stl/inc/chrono#L823-L959) (with the same "variables" as https://howardhinnant.github.io/date_algorithms.html#civil_from_days).
The time is displayed as hh:mm:ss.ssssss. Additionally, the visualizer shows `ns/100` (the actual value of the time-point), `us`, `ms`, and `s`.

Unfortunately, the [format specifiers](https://learn.microsoft.com/en-us/visualstudio/debugger/format-specifiers-in-cpp?view=vs-2022#BKMK_Visual_Studio_2012_format_specifiers) don't provide a way of formatting numbers with padding, so they're formatted "manually" (e.g. a 2-digit `v` is formatted as `{v/10}{v%10}`).

See also: https://github.com/microsoft/STL/issues/314

### DIFF
--- a/stl/debugger/STL.natvis
+++ b/stl/debugger/STL.natvis
@@ -2312,4 +2312,57 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
     </Expand>
   </Type>
 
+  <Type Name="std::chrono::time_point&lt;std::chrono::system_clock,std::chrono::duration&lt;__int64,std::ratio&lt;1,10000000&gt; &gt; &gt;">
+    <!--
+      Same computation as in std::chrono::year_month_day::_Civil_from_days
+      and https://howardhinnant.github.io/date_algorithms.html#civil_from_days.
+    -->
+    <Intrinsic Name="z"     Expression="(_MyDur._MyRep / (24 * 60 * 60 * 10000000ull)) + 719468"/>
+    <Intrinsic Name="era"   Expression="(z() &gt;= 0 ? z() : z() - 146096) / 146097"/>
+    <Intrinsic Name="doe"   Expression="(unsigned)(z() - era() * 146097)"/>
+    <Intrinsic Name="yoe"   Expression="(doe() - doe()/1460 + doe()/36524 - doe()/146096) / 365"/>
+    <Intrinsic Name="doy"   Expression="doe() - (365*yoe() + yoe()/4 - yoe()/100)"/>
+    <Intrinsic Name="mp"    Expression="(5*doy() + 2)/153"/>
+    <Intrinsic Name="day"   Expression="doy() - (153*mp()+2)/5 + 1"/>
+    <Intrinsic Name="month" Expression="mp() &lt; 10 ? mp()+3 : mp()-9"/>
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
* Can you show a screenshot?
* Because you're manually formatting numbers, is this totally wrecked by Hexadecimal Display? Can that be avoided with a decimal format specifier?

Nerixyz:
> * Can you show a screenshot?

| | |
|---|---|
| VS |   |
| VS (hex) |  |
| VS Code|  |

> * Because you're manually formatting numbers, is this totally wrecked by Hexadecimal Display? Can that be avoided with a decimal format specifier?

Yes - I updated it to use decimal formatting.

stl/debugger/STL.natvis
```diff
@@ -2312,4 +2312,57 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
...
+  <Type Name="std::chrono::time_point&lt;std::chrono::system_clock,std::chrono::duration&lt;__int64,std::ratio&lt;1,10000000&gt; &gt; &gt;">
+    <!--
+      Same computation as in std::chrono::year_month_day::_Civil_from_days
+      and https://howardhinnant.github.io/date_algorithms.html#civil_from_days.
+    -->
+    <Intrinsic Name="z"     Expression="(_MyDur._MyRep / (24 * 60 * 60 * 10000000ull)) + 719468"/>
+    <Intrinsic Name="era"   Expression="(z() &gt;= 0 ? z() : z() - 146096) / 146097"/>
+    <Intrinsic Name="doe"   Expression="(unsigned)(z() - era() * 146097)"/>
+    <Intrinsic Name="yoe"   Expression="(doe() - doe()/1460 + doe()/36524 - doe()/146096) / 365"/>
+    <Intrinsic Name="doy"   Expression="doe() - (365*yoe() + yoe()/4 - yoe()/100)"/>
```
StephanTLavavej, line 2324: No change requested: Stylistically, I would have put spaces around all binary operators, as we do in product code (enforced by clang-format). But I can see an argument for the choices here, and I'm not picky enough about visualizer syntax for it to be worth resetting testing.

StephanTLavavej (approved):

StephanTLavavej:
Thanks, this is awesome! :heart_eyes_cat:

It didn't matter for this PR, but for the future - we recommend against force-pushing after review has begun, as GitHub makes it more difficult to see what's incrementally changed. (Force-pushing is okay if there's a solid reason for it, like reorganizing commits to make them comprehensible, or if review hasn't started yet.)

CaseyCarter (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for this significant improvement - and congratulations on your first microsoft/STL commit! :tada: :heart_eyes_cat: :rocket:

I've filed a followup issue because we need to "triple-mirror" this change into the VS repo in order for it to actually take effect in the IDE.

PR #4934, opened 2024-09-04
Vectorize `basic_string::find_last_of`

AlexGuteniev:
# 🧭 Overview

 * Same `pcmpestri` approach as in `find_first_of` for 1 or 2 byte element. Did not bother making anything for larger characters.
 * The existing `find_first_of` was a bit copypasta reduced. Still the `_first_` and the `_last_` don't share the implementation, it would have been very confusing to iterate in different directions in a single loop.
 * The new functions have `_pos` in names to indicate being position based rather than pointer based. They are used in strings only, so pointers would have been extra conversion for no gain. We might add position-based `find_first_of` or pointer based `find_last_of` someday, so the naming scheme has provision for this.
 * Also the vectorization threshold was tuned (even for 8-bytes character, you'll hate it).
 * Test coverage was also expanded to cover the bug of having `basic_string<unsigned long long>`.

#  ⏱️ Benchmark results

Benchmark                                                   |       main    |   this
------------------------------------------------------------|---------------|-------------
bm<AlgType::str_member_last, char>/2/3                      |    7.79 ns    |   5.56 ns   
bm<AlgType::str_member_last, char>/7/4                      |    22.1 ns    |   12.5 ns   
bm<AlgType::str_member_last, char>/9/3                      |    23.8 ns    |   11.2 ns   
bm<AlgType::str_member_last, char>/22/5                     |    37.2 ns    |   11.9 ns   
bm<AlgType::str_member_last, char>/58/2                     |    63.3 ns    |   13.2 ns   
bm<AlgType::str_member_last, char>/102/4                    |    54.0 ns    |   16.3 ns   
bm<AlgType::str_member_last, char>/325/1                    |     119 ns    |   36.8 ns   
bm<AlgType::str_member_last, char>/400/50                   |     156 ns    |    178 ns   
bm<AlgType::str_member_last, char>/1011/11                  |     321 ns    |    100 ns   
bm<AlgType::str_member_last, char>/1502/23                  |     605 ns    |    308 ns   
bm<AlgType::str_member_last, char>/3056/7                   |     916 ns    |    284 ns   
bm<AlgType::str_member_last, wchar_t>/2/3                   |    5.82 ns    |   5.99 ns   
bm<AlgType::str_member_last, wchar_t>/7/4                   |    11.0 ns    |   11.6 ns   
bm<AlgType::str_member_last, wchar_t>/9/3                   |    11.2 ns    |   11.9 ns   
[...]

PR #5006, opened 2024-10-10
Simplify control flow in `_Traits_find_MEOW_not_of`

StephanTLavavej:
This simplifies `_Traits_find_first_not_of` and `_Traits_find_last_not_of`, following the patterns in #4744 and #4934, which will make them easier to understand and easier to add vectorized codepaths. These are lossless transformations that don't affect behavior.

* Replace unnecessarily confusing `break`s with immediate `return`s.
  + Except in `_Traits_find_last_of`, which is being changed by #4934.
* Add a `return` instead of relying on fall-through.
  + This will simplify the following transformation.
* Flip `!_Matches._Mark` tests.
  + Both sides unconditionally return, allowing us to flip them. (One case is a forever loop where the only ways out are returns.)
* Transform "call self with `_Special=false`" to fall-through.
* Extract `_Elem`.
* Extract `_Hay_start`, `_Hay_end`.
* Fuse `bool _Special` into its usage.


AlexGuteniev (changes requested):
I was going to do most of this in `not` vectorization PR, but maybe it is indeed cleaner not to mix.

As we no longer have tail calls here, remove `bool _Special = _Is_implementation_handled_char_traits<_Traits>` template parameter, as @frederick-vs-ja suggested and attempted.

StephanTLavavej:
> I was going to do most of this in not vectorization PR, but maybe it is indeed cleaner not to mix.

Yeah, it's difficult for me to follow two major changes combined, and I have a pretty high tolerance for review complexity.

> As we no longer have tail calls here, remove `bool _Special = _Is_implementation_handled_char_traits<_Traits>` template parameter, as @frederick-vs-ja suggested and attempted.

I agree that this should be done after #4934 and this PR land.

stl/inc/__msvc_string_view.hpp
```diff
@@ -635,7 +635,7 @@ constexpr size_t _Traits_rfind(_In_reads_(_Hay_size) const _Traits_ptr_t<_Traits
             }
 
             if (_Match_try == _Haystack) {
-                break; // at beginning, no more chance for match
+                return static_cast<size_t>(-1); // at beginning, no more chance for match
```
CaseyCarter: It seems odd to me that we're happy to early-return in this function, but we don't do so on line 631.

stl/inc/__msvc_string_view.hpp
```diff
@@ -654,7 +654,7 @@ constexpr size_t _Traits_rfind_ch(_In_reads_(_Hay_size) const _Traits_ptr_t<_Tra
             }
```
CaseyCarter: Ditto "early return on 650 could save us an indent".

stl/inc/__msvc_string_view.hpp
```diff
@@ -723,7 +723,7 @@ constexpr size_t _Traits_find_first_of(_In_reads_(_Hay_size) const _Traits_ptr_t
         const auto _Hay_start = _Haystack + _Start_at;
```
CaseyCarter: Ditto "early return on 722 could save us an indent".

stl/inc/__msvc_string_view.hpp
```diff
@@ -814,32 +814,33 @@ constexpr size_t _Traits_find_last_of(_In_reads_(_Hay_size) const _Traits_ptr_t<
     return static_cast<size_t>(-1); // no match
 }
 
-template <class _Traits, bool _Special = _Is_implementation_handled_char_traits<_Traits>>
+template <class _Traits>
 constexpr size_t _Traits_find_first_not_of(_In_reads_(_Hay_size) const _Traits_ptr_t<_Traits> _Haystack,
     const size_t _Hay_size, const size_t _Start_at, _In_reads_(_Needle_size) const _Traits_ptr_t<_Traits> _Needle,
     const size_t _Needle_size) noexcept {
     // in [_Haystack, _Haystack + _Hay_size), look for none of [_Needle, _Needle + _Needle_size), at/after _Start_at
     if (_Start_at < _Hay_size) { // room for match, look for it
```
CaseyCarter: Ditto "Consider early return".

stl/inc/__msvc_string_view.hpp
```diff
@@ -863,37 +864,39 @@ constexpr size_t _Traits_find_not_ch(_In_reads_(_Hay_size) const _Traits_ptr_t<_
     return static_cast<size_t>(-1); // no match
 }
 
-template <class _Traits, bool _Special = _Is_implementation_handled_char_traits<_Traits>>
+template <class _Traits>
 constexpr size_t _Traits_find_last_not_of(_In_reads_(_Hay_size) const _Traits_ptr_t<_Traits> _Haystack,
     const size_t _Hay_size, const size_t _Start_at, _In_reads_(_Needle_size) const _Traits_ptr_t<_Traits> _Needle,
     const size_t _Needle_size) noexcept {
     // in [_Haystack, _Haystack + _Hay_size), look for none of [_Needle, _Needle + _Needle_size), before _Start_at
     if (_Hay_size != 0) { // worth searching, do it
```
CaseyCarter: And one last recommendation for early-return.

CaseyCarter (approved):
This is such a nice simplification, I thought I'd push another commit to do some early-return outdenting.

CaseyCarter (approved):

StephanTLavavej:
:heart_eyes_cat: Looks purr to me!

StephanTLavavej:

CaseyCarter:
A merge conflict? Couldn't be me.

StephanTLavavej:
Remember, I named you Casey "Conflict" Carter via office post-it note almost a decade ago! :dart: :100: :crystal_ball:

PR #5008, opened 2024-10-12
`<bitset>`: optimize streaming `operator >>` 

AlexGuteniev:
# 📝  What it does

 * Instead of calling `push_back` in a loop for temporary `string`, use temporary buffer with integer index
    - Have it on stack if small enough
    - Lazy allocate to have exceptions handled as they were before
    - Index is also reused as `_Changed`, there's no need for explicit variable for that
    - We also don't need to check for `string::max_size()` then
    - Custom RAII for that buffer if it is large to avoid dragging in `<memory>`
 * Also restore not calling `Traits::length` in `operator <<`. The previous optimization avoided allocation, but introduced implicit length, which is good in total effect, but could have been better

##  More conservative alternative

`string::_Resize_and_overwrite` could have been used. 
It is slightly worse due to not avoiding the allocation for mid-sized `bitset`.
The benchmark results are listed too.

# 🔍 Where's coverage

Not a vectorization PR, so no need to stress. Existing coverage ought to be enough:
```
> python tests\utils\stl-lit\stl-lit.py ..\..\llvm-project\libcxx\test ..\..\tests\std --filter=bitset -v
-- Testing: 207 of 53614 tests, 12 workers --
```

# ⏱️ How much of improvement

Benchmark                          |     main     | resize_and_overwrite | custom buffer (current)
-----------------------------------|--------------|----------------------|---------------
bitset_from_stream<15, char>       |   22858 ns   |     24288 ns         |   20708 ns 
bitset_from_stream<16, char>       |   26394 ns   |     26224 ns         |   19523 ns 
bitset_from_stream<36, char>       |   17921 ns   |     14053 ns         |   10970 ns 
bitset_from_stream<64, char>       |   14559 ns   |      9911 ns         |    8045 ns 
bitset_from_stream<512, char>      |    8799 ns   |      5241 ns         |   10640 ns 
bitset_from_stream<2048, char>     |    9396 ns   |      4801 ns         |    4619 ns 
bitset_from_stream<15, wchar_t>    |   24327 ns   |     26221 ns         |   21511 ns 
bitset_from_stream<16, wchar_t>    |   27978 ns   |     25502 ns         |   20432 ns 
bitset_from_stream<36, wchar_t>    |   18754 ns   |     14578 ns         |   11391 ns 
bitset_from_stream<64, wchar_t>    |   15853 ns   |      9798 ns         |    8462 ns 
bitset_from_stream<512, wchar_t>   |    9609 ns   |      5244 ns         |    5014 ns 
bitset_from_stream<2048, wchar_t>  |    8348 ns   |      5109 ns         |    4334 ns 

# 🔜 What's up next

## `operator>>` non-performance

~~The loop structure can be improved.~~
~~It can avoid unnecessary `_Istr.rdbuf()->snextc()` on the final counter increment.~~

I'm calling that back< I realized that it is an intention.

## `operator>>` performance

Sigh... locales and streams.
Part of it is is in CRT which means it cannot be fixed.

So overall performance cannot reach the performance of from string construction.

image


stl/inc/bitset
```diff
@@ -40,6 +40,36 @@ __declspec(noalias) bool __stdcall __std_bitset_from_string_2(void* _Dest, const
 #endif // _USE_STD_VECTOR_ALGORITHMS
 
 _STD_BEGIN
+template <class _Ty, size_t _Size, bool _Dynamic = (_Size * sizeof(_Ty) > _STD_BITSET_TO_STREAM_STACK_RESERVATION)>
```
StephanTLavavej, line 43: One of the things I always say about template programming is that (especially prior to concepts) template parameters have no type system of their own, so it's especially important to rely on naming conventions to avoid surprises. In this case, it is deeply confusing (to me at least) for `_Ty` to appear here, when that has a very different meaning to `bitset` (i.e. the storage unit, either 32-bit or 64-bit). The usage is actually `_Buffer_for_bitset<_Elem, _Bits>`, so having `_Elem` "change name" is what's confusing.

Unless there's a very good reason to rename the parameters, I think we should keep the same names as the usage. This sounds like a superficial naming nitpick but it's one I really really care about.

StephanTLavavej, line 43: The `_Dynamic` criterion is improper.

The stack buffer is `_Ty _Buf[_STD_BITSET_TO_STREAM_STACK_RESERVATION];` and the dynamic option is `new _Ty[_Size];`. You should switch to dynamic when `_Size` exceeds the stack reservation constant, i.e. the `sizeof` shouldn't factor into the calculation. (If you wanted the constant to be a number of bytes, then the stack buffer's size is wrong. Element/byte confusion is dangerous, so proper unit naming is critical if you're crossing between domains.)

This was clearer to me after renaming the parameters.

StephanTLavavej:
Thanks! I pushed a conflict-free merge with `main` and several changes, including an important rename and a fix for the `_Dynamic` criterion, please double-check.

Final results on my 5950X:

Benchmark                           |   Before |    After | Speedup
:-----------------------------------|---------:|---------:|----:
`bitset_from_stream<15, char>`      | 24527 ns | 16516 ns | 1.49
`bitset_from_stream<16, char>`      | 26691 ns | 15469 ns | 1.73
`bitset_from_stream<36, char>`      | 18116 ns |  9761 ns | 1.86
`bitset_from_stream<64, char>`      | 14495 ns |  7542 ns | 1.92
`bitset_from_stream<512, char>`     |  8739 ns |  5392 ns | 1.62
`bitset_from_stream<2048, char>`    |  9585 ns |  5120 ns | 1.87
`bitset_from_stream<15, wchar_t>`   | 26066 ns | 18357 ns | 1.42
`bitset_from_stream<16, wchar_t>`   | 28599 ns | 16726 ns | 1.71
`bitset_from_stream<36, wchar_t>`   | 18729 ns | 10308 ns | 1.82
`bitset_from_stream<64, wchar_t>`   | 14788 ns |  8018 ns | 1.84
`bitset_from_stream<512, wchar_t>`  |  9561 ns |  5909 ns | 1.62
`bitset_from_stream<2048, wchar_t>` |  9118 ns |  5181 ns | 1.76

stl/inc/bitset
AlexGuteniev: I wanted the constant stack size (consistent with `basic_string` SSO)

AlexGuteniev:
> an important rename and a fix for the `_Dynamic` criterion

I don't agree. I wanted the constant stack size (like `basic_string` SSO, and like it was for `to_string` before PR), and I don't see how array size multiplied by element size wouldn't get that.

stl/inc/bitset
AlexGuteniev: Oh, I see now. What I intended was not complete. `_Bits` should have been passed there instead of `_STD_BITSET_TO_STREAM_STACK_RESERVATION`. I still think it is better to do that, than have constant element reservation.

stl/inc/bitset
```diff
@@ -608,14 +635,15 @@ basic_istream<_Elem, _Tr>& operator>>(basic_istream<_Elem, _Tr>& _Istr, bitset<_
...
+    size_t _Count                    = 0;
+    _Buffer_for_bitset<_Elem, _Bits> _Buf;
     const typename _Istr_t::sentry _Ok(_Istr);
 
     if (_Ok) { // valid stream, extract elements
         _TRY_IO_BEGIN
+        _Buf._Allocate();
         typename _Tr::int_type _Meta = _Istr.rdbuf()->sgetc();
-        for (size_t _Count = _Right.size(); 0 < _Count; _Meta = _Istr.rdbuf()->snextc(), (void) --_Count) {
+        for (; _Count < _Bits; (void) ++_Count, _Meta = _Istr.rdbuf()->snextc()) {
```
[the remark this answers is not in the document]
AlexGuteniev: > Or you could move the `_Count` update to the right, as it was originally

I've intentionally changed that in 26a55c3f7bf3fa80dca4633a6fee2eb8fe687fc1 

I was afraid that `snextc()` might throw, and the exception may be caught, and it this case we would need the count of successfully extracted characters.

stl/inc/bitset
AlexGuteniev: I mean like `_Elem _Buf[_Bits ? _Bits : 1];`

StephanTLavavej:
Ah - yes, at a minimum, it would make sense to not consume more stack space than necessary. (I think I missed that at first because the parameter wasn't named `_Bits`, and I didn't fully rethink the meaning after doing the rename).

If you additionally want to adjust the criterion so that we switch to dynamic once the size would exceed a number of bytes instead of a number of elements, go ahead - I'll move this back to WIP. Thanks!

AlexGuteniev:
Applied changes, verified that benchmark results are about the same (although there is seemingly random variation)

StephanTLavavej:

StephanTLavavej:
Thanks for optimizing this operator! :rocket: :stopwatch: :tada:

PR #5010, opened 2024-10-12
Auto-vectorize  minmax instead of manual vectorization

### LINKED ISSUE #4453: Should `ranges::minmax` be auto-vectorized instead of manual vectorization.
I now see that manually-vectorizing `ranges::minmax` in #4384 was not necessarily a good idea.
Original implementation was implemented in terms of `minmax_element` and was not vectorized. 
But if it was implemented directly, it would have been vectorized.
Consider the following:
```c++
unsigned int* max_element(unsigned int*b, unsigned int* e){
    unsigned int m = *b;
    unsigned int* r = b;
    for (++b; b != e; ++b){
        if (m < *b) {
            m = *b;
            r = b;
        }
    }
    return b;
}

unsigned int max1(unsigned int*b, unsigned int* e){
    return *max_element(b, e);
}

unsigned int max2(unsigned int *b, unsigned int* e){
    unsigned int m = *b;
    for (++b; b != e; ++b){
        if (m < *b) {
            m = *b;
        }
    }
    return m;
}
```
`max1` is not vectorized, but `max2` is. see https://godbolt.org/z/fG6Yzfz4v

So we could just reimplement `ranges::max_element` in headers and enjoy vectorization with more context.

Question how should we proceed:
 * Keep manual vectorization forever
 * Drop it in vNext
 * Drop it any time

AlexGuteniev:
Closes #4453

# 🤔 Should this be done?

I'm not sure.

In #4453 there's a decision we should.

#4673 was closed without merging.

The benchmark results are ambiguous.

# 🏁 Benchmark results

All benchmarks were ran with `/fp:fast` to make the floating versions affected too. (This optimization is only engaged with `/fp;fast`)

Since auto-vectorization doesn't do runtime dispatch, it was separately ran with `/arch:AVX2` and with default options

Benchmark                           | main + /fp:fast | this + /fp:fast| this + /fp:fast+ AVX2
------------------------------------|-----------------|----------------|-------------------
bm<uint8_t, Op::Min_val>/8021       |  68.2 ns        |  93.5 ns       |   72.2 ns    
bm<uint8_t, Op::Min_val>/63         |  4.52 ns        |  15.3 ns       |   22.3 ns    
bm<uint8_t, Op::Max_val>/8021       |  77.9 ns        |  90.9 ns       |   70.9 ns    
bm<uint8_t, Op::Max_val>/63         |  4.41 ns        |  17.9 ns       |   22.3 ns    
bm<uint8_t, Op::Both_val>/8021      |  94.9 ns        |   143 ns       |    145 ns    
bm<uint8_t, Op::Both_val>/63        |  5.86 ns        |  21.2 ns       |   39.3 ns    
bm<uint16_t, Op::Min_val>/8021      |   129 ns        |   177 ns       |    135 ns    
bm<uint16_t, Op::Min_val>/31        |  3.61 ns        |  6.93 ns       |   12.5 ns    
bm<uint16_t, Op::Max_val>/8021      |   147 ns        |   157 ns       |    137 ns    
bm<uint16_t, Op::Max_val>/31        |  3.59 ns        |  12.6 ns       |   12.4 ns    
bm<uint16_t, Op::Both_val>/8021     |   166 ns        |   244 ns       |    276 ns    
bm<uint16_t, Op::Both_val>/31       |  4.53 ns        |  13.2 ns       |   24.3 ns    
bm<uint32_t, Op::Min_val>/8021      |   267 ns        |   310 ns       |    224 ns    
bm<uint32_t, Op::Min_val>/15        |  3.18 ns        |  4.23 ns       |   5.94 ns    
bm<uint32_t, Op::Max_val>/8021      |   243 ns        |   334 ns       |    238 ns    
bm<uint32_t, Op::Max_val>/15        |  3.13 ns        |  4.11 ns       |   7.77 ns    
bm<uint32_t, Op::Both_val>/8021     |   334 ns        |   489 ns       |    422 ns    
bm<uint32_t, Op::Both_val>/15       |  3.93 ns        |  6.25 ns       |   8.69 ns    
bm<uint64_t, Op::Min_val>/8021      |  2829 ns        |  3840 ns       |   3793 ns    
bm<uint64_t, Op::Min_val>/7         |  4.59 ns        |  3.10 ns       |   3.34 ns    
bm<uint64_t, Op::Max_val>/8021      |  2849 ns        |  3835 ns       |   3791 ns    
bm<uint64_t, Op::Max_val>/7         |  4.66 ns        |  3.10 ns       |   3.35 ns    
bm<uint64_t, Op::Both_val>/8021     |  3027 ns        |  5719 ns       |   5761 ns    
bm<uint64_t, Op::Both_val>/7        |  9.24 ns        |  4.50 ns       |   4.47 ns    
bm<int8_t, Op::Min_val>/8021        |  76.4 ns        |  90.2 ns       |   69.6 ns    
bm<int8_t, Op::Min_val>/63          |  4.63 ns        |  15.1 ns       |   20.1 ns    
bm<int8_t, Op::Max_val>/8021        |  69.0 ns        |   106 ns       |   72.1 ns    
bm<int8_t, Op::Max_val>/63          |  4.59 ns        |  11.2 ns       |   24.2 ns    
bm<int8_t, Op::Both_val>/8021       |  87.2 ns        |   141 ns       |    142 ns    
bm<int8_t, Op::Both_val>/63         |  7.11 ns        |  18.3 ns       |   34.6 ns    
bm<int16_t, Op::Min_val>/8021       |   124 ns        |   163 ns       |    137 ns    
bm<int16_t, Op::Min_val>/31         |  4.06 ns        |  7.83 ns       |   12.4 ns    
bm<int16_t, Op::Max_val>/8021       |   132 ns        |   154 ns       |    137 ns    
bm<int16_t, Op::Max_val>/31         |  4.07 ns        |  7.85 ns       |   12.4 ns    
bm<int16_t, Op::Both_val>/8021      |   176 ns        |   248 ns       |    271 ns    
bm<int16_t, Op::Both_val>/31        |  5.11 ns        |  12.6 ns       |   20.1 ns    
bm<int32_t, Op::Min_val>/8021       |   247 ns        |   332 ns       |    235 ns    
bm<int32_t, Op::Min_val>/15         |  3.62 ns        |  3.78 ns       |   6.83 ns    
bm<int32_t, Op::Max_val>/8021       |   263 ns        |   306 ns       |    233 ns    
bm<int32_t, Op::Max_val>/15         |  3.70 ns        |  3.97 ns       |   7.15 ns    
bm<int32_t, Op::Both_val>/8021      |   328 ns        |   486 ns       |    433 ns    
bm<int32_t, Op::Both_val>/15        |  4.18 ns        |  5.89 ns       |   8.47 ns    
bm<int64_t, Op::Min_val>/8021       |  2866 ns        |  3837 ns       |   3789 ns    
bm<int64_t, Op::Min_val>/7          |  3.83 ns        |  3.21 ns       |   3.46 ns    
bm<int64_t, Op::Max_val>/8021       |  2821 ns        |  3792 ns       |   3786 ns    
bm<int64_t, Op::Max_val>/7          |  3.83 ns        |  3.15 ns       |   3.42 ns    
bm<int64_t, Op::Both_val>/8021      |  3105 ns        |  3802 ns       |   3796 ns    
bm<int64_t, Op::Both_val>/7         |  8.94 ns        |  4.01 ns       |   4.27 ns    
bm<float, Op::Min_val>/8021         |   864 ns        |   906 ns       |    440 ns    
bm<float, Op::Min_val>/15           |  3.35 ns        |  4.39 ns       |   6.50 ns    
bm<float, Op::Max_val>/8021         |   872 ns        |   911 ns       |    446 ns    
bm<float, Op::Max_val>/15           |  3.34 ns        |  3.76 ns       |   4.27 ns    
bm<float, Op::Both_val>/8021        |  1062 ns        |   946 ns       |    476 ns    
bm<float, Op::Both_val>/15          |   111 ns        |  4.97 ns       |   6.49 ns    
bm<double, Op::Min_val>/8021        |  1819 ns        |  1872 ns       |    967 ns    
bm<double, Op::Min_val>/7           |  3.21 ns        |  2.60 ns       |   2.67 ns    
bm<double, Op::Max_val>/8021        |  1826 ns        |  1851 ns       |    939 ns    
bm<double, Op::Max_val>/7           |  3.24 ns        |  2.61 ns       |   2.66 ns    
bm<double, Op::Both_val>/8021       |  1997 ns        |  1890 ns       |   1166 ns    
bm<double, Op::Both_val>/7          |   113 ns        |  3.32 ns       |   3.43 ns    

# 🥈 Results observations

On large arrays for integer types manual vectorization of `min` and `max` has significant advantage for 64-bit, and some disadvantage for 32-bit. For `minmax`, manual vectorization is clearly better.

Small tails cases are way better for manual vectorization most of the time.

However, auto vectorization does surprisingly well on floats. 

Very odd anomaly is seen on `bm<float, Op::Both_val>/15` and `bm<double, Op::Both_val>/7` cases, I double checked that it is a consistent result.

### DIFF
--- a/stl/inc/__msvc_minmax.hpp
+++ b/stl/inc/__msvc_minmax.hpp
@@ -21,61 +21,6 @@ struct _Min_max_element_t {
     const void* _Min;
     const void* _Max;
 };
-
-struct _Min_max_1i {
-    int8_t _Min;
-    int8_t _Max;
-};
-
-struct _Min_max_1u {
-    uint8_t _Min;
-    uint8_t _Max;
-};
-
-struct _Min_max_2i {
-    int16_t _Min;
-    int16_t _Max;
-};
-
-struct _Min_max_2u {
-    uint16_t _Min;
-    uint16_t _Max;
-};
-
-struct _Min_max_4i {
-    int32_t _Min;
-    int32_t _Max;
-};
-
-struct _Min_max_4u {
-    uint32_t _Min;
-    uint32_t _Max;
-};
-
-struct _Min_max_8i {
-    int64_t _Min;
-    int64_t _Max;
-};
-
-struct _Min_max_8u {
-    uint64_t _Min;
-    uint64_t _Max;
-};
-
-struct _Min_max_f {
-    float _Min;
-    float _Max;
-};
-
-struct _Min_max_d {
-    double _Min;
-    double _Max;
-};
-
-struct _Min_max_p {
-    void* _Min;
-    void* _Max;
-};
 } // extern "C"
 
 #pragma pop_macro("new")
--- a/stl/inc/algorithm
+++ b/stl/inc/algorithm
@@ -59,17 +59,6 @@ const void* __stdcall __std_find_last_trivial_2(const void* _First, const void* 
 const void* __stdcall __std_find_last_trivial_4(const void* _First, const void* _Last, uint32_t _Val) noexcept;
 const void* __stdcall __std_find_last_trivial_8(const void* _First, const void* _Last, uint64_t _Val) noexcept;
 
-__declspec(noalias) _Min_max_1i __stdcall __std_minmax_1i(const void* _First, const void* _Last) noexcept;
-__declspec(noalias) _Min_max_1u __stdcall __std_minmax_1u(const void* _First, const void* _Last) noexcept;
-__declspec(noalias) _Min_max_2i __stdcall __std_minmax_2i(const void* _First, const void* _Last) noexcept;
-__declspec(noalias) _Min_max_2u __stdcall __std_minmax_2u(const void* _First, const void* _Last) noexcept;
-__declspec(noalias) _Min_max_4i __stdcall __std_minmax_4i(const void* _First, const void* _Last) noexcept;
-__declspec(noalias) _Min_max_4u __stdcall __std_minmax_4u(const void* _First, const void* _Last) noexcept;
-__declspec(noalias) _Min_max_8i __stdcall __std_minmax_8i(const void* _First, const void* _Last) noexcept;
-__declspec(noalias) _Min_max_8u __stdcall __std_minmax_8u(const void* _First, const void* _Last) noexcept;
-__declspec(noalias) _Min_max_f __stdcall __std_minmax_f(const void* _First, const void* _Last) noexcept;
-__declspec(noalias) _Min_max_d __stdcall __std_minmax_d(const void* _First, const void* _Last) noexcept;
-
 // TRANSITION, DevCom-10610477
 __declspec(noalias) void __stdcall __std_replace_4(
     void* _First, void* _Last, uint32_t _Old_val, uint32_t _New_val) noexcept;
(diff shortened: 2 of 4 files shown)


AlexGuteniev:
> However, auto vectorization does surprisingly well on floats.
> 
> Very odd anomaly is seen on `bm<float, Op::Both_val>/15` and `bm<double, Op::Both_val>/7` cases, I double checked that it is a consistent result.

### 🐞 First of all, this is a severe compiler bug that affects manually vectorized code. The compiler fails to place `vzeroupper` properly.

I'm afraid I don't have a workaround.

Also another thing smaller regarding codegen that I wasn't able to fix from my side.

AlexGuteniev:
> ### 🐞 First of all, this is a severe compiler bug that affects manually vectorized code. The compiler fails to place `vzeroupper` properly.

Reported DevCom-10767462

AlexGuteniev:
> Reported DevCom-10767462

This problems get unnoticed thus far because:
 * The original PR only had a benchmark for large array, where vectorization benefit outweighs the misplaced `vzeroupper` penalty
 * The recently merged masks PR added small arrays to benchmark masks, but it wasn't benchmarked with `/fp:fast` where floats+values case is executed
 * The test coverage doesn't catch this, as it is only perf problem that doesn't affect correcness

AlexGuteniev:
> manual vectorization of `min` and `max` has significant advantage for 64-bit

This is obvious without even looing into it. The compiler doesn't attempt to vectorize 64-bit min/max. There are no 64-bit min/max instructions, and the compiler does not try `cmpgt` + `blendv`. 

The benefit from manual vectorization here  is small, less than x2, so maybe it doesn't worth a compiler optimization.

AlexGuteniev:
Note that the whole problem with floating `minmax`, while being severe, seems rather rare:
 * It only affects `ranges::minmax`. The classic algorithms have only `_element` which is not affected.
 * It only affects `/fp:fast`. With other floating modes, including the default, floating `minmax` uses `_element` implementation to deal properly with negative zeros.

AlexGuteniev:
> Also another thing smaller regarding codegen that I wasn't able to fix from my side.

Perhaps this would be fixable if we determine on keeping manual vectorization somehow; I've managed to generate proper insns sequence with intrinsics on Godbolt.

StephanTLavavej:
We talked about this at the weekly maintainer meeting. Ignoring the enormous sunk cost of having reviewed all of the manual vectorization PRs for minmax, it would be generous to say that the performance results of auto-vectorization here are mixed - it sounds like manual vectorization is generally superior. As a result, this PR would be perceived as a performance regression. At this time, we believe that keeping the current approach is superior. Thanks for looking into this!

AlexGuteniev:
I have looked into the cases where the auto vectorization wins. Unlike the manual vectorization, the auto one does the following things:
 * Loop unrolling by two iterations
 * Memory operand for min/max instruction instead of separate load and min/max

Looks like the speedup is contributed by entirely the first factor.

Not sure if manual vectorization should unroll loops.

PR #5013, opened 2024-10-13
Generalize vectorized `bitset::to_string` implementation

AlexGuteniev:
Not an optimization, rather maintenance change.

Experience with bitset from string and others has shown that while traits-based approach is a bit harder to read, it is less error prone.

The benchmark results are within the same.

The only semantic change is that now `memcpy` is always used for tail, instead of just one specific case where tail can be long.

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -3845,7 +3845,6 @@ __declspec(noalias) void __stdcall __std_replace_8(
 
 } // extern "C"
 
-#ifndef _M_ARM64EC
 namespace {
     namespace __std_bitset_to_string {
 #ifdef _M_ARM64EC
@@ -3944,7 +3943,6 @@ namespace {
                 _mm_storeu_si128(static_cast<__m128i*>(_Dest), _Elems);
             }
         };
-#endif // ^^^ !defined(_M_ARM64EC) ^^^
 
         template <class _Traits, class _Elem>
         void __stdcall _Impl(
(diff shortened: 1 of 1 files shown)


stl/src/vector_algorithms.cpp
```diff
@@ -3847,189 +3847,172 @@ __declspec(noalias) void __stdcall __std_replace_8(
...
+                _Elem _Tmp[_Step_size_bits];
+                _Traits::_Out(_Tmp, _Elems);
+                const _Elem* const _Tmpd = _Tmp + (_Step_size_bits - _Size_bits);
+                memcpy(_Dest, _Tmpd, _Size_bits * sizeof(_Elem));
             }
         }
-    } else
+
+        template <class _Avx_traits, class _Sse_traits, class _Elem>
+        void __stdcall _Dispatch(_Elem* const _Dest, const void* const _Src, size_t _Size_bits, const _Elem _Elem0,
```
StephanTLavavej, line 3983: `_Size_bits` can be `const` in these refactored wrappers - only the inner implementation modifies it.

Ditto for `_Src` below.

stl/src/vector_algorithms.cpp
```diff
@@ -3847,189 +3847,172 @@ __declspec(noalias) void __stdcall __std_replace_8(
...
+        template <class _Traits, class _Elem>
+        void __stdcall _Impl(
+            _Elem* const _Dest, const void* _Src, size_t _Size_bits, const _Elem _Elem0, const _Elem _Elem1) noexcept {
+            constexpr size_t _Step_size_bits = sizeof(_Traits::_Value_type) * 8;
+
+            const auto _Px0 = _Traits::_Set(_Elem0);
+            const auto _Px1 = _Traits::_Set(_Elem1);
+            if (_Size_bits >= _Step_size_bits) {
+                _Elem* _Pos = _Dest + _Size_bits;
+                _Size_bits &= (_Step_size_bits - 1);
```
StephanTLavavej, line 3958: Style nitpick: Parentheses here are unnecessary, as this is a compound assignment. Bitwise operators are squirrely by themselves, but not in compound assignments.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for refactoring this code! :desktop_computer: :gear: :magic_wand:

PR #5016, opened 2024-10-14
Fix floating `minmax` performance

AlexGuteniev:
Work around DevCom-10767462

Looks fragile (how can we be sure that the compiler won't misplace another `vzeroupper` too?) but seems to work.

Alternative to #5010

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -2222,7 +2222,7 @@ namespace {
         }
     }
 
-    // TRNSITION, DevCom-10767462
+    // TRANSITION, DevCom-10767462
     template <_Min_max_mode _Mode, class _Traits, bool _Sign>
     auto __std_minmax_impl_wrap(const void* const _First, const void* const _Last) {
         auto _Rx = __std_minmax_impl<_Mode, _Traits, _Sign>(_First, _Last);
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -2222,6 +2222,7 @@ namespace {
         }
     }
 
+#ifndef _M_ARM64EC
     // TRANSITION, DevCom-10767462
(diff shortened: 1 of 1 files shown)


stl/src/vector_algorithms.cpp
```diff
@@ -2222,11 +2222,23 @@ namespace {
         }
     }
 
+    // TRNSITION, DevCom-10767462
```
StephanTLavavej, line 2225: Typo.

stl/src/vector_algorithms.cpp
```diff
@@ -2222,11 +2222,23 @@ namespace {
         }
     }
 
+    // TRNSITION, DevCom-10767462
+    template <_Min_max_mode _Mode, class _Traits, bool _Sign>
+    auto __std_minmax_impl_wrap(const void* const _First, const void* const _Last) {
```
StephanTLavavej, line 2227: Missing `noexcept`.

stl/src/vector_algorithms.cpp
```diff
@@ -2222,11 +2222,23 @@ namespace {
         }
     }
 
+    // TRNSITION, DevCom-10767462
+    template <_Min_max_mode _Mode, class _Traits, bool _Sign>
+    auto __std_minmax_impl_wrap(const void* const _First, const void* const _Last) {
+        auto _Rx = __std_minmax_impl<_Mode, _Traits, _Sign>(_First, _Last);
+        _mm256_zeroupper();
```
StephanTLavavej, line 2229: Missing `_M_ARM64EC` guard.

StephanTLavavej (approved):

StephanTLavavej:
Thanks for figuring out this problem and how to work around it! I pushed a conflict-free merge with `main`, and an ARM64EC fix.

StephanTLavavej:

StephanTLavavej:
:adhesive_bandage: :heart_eyes_cat: :tada:
===

PR #5027, opened 2024-10-18
Don't throw arbitrary iterators at `sized_sentinel_for`

CaseyCarter:
* Split `_Idl_distance` into Ranges and pre-Ranges flavors. Who put their Ranges in my peanut butter??!?
* Move `_STD _Idl_dist_min` into `ranges` for consistency.

Fixes VSO-2285862 and VSO-2285916. (Currently validating.)

This is a mirror of MSVC-PR-586447. (I created the mirror PR early so I can verify this fixes things that #5015 broke in RWC.)

### DIFF
--- a/stl/inc/algorithm
+++ b/stl/inc/algorithm
@@ -1576,7 +1576,7 @@ namespace ranges {
             auto _UFirst = _RANGES _Unwrap_iter<_Se1>(_STD move(_First));
             auto _ULast  = _RANGES _Get_final_iterator_unwrapped<_It1>(_UFirst, _STD move(_Last));
             _STD _Seek_wrapped(_First, _ULast);
-            const auto _Count = _STD _Idl_distance<_It1>(_UFirst, _ULast);
+            const auto _Count = _RANGES _Idl_distance<_It1>(_UFirst, _ULast);
             auto _UOutput     = _STD _Copy_backward_unchecked(
                     _STD move(_UFirst), _STD move(_ULast), _STD _Get_unwrapped_n(_STD move(_Output), -_Count));
             _STD _Seek_wrapped(_Output, _STD move(_UOutput));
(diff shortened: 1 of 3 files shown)


StephanTLavavej:
Hahaha - I noticed the missing namespace-closing comment in the internal PR without having run clang-format outside of my brain. :brain: :100: :mage_man:

CaseyCarter:
> Hahaha - I noticed the missing namespace-closing comment in the internal PR without having run clang-format outside of my brain. 🧠 💯 🧙‍♂️

VS Code has been driving me nuts the last couple of updates deciding to not format-on-save seemingly at random.

stl/inc/algorithm
```diff
@@ -7404,13 +7404,16 @@ template <class _Diff1, class _Diff2>
...
-        return _Distance_unknown{};
-    } else {
-        return _Lhs + _Rhs;
-    }
+#if _HAS_CXX20
+    if constexpr (is_same_v<_Diff1, ranges::_Distance_unbounded> || is_same_v<_Diff2, ranges::_Distance_unbounded>) {
+        return ranges::_Distance_unbounded{};
+    } else
+#endif // _HAS_CXX20
+        if constexpr (is_same_v<_Diff1, _Distance_unknown> || is_same_v<_Diff2, _Distance_unknown>) {
```
StephanTLavavej, line 7412: This `else` is weird - we don't have this pattern anywhere else. If a statement were inserted here, it would lead to non-braced control flow problems. I won't demand a testing reset, but I will clean it up later if you don't get to it first.

StephanTLavavej (approved):

StephanTLavavej:
Thanks for fixing this up! :hammer_and_wrench: :magic_wand: :black_cat:

stl/inc/algorithm
CaseyCarter: Pushed a cleanup for this into my `misc` branch.

PR #5040, opened 2024-10-25
Add `.mailmap` file

CaseyCarter:
Git will use this file to determine a canonical name and email address for each committer and/or author email address to display in the output of `git log` and `git blame`. This is useful when people change names or emails or if someone wants to map their different accounts to a single canonical set of info.

See https://git-scm.com/docs/gitmailmap for details.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,10 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# This file maps author and committer email addresses to canonical names and email
+# addresses for display in the output of `git log` and `git blame`.
+# Format:
+# Canonical Name <canonical@address.dns> <actual@address.dns>
+
+Amy Wishnousky <amyw@microsoft.com> <stwish@microsoft.com>
+Anju del Moral Gonzalez <delMoralGonzalez.Anju@microsoft.com> <judelmor@microsoft.com>
--- a/.mailmap
+++ b/.mailmap
@@ -4,7 +4,8 @@
 # This file maps author and committer email addresses to canonical names and email
 # addresses for display in the output of `git log` and `git blame`.
 # Format:
-# Canonical Name <canonical@example.com> <actual@example.com>
+# Canonical Name <canonical@example.com> <commit@example.com>
+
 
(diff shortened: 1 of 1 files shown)


.mailmap
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# This file maps author and committer email addresses to canonical names and email
+# addresses for display in the output of `git log` and `git blame`.
+# Format:
+# Canonical Name <canonical@address.dns> <actual@address.dns>
```
StephanTLavavej, line 7: Is `address.dns` actually reserved? I think this should be either `example.com` or anything ending in `.example`.

StephanTLavavej (changes requested):

.mailmap
StephanTLavavej, line 7: Additionally, I think we should follow the example of https://git-scm.com/docs/gitmailmap and say "commit" instead of "actual" for the second address.

StephanTLavavej (changes requested):

.mailmap
CaseyCarter: "Contoso" would be funny. But yes, I think `example.com` makes sense.
```suggestion
# Canonical Name <canonical@example.com> <actual@example.com>
```

.mailmap
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# This file maps author and committer email addresses to canonical names and email
+# addresses for display in the output of `git log` and `git blame`.
+# Format:
+# Canonical Name <canonical@example.com> <actual@example.com>
```
CaseyCarter, line 7:
```suggestion
# Canonical Name <canonical@example.com> <commit@example.com>

```

StephanTLavavej (approved):

.mailmap
StephanTLavavej: Arguably `contoso.example` and `fabrikam.example` would be even better by depicting that the domain can change, but this is of course fine!

StephanTLavavej:

PR #4943, opened 2024-09-09
Vectorize `find_end `

AlexGuteniev:
#  📜 Overview
 * Similar to #4745 in the opposite direction
 * The algorithm is not very similar though. I have not attempted the one-index `pcmpestri` instruction and variable step, but it looked like multiple-indices `pcmpestrm` and fixed step would work better for this direction, because the former would return more partial false matches than in forward direction, as the match with higher index is more likely to be partial. 
    - If that's not convincing enough, I can try the other approach
    - `pcmpestrm` approach could have been made more efficient if DevCom-10689455 is fixed, but honestly I don't hope for that much
 * Also restructured control flow a vit in `__std_search_impl`, the place where `_Match_1st_16` is introduced. I basically implemented the same thing from scratch, and the new attempt gave clearer control flow implementation. They are not shared though, as `_First1 > _Stop1` check is needed for variable step, but not needed for fixed step, so here's still some variation.
 * The types larger than 2 bytes were excluded from test and benchmark -- same way as we don't test `replace` with smaller elements.
 * The benchmark patterns list is expanded to test small and large needle closer to the other end. I deliberately left coverage for `search` + closer to start and `find_end` + closer to end, It is useful to test the case where startup cost contributes more.
 * Also added couple of "evil" patterns in benchmark. The goal for them is not to make them too much worse. I leave up to maintainers to decide if this goal is achieved

# 👿 The Evil case  

Unlike many other algorithms, `search` algorithm run time highly depends on the data. It is possible to craft data which takes way more time than typical data of the same length.

The worst case could be haystack of a single repeating value, and the needle with the same repeating values and another value in the end. This is bad for plain `search`, although Boyer-Moore or similar algorithms are expected to handle that better.

**⚠️  The results for `find_end` are much worse for that case ⚠️**

**⚠️ The results for `search` are also worse after vectorization for such case, but not that much as `find_end` ⚠️**

It look possible to detect the evil case (by the amount of matched beginnings per some amount of data) and switch to another algorithm, or to modify the whole algorithm to be less affected. It can be done in this or subsequent PR.

I doubt how much the Evil case is important.

# 🏁 Benchmark results

Legend for the benchmark parameter:
https://github.com/microsoft/STL/blob/60825c4511144238c0b3485c9fc788b94a4244b2/benchmarks/src/search.cpp#L71-L77

Benchmark                                  |    main     |    this   
-------------------------------------------|-------------|--------------
[...]

PR #5042, opened 2024-10-26
Vectorize `find_end`, make sure ASan passes

AlexGuteniev:
Re-try of #4943 

Reverts #5041 and fixes the bug that triggered ASan.

# 🦠 The Bug

## The overall structure

We do SSE4.2 checks that search for up to 16 bytes sequence in a 16 bytes sequence. Such search can result in a match if it is a full match, or if it is a match of the beginning, and can be a match or mismatch when looking into further bytes.

We have two different code branches for needles within 16 bytes and longer needles. The bug is in the longer needles branch.

## Short needle branch

The 16 or less bytes branch can have SSE4.2 match as confirmed, when the needle fits fully, or a match that needs further confirmation when beginning of the needle matched. 

The search starts from the 16 bytes before the end, to have data for the first SSE4.2 instruction.

The whole haystack is split into three parts:
1. The last part, that is 16 bytes closest to the end. Here we can have only confirmed match, as whatever doesn't fit, doesn't have any remaining part.
2. The middle part 16 bytes sequences further on. Here we can have match that needs confirmation. In this case, 16 bytes are loaded from the potential match offset, and compared against the needle, using valid bytes mask. The load is safe, as we already passed the last part here, so we certainly have more than 16 bytes available.
3. The first part. Any remaining bytes that are less than 16 bytes. The same as middle part, except that we know that some matches are not possible. Mask them out to avoid checking again.

## Long needle branch

Any match needs further confirmation.

The search starts from the byte offset from the end that can have first match. This offset is greater than 16 bytes.

The whole haystack  split into two parts:
1. The main part. For any match starting from nonzero offset we load 16 bytes and compare with needle beginning, then `memcmp` the remaining. For match starting from zero offset we can go directly to `memcmp`, as first 16 bytes are already checked
2. The first part. Again, any remaining bytes, again, no difference, except the mask.

Comparing first 16 bytes before going to `memcmp` is an optimization to check first 16 bytes faster. We already have the needle in a vector register, and we know that there are more than 16 bytes, so we can compare them faster, and skip `memcmp` at all, if there's a mismatch.

## Meet the bug

On the first iteration of main part, we have a possibility that a match has nonzero offset. In this case, either the 16 bytes check or `memcmp` would do out-of-range read.

# 💥 The Impact

* ASan failure when building STL with ASan
* Potentially unallocated memory read
* False match when the characters beyond the end happen to match

# 💊 The Fix

Here are two ways possible:
1. Make the last part handled separately, like in short needles branch, with only zero offset match being valid
2. Start search from the earlier offset, so that any match is valid

Whereas the second approach is more appealing from the performance perspective, it is harder to reason about. One of the difficulties of the second approach is that skipping more bytes than 16 may result in offset before the beginning. In this case we have to do one matching, with beginning offset, and apply proper mask to the result of such matching.

So, this fix implements the first approach. It uses `pxor`/`ptest` for matching, as to match only first offset we don't need `pcmp*str*`, and can use faster instructions, and if they confirm, uses `memcmp` for the rest.

# ⏱️ The Fix Impact

About the same results in the benchmark. Within the usual variation.

### DIFF
--- a/stl/inc/algorithm
+++ b/stl/inc/algorithm
@@ -59,6 +59,11 @@ const void* __stdcall __std_find_last_trivial_2(const void* _First, const void* 
 const void* __stdcall __std_find_last_trivial_4(const void* _First, const void* _Last, uint32_t _Val) noexcept;
 const void* __stdcall __std_find_last_trivial_8(const void* _First, const void* _Last, uint64_t _Val) noexcept;
 
+const void* __stdcall __std_find_end_1(
+    const void* _First1, const void* _Last1, const void* _First2, size_t _Count2) noexcept;
+const void* __stdcall __std_find_end_2(
+    const void* _First1, const void* _Last1, const void* _First2, size_t _Count2) noexcept;
+
 __declspec(noalias) _Min_max_1i __stdcall __std_minmax_1i(const void* _First, const void* _Last) noexcept;
 __declspec(noalias) _Min_max_1u __stdcall __std_minmax_1u(const void* _First, const void* _Last) noexcept;
 __declspec(noalias) _Min_max_2i __stdcall __std_minmax_2i(const void* _First, const void* _Last) noexcept;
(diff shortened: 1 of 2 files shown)


StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for figuring out how to fix this! :hammer_and_wrench: :brain: :magic_wand:

PR #5046, opened 2024-10-27
Refactor `_Copy_memmove` and `_Copy_memmove_n`

CaseyCarter:
* BUG: `_Copy_memmove_n` adds a `size_t` directly to an iterator.
* BUG: `ranges::copy_n` converts `n` to `size_t` without verifying it is non-negative.
* There's a redundant recalculation of `n` in our existing implementation of `_Copy_memmove_n` which calls `_Copy_memmove`. Eliminate that redundancy by extracting the commonality into a new function `_Copy_memmove_tail` called by both `_Copy_memmove` and `_Copy_memmove_n`. 
* `_Copy_n_unchecked4` should verify its precondition that `n >= 0`. Drive-by: We can unconditionally `_STL_INTERNAL_STATIC_ASSERT(_Integer_like<_SizeTy>)` since #5015 defined `_Integer_like` in pre-20 modes

### DIFF
--- a/stl/inc/xutility
+++ b/stl/inc/xutility
@@ -4763,9 +4763,8 @@ template <class _InIt, class _SizeTy, class _OutIt>
 _CONSTEXPR20 _OutIt _Copy_n_unchecked4(_InIt _First, _SizeTy _Count, _OutIt _Dest) {
     // copy _First + [0, _Count) to _Dest + [0, _Count), returning _Dest + _Count
     // note: has callers outside the copy family
-#if _HAS_CXX20
     _STL_INTERNAL_STATIC_ASSERT(_Integer_like<_SizeTy>);
-#endif // _HAS_CXX20
+    _STL_INTERNAL_CHECK(_Count >= 0);
 
     if constexpr (_Iter_copy_cat<_InIt, _OutIt>::_Bitcopy_assignable) {
 #if _HAS_CXX20
--- a/stl/inc/xutility
+++ b/stl/inc/xutility
@@ -4708,32 +4708,40 @@ _CONSTEXPR20 void _Verify_ranges_do_not_overlap(const _Iter1& _First1, const _Se
 #endif // _ITERATOR_DEBUG_LEVEL != 2 ^^^
 }
 
-template <class _CtgIt, class _OutCtgIt>
-_OutCtgIt _Copy_memmove(_CtgIt _First, _CtgIt _Last, _OutCtgIt _Dest) {
-    auto _FirstPtr              = _STD _To_address(_First);
-    auto _LastPtr               = _STD _To_address(_Last);
-    auto _DestPtr               = _STD _To_address(_Dest);
-    const char* const _First_ch = const_cast<const char*>(reinterpret_cast<const volatile char*>(_FirstPtr));
-    const char* const _Last_ch  = const_cast<const char*>(reinterpret_cast<const volatile char*>(_LastPtr));
-    char* const _Dest_ch        = const_cast<char*>(reinterpret_cast<const volatile char*>(_DestPtr));
-    const auto _Count           = static_cast<size_t>(_Last_ch - _First_ch);
-    _CSTD memmove(_Dest_ch, _First_ch, _Count);
+template <class _OutCtgIt>
+_OutCtgIt _Copy_memmove_tail(
+    const char* const _First_ch, _OutCtgIt const _Dest, const size_t _Byte_count, const size_t _Object_count) {
+    _STL_INTERNAL_CHECK(_Byte_count == _Object_count * sizeof(*_Dest));
+    const auto _DestPtr  = _STD _To_address(_Dest);
+    char* const _Dest_ch = const_cast<char*>(reinterpret_cast<const volatile char*>(_DestPtr));
+    _CSTD memmove(_Dest_ch, _First_ch, _Byte_count);
     if constexpr (is_pointer_v<_OutCtgIt>) {
+        (void) _Object_count;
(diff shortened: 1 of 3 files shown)


CaseyCarter:
Bah, I've once again forgotten that I can't queue a draft PR and manually trigger a test run.

StephanTLavavej:
This still needs regression tests?

CaseyCarter:
> This still needs regression tests?

Yes. I'll keep it in WIP until I find some time to add them.

stl/inc/xutility
```diff
@@ -4708,32 +4708,40 @@ _CONSTEXPR20 void _Verify_ranges_do_not_overlap(const _Iter1& _First1, const _Se
...
-_OutCtgIt _Copy_memmove_n(_CtgIt _First, const size_t _Count, _OutCtgIt _Dest) {
-    const auto _Result = _STD _Copy_memmove(_First, _First + _Count, _Dest);
-    if constexpr (is_pointer_v<_OutCtgIt>) {
-        return _Result;
-    } else { // _Result is unused so the compiler can optimize it away
-        return _Dest + static_cast<_Iter_diff_t<_OutCtgIt>>(_Count);
-    }
+_OutCtgIt _Copy_memmove(_CtgIt _First, _CtgIt _Last, _OutCtgIt _Dest) {
+    _STL_INTERNAL_CHECK(_First <= _Last);
+    const auto _FirstPtr        = _STD _To_address(_First);
```
CaseyCarter, line 4730: Idle thought: Should we use `const auto*` as the type of the results of `to_address` to fail quickly if someone implements `pointer_traits<meow>::to_address` as returning a non-pointer type? Should we `static_assert(is_pointer_v)` in `to_address` instead?

StephanTLavavej: I've never liked `const auto*` and I've eradicated several occurrences that have accumulated. A `static_assert` would be clearer.

tests/std/tests/GH_000431_copy_move_family/test.cpp
```diff
@@ -641,6 +642,85 @@ void test_algorithms(CopyFn copy_fn) {
     }
 }
 
+#if _HAS_CXX20
+struct MyIterator { // A contiguous iterator with a weirdly narrow difference type
+    using iterator_concept = std::contiguous_iterator_tag;
+    using value_type       = int;
+    using difference_type  = short;
+
+    int* ptr = nullptr;
```
StephanTLavavej, line 651: Suggestion: Everywhere else consistently uses `value_type` and you've made changes in other PRs to be more consistent about using typedefs. It would seem reasonable to replace `int` here too, and in the constructor.

tests/std/tests/GH_000431_copy_move_family/test.cpp
```diff
@@ -641,6 +642,85 @@ void test_algorithms(CopyFn copy_fn) {
...
+        auto tmp = *this;
+        --ptr;
+        return tmp;
+    }
+
+    bool operator==(const MyIterator&) const  = default;
+    auto operator<=>(const MyIterator&) const = default;
+
+    value_type& operator[](difference_type i) const {
+        return ptr[i];
```
StephanTLavavej, line 687: Extreme nitpick: All of your other indices are `n`, and below you use `i` for an iterator. This one would be more consistent if named `n`.

tests/std/tests/GH_000431_copy_move_family/test.cpp
```diff
@@ -641,6 +642,85 @@ void test_algorithms(CopyFn copy_fn) {
...
+static_assert(std::contiguous_iterator<MyIterator>);
+
+void test_copy_n_regressions() {
+    // _Copy_memmove_n was adding a size_t to an iterator without converting to its difference_type:
+    //  warning C4267: 'argument': conversion from 'size_t' to 'MyIterator::difference_type', possible loss of data
+    std::copy_n(MyIterator{}, -42, MyIterator{});
+
+    // ranges::copy_n wasn't guarding against negative n when calling the memmove optimization
+    int x = 42;
+    std::ranges::copy_n(&x, -42, &x);
```
StephanTLavavej, line 720: Nitpick: In test code, I always recommend "unique numbers for unique purposes". Here, there is no connection between the `int`'s value of `42` and the count's value of `-42`, but using the same absolute value is very slightly confusing.

(Using `-42` for both negative counts is fine - same purpose, no interaction.)

StephanTLavavej, line 720: Suggestion: You've gone to all the trouble of defining an actually-functioning iterator (instead of just compile-only declared-but-not-defined). We could make this test more comprehensive with very little additional code:

* Let's test negative, zero, and positive cases.
* We can use `MyIterator{}` with `ranges::copy_n` too, which would exercise the `difference_type`.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: I pushed various nitpicks and a significant expansion of test coverage - please meow if you have any objections, I was a bit more daring than usual.

tests/std/tests/GH_000431_copy_move_family/test.cpp
```diff
@@ -17,6 +17,7 @@
 #endif // _HAS_CXX17
 
 #if _HAS_CXX20
+#include <compare>
```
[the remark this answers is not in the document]
CaseyCarter: I wouldn't have made this change had I remembered, but I don't think it's worth any effort to remove.

StephanTLavavej:

StephanTLavavej:
Thanks for memmoving these bugs out of our implementation! :lady_beetle: :truck: :joy_cat:

PR #5060, opened 2024-10-31
Internally disable `vformat.pass.cpp`

CaseyCarter:
Compiling this test is reliably taking over ten minutes during MSVC-internal STL ASan testing. Surprisingly enough the nearly identical `vformat.locale.pass.cpp` is not. Let's disable the test internally until we can figure out what's going on.


StephanTLavavej (approved):

CaseyCarter:
Sorry for the noise, I've decided to address this by increasing the test timeout. We've traditionally kept the timeout fairly tight on our test pipelines, and I realized that I used the same timeouts when creating the ASan pipelines. Both building and running a test with ASan enabled are going to take longer, and we should account for that.

Issue #933, opened 2020-06-27
<algorithm>: randomized test coverage for std::search_n and ranges::search_n


CaseyCarter:
We've been bitten by bugs in the deceptively simple `search_n` algorithm in the past, and `ranges::search_n` makes some slightly different implementation choices from `std::search_n`. It would raise confidence in the correctness of those choices if we had a test that searched random strings of binary digits for sequences of `0`s or `1`s of randomly-chosen lengths and validated that `ranges::search_n` and `std::search_n` produce identical correct results. 

(Such a randomized test must ensure reproducibility and could do so by following the model of [`tests/std/tests/P0067R5_charconv/test.cpp`](https://github.com/microsoft/STL/blob/1e312b38db8df1dfbea17adc344454feb8d00dd9/tests/std/tests/P0067R5_charconv/test.cpp#L66-L122).)

AlexGuteniev:
With #5352 we have progress towards this one. We have some random coverage, with the following differences:
1. We don't pretty print the failed case
2. We don't have seed as command-line arg
3. We only have coverage for random iterators, forward iterators branch is not covered
4. We have more random strings than just binary 0 and 1

I expect that 4 is fine. Are all other differences still required to be addressed to resolve this issue?

StephanTLavavej:
As long as we record the seed data, printing the failed case isn't absolutely necessary. We can also live without a convenience mechanism to pass a seed on the command line (since we can always edit the code to pass the seed). However, if the algorithm has substantially different codepaths for RanIt vs. FwdIt, covering both would be a good idea.

PR #5063, opened 2024-11-03
Partition vector algorithms test: move out lex compare family

AlexGuteniev:
`VSO_0000000_vector_algorithms` tends to take longer with more algorithms vectorized.

Think we need to partition it.

Starting with lex compare family. They are taking about 30% of total run time, 
Also the lex compare direction looks complete, and unlikely to have overlaps with other algorithms.


StephanTLavavej:
Would it be better to keep everything in a single file, but controlled via macros?

https://github.com/microsoft/STL/blob/a1bc1261795d4097cf7c12cfd0b5e2091809f281/tests/std/tests/P2321R2_views_zip/env.lst#L9-L13

With the `variant_msvc` tests, I permanently split them into a separate file because they're developed totally separately from the LLVM-derived tests. With the vector algorithms tests, I think keeping them in a single file might be better, because that way we can easily re-group how much we run each time.

AlexGuteniev:
> Would it be better to keep everything in a single file, but controlled via macros?

Is there an easy way to select a group using some extra parameter `python tests\utils\stl-lit\stl-lit.py ..\..\tests\std\tests\VSO_0000000_vector_algorithms -v` command?

AlexGuteniev:
I'm not yet concerned about run time of a single configuration. The total run time seems too long.

AlexGuteniev:
Also, partitioning algorithms by algorithm type seems more natural than partitioning views by iterator type.
Algorithms implementations are unrelated (not going to split where they are related).

StephanTLavavej:
We could use tags (like with ASAN), but since the algorithms are unrelated and there's no commonality between the test support machinery, having a separate test does make sense. Let's keep this as-is, thanks!

AlexGuteniev:
> there's no commonality between the test support machinery

There is no new commonality extracted during this separation, as all the commonality, which is mostly the randomness initialization, was previously extracted in #4734 into `/tests/std/include/test_vector_algorithms_support.hpp` to support separate floating minmax testing.  

This specific commonality, I think, is not an indication that these tests should be together. 
Instead, it seems to be something potentially reusable in more separate tests. 
`<charconv>` test mentioned in #933 is a candidate:
https://github.com/microsoft/STL/blob/1e312b38db8df1dfbea17adc344454feb8d00dd9/tests/std/tests/P0067R5_charconv/test.cpp#L66-L122

AlexGuteniev:
Actually I'm floating the separation idea with this PR. 

When thinking of `search_n` I concluded that comprehensive enough test would have a cubic run time: O(H*N*F) where H is haystack length, N is needle length and F is the frequency of a match. And it has no relationship to anything else. So I think it deserves the whole separate `.cpp`

StephanTLavavej (approved):

StephanTLavavej:
> When thinking of `search_n` I concluded that comprehensive enough test would have a cubic run time

If a comprehensive test would have a long run time, then I'd recommend having an off-by-default "EXHAUSTIVE" mode that allows the whole space to be manually run, but having automated runs select a randomized subset of cases.

Expensive tests are problematic both from a machine resourcing and timeout perspective, so we need to be careful.

StephanTLavavej:
Thanks for keeping an eye on this ever-growing test! :hatching_chick: :baby_chick: :chicken:

PR #5071, opened 2024-11-09
`<filesystem>`: Improve `symlink_status` performance

nikola-sh:
`_Get_any_status(__Path, __std_fs_stats_flags::_Attributes | __std_fs_stats_flags::_Reparse_tag)` calls to many WinAPI functions for regular files or directories. It calls `GetFileAttributesEx` + `CreateFile` + `GetFileInformationByHandleEx(FileBasicInfo)` but `GetFileAttributesEx` is enough to get all needed data if it is known that a file is not a reparse point.
This problem affects `symlink_status` performance.

Benchmark `symlink_status` results before changes:
```
---------------------------------------------------------
Benchmark               Time             CPU   Iterations
---------------------------------------------------------
symlink_status      15376 ns        15346 ns        44800
```
Benchmark `symlink_status` results after changes:
```
---------------------------------------------------------
Benchmark               Time             CPU   Iterations
---------------------------------------------------------
symlink_status       2219 ns         2148 ns       320000
```


stl/src/filesystem.cpp
```diff
@@ -899,6 +899,13 @@ namespace {
 
                 _Flags &= ~_Get_file_attributes_data;
             }
+
+            if (!_STD _Bitmask_includes_any(_Attributes, __std_fs_file_attr::_Reparse_point)
+                && _STD _Bitmask_includes_any(_Flags, __std_fs_stats_flags::_Reparse_tag)) {
+                // if reparse tag is requested by caller but the file is not a reparse point, set tag to _None
+                _Stats->_Reparse_point_tag = __std_fs_reparse_tag::_None;
+                _Flags &= ~__std_fs_stats_flags::_Reparse_tag;
+            }
```
AlexGuteniev, line 908: We have a check in this function below:
```C++
            // Calling GetFileInformationByHandleEx with FileAttributeTagInfo fails on FAT file system with
            // ERROR_INVALID_PARAMETER. We avoid calling this for non-reparse-points.
            if ((_Info.FileAttributes & FILE_ATTRIBUTE_REPARSE_POINT) != 0u) {
```

I think we should drop that check after checking it here, as that one will always succeed then.

benchmarks/src/filesystem.cpp
```diff
@@ -0,0 +1,19 @@
...
+#include <benchmark/benchmark.h>
+#include <filesystem>
+
+void symlink_status(benchmark::State& state) {
+    const auto path = std::filesystem::temp_directory_path();
+
+    for (auto _ : state) {
+        std::error_code ec;
+        const auto status = std::filesystem::symlink_status(path, ec);
+        benchmark::DoNotOptimize(status);
```
AlexGuteniev, line 13: Not that it would matter much, but I would call `DoNotOptimize` for both `status` and `ec`

stl/src/filesystem.cpp
nikola-sh: Actually, if `_Follow_symlinks` is true `_Info.FileAttributes` can be attributes for another object and can differ from `_Data.dwFileAttributes`. So it is ok to perform this check.

benchmarks/src/filesystem.cpp
nikola-sh: No problem, one more DoNotOptimize has been added.

benchmarks/src/filesystem.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include <benchmark/benchmark.h>
+#include <filesystem>
+
+void symlink_status(benchmark::State& state) {
+    const auto path = std::filesystem::temp_directory_path();
+
+    for (auto _ : state) {
+        std::error_code ec;
+        const auto status = std::filesystem::symlink_status(path, ec);
```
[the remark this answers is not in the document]
StephanTLavavej: I checked the full matrix of 4 results (without/with `DoNotOptimize(path)`, without/with product code changes) and this doesn't materially affect the results.

StephanTLavavej:
Thanks! On my system (5950X, 24H2, Samsung SSD 980 PRO), I observe much less of a performance improvement, but still major:  21126 ns => 10098 ns, for a speedup of 2.1x.

StephanTLavavej:

StephanTLavavej:
Thanks again, and congratulations on your first microsoft/STL commit! :heart_eyes_cat: :tada: :floppy_disk:

This change is expected to ship in VS 2022 17.13 Preview 3.

PR #5072, opened 2024-11-10
`<regex>`: Silence CodeQL false positive warning

StephanTLavavej:
Windows encountered internal OS-50456631 "\[CodeQL:Warning\]: SM02323: cpp/infiniteloop (in packages/\[...\]/regex)", reported to me by @DefaultRyan. According to the [internal doc](https://liquid.microsoft.com/Web/Object/Read/scanningtoolwarnings/Requirements/CodeQL.SM02323#Zguide), this is:

> \[SM02323\] Comparison of narrow type with wide type in loop condition
> Comparisons between types of different widths in a loop condition can cause the loop to fail to terminate.

In this case, the warning is a false positive, and my suppression comment explains why.

Alternatively, we could widen the `unsigned int` to `size_t`, either with a `static_cast` or a named variable, but that'd be unusual enough to still merit a comment, so I chose the less invasive approach.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -3567,6 +3567,7 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Match_pat(_Node_base* _Nx) { // c
             { // record current position
                 _Node_capture* _Node                 = static_cast<_Node_capture*>(_Nx);
                 _Tgt_state._Grps[_Node->_Idx]._Begin = _Tgt_state._Cur;
+                // CodeQL [SM02323] Comparing unchanging unsigned int _Node->_Idx to decreasing size_t _Idx is safe.
                 for (size_t _Idx = _Tgt_state._Grp_valid.size(); _Node->_Idx < _Idx;) {
                     _Tgt_state._Grp_valid[--_Idx] = false;
                 }


stl/inc/regex
```diff
@@ -3567,6 +3567,7 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Match_pat(_Node_base* _Nx) { // c
             { // record current position
                 _Node_capture* _Node                 = static_cast<_Node_capture*>(_Nx);
                 _Tgt_state._Grps[_Node->_Idx]._Begin = _Tgt_state._Cur;
+                // CodeQL [SM02323] Comparing unchanging unsigned int _Node->_Idx to decreasing size_t _Idx is safe.
```
CaseyCarter, line 3570: Would it be less noisy to `static_cast<size_t>(_Node->_Idx)` below and avoid the direct suppression?

CaseyCarter (approved):
This is "Approve with Suggestions".

stl/inc/regex
CaseyCarter: ... and I see you commented on this in the description that I didn't read. Fair enough.

StephanTLavavej:

PR #5073, opened 2024-11-10
Properly null-terminate output buffer in `basic_istream::get[line]`

### LINKED ISSUE #5070: `<istream>`: `basic_istream::get[line](char_type* s, std::streamsize n, char_type delim)` do not null-terminate the output buffer correctly
# Describe the bug
`basic_stringstream::get[line]()` do not handle null-termination correctly:

* If an output buffer of size 0 is passed in, a null character is still written into the buffer, resulting in a buffer overrun.
* In many cases, the output buffer is not null-terminated when an exception is raised.

The standard says (N4993 [istream.unformatted]/9 and 21):
> In any case, if n is greater than zero it then stores a null character into the next successive location of the array.

So the output buffer must always be null-terminated except if the buffer has size 0.

# Test case
```
#include <sstream>
#include <print>
#include <cassert>

using namespace std;

class throwing_buffer : public basic_streambuf<char> {
    virtual int_type underflow() override {
        throw 0;
    }
};

void test_get() {
    // output buffer of size 0 must not be null-terminated
    istringstream stream1("lorem");
    char buf1[] = "meow";
    stream1.get(buf1 + 1, 0, '\n');
    println("{}", buf1);

    // output buffers nust be null-terminated
    // even when exceptions are raised

    // exception raised due to eofbit being set
    istringstream stream2("o");
    char buf2[] = "woof";
    stream2.exceptions(ios_base::eofbit);
    try {
        stream2.get(buf2, sizeof(buf2), '\n');
    } catch (ios_base::failure&) {
    }
    println("{}", buf2);

    // exception raised in sentry constructor
    istringstream stream3("p");
    stream3.exceptions(ios_base::failbit);
    char buf3[] = "neigh";
    stream3.get(buf3, sizeof(buf3), '\n');
    char buf4[] = "bark";
    try {
        stream3.get(buf3 + 1, 1, '\n');
    } catch (ios_base::failure&) {
    }
    println("{}", buf4);

    // exception raised in streambuf
    // when extracting a character
    throwing_buffer strbuf;
    istream stream4(&strbuf);
    stream4.exceptions(ios_base::badbit);
    char buf5[] = "ribbit";
    try {
        stream4.get(buf5 + 1, 1, '\n');
    } catch (int) {
    }
    println("{}", buf5);
}

void test_getline() {
    // output buffer of size 0 must not be null-terminated
    istringstream stream1("lorem");
    char buf1[] = "meow";
    stream1.getline(buf1 + 1, 0, '\n');
    println("{}", buf1);

    // output buffers nust be null-terminated
    // even when exceptions are raised

    // exception raised due to eofbit being set
    istringstream stream2("o");
    char buf2[] = "woof";
    stream2.exceptions(ios_base::eofbit);
    try {
        stream2.getline(buf2, sizeof(buf2), '\n');
    } catch (ios_base::failure&) {
    }
    println("{}", buf2);

    // exception raised in sentry constructor
    istringstream stream3("p");
    stream3.exceptions(ios_base::failbit);
    char buf3[] = "neigh";
    stream3.get(buf3, sizeof(buf3), '\n');
    char buf4[] = "bark";
    try {
        stream3.getline(buf3 + 1, 1, '\n');
    } catch (ios_base::failure&) {
    }
    println("{}", buf4);

    // exception raised in streambuf
    // when extracting a character
    throwing_buffer strbuf;
    istream stream4(&strbuf);
    stream4.exceptions(ios_base::badbit);
    char buf5[] = "ribbit";
    try {
        stream4.getline(buf5 + 1, 1, '\n');
    } catch (int) {
    }
    println("{}", buf5);
}

int main() {
    test_get();
    println("----");
    test_getline();
    return 0;
}
```
https://godbolt.org/z/zfKqnePGh

# Expected behavior
The program should print
```
meow
o
b
r
----
meow
o
b
r
```

# Actual behavior
The program actually prints
```
m
ooof
bark
ribbit
----
m
o
bark
ribbit
```

# Additional remarks
I discovered these bugs while analyzing why the libcxx test std/input.output/iostream.format/input.streams/istream.unformatted/get_pointer_size.pass.cpp fails. (An assert fails because `get()` does not null-terminate when an exception is raised on EOF.)

libc++ and libstdc++ do not null-terminate as well when a streambuf member function or the `sentry` constructor throw. However, the standard mandates that the output buffer must be null-terminated "in any case". Moreover, there is additional language in [istream.unformatted]/1 that the output buffer must be null-terminated when the `sentry` constructor throws:

> Otherwise, if the sentry constructor exits by throwing an exception or if the sentry object produces false, when converted to a value of type bool, the function returns without attempting to obtain any input. In either case the number of extracted characters is set to 0; unformatted input functions taking a character array of nonzero size as an argument shall also store a null character (using charT()) in the first location of the array.

muellerj2:
Fixes #5070.

The libcxx tests istream.unformatted/get_pointer_size.pass.cpp and get_pointer_size_chart.pass.cpp pass now.

Since this issue is mostly about correct null-termination in the presence of exceptions, I added the tests to GH_001858_iostream_exception. They should cover the relevant cases:

* output buffer of size zero with and without exception on `failbit`
* `sentry` failure with and without exception on `failbit`
* exception during `streambuf` character extraction with and without `badbit` exception propagation
* encountering EOF with and without exception on `eofbit`

### DIFF
--- a/tests/std/tests/GH_001858_iostream_exception/test.cpp
+++ b/tests/std/tests/GH_001858_iostream_exception/test.cpp
@@ -323,6 +323,7 @@ void test_gh5070_istream_get_null_termination_under_exceptions() {
         assert(!is.bad());
         try {
             is.getline(buf.data(), 0);
+            assert(false);
         } catch (const ios_base::failure&) {
             // Expected case
         }


muellerj2:

tests/std/tests/GH_001858_iostream_exception/test.cpp
```diff
@@ -148,6 +149,254 @@ void test_istream_exceptions() {
...
+constexpr array<wchar_t, 5> meow_array<wchar_t> = {L"meow"};
+
+// testing GH-5070: basic_istream::get[line](char_type* s, std::streamsize n, char_type delim)
+// do not null-terminate the output buffer correctly
+template <class CharT>
+void test_gh5070_istream_get_nulltermination_under_exceptions() {
+    throwing_buffer<CharT> buffer;
+    const basic_string<CharT> stream_content(1U, meow_array<CharT>[2]);
+
+    { // get, exception during input extraction
```
StephanTLavavej, line 166: The `get` and `getline` comments are highly systematic, which is great, except that there's unnecessary variation in the first two, where `getline` is a bit more detailed:
```cpp
// getline, exception during input extraction, no exception rethrow
// getline, exception during input extraction, exception rethrow enabled
```

tests/std/tests/GH_001858_iostream_exception/test.cpp
```diff
@@ -148,6 +149,254 @@ void test_istream_exceptions() {
...
+    { // getline, empty output buffer, exception raised on failbit
+        basic_istream<CharT> is(buffer.to_buf());
+        auto buf = meow_array<CharT>;
+        is.exceptions(ios_base::failbit);
+        assert(!is.bad());
+        try {
+            is.getline(buf.data(), 0);
+        } catch (const ios_base::failure&) {
+            // Expected case
+        }
```
StephanTLavavej, line 327: Diffing `get` versus `getline` reveals that there's an `assert(false)` missing from the no-exception-thrown codepath here.

StephanTLavavej (approved):

StephanTLavavej:
Thanks, this is fantastic - and incredible for a first PR! :heart_eyes_cat: I've pushed several minor nitpicks, a product code simplification, and a test code improvement. I appreciate the extremely detailed test coverage.

We merge PRs simultaneously to our GitHub and MSVC-internal repos, mirrored in a semi-manual process and batched up to save time. Your PR will be part of the next batch, very likely this week.

StephanTLavavej:

StephanTLavavej:
Thanks again, and congratulations on your first microsoft/STL commit! :rocket: :smile_cat: :partying_face:

This change is expected to ship in VS 2022 17.13 Preview 3.

PR #5092, opened 2024-11-16
Vectorize `unique`

AlexGuteniev:
Not really unique, modelled on #4987

# ⏬ Double load

To compare adjacent values, the same memory is loaded twice with an element shift. 

It is possible to reuse the previous vector part, and mix it with the current, to save one load, but have some extra instructions to mix values, and a loop-carried dependency. On SSE path it is possible with `_mm_alignr_epi8` (except for 8-bit elements). For AVX it would be way more complex due to AVX lanes.

Benchmarking shows that double load is faster than any reuse attempt. To some extent such a result overlaps with #4958

# ⏱️ Benchmark results

Benchmark                             |      main  |   this     
--------------------------------------|------------|------------
u<alg_type::std_fn, std::uint8_t>     |   1166 ns  | 190 ns     
u<alg_type::std_fn, std::uint16_t>    |   1222 ns  | 247 ns     
u<alg_type::std_fn, std::uint32_t>    |   1555 ns  | 310 ns     
u<alg_type::std_fn, std::uint64_t>    |   1470 ns  | 665 ns     
u<alg_type::rng, std::uint8_t>        |   1230 ns  | 187 ns     
u<alg_type::rng, std::uint16_t>       |   1204 ns  | 233 ns     
u<alg_type::rng, std::uint32_t>       |   1268 ns  | 308 ns     
u<alg_type::rng, std::uint64_t>       |   1505 ns  | 665 ns


stl/inc/algorithm
```diff
@@ -207,6 +212,27 @@ __declspec(noalias) void _Replace_vectorized(
...
 }
 
+template <class _Ty>
+_Ty* _Unique_vectorized(_Ty* const _First, _Ty* const _Last) noexcept {
+    if constexpr (is_pointer_v<_Ty>) {
+#ifdef _WIN64
+        return reinterpret_cast<_Ty*>(::__std_unique_8(_First, _Last));
+#else
+        return reinterpret_cast<_Ty*>(::__std_unique_4(_First, _Last));
+#endif
```
StephanTLavavej, line 222: Unlike other vectorized algorithms, it looks like these cases aren't necessary. I'd recommend dropping them, and ensuring that we have test coverage of unique-ifying pointer elements (it doesn't need to be exhaustive, just enough to verify that it works, then the normal test coverage for 4 and 8-byte integers will suffice).

stl/inc/algorithm
```diff
@@ -221,6 +247,17 @@ template <class _Iter, class _Ty1, class _Ty2>
...
+
+// Can we activate the vector algorithms for unique?
+template <class _Iter, class _Pr>
+constexpr bool _Vector_alg_in_unique_is_safe =
+    _Iterator_is_contiguous<_Iter> && !_Iterator_is_volatile<_Iter> // Contiguous nonvolatile iterator
+    && _Is_any_of_v<_Pr,
+#if _HAS_CXX20
+        _RANGES equal_to,
+#endif // _HAS_CXX20
+        _STD equal_to<>> // default comparison
```
StephanTLavavej, line 259: This rejects `equal_to<T>`.

Would `_Can_memcmp_elements_with_pred` be applicable here? It feels like we've already solved this problem.

AlexGuteniev: That seems using too generic tool for simpler problem (we have only one iterator type here), but still looks like a good idea.

CaseyCarter:
> Not really unique, modelled on #4987

🤦

StephanTLavavej (changes requested):

StephanTLavavej:
Thanks! (And sorry for taking a while to get to this. Partial review: still need to finish reviewing `test.cpp` and `vector_algorithms.cpp`.)

I pushed a conflict-free merge with `main` followed by tiny nitpicks, but then noticed a bug in the test.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -696,6 +697,81 @@ void test_remove(mt19937_64& gen) {
...
+void test_unique(mt19937_64& gen) {
+    shared_ptr<void> tmp_array;
+
+    constexpr int number_of_values = 5;
+
+    using TD = conditional_t<sizeof(T) == 1 || is_pointer_v<T>, int, T>;
+    binomial_distribution<TD> dis(number_of_values);
+
+    if constexpr (is_pointer_v<T>) {
+        tmp_array = std::make_unique<remove_pointer<T>[]>(5);
```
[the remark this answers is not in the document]
AlexGuteniev: Done

StephanTLavavej (approved):

StephanTLavavej:
The speedups are less significant on my 5950X, but good across the board with no regressions:

Benchmark                            |  Before |   After | Speedup
-------------------------------------|--------:|--------:|--------
`u<alg_type::std_fn, std::uint8_t>`  | 1681 ns |  494 ns | 3.40
`u<alg_type::std_fn, std::uint16_t>` | 1566 ns |  503 ns | 3.11
`u<alg_type::std_fn, std::uint32_t>` | 1319 ns |  657 ns | 2.01
`u<alg_type::std_fn, std::uint64_t>` | 1466 ns | 1138 ns | 1.29
`u<alg_type::rng, std::uint8_t>`     | 1261 ns |  497 ns | 2.54
`u<alg_type::rng, std::uint16_t>`    | 1691 ns |  584 ns | 2.90
`u<alg_type::rng, std::uint32_t>`    | 1765 ns |  601 ns | 2.94
`u<alg_type::rng, std::uint64_t>`    | 1403 ns | 1139 ns | 1.23

StephanTLavavej (approved):

AlexGuteniev:
I've discovered that something is missing. 
But I think it can wait for a follow up PR.

----

Unlike `remove` algorithm, this one doesn't have the search for the first duplicate before the main vectorization loop. The scalar implementations in headers have that part, the vectorized one currently doesn't.

For performance it is clearly a missed opportunity. Though the vectorization improvement should be bigger than the negative effect of extra writes.

For correctness, I'm not sure. [[algorithms.requirements]](https://eel.is/c++draft/algorithms.requirements#3)/3 says:
> For purposes of determining the existence of data races, algorithms shall not modify objects referenced through an iterator argument unless the specification requires such modification[.](https://eel.is/c++draft/algorithms.requirements#3.sentence-1)

However as the writes write equal integer values, this is not observable for concurrent writes, even if container violates alignment requirements and the write is not atomic.

The only thing where extra writes can be observable is running this algorithm on a read-only data without adjacent duplicates. But this is a very silly use case.

----

It is easily fixable with `adjacent_find`, vectorized in another PR here.

StephanTLavavej:
Thanks for the one-of-a-kind PR! :joy_cat: :rocket: :tada:

PR #5097, opened 2024-11-18
Don't use `__restrict__` for CUDA

StephanTLavavej:
#5079 was a nice try, but it doesn't work. We need to revert this before 17.13 Preview 3 branches for release.

It broke NVIDIA's Cutlass again. @CaseyCarter observed:

> In both the old and new preprocessed repros, our `_RESTRICT`s in the declaration of `_Adjacent_difference_no_overlap` are being expanded to plain `restrict`. If I had to guess, I'd say that CUDA is replacing `__restrict__` with `restrict` under the expectation that `__restrict__` is only used in C mode.

Fixes VSO-2295101 / AB#2295101 again.


CaseyCarter (approved):

StephanTLavavej:

PR #5098, opened 2024-11-19
Maybe Clang _does_ work with `ALTERNATENAME`

CaseyCarter:
Enable our `ALTERNATENAME` machinery when the compiler is Clang. Add Clang non-ASan coverage to `GH_002030_asan_annotate_{string,vector}`, which lets us simplify their matrices by extracting the ASan/not-ASan distinction into a separate `RUNALL_CROSSLIST`.

Let's hold this change for early in the 17.14 cycle instead of merging it now, super late in 17.13. I'd like to do some extra internal validation before merging as well:
* [x] STL-ASan-CI: https://dev.azure.com/vclibs/STL/_build/results?buildId=17968&view=results
* [ ] stl x86
* [ ] stl x64
* [ ] stl arm64? (IIRC we need changes to enable Clang ARM64 run.pl coverage.)


CaseyCarter:
Let's hold this change for early in the 17.14 cycle instead of merging it now, super late in 17.13.

StephanTLavavej:
I'm not eager to mess with this, but I suppose if you break it, you buy it. :joy_cat:

tests/std/tests/GH_002030_asan_annotate_string/env.lst
```diff
@@ -8,50 +8,30 @@ RUNALL_INCLUDE ..\prefix.lst
...
-PM_CL="-fsanitize=address /EHsc /MTd /std:c++latest /permissive- /fp:strict"
-PM_CL="-fsanitize=address /EHsc /MTd /std:c++latest /permissive-"
-PM_CL="/D_ANNOTATE_STRING /BE /c /EHsc /MD /std:c++14"
-PM_CL="/D_ANNOTATE_STRING /BE /c /EHsc /MDd /std:c++17 /permissive-"
-PM_CL="/D_ANNOTATE_STRING /BE /c /EHsc /MT /std:c++20 /permissive-"
-PM_CL="/D_ANNOTATE_STRING /BE /c /EHsc /MTd /std:c++latest /permissive-"
-PM_CL="/D_ANNOTATE_STRING /EHsc /MD /std:c++14"
-PM_CL="/D_ANNOTATE_STRING /EHsc /MD /std:c++14 /Zc:char8_t"
-PM_CL="/D_ANNOTATE_STRING /EHsc /MD /std:c++17"
-PM_CL="/D_ANNOTATE_STRING /EHsc /MD /std:c++17 /Zc:char8_t"
```
StephanTLavavej, line 38: You are losing these 14/17 with `/Zc:char8_t` configurations, due to pre-existing divergence between `-fsanitize=address` and `/D_ANNOTATE_STRING`. Intentional?

StephanTLavavej (changes requested):

tests/std/tests/GH_002030_asan_annotate_string/env.lst
CaseyCarter: Not intentional. I should have known better than to trust the description.

StephanTLavavej (approved):

StephanTLavavej:
"I have many names." :eye: :fire: :smiling_imp:

PR #5099, opened 2024-11-19
More specific assertion for unlocking mutex not owned by the current thread

AlexGuteniev:
Inspired by https://github.com/microsoft/STL/issues/4418#issuecomment-2484715880

### DIFF
--- a/stl/src/mutex.cpp
+++ b/stl/src/mutex.cpp
@@ -147,8 +147,9 @@ static _Thrd_result mtx_do_lock(_Mtx_t mtx, const _timespec64* target) noexcept 
 }
 
 _CRTIMP2_PURE _Thrd_result __cdecl _Mtx_unlock(_Mtx_t mtx) noexcept { // unlock mutex
+    _THREAD_ASSERT(1 <= mtx->_Count, "unlock of unowned mutex");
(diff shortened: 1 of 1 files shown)


stl/src/mutex.cpp
```diff
@@ -147,8 +147,9 @@ static _Thrd_result mtx_do_lock(_Mtx_t mtx, const _timespec64* target) noexcept
 }
 
 _CRTIMP2_PURE _Thrd_result __cdecl _Mtx_unlock(_Mtx_t mtx) noexcept { // unlock mutex
+    _THREAD_ASSERT(1 <= mtx->_Count, "unlock of unowned mutex");
```
AlexGuteniev, line 150: I'd prefer having `mtx->_Count > 0` here and in another debug check in `_Mtx_destroy_in_situ`.

`mtx->_Count == 0` is also a straightforward form, but `<` also checks the invariant of non-negative count for free.

CaseyCarter (approved):

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for improving this diagnostic! :lock: :key: :unlock:

PR #3260, opened 2022-12-03
Fallback to language id 0 with `FormatMessageA` if the error messages are not available in system locale

cppdev123:
After #2669, `std::system_category` used the locale system language with `FormatMessageA` to get the error message which fixed a problem and introduced another one. As reported in #3254 the error messages may not be available in the system locale, but they are available in en-US locale shipped with windows so `FormatMessageA` failed with `ERROR_MUI_FILE_NOT_FOUND` and `std::system_category` returned hard coded "unknown error". It was working before because if the language id is 0 it will eventually fall back to en-US locale and return an error message in that locale.

I reproduced the problem on following windows installations:
- Windows 10 and Windows 11 with locale set to ar-EG and the Arabic langauge pack installed but `FormatMessageA` failed with `ERROR_MUI_FILE_NOT_FOUND`. Seems that there are no error messages in the Arabic mui because windows apps, explorer and login screen can show properly in arabic and right to left layout
- Windows 11 installed with locale en-GB set during installation by mistake then I removed the English UK keyboard and added English US and left the system locale set to en-GB. I was getting "unknown error" until I changed the locale back to "en-US"

This should fix https://github.com/microsoft/STL/issues/3254

PR #2669, opened 2022-04-24
<system_error>: explicitly pass the system locale ID for system_category messages

fsb4000:
Fixes #2451 

```c++
type main.cpp
#include <clocale>
#include <iostream>
#include <string>

#include <Windows.h>

int main() {
    setlocale(LC_ALL, "");
    auto file_handle(CreateFile(
        L"D:\\non_exist_file.txt", GENERIC_READ, FILE_SHARE_READ, NULL, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL));
    if (file_handle == INVALID_HANDLE_VALUE) {
        const auto error_code = GetLastError();
        std::string errmsg    = std::system_category().message(error_code); // Sometimes the string only contains "???"
        std::cout << errmsg << '\n';
    }
}
```

If we have different “user language id” and “system language id” (this is how we can reach it: https://github.com/microsoft/STL/issues/2451#issuecomment-1107744166) FormatMessageA can choose “user language id” instead of “system language id” if we pass 0.

Before
```
>main
??????? ?? ??????? ????? ????????? ????.
```
After
```
>main
The system cannot find the path specified.
```

Issue #2451, opened 2022-01-01
<system_error>: FormatMessageA needs be used in a different way


snnn:
**Describe the bug**

Windows settings have “user language id” and “system language id”.  They can be different.  std::system_category().message() returns "???" when they are different.

By default FormatMessage returns a string in “user language”. The Unicode version of FormatMessage works good in all cases. But the ANSI version, like all other ANSI version Windows APIs, would try to convert the result string to a multibyte string based on “system language id”. It goes wrong when the two ids are different. For example, If your user language is Chinese but the system language is English,  it can only returns “????”, because you can’t encode a Chinese string in ISO-8859-1. 

So if you need to return the error message in std::string type, there are two solutions:

1.	Pass the system language id to FormatMessageA
2.	If you use FormatMessageW, you can set the language id to the system language or thread ACP(the user language). Then when you use WideCharToMultiByte, use CP_ACP for the system language and CP_THREAD_CP for the user language.

**Command-line test case**
```c++

int main() {
    setlocale(LC_ALL, "");
	LANGID user_language_id = GetUserDefaultLangID();
	WCHAR buf[1024];
	int ret = GetUserDefaultLocaleName(buf,sizeof(buf)/sizeof(buf[0]));
	assert(ret != 0);
	std::wcout << user_language_id << L" " << buf << "\n";
	ret = GetSystemDefaultLocaleName(buf, sizeof(buf) / sizeof(buf[0]));
	assert(ret != 0);
	LANGID system_language_id = GetSystemDefaultLangID();
	std::wcout << system_language_id << L" " << buf << std::endl;
    
    wil::unique_hfile file_handle(CreateFile(L"D:\\non_exist_file.txt", GENERIC_READ, FILE_SHARE_READ, NULL, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL));
    if (file_handle.get() == INVALID_HANDLE_VALUE) {
        const auto error_code = GetLastError();
        std::string errmsg = std::system_category().message(error_code); //Sometimes the string only contains "???"
    }
  return 0;
}
```
To compile it:
C:\Temp>cl /EHsc .\repro.cpp

To reproduce the error:
I assume you have a clean Windows installation. Then please create a new user without Administrator privilege. Then login to the user, go to settings and change the display language to something different. Like, Chinese.  In such a case, you only changed the user's locale, not the system. Because the user doesn't have the privilege to change system level settings. 

[...]

PR #5104, opened 2024-11-20
Prefer US English for `system_category()` messages, fall back to system locale, then ID 0

### LINKED ISSUE #3254: Visual Studio 2022 `std::system_category` returns "unknown error" if system locale is not en-US
I download visual studio 2022 17.4.1 and was surprised to see that "unknown error" is returned from `std::error_code` using `std::system_categoey()` as the error category while it was reporting the correct error message before.
After searching on the internet, I found someone complaining about the same issue here : ["unknown error" from std::error_code on Windows](https://stackoverflow.com/questions/73584099/unknown-error-from-stderror-code-on-windows)
The solution was to set the system locale to en-US from "administrative language settings".

I tested it on two windows installations:
Windows 11 with locale set to "en-GB" and worked well when changed to "en-US"
Windows 10 with locale set to "ar-EG" and worked well when changed to "en-US"

I have also visual studio 2019 installed and the binary produced by it works as expected without needing to change the system locale

The issue seems to have appeared in `__std_system_error_allocate_message` in `syserror_import_lib.cpp` after this commit [<system_error>: explicitly pass the system locale ID for system_categ…](https://github.com/microsoft/STL/commit/fa51112a5aa66137cc1aa24b3b35131608cd027d)

To mention also: both windows installs have "English (United States)" as the "Windows display language" but had system locale different from "en-US"

EDIT: I set both display language and system locale to "ar-EG" but still "unknown error" is returned. Only when system locale is set to en-US the correct message is returned disregarding the display language

Also tracked by internal VSO-1735227 / AB#1735227 .

StephanTLavavej:
Fixes #3254, a regression introduced by #2669 in VS 2022 17.3.

# :heart_eyes_cat: Thanks

Thanks to:

* @cppdev123, who first reported the regression and whose PR #3260 inspired this alternative approach,
* @fsb4000 for noting an issue with that PR (in https://github.com/microsoft/STL/pull/3260#issuecomment-1336299026),
* @sylveon for suggesting the US English approach (in https://github.com/microsoft/STL/pull/3260#issuecomment-1336337294),
* @davemcincork for providing an extra-simple repro (in https://github.com/microsoft/STL/issues/3254#issuecomment-1991532148),
* @muellerj2 and @TheStormN for checking single-language OS installations,
* and everyone else who commented on these issues and PRs.

# :scroll: Overview

The original (pre-#2669) strategy of using language ID 0 to request [`FormatMessageA`'s multi-step behavior](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-formatmessagea#parameters) was problematic because it tries the user's locale before the system locale, but the user's locale might not result in an intelligible `char*`, so we could get strings filled with question marks.

The status quo (post-#2669) strategy of using the system locale was problematic because the system locale might not have error message strings. For example, this is the case when the system locale is English (United Kingdom).

We also have to worry about single-language installations of Windows, which don't have US English error message strings.

After considering all of these scenarios and talking to contributors, I believe that the best strategy is to attempt 3 things in this order:

1. Prefer US English.
   * This should succeed on multi-language installations of Windows, which is the common case.
   * This is consistent with `generic_category()`, as my new comment explains.
   * Instead of dynamically querying for what "en-US" is, or using the (hilariously mega-deprecated in `winnt.h`) macro `MAKELANGID` to construct it, I'm just hardcoding the language ID.
2. Fall back to the system locale, if we can obtain its language ID.
   * This is what we're currently doing in `main` (i.e. what #2669 did).
   * VS 2022 17.3 shipped in Aug 2022, so we have two years of experience with this behavior.
   * If US English isn't available, we're presumably dealing with a single-language OS, so our best bet is the system locale.
3. As an ultimate fallback, if we can't get the system locale or it doesn't have an error message string, then use language ID 0.
   * This is what `main` does if we can't get the system locale, and what we always did before #2669, so we have many years of experience with this codepath.
   * If this doesn't result in something intelligible (e.g. US English isn't available, the system locale doesn't have an error message string for us, and language ID 0 behavior picks up the user's Windows display language, but its encoding is radically different from what the programmer-user expects for a `char*`), well, we tried. At this time, I don't want to explore novel logic like calling `FormatMessageW` and converting to UTF-8. If this scenario turns out to occur in practice, we can explore changes in the future.

# :computer: Test Code
```
C:\temp>type meow.cpp
```
```cpp
#include <limits>
#include <print>
#include <string>
#include <system_error>
#include <Windows.h>
using namespace std;

void print_system_locale_name() {
    wchar_t buf[LOCALE_NAME_MAX_LENGTH]{};
    if (GetLocaleInfoEx(
        LOCALE_NAME_SYSTEM_DEFAULT,
        LOCALE_SNAME,
        buf,
        LOCALE_NAME_MAX_LENGTH
    ) == 0) {
        println("GetLocaleInfoEx failed!");
    }
    string narrow;
    for (const auto& w : buf) {
        if (w == L'\0') {
            break;
        } else if (w > (numeric_limits<char>::max)()) {
            narrow.push_back('#');
        } else {
            narrow.push_back(static_cast<char>(w));
        }
    }
    println(R"(  System locale name: "{}")", narrow);
}

void print_system_language_id() {
    DWORD lang_id = 0;
    if (GetLocaleInfoEx(
        LOCALE_NAME_SYSTEM_DEFAULT,
        LOCALE_ILANGUAGE | LOCALE_RETURN_NUMBER,
        reinterpret_cast<LPWSTR>(&lang_id),
        sizeof(lang_id) / sizeof(wchar_t)
    ) == 0) {
        println("GetLocaleInfoEx failed!");
    }
    println("  System language ID: {0} decimal, 0x{0:04x} hex", lang_id);
}

int main() {
    print_system_locale_name();
    print_system_language_id();

    // Repro derived from @davemcincork's example.
    println(R"(ERROR_FILE_NOT_FOUND: "{}")", system_category().message(2));
}
```

# :scientist: Initial Testing
On a Dev Box (Win11 24H2, en-US), I left the "Windows display language" unchanged:

Settings > Time & language > Language & region > Windows display language > English (United States)

But I changed the system locale to en-GB and rebooted:

Settings > Time & language > Language & region > Administrative language settings > Change system locale... > English (United Kingdom)

Before this PR:

```
C:\temp>cl /EHsc /nologo /W4 /std:c++latest /MTd /Od meow.cpp && meow
meow.cpp
  System locale name: "en-GB"
  System language ID: 2057 decimal, 0x0809 hex
ERROR_FILE_NOT_FOUND: "unknown error"
```

After this PR:

```
C:\temp>cl /EHsc /nologo /W4 /std:c++latest /MTd /Od meow.cpp && meow
meow.cpp
  System locale name: "en-GB"
  System language ID: 2057 decimal, 0x0809 hex
ERROR_FILE_NOT_FOUND: "The system cannot find the file specified."
```

Retested with the 3-step loop successfully.

## :test_tube: More Testing

Then, I verified that this PR isn't causing #2451 to reappear (this was the original bug that #2669 attempted to fix).

Following https://github.com/microsoft/STL/issues/2451#issuecomment-1107744166, I set my Dev Box's Windows display language to Chinese (installing all of the language pack stuff) and the system locale back to en-US. Then, after slightly altering the repro to use `format()` instead of `print()`, I verified that the side-by-side install of VS 2019 16.11 reproed the original bug, VS 2022 17.13 Preview 2 (internal) worked with #2669's changes (the status quo of `main`), and this PR worked as well.

Retested with the 3-step loop successfully.

## :one: Single-Language OS Testing

In https://github.com/microsoft/STL/pull/5104#pullrequestreview-2452756188, @muellerj2 verified that

* de-DE single-language system with de-DE system locale
* de-DE single-language system with de-CH system locale

succeed with German strings. These are exercising the system locale and ID 0 fallbacks, respectively.

### DIFF
--- a/stl/src/syserror_import_lib.cpp
+++ b/stl/src/syserror_import_lib.cpp
@@ -41,6 +41,12 @@ extern "C" {
     // __std_system_error_deallocate_message should be called even if 0 is returned
     // pre: *_Ptr_str == nullptr
 
+    // We start by requesting US English for system_category() messages. (See GH-2451 and GH-3254 for the history.)
+    // This is consistent with generic_category(), which uses a table of US English strings in the STL.
+    // In general, system_error messages aren't directly useful to end-users - they're meant for programmer-users.
+    // Of course, the programmer-user might not speak US English, but machine translation of the message
+    // (and the numeric value of the error code) should help them understand the error.
+
     constexpr auto _Flags = FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS;
 
     DWORD _Lang_id = 0x0409; // 1033 decimal, "en-US" locale
@@ -48,11 +54,15 @@ extern "C" {
     auto _Chars = FormatMessageA(_Flags, nullptr, _Message_id, _Lang_id, reinterpret_cast<char*>(_Ptr_str), 0, nullptr);
 
     if (_Chars == 0) {
+        // If FormatMessageA() failed for any reason, cleanup and restore this function's precondition.
         LocalFree(*_Ptr_str);
         *_Ptr_str = nullptr;
 
+        // Attempt to get the system locale's language ID.
         const int _Ret = GetLocaleInfoEx(LOCALE_NAME_SYSTEM_DEFAULT, LOCALE_ILANGUAGE | LOCALE_RETURN_NUMBER,
             reinterpret_cast<LPWSTR>(&_Lang_id), sizeof(_Lang_id) / sizeof(wchar_t));
+
+        // If we can't get the system locale, the final fallback is FormatMessageA()'s behavior for language ID 0.
         if (_Ret == 0) {
             _Lang_id = 0;
         }
--- a/stl/src/syserror_import_lib.cpp
+++ b/stl/src/syserror_import_lib.cpp
@@ -47,6 +47,19 @@ extern "C" {
 
     auto _Chars = FormatMessageA(_Flags, nullptr, _Message_id, _Lang_id, reinterpret_cast<char*>(_Ptr_str), 0, nullptr);
 
+    if (_Chars == 0) {
+        LocalFree(*_Ptr_str);
+        *_Ptr_str = nullptr;
+
+        const int _Ret = GetLocaleInfoEx(LOCALE_NAME_SYSTEM_DEFAULT, LOCALE_ILANGUAGE | LOCALE_RETURN_NUMBER,
+            reinterpret_cast<LPWSTR>(&_Lang_id), sizeof(_Lang_id) / sizeof(wchar_t));
+        if (_Ret == 0) {
+            _Lang_id = 0;
+        }
+
+        _Chars = FormatMessageA(_Flags, nullptr, _Message_id, _Lang_id, reinterpret_cast<char*>(_Ptr_str), 0, nullptr);
+    }
+
     return _CSTD __std_get_string_size_without_trailing_whitespace(*_Ptr_str, _Chars);
 }
 
--- a/stl/src/syserror_import_lib.cpp
+++ b/stl/src/syserror_import_lib.cpp
@@ -56,7 +56,7 @@ extern "C" {
     constexpr auto _Flags = FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS;
     constexpr auto _Lang_id = 0x0409; // 1033 decimal, "en-US" locale
 
-    const unsigned long _Chars =
+    const auto _Chars =
         FormatMessageA(_Flags, nullptr, _Message_id, _Lang_id, reinterpret_cast<char*>(_Ptr_str), 0, nullptr);
 
     return _CSTD __std_get_string_size_without_trailing_whitespace(*_Ptr_str, _Chars);
--- a/stl/src/syserror_import_lib.cpp
+++ b/stl/src/syserror_import_lib.cpp
@@ -40,8 +40,22 @@ extern "C" {
     // convert to name of Windows error, return 0 for failure, otherwise return number of chars in buffer
     // __std_system_error_deallocate_message should be called even if 0 is returned
     // pre: *_Ptr_str == nullptr
+
+    // We always request US English for system_category() messages.
+    // This is consistent with generic_category(), which uses a table of US English strings in the STL.
+    // See GH-2451 and GH-3254 for the history here - we previously tried to localize system_category() messages,
+    // but attempting to use FormatMessageA's behavior for language ID 0 and attempting to use the system locale
+    // had various failure scenarios.
+    // Using US English (which is FormatMessageA's final fallback for language ID 0)
+    // is likely to succeed with whatever the end-user's system configuration is.
+    // In general, system_error messages aren't directly useful to end-users - they're meant for programmer-users.
+    // Of course, the programmer-user might not speak US English, but machine translation of the message
+    // (and the numeric value of the error code) should help them understand the error.
+    // The previous failure scenarios of "unknown error" or a string of question marks were completely unhelpful.
+
     constexpr auto _Flags = FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS;
     constexpr auto _Lang_id = 0x0409; // 1033 decimal, "en-US" locale
+
     const unsigned long _Chars =
         FormatMessageA(_Flags, nullptr, _Message_id, _Lang_id, reinterpret_cast<char*>(_Ptr_str), 0, nullptr);
 
--- a/stl/src/syserror_import_lib.cpp
+++ b/stl/src/syserror_import_lib.cpp
@@ -46,9 +46,9 @@ extern "C" {
     if (_Ret == 0) {
         _Lang_id = 0;
     }
+    constexpr auto _Flags = FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS;
     const unsigned long _Chars =
-        FormatMessageA(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
-            nullptr, _Message_id, _Lang_id, reinterpret_cast<char*>(_Ptr_str), 0, nullptr);
+        FormatMessageA(_Flags, nullptr, _Message_id, _Lang_id, reinterpret_cast<char*>(_Ptr_str), 0, nullptr);
 
     return _CSTD __std_get_string_size_without_trailing_whitespace(*_Ptr_str, _Chars);
 }
--- a/stl/src/syserror_import_lib.cpp
+++ b/stl/src/syserror_import_lib.cpp
@@ -40,13 +40,8 @@ extern "C" {
     // convert to name of Windows error, return 0 for failure, otherwise return number of chars in buffer
     // __std_system_error_deallocate_message should be called even if 0 is returned
     // pre: *_Ptr_str == nullptr
-    DWORD _Lang_id;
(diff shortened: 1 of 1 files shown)


TheStormN:
I think you missed one case in your verifications steps. Installations which do not have English at all.

muellerj2:
de-DE-only install:

```
  System locale name: "de-DE"
  System language ID: 1031 decimal, 0x0407 hex
ERROR_FILE_NOT_FOUND: "unknown error"
```

StephanTLavavej:
Yes - I'll follow @cppdev123's fallback technique (with different code) and credit them as a co-author.

jovibor:
I believe the steps should be (pseudo code):
```cpp
FormatMessageA(SYSTEM_LANG_ID); //Not CURRENT_USER_LANG_ID
if(_Last_error == ERROR_MUI_FILE_NOT_FOUND || _Last_error == ERROR_RESOURCE_LANG_NOT_FOUND) {
    FormatMessageA(ENGLISH_HARDCODED_LANG_ID);
    if(_Last_error == ERROR_MUI_FILE_NOT_FOUND || _Last_error == ERROR_RESOURCE_LANG_NOT_FOUND) {
        FormatMessageA(ZERO_LANG_ID); //The last resort try.
    }
}
```

StephanTLavavej:
@jovibor Preferring the system locale wouldn't necessarily be wrong, but I think that the "`generic_category()` is always US English" argument is a reasonable justification for preferring US English.

I felt it was simpler and more robust to activate fallbacks for any failures, instead of checking specific error codes.

As I mentioned in my updated PR description (that I was typing while you commented), having a third call to ultimate-fallback to 0, even when the system locale can be retrieved, could potentially be useful, but I'd want to see what kind of exotic configuration doesn't have US English and doesn't have an error message string in the system locale, but *does* have something useful returned by language ID 0.

I am always willing to be convinced by an explicit example of a system configuration where this PR results in question marks or "unknown error", but another strategy succeeds without regressing behavior anywhere else!

stl/src/syserror_import_lib.cpp
```diff
@@ -40,15 +40,35 @@ extern "C" {
...
+
+    constexpr auto _Flags = FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS;
+
+    DWORD _Lang_id = 0x0409; // 1033 decimal, "en-US" locale
+
+    auto _Chars = FormatMessageA(_Flags, nullptr, _Message_id, _Lang_id, reinterpret_cast<char*>(_Ptr_str), 0, nullptr);
+
+    if (_Chars == 0) {
+        // If FormatMessageA() failed for any reason, cleanup and restore this function's precondition.
+        LocalFree(*_Ptr_str);
```
sylveon, line 58: I'd be surprised if FormatMessage has a failure mode where it requires the user to clean up allocated memory.

stl/src/syserror_import_lib.cpp
```diff
@@ -40,15 +40,35 @@ extern "C" {
...
+        *_Ptr_str = nullptr;
+
+        // Attempt to get the system locale's language ID.
+        const int _Ret = GetLocaleInfoEx(LOCALE_NAME_SYSTEM_DEFAULT, LOCALE_ILANGUAGE | LOCALE_RETURN_NUMBER,
+            reinterpret_cast<LPWSTR>(&_Lang_id), sizeof(_Lang_id) / sizeof(wchar_t));
+
+        // If we can't get the system locale, the final fallback is FormatMessageA()'s behavior for language ID 0.
+        if (_Ret == 0) {
+            _Lang_id = 0;
```
sylveon, line 67: I don't believe this handles the case where we can get the system locale, but calling FormatMessageA would fail.

sylveon: For example, what happens in a system with a `fr-CA` system locale, but `fr-FR` Windows display language? Is this a scenario that one could trigger on a single language install?

stl/src/syserror_import_lib.cpp
StephanTLavavej: It's probably unnecessarily paranoid, but [`LocalFree`](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-localfree) is a no-op for null.

stl/src/syserror_import_lib.cpp
StephanTLavavej: I could add logic, but I'd really want it to be motivated by an actual test run.

muellerj2 (approved):
Looks good on single-language de-DE:

```
  System locale name: "de-DE"
  System language ID: 1031 decimal, 0x0407 hex
ERROR_FILE_NOT_FOUND: "Das System kann die angegebene Datei nicht finden."
```

stl/src/syserror_import_lib.cpp
StephanTLavavej: Hmm, I think I'll omit this, thanks.

stl/src/syserror_import_lib.cpp
StephanTLavavej: I'll change to a three-step process, thanks.

muellerj2 (approved):
de-DE single-language system with de-DE system locale:

```
  System locale name: "de-DE"
  System language ID: 1031 decimal, 0x0407 hex
ERROR_FILE_NOT_FOUND: "Das System kann die angegebene Datei nicht finden."
```

de-DE single-language system with de-CH system locale:
```
  System locale name: "de-CH"
  System language ID: 2055 decimal, 0x0807 hex
ERROR_FILE_NOT_FOUND: "Das System kann die angegebene Datei nicht finden."
```

TheStormN (approved):

sylveon:
Consider the following extreme scenario: an OS where the system locale is set to Chinese (without the corresponding language pack), the user display language is Russian, and no English language pack is installed.

In this case, the first attempt for English (US) would fail.
The second attempt would try to find a Chinese string, and would fail.
The last attempt would find the Russian string, but then attempt to encode it in Chinese (the active code page), giving just repeated question marks (pre-17.3 behavior).

The only fully correct fix would be to use FormatMessageW and convert that down to UTF-8 instead of the active code page. This is obviously a breaking runtime API change, so was not considered here.

I would argue such a change should be considered for an ABI break or even earlier. Since the code lives in the import library, the breaking change happen based on the version of Visual Studio used to compile, and not the version of the runtime installed on the end-user computer.

StephanTLavavej:
As I mentioned in the PR description, we can consider pathological scenarios in the future, if they turn out to happen with any significant frequency. I think I would actually recommend manually `LocalAlloc`ing a buffer and stringizing the error code value to produce `"system error code 1729"` as a fallback.

stl/src/syserror_import_lib.cpp
```diff
@@ -40,15 +40,33 @@ extern "C" {
...
+                reinterpret_cast<LPWSTR>(&_Lang_id), sizeof(_Lang_id) / sizeof(wchar_t));
+            if (_Ret == 0) {
+                continue; // If we can't get the system locale's language ID, skip this attempt
+            }
+        } else {
+            _Lang_id = 0;
+        }
+
+        _Chars = FormatMessageA(_Flags, nullptr, _Message_id, _Lang_id, reinterpret_cast<char*>(_Ptr_str), 0, nullptr);
     }
```
CaseyCarter, line 69: No change requested, but I see the point of the discussion in Discord about the looping structure obfuscating things a bit. Did you consider extracting the `FormatMessageA` call and making this non-looping a la:
```suggestion
    const auto _Format_msg = [=](DWORD _Lang_id) {
        constexpr auto _Flags = FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS;
        return FormatMessageA(_Flags, nullptr, _Message_id, _Lang_id, reinterpret_cast<char*>(_Ptr_str), 0, nullptr);
    };
    DWORD _Chars = _Format_msg(0x0409); // 1033 decimal, "en-US" locale
    if (_Chars == 0) {
        DWORD _Lang_id;
        const int _Ret = GetLocaleInfoEx(LOCALE_NAME_SYSTEM_DEFAULT, LOCALE_ILANGUAGE | LOCALE_RETURN_NUMBER,
            reinterpret_cast<LPWSTR>(&_Lang_id), sizeof(_Lang_id) / sizeof(wchar_t));
        if (_Ret != 0) {
            _Chars = _Format_msg(_Lang_id);
        }
    }
    if (_Chars == 0) {
        _Chars = _Format_msg(0);
    }
```
I like the control flow better, but I'm not sure that the lambda is any less obfuscating.

CaseyCarter (approved):

StephanTLavavej:

PR #5105, opened 2024-11-21
Move `system_clock` from `<__msvc_chrono.hpp>` to `<chrono>`

cpplearner:
After #4457 removed `_To_timespec64_sys_10_day_clamped`, `<mutex>` and friends no longer depend on `system_clock`, so `system_clock` can be moved back to `<chrono>`.

This PR also moves `high_resolution_clock` and `chrono_literals` to `<chrono>` since they are not used in STL headers.

### DIFF
--- a/stl/inc/__msvc_chrono.hpp
+++ b/stl/inc/__msvc_chrono.hpp
@@ -646,33 +646,6 @@ namespace chrono {
         return time_point<_Clock, _To>(_CHRONO round<_To>(_Time.time_since_epoch()));
     }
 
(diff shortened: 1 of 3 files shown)


StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for this throughput improvement! It's going to break some source code (we found 6 projects in our Real World Code test suite that were affected), so never let it be said that we're unwilling to make disruptive but righteous changes! :joy_cat: :computer: :alarm_clock:

PR #5130, opened 2024-11-24
Fix false positives by `filesystem::equivalent` on file systems with transient file IDs

### LINKED ISSUE #3571: <filesystem>: equivalent() gives false positives on Remote Desktop shared folders
**Describe the bug**
When connecting to a Windows machine with Remote Desktop and sharing a local folder with the remote machine, filesystem operations on the remote mounted folder can give spurious results for `equivalent()`.

**Command-line test case**
```
C:\Temp>type fs-equivalent.cpp
#include <filesystem>
#include <iostream>

int main(int argc, char *argv[]) {
    const char *dir = ".";
    if (argc > 1)
        dir = argv[1];
    std::cout << "Iterating over " << dir << std::endl;
    std::filesystem::path firstPath;
    int duplicates = 0;
    int unique = 0; 
    for (const auto &entry : std::filesystem::directory_iterator(dir)) {
        if (firstPath.empty()) {
            firstPath = entry.path();
            unique++; 
        } else {
            if (std::filesystem::equivalent(firstPath, entry.path())) {
                std::cout << firstPath << " and " << entry.path() << " seem equivalent" << std::endl;
                duplicates++;
            } else {
                unique++;
            }   
        }
    } 
    std::cout << "Iterated over " << unique << " unique files, found " << duplicates << " duplicates" << std::endl;
    return 0;
}

C:\Temp>cl /std:c++17 /EHsc /W4 /WX .\fs-equivalent.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.36.32323 for x64

fs-equivalent.cpp
Microsoft (R) Incremental Linker Version 14.36.32323.0

/out:fs-equivalent.exe  
fs-equivalent.obj 

C:\Temp>.\fs-equivalent.exe z:\directory
"z:\\directory\\file1" and "z:\\directory\\file42" seem equivalent
[...]
Iterated over 830 unique files, found 88 duplicates
```

**Expected behavior**
`std::filesystem::equivalent` shouldn't give false positives on distinct different files. When run on a directory with distinct files, without symlinks/hardlinks, the tool should print something like this: `Iterated over 918 unique files, found 0 duplicates`

**STL version**
Tested with MSVC 17.6 preview 1.

**Additional context**
To test this, connect to a remote Windows machine with Microsoft Remote Desktop (tested on macOS). In Remote Desktop, choose to share a local folder with the remote machine. Pick a directory that contains many distinct files (no symlinks, no hardlinks, etc). For the purposes of this test, I've tested with https://github.com/mstorsjo/llvm-mingw/releases/download/20230130/llvm-mingw-20230130-ucrt-x86_64.zip unpacked, iterating over files in the `include` or `x86_64-w64-mingw32\lib` subdirectories - but any directory with more than a few dozens files should do.

The reason for the issue, is that the unique file ID as returned by `GetFileInformationByHandle` in the `nFileIndexHigh` / `nFileIndexLow` fields actually aren't properly unique on some filesystem mounts. This seems to be the case with shared folders mounted by Remote Desktop and folders shared to a VM in VirtualBox.

Is it the case that the supposedly unique file ID fields from `GetFileInformationByHandle` are supposed to be unique only as long as the handles are open?

This same root cause (file IDs from `GetFileInformationByHandle` aren't unique on such filesystems) is hitting other tools such as LLVM, as reported in https://github.com/llvm/llvm-project/issues/22079 and https://github.com/llvm/llvm-project/issues/61401.

I.e. the issue can be reframed into this: Is STL's use of `GetFileInformationByHandle` in `__std_fs_get_file_id` (used later in `std::filesystem::equivalent`) correct (opening a handle, inspecting it and grabbing the unique ID, closing the handle, then later comparing the IDs)? Or are the file IDs only valid as long as the corresponding HANDLE is open? I.e. is this a fault in the users of `GetFileInformationByHandle` (STL and LLVM), or in the host of the shared directories, giving non-unique IDs?

muellerj2:
Fixes #3571 for `<filesystem>` and `<experimental/filesystem>`.

The fix for `<experimental/filesystem>` affects the redist because it changes the implementation of `_Equivalent` in `filesys.cpp` that is exported by msvcp140.dll.

There is no additional test coverage because a reliable test for this issue would involve setting up a drive commonly producing such transient file IDs. But I verified manually with a drive shared by Remote Desktop that (a) false positives by `std::filesystem::equivalent` and `std::experimental::filesystem::equivalent` are present without this PR and (b) are gone with it. I also added comments at the top of `_Equivalent` and `__std_fs_equivalent` that reference #3571 and serve as a reminder why the file handles can't be closed before file IDs for both files have been obtained.

### DIFF
--- a/stl/inc/xfilesystem_abi.h
+++ b/stl/inc/xfilesystem_abi.h
@@ -232,11 +232,6 @@ struct __std_fs_convert_result {
     __std_win_error _Err;
 };
 
-struct __std_fs_file_id { // typedef struct _FILE_ID_INFO {
-    unsigned long long _Volume_serial_number; //    ULONGLONG VolumeSerialNumber;
-    unsigned char _Id[16]; //    FILE_ID_128 FileId;
-}; // } FILE_ID_INFO, ...;
-
 enum class __std_fs_copy_options {
     _None = 0x0,
 
@@ -303,8 +298,13 @@ _NODISCARD __std_fs_convert_result __stdcall __std_fs_convert_wide_to_narrow_rep
     _In_ __std_code_page _Code_page, _In_reads_(_Input_len) const wchar_t* _Input_str, _In_ int _Input_len,
     _Out_writes_opt_(_Output_len) char* _Output_str, _In_ int _Output_len) noexcept;
 
-_NODISCARD _Success_(return == __std_win_error::_Success) __std_win_error
-    __stdcall __std_fs_get_file_id(_Out_ __std_fs_file_id* _Id, _In_z_ const wchar_t* _Path) noexcept;
+struct __std_fs_equivalent_result {
+    bool _Equivalent;
+    __std_win_error _Error;
+};
+
+_NODISCARD __std_fs_equivalent_result __stdcall __std_fs_equivalent(
+    _In_z_ const wchar_t* _Path1, _In_z_ const wchar_t* _Path2) noexcept;
 
 _NODISCARD __std_win_error __stdcall __std_fs_set_last_write_time(
     _In_ long long _Last_write_filetime, _In_z_ const wchar_t* _Path) noexcept;
--- a/stl/src/filesystem.cpp
+++ b/stl/src/filesystem.cpp
@@ -455,7 +455,7 @@ struct __std_fs_file_id { // typedef struct _FILE_ID_INFO {
 }
 
 [[nodiscard]] __std_fs_equivalent_result __stdcall __std_fs_equivalent(
-    _In_z_ const wchar_t* _Left_path, _In_z_ const wchar_t* _Right_path) noexcept {
+    _In_z_ const wchar_t* const _Left_path, _In_z_ const wchar_t* const _Right_path) noexcept {
     // See GH-3571: File IDs are only guaranteed to be unique and stable while handles remain open
     __std_win_error _Last_error;
     const _STD _Fs_file _Left_handle(
--- a/stl/src/filesystem.cpp
+++ b/stl/src/filesystem.cpp
@@ -434,6 +434,26 @@ void __stdcall __std_fs_directory_iterator_close(_In_ const __std_fs_dir_handle 
     return __vcp_Copyfile(_Source, _Target, /* _Fail_if_exists = */ false);
 }
 
+struct __std_fs_file_id { // typedef struct _FILE_ID_INFO {
+    unsigned long long _Volume_serial_number; //    ULONGLONG VolumeSerialNumber;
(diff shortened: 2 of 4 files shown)


stl/inc/xfilesystem_abi.h
```diff
@@ -303,8 +298,13 @@ _NODISCARD __std_fs_convert_result __stdcall __std_fs_convert_wide_to_narrow_rep
     _In_ __std_code_page _Code_page, _In_reads_(_Input_len) const wchar_t* _Input_str, _In_ int _Input_len,
     _Out_writes_opt_(_Output_len) char* _Output_str, _In_ int _Output_len) noexcept;
 
-_NODISCARD _Success_(return == __std_win_error::_Success) __std_win_error
-    __stdcall __std_fs_get_file_id(_Out_ __std_fs_file_id* _Id, _In_z_ const wchar_t* _Path) noexcept;
```
muellerj2, line 307: `__std_fs_get_file_id` is inherently broken because it's based on the assumption that file IDs are unique and stable, so I removed it. I *think* this removal is fine ABI-wise, but I would like someone to confirm because it seems to be the first time that a function is removed from `<xfilesystem_abi.h>`.

stl/src/filesystem.cpp
```diff
@@ -434,18 +434,35 @@ void __stdcall __std_fs_directory_iterator_close(_In_ const __std_fs_dir_handle
     return __vcp_Copyfile(_Source, _Target, /* _Fail_if_exists = */ false);
 }
 
-_Success_(return == __std_win_error::_Success) __std_win_error
-    __stdcall __std_fs_get_file_id(_Out_ __std_fs_file_id* const _Id, _In_z_ const wchar_t* const _Path) noexcept {
```
AlexGuteniev, line 438: I don't think `__std_fs_get_file_id` can be removed from import library; it has to be preserved due to ABI.
See `TRANSITION, ABI: preserved for binary compatibility` comment in many places here

muellerj2: Yeah, I think you are right. I restored the function in the cpp file and also moved the definition of the type `__std_fs_file_id` there.

StephanTLavavej (approved):

StephanTLavavej:
Thanks, this is awesome! :heart_eyes_cat: I pushed cosmetic changes.

StephanTLavavej:

StephanTLavavej:
Thanks again for figuring out how to fix this runtime correctness bug! :page_facing_up: :heavy_equals_sign: :scroll:

Issue #994, opened 2020-07-05
<regex>: Regex erroneously returns a match


AlexGuteniev:
**Describe the bug**
Regex returns a match where it is not expected to return a match.

Note that the pattern has mismatched `[` and `]`.
Need to clarify if it should actually throw, or the behavior is undefined.

**Command-line test case**
```
d:\Temp2>type repro.cpp
#include <iostream>
#include <regex>
#include <string>

int main()
{
        try
        {
                std::string Regex = "[[.(.]a[a]";
                std::string MatchString = "v";
                std::regex std_regex(Regex);
                bool Result = std::regex_match(MatchString, std_regex);

                std::cout << "Is matched:" << Result;
        }
        catch (std::regex_error& err)
        {
                std::cout << err.code() << '\n' << err.what() << '\n';
        }
}
d:\Temp2>cl /EHsc /W4 /WX .\repro.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.27.29009.1 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.

repro.cpp
Microsoft (R) Incremental Linker Version 14.27.29009.1
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:repro.exe
repro.obj

d:\Temp2>.\repro.exe
Is matched:1
```

**Expected behavior**
DevCom-306176 reporter expects:
```
Is matched:0
```
I expect an exception.

**STL version**
```
Microsoft Visual Studio Professional 2019 Preview
Version 16.7.0 Preview 3.1
```

**Additional context**
This item is also tracked on Developer Community as DevCom-306176 and by Microsoft-internal VSO-660624 / [AB#660624](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/660624).

Amaroker:
If I understand correctly, std::regex_match would never be invoked here because the constructor of std::regex has to throw if the regular expression is not valid. It should throw a std::regex_error with code std::regex_constants::error_brack.

achabense:
This is valid, `'['` is treated as a normal character in []:
```cpp
#include<regex>
#include<iostream>

int main() {
    std::regex regex("[a[a]*");
    std::string str("[[[[[[aaaa");
    std::cout << std::regex_match(str, regex);//1
}
```

[...]

PR #5158, opened 2024-11-29
`<regex>`: Always reject character ranges with set limits

### LINKED ISSUE #4995: `<regex>`: `R"([\d-e])"` should be rejected
# Describe the bug
The regex [\d-e] (character class containing the range \d to e) is accepted (treated as \d, and the literal characters - and e), contrary to [the ECMA-262 spec](https://262.ecma-international.org/5.1/#sec-15.10.2.15) (\d isn't a single character, so it can't be used like that)

# Command-line test case
```
#include <regex>

int main()
{
    try {
        std::regex r("[\\d-e]");
        puts("it's legal");
    } catch (std::exception& e) {
        puts(e.what());
    }
    try {
        std::regex r("[b-a]");
        puts("it's legal");
    } catch (std::exception& e) {
        puts(e.what());
    }
}
```
https://godbolt.org/z/oMvEr5YTs

# Expected behavior
Both should be illegal (currently, only the latter is rejected)

# STL version
Ask Godbolt

# Additional context
Feel free to close this one as wontfix, if you feel it's ossified into a vendor extension. As long as it's a conscious choice, I'm fine with whichever outcome.

muellerj2:
Fixes #4995.

Unblocks libcxx test `std/re/re.regex/re.regex.construct/bad_range.pass.cpp`.

Note that I did not add test coverage for regular expressions like `[[.a.]-e]`: After this change, this regex is rejected as well (rather than parsed incorrectly). However, such regular expressions should actually be accepted (!), because `[.xxx.]` is not a set of characters, but (kind of) a single character if `xxx` is the name of a valid collating element. But support of collating symbols is currently so terribly broken that it is basically guaranteed to produce wrong results (see https://github.com/microsoft/STL/issues/994#issuecomment-1603697429), so I think it would be a major disservice to users if we actually accepted this regex without a major overhaul of collation support in the implementation.

### DIFF
--- a/tests/std/tests/VSO_0000000_regex_use/test.cpp
+++ b/tests/std/tests/VSO_0000000_regex_use/test.cpp
@@ -582,6 +582,19 @@ void test_gh_993() {
     }
 }
 
+void test_gh_4995() {
+    // GH-4995: R"([\d-e])" should be rejected
+    g_regexTester.should_throw("[\\d-e]", error_range);
+    g_regexTester.should_throw("[e-\\d]", error_range);
+    g_regexTester.should_throw("[\\w-\\d]", error_range);
+    g_regexTester.should_throw("[[:digit:]-e]", error_range);
+    g_regexTester.should_throw("[e-[:digit:]]", error_range);
+    g_regexTester.should_throw("[[:alpha:]-[:digit:]]", error_range);
+    g_regexTester.should_throw("[[=a=]-e]", error_range, ECMAScript | regex::collate);
+    g_regexTester.should_throw("[e-[=a=]]", error_range, ECMAScript | regex::collate);
+    g_regexTester.should_throw("[[=a=]-[=b=]]", error_range, ECMAScript | regex::collate);
+}
+
 void test_gh_5058() {
     // GH-5058 "<regex>: Small cleanups" changed some default constructors to be defaulted.
     // Verify that <regex> types are still const-default-constructible (N4993 [dcl.init.general]/8).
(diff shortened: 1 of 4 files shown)


tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -582,6 +582,19 @@ void test_gh_993() {
...
+    g_regexTester.should_throw("[\\d-e]", error_range);
+    g_regexTester.should_throw("[e-\\d]", error_range);
+    g_regexTester.should_throw("[\\w-\\d]", error_range);
+    g_regexTester.should_throw("[[:digit:]-e]", error_range);
+    g_regexTester.should_throw("[e-[:digit:]]", error_range);
+    g_regexTester.should_throw("[[:alpha:]-[:digit:]]", error_range);
+    g_regexTester.should_throw("[[=a=]-e]", error_range, ECMAScript | regex::collate);
+    g_regexTester.should_throw("[e-[=a=]]", error_range, ECMAScript | regex::collate);
+    g_regexTester.should_throw("[[=a=]-[=b=]]", error_range, ECMAScript | regex::collate);
+}
```
StephanTLavavej, line 596: This is great test coverage of the error cases. As you're adjusting the control flow, let's also add positive test coverage of sets that should still be accepted.

StephanTLavavej (approved):

StephanTLavavej:
Thanks for fixing this complicated control flow! :heart_eyes_cat: I pushed minor stylistic changes and test improvements after debugging through the code and convincing myself that your changes were correct.

StephanTLavavej:

StephanTLavavej:
Thanks for figuring out how this code was broken and how to fix it! :heart_eyes_cat: :yarn: :hammer_and_wrench:

PR #5163, opened 2024-12-05
Globally suppress C5278

CaseyCarter:
This is a new warning MSVC is adding to steer users away from attempting to specialize the type traits as forbidden by N4993 [meta.rqmts]/4. Since MSVC implements many of the type traits in the compiler, such specializations are generally ignored to the confusion of users who trigger the UB.

This is a mirror of the STL portion of MSVC-PR-596210.


StephanTLavavej (approved):

StephanTLavavej:
Verified that this works with the older Preview compiler as we've already silenced ~~mortals~~ "warning C4619: #pragma warning: there is no warning number 'number' (/Wall)"

CaseyCarter:
Merged without Changelogging - I doubt this change is interesting to anyone not an STL dev.

PR #5164, opened 2024-12-05
`<regex>`: Fix character range bounds in case-insensitive regexes

muellerj2:
This PR deals with three related problems for character ranges in case-insensitive mode:
* `_Builder::_Add_range()` casts the character bounds to `unsigned int`. As a consequence, characters with negative numeric values are not added to the bitmap, but rather to the `_Large` list of characters. This means that these characters are not found during matching. (Note the suspiciously different casts in the else branch for the case-sensitive case.)
* The parser fails to reject some empty ranges in case-insensitive mode such as `[Z-a]` (= `[z-a]`).
* When both the `collate` and `icase` flags are set, there is an unnecessary call to translate the bounds by `_Traits.translate()` first before passing them to `_Traits.translate_nocase()`. The standard says in [\[re.grammar\]/14.1 and 14.2](https://eel.is/c++draft/re.grammar#14) that it is sufficient to call `translate_nocase()` only. (See also `_Builder::_Add_char()`, which already follows the Standard in this regard.)

The PR moves the entire character translation into the parser so that empty ranges can be reliably diagnosed there in case-insensitive mode as well. It also fixes the unsigned cast and removes the unnecessary `translate()` call.

The test deliberately does not use any manual signed/unsigned casts, but leaves all of these casts to `char_traits` to avoid getting the casts similarly wrong in `<regex>` and the test.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -1499,7 +1499,7 @@ public:
     void _Add_char(_Elem _Ch);
     void _Add_class();
     void _Add_char_to_class(_Elem _Ch);
-    void _Add_range(_Elem _Ex0, _Elem _Ex1);
+    void _Add_range2(_Elem, _Elem);
     void _Add_named_class(_Regex_traits_base::char_class_type, bool = false);
     void _Add_equiv(_FwdIt, _FwdIt, _Difft);
     void _Add_coll(_FwdIt, _FwdIt, _Difft);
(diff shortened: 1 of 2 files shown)


stl/inc/regex
```diff
@@ -4113,16 +4106,22 @@ void _Parser<_FwdIt, _Elem, _RxTraits>::_ClassRanges() { // check for valid clas
...
+            if (_Flags & regex_constants::icase) {
+                _Chr1 = _Traits.translate_nocase(_Chr1);
+                _Chr2 = _Traits.translate_nocase(_Chr2);
+            } else if (_Flags & regex_constants::collate) {
                 _Chr1 = _Traits.translate(_Chr1);
+                _Chr2 = _Traits.translate(_Chr2);
             }
 
-            if (static_cast<typename _RxTraits::_Uelem>(_Val) < static_cast<typename _RxTraits::_Uelem>(_Chr1)) {
+            if (static_cast<typename _RxTraits::_Uelem>(_Chr2) < static_cast<typename _RxTraits::_Uelem>(_Chr1)) {
```
muellerj2, line 4120: Note that there is still a (pre-existing) bug in this error check when the `collate` flag is set: The bounds should be transformed by `_Traits.transform()` before performing the comparison.

However, the matcher fails to do this `transform()` dance as well, so I think this should rather be fixed for both matcher and parser in a coordinated fashion in a follow-up PR. (But I will adjust the PR if you rather prefer to fix this check completely now.)

StephanTLavavej: A follow-up PR will be great.

StephanTLavavej (approved):

StephanTLavavej:
Thanks - I really appreciate the detailed explanations in your PRs! :cat:

StephanTLavavej:

StephanTLavavej:
Thanks! :heart_eyes_cat:+

stl/inc/regex
StephanTLavavej: For the record, I believe this was done in #5238.

muellerj2: Yes, it was.

PR #5167, opened 2024-12-06
`<regex>`: Limit backreference parsing to single digit for basic regular expressions

muellerj2:
According to [Sections 9.3.6 and 9.5.1 of the POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/basedefs/V1_chap09.html#tag_09_03_06), a backreference is a backslash followed by a single digit, so `\1` to `\9`. Thus, `\(a\)\10` is a valid basic regular expression matching the string `aa0`, because `\10` is a backreference to capture group 1 (`\(a\)`) plus the digit 0. However, the regex parser currently rejects this regular expression because it interprets `\10` as a backreference to capture group 10.

After this PR, the parser only reads a single digit in basic regular expression (and grep) mode, so `\10` is parsed correctly as a backreference to capture group 1 plus digit 0.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -1187,8 +1187,6 @@ _NODISCARD bool operator!=(const match_results<_BidIt, _Alloc>& _Left, const mat
 }
 #endif // !_HAS_CXX20
 
-_INLINE_VAR constexpr int _BRE_MAX_BACKREF_DIGITS = 1;
-
 _INLINE_VAR constexpr unsigned int _Bmp_max   = 256U; // must fit in an unsigned int
 _INLINE_VAR constexpr unsigned int _Bmp_shift = 3U;
 _INLINE_VAR constexpr unsigned int _Bmp_chrs  = 1U << _Bmp_shift; // # of bits to be stored in each char
@@ -4333,9 +4331,10 @@ bool _Parser<_FwdIt, _Elem, _RxTraits>::_CharacterEscape() { // check for valid 
 
 template <class _FwdIt, class _Elem, class _RxTraits>
 void _Parser<_FwdIt, _Elem, _RxTraits>::_AtomEscape() { // check for valid atom escape
+    constexpr int _Bre_max_backref_digits = 1;
     if ((_L_flags & _L_bckr)
         && _DecimalDigits2(regex_constants::error_backref,
-            (_L_flags & _L_lim_bckr) ? _BRE_MAX_BACKREF_DIGITS : INT_MAX)) { // check for valid back reference
+            (_L_flags & _L_lim_bckr) ? _Bre_max_backref_digits : INT_MAX)) { // check for valid back reference
         if (_Val == 0) { // handle \0
             if (!(_L_flags & _L_bzr_chr)) {
                 _Error(regex_constants::error_escape);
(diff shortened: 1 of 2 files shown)


stl/inc/regex
```diff
@@ -1187,7 +1187,8 @@ _NODISCARD bool operator!=(const match_results<_BidIt, _Alloc>& _Left, const mat
 }
 #endif // !_HAS_CXX20
 
-_INLINE_VAR constexpr unsigned int _BRE_MAX_GRP = 9U;
+_INLINE_VAR constexpr unsigned int _BRE_MAX_GRP   = 9U;
```
StephanTLavavej, line 1190: Now that you're consuming at most 1 digit, do we still need the only usage of `_BRE_MAX_GRP` here?

https://github.com/microsoft/STL/blob/89ca073a866d24db42b3f4c5e406be4b28a148f4/stl/inc/regex#L4343-L4346

StephanTLavavej (changes requested):

stl/inc/regex
muellerj2: I guess it effectively only defends against questionable implementations of `value` method provided by the traits class now. That doesn't seem like a reason good enough to keep it, so changed in the next commit.

stl/inc/regex
```diff
@@ -3950,9 +3950,9 @@ int _Parser<_FwdIt, _Elem, _RxTraits>::_Do_digits(
 }
 
 template <class _FwdIt, class _Elem, class _RxTraits>
-bool _Parser<_FwdIt, _Elem, _RxTraits>::_DecimalDigits(
-    regex_constants::error_type _Error_type) { // check for decimal value
-    return _Do_digits(10, INT_MAX, _Error_type) != INT_MAX;
+bool _Parser<_FwdIt, _Elem, _RxTraits>::_DecimalDigits2(
+    const regex_constants::error_type _Error_type, const int _Count /*= INT_MAX */) { // check for decimal value
```
StephanTLavavej, line 3954: Hyper-nitpick: When we comment default args like this, we put a space after the `/*`.

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -669,6 +669,20 @@ void test_gh_5160() {
     neg_regex.should_search_fail(L"xxxYxx\x2009xxxZxxx"); // U+2009 THIN SPACE
 }
 
+void test_gh_5167() {
+    // GH-5167: Limit backreference parsing to single digit for basic regular expressions
+    g_regexTester.should_match("abab0", R"x(\(ab*\)\10)x", basic);
```
StephanTLavavej, line 674: Style: As the raw string literal doesn't contain double quotes, the extra `x` delimiter is unnecessary and adds visual noise. Other raw string literals in this test don't do this.

StephanTLavavej:
Excellent, thank you! :green_heart: I pushed small stylistic nitpicks.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for properly implementing this grammar! :one: :nine: :cat:

PR #5169, opened 2024-12-06
Stop using ctest

CaseyCarter:
It's pointless to use ctest to run one of two different LIT commands when we can simply run LIT directly. This will make it easier to add future tests: we can simply add more targets, we won't need to jump through hoops to tell ctest not to run the new test all the time. This change will ease adding a new EDG-only test target for the IntelliSense team.

Detailed changes:
* In `tests/CMakeLists.txt`:
  * Replace `add_test` targets with `add_custom_target`s. The `stl` and `stlasan` tests are now `test` and `extra-asan-tests` targets, respectively.
  * Replace `list(APPEND unset_variable_name a b c d)` with the equivalent `set(unset_variable_name a b c d)` which makes it clear that we expect `unset_variable_name` to be undefined.
  * Merge `${Python_EXECUTABLE}` into `STL_LIT_COMMAND`, and extract `${STL_LIT_TEST_DIRS}`. The makes it easy to reuse `${STL_LIT_COMMAND}` and `${STL_LIT_TEST_DIRS}` with target-specific LIT options.
  * Note that `USES_TERMINAL` in the `add_custom_target`s makes ninja run those targets serially, maintaining the observable effects of the removed `set_tests_properties(stl stlasan PROPERTIES RUN_SERIAL ON)`.
* Remove `enable_testing` from the top-level `CMakeLists.txt`; it does nothing now that we have no `add_test`s. We still use `BUILD_TESTING` to control including the test targets in the generated build system, but I'm not certain there's a good reason to do so. Maybe to enable folks without python to build but not test the STL?
* In `run-tests.yml`, replace `ctestOptions` parameter with a `testTargets` parameter, a list of targets to build via `cmake --build meow --target woof quack`. This needs to be set in `asan-pipeline.yml` and plumbed through `build-and-test.yml` with a default of `test` to run the vanilla STL tests.
* Remove/Replace the `ctest` discussion in `README.md`.
* Drive-by: Report that the `psutil` python module is unavailable at most once per `stl-lit` run. Seeing three repetitions in the CI logs bothered me.


azure-devops/run-tests.yml
```diff
@@ -6,15 +6,15 @@ parameters:
...
+- name: testTargets
   type: string
 - name: skipTesting
   type: boolean
 steps:
 - script: |
     call "%ProgramFiles%\Microsoft Visual Studio\2022\Preview\Common7\Tools\VsDevCmd.bat" ^
     -host_arch=${{ parameters.hostArch }} -arch=${{ parameters.targetArch }} -no_logo
-    ctest --verbose ${{ parameters.ctestOptions }}
+    ninja ${{ parameters.testTargets }}
```
CaseyCarter, line 17: Should we use `ctest --build` here instead of invoking `ninja` directly to future-proof this?

AlexGuteniev (changes requested):
`README.md` should be updated within the same change I think

azure-devops/run-tests.yml
```diff
@@ -6,19 +6,18 @@ parameters:
...
+- name: testTargets
   type: string
 - name: skipTesting
   type: boolean
 steps:
 - script: |
     call "%ProgramFiles%\Microsoft Visual Studio\2022\Preview\Common7\Tools\VsDevCmd.bat" ^
     -host_arch=${{ parameters.hostArch }} -arch=${{ parameters.targetArch }} -no_logo
-    ctest --verbose ${{ parameters.ctestOptions }}
+    cmake --build "$(buildOutputLocation)" --target ${{ parameters.testTargets }}
```
CaseyCarter, line 17: Bug: this isn't going to run multiple test targets (as intended) if tests in the early targets fail. I think we need to run `ninja` directly so we can add `-k0` to get the desired behavior.

CMakeLists.txt
```diff
@@ -118,7 +118,6 @@ set(VCLIBS_DEBUG_OPTIONS "$<$<COMPILE_LANGUAGE:CXX>:/Od>")
 set(VCLIBS_RELEASE_OPTIONS "$<$<COMPILE_LANGUAGE:CXX>:/O2;/Os>")
 
 if(BUILD_TESTING)
-    enable_testing()
     add_subdirectory(tests)
 endif()
```
StephanTLavavej, line 122: No change requested:

> We still use `BUILD_TESTING` to control including the test targets in the generated build system, but I'm not certain there's a good reason to do so. Maybe to enable folks without python to build but not test the STL?

Keeping the `BUILD_TESTING` option seems like a good idea for that reason. The Python requirement is indeed limited to the `tests` subdirectory:

https://github.com/microsoft/STL/blob/89ca073a866d24db42b3f4c5e406be4b28a148f4/tests/CMakeLists.txt#L26

tests/CMakeLists.txt
```diff
@@ -26,19 +26,12 @@ add_subdirectory(utils/stl-lit)
...
-list(APPEND STLASAN_LIT_COMMAND "${STL_LIT_OUTPUT}"
-                                "${LIT_FLAGS}"
-                                "-D" "tags=ASAN"
-                                "${STL_LIT_TEST_DIRS}")
-
-add_test(NAME stl COMMAND ${Python_EXECUTABLE} ${STL_LIT_COMMAND} COMMAND_EXPAND_LISTS)
-add_test(NAME stlasan COMMAND ${Python_EXECUTABLE} ${STLASAN_LIT_COMMAND} COMMAND_EXPAND_LISTS)
-set_tests_properties(stl stlasan PROPERTIES RUN_SERIAL ON)
+
+add_custom_target(test COMMAND ${STL_LIT_COMMAND} -Dnotags=ASAN ${STL_LIT_TEST_DIRS} USES_TERMINAL)
```
StephanTLavavej, line 35: `test` is essentially impossible to search for - it's really, really important to be able to trace through build systems and figure out what links with what. I strongly recommend renaming this to something like `normal-tests` or `vanilla-tests` (for symmetry with `extra-asan-tests`). Then I believe you will need to update:

* `azure-devops/build-and-test.yml`
  + `default: 'test'`
* `README.md`
  + ``the plain `test` target to run the "normal" set of tests``
  + `` `ninja test extra-asan-tests` ``

README.md
```diff
@@ -305,31 +308,9 @@ control over what category of tests you would like to run. The following will ru
...
-
-CTest will output everything that was sent to stderr for each of the failed test suites, which can be used to identify
-which individual test within the test suite failed. It can sometimes be helpful to run CTest with the `-V` option in
-order to see the stdout of the tests.
-
-### stl-lit
-
-When running the tests directly via the generated `stl-lit.py` script the result of each test will be printed. The
-format of each result is `{Result Code}: {Test Suite Name} :: {Test Name}:{Configuration Number}`.
+Whether running the tests via building the cmake test targets or directly via the generated `stl-lit.py` script, the
```
StephanTLavavej, line 311: We consistently capitalize CMake except in commands.

README.md
```diff
@@ -286,13 +289,13 @@ under a category in libcxx, or running a single test in `std` and `tr1`.
 These examples assume that your current directory is `C:\Dev\STL\out\x64`.
 
 * This command will run all of the test suites with verbose output.
-  + `ctest -V`
-* This command will also run all of the test suites.
+  + `ninja test extra-asan-tests`
```
StephanTLavavej, line 292: This still says "with verbose output", but is that true? `--verbose` is not being passed to `ninja` here, but we do in `azure-devops/run-tests.yml`. (I haven't checked what this does.)

StephanTLavavej, line 292: This is missing `-k 0` (as seen in `azure-devops/run-tests.yml`), so presumably it will stop if the normal tests fail. This is unlike the following stl-lit command, which will keep going. I see several options here:

* Mention this behavioral difference.
* Avoid this behavioral difference by adding `-k 0` here.
* Avoid this behavioral difference by adding an `all-tests` target that runs both normal and ASan.
  + Running `-Dnotags=ASAN` followed by `-Dtags=ASAN` serially is more complicated and slightly less efficient than entirely omitting the tags options.
* My recommendation: Change *all* of these listed commands to omit ASan (i.e. mention only the normal test target here, and add `-Dnotags=ASAN` to all stl-lit invocations). This aligns with our PR checks and our normal development practice.

StephanTLavavej (changes requested):

README.md
StephanTLavavej: In that case, do we need `--verbose` in `azure-devops/run-tests.yml`?

CaseyCarter: Ninja abbreviates the command for custom targets without verbose:
```
[0/1] C:\WINDOWS\system32\cmd.exe /C "cd /D C:\STL\out\x64...xx/test/std C:/STL/tests/std/tests C:/STL/tests/tr1/tests"-- Testing: 58732 tests, 32 workers --
SKIPPED: std :: tests/GH_003617_vectorized_meow_element:1 (1 of 58732)
SKIPPED: std :: tests/P2465R3_standard_library_modules:01 (2 of 58732)
```
which is fine for typical command-line use where you don't really care to see the full expansion, but for CI logs I prefer to see the full command:
```
C:\STL\out\x64>ninja -v test
[0/1] C:\WINDOWS\system32\cmd.exe /C "cd /D C:\STL\out\x64\tests && "C:\Program Files\Python313\python.exe" C:/STL/out/x64/tests/utils/stl-lit/stl-lit.py -o C:/STL/out/x64/tests/test_results.json -Dnotags=ASAN C:/STL/llvm-project/libcxx/test/std C:/STL/tests/std/tests C:/STL/tests/tr1/tests"
-- Testing: 58732 tests, 32 workers --
SKIPPED: std :: tests/GH_003617_vectorized_meow_element:1 (1 of 58732)
SKIPPED: std :: tests/P2465R3_standard_library_modules:01 (2 of 58732)
```

StephanTLavavej: That's an excellent reason, thanks. Abbreviated logs are the worst.

tests/CMakeLists.txt
CaseyCarter: Given that we never use these directly - they exist solely for the CI pipelines - why not simply name them `STL-CI` and `STL-ASan-CI`, and skip mentioning them in the `README` at all?

StephanTLavavej: That sounds good to me. We aren't currently using those names in the infra, so they will be searchable.

tests/CMakeLists.txt
```diff
@@ -26,19 +26,12 @@ add_subdirectory(utils/stl-lit)
...
+add_custom_target(extra-asan-tests COMMAND ${STL_LIT_COMMAND} -Dtags=ASAN ${STL_LIT_TEST_DIRS} USES_TERMINAL)
+# add_custom_target(only-edg-tests COMMAND ${STL_LIT_COMMAND} -Dnotags=ASAN --just-BE-things ${STL_LIT_TEST_DIRS} USES_TERMINAL)
```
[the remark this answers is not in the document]
CaseyCarter: This was a "----> make changes here <----" for the IntelliSense team that I expected to be cleaned up in a week or two, but the file is small enough I can point them at it and they'll work it out.

README.md
CaseyCarter: Resolved by removing all mention of running the custom test targets.

README.md
```diff
@@ -269,34 +269,33 @@ C:\Users\username\Desktop>dumpbin /DEPENDENTS .\example.exe | findstr msvcp
...
+and `tr1`.
 
-## Running A Subset Of The Tests
-
-`${PROJECT_BINARY_DIR}\tests\utils\stl-lit\stl-lit.py` can be invoked on a subdirectory of a test suite and will execute
-all the tests under that subdirectory. This can mean executing the entirety of a single test suite, running all tests
-under a category in libcxx, or running a single test in `std` and `tr1`.
+Some useful `stl-lit.py` options:
+* `-v` (verbose) tells `stl-lit.py` to show us output from failed test cases.
+* `-Dnotags=ASAN` disables the "extra ASAn configs" that we typically run only in CI. This is useful to limit runtime
```
StephanTLavavej, line 282: ASAn capitalization, occurs below.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:stop_sign: :heart_eyes_cat: :green_heart:
===

PR #5170, opened 2024-12-07
`<xmemory>`: Tolerate bogus const-overloading in iterators passed to `uninitialized_meow()`

StephanTLavavej:
* Followup to #5135.
* Reported llvm/llvm-project#119084.
* Fixes VSO-2318081 / AB#2318081 "\[RWC\]\[prod/fe\]\[Regression\] 4 projects failed with error C2440: 'initializing': cannot convert from `'const llvm::Value *'` to `'_Ty'`".

In an ideal world with infinite capacity, we would push through RWC fixes as we did with #5105. However:

* :santa:
* These fixes will be more intrusive than adding `#include <chrono>`.
* I suspect that a fair amount of code will be affected and LLVM is just the first canary, even though `uninitialized_meow()` is relatively obscure.

### DIFF
--- a/stl/inc/xmemory
+++ b/stl/inc/xmemory
@@ -1602,7 +1602,7 @@ void _Return_temporary_buffer(_Ty* const _Pbuf) noexcept {
 }
 
 template <class _Ty, class _InIt>
-void _Construct_in_place_by_deref(_Ty& _Val, const _InIt& _Iter)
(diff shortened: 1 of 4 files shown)


CaseyCarter (approved):

StephanTLavavej:

PR #5177, opened 2024-12-10
`STL.natvis`: Simplify visualization for `string`

pps83:
It appears that debugger visualizer accepts `na` regardless if underlying char type

### DIFF
--- a/stl/debugger/STL.natvis
+++ b/stl/debugger/STL.natvis
@@ -1005,13 +1005,11 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
    * wchar_t / unsigned short / char16_t = su format
    * char32_t = s32 format
    -->
-  <Type Name="std::basic_string&lt;char,*&gt;">
-    <AlternativeType Name="std::basic_string&lt;char8_t,*&gt;" />
-
+  <Type Name="std::basic_string&lt;*,*&gt;">
     <Intrinsic Name="size" Expression="_Mypair._Myval2._Mysize" />
     <Intrinsic Name="capacity" Expression="_Mypair._Myval2._Myres" />
-    <!-- _BUF_SIZE = 16 / sizeof(char) &lt; 1 ? 1 : 16 / sizeof(char) == 16 -->
-    <Intrinsic Name="bufSize" Expression="16" />
+    <!-- _BUF_SIZE = 16 / sizeof($T1) &lt; 1 ? 1 : 16 / sizeof($T1) -->
+    <Intrinsic Name="bufSize" Expression="16 / sizeof($T1) &lt; 1 ? 1 : 16 / sizeof($T1)" />
     <Intrinsic Name="isShortString" Expression="capacity() &lt; bufSize()" />
     <Intrinsic Name="isLongString" Expression="capacity() &gt;= bufSize()" />
     <DisplayString Condition="isShortString()">{_Mypair._Myval2._Bx._Buf,na}</DisplayString>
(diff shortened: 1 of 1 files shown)


pps83:
here's debugger view:

```
    std::string s15 = "012345678901234";
    std::string s16 = "0123456789012345";
    std::string::iterator s15i = s15.begin() + 1;
    std::string::const_iterator s16ic = s16.cbegin() + 1;

    std::wstring w7 = L"0123456";
    std::wstring w8 = L"01234567";
    std::wstring::iterator w7i = w7.begin() + 1;
    std::wstring::const_iterator w8ic = w8.cbegin() + 1;

    std::u32string u3 = U"012";
    std::u32string u4 = U"0123";
    std::u32string::iterator u3i = u3.begin() + 1;
    std::u32string::const_iterator u4ic = u4.cbegin() + 1;
```

image

pps83:
Note, it does break `std::basic_string<unsigned short>` case:
```
    std::wstring w7 = L"0123456";
    std::wstring w8 = L"01234567";

    std::basic_string<unsigned short> x7 = (unsigned short*)w7.c_str();
    std::basic_string<unsigned short> x8 = (unsigned short*)w8.c_str();
    std::basic_string<unsigned short>::iterator x7i = x7.begin() + 1;
    std::basic_string<unsigned short>::const_iterator x8ic = x8.cbegin() + 1;
```

image

stl/debugger/STL.natvis
```diff
@@ -1030,85 +1028,14 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
     </Expand>
   </Type>
 
-  <Type Name="std::basic_string&lt;wchar_t,*&gt;">
-    <AlternativeType Name="std::basic_string&lt;unsigned short,*&gt;" />
```
CaseyCarter, line 1034: I think we should restore this special case but only for `unsigned short` to preserve the `/Zc:wchar_t-` user experience. We've deprecated `/Zc:wchar_t-` a bit in the STL, but I see no reason to regress debugging for the poor folks who have to maintain ancient code that needs it.

stl/debugger/STL.natvis
```diff
@@ -1030,85 +1028,14 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
...
       <SmartPointer Usage="Indexable">_Ptr,na</SmartPointer>
       <Expand>
           <Item Name="[ptr]">_Ptr</Item>
       </Expand>
   </Type>
 
-  <Type Name="std::_String_iterator&lt;std::_String_val&lt;std::_Simple_types&lt;wchar_t&gt; &gt; &gt;">
-      <AlternativeType Name="std::_String_const_iterator&lt;std::_String_val&lt;std::_Simple_types&lt;wchar_t&gt; &gt; &gt;" />
-      <AlternativeType Name="std::_String_iterator&lt;std::_String_val&lt;std::_Simple_types&lt;unsigned short&gt; &gt; &gt;" />
-      <AlternativeType Name="std::_String_const_iterator&lt;std::_String_val&lt;std::_Simple_types&lt;unsigned short&gt; &gt; &gt;" />
```
CaseyCarter, line 1095: Ditto "restore `unsigned short` only"

CaseyCarter (changes requested):

stl/debugger/STL.natvis
pps83: this whole PR is weird. Somehow `na` works for all "normal" string types, but not for `unsigned short`, which requires explicit `su`. Feels like it's some sort of bug in VS that made it work. Is it possible for you to check internally why `na` works for other types? Otherwise, with this PR, visuzlizers might be broken eventually if this whole thing worked due to internal VS bug

CaseyCarter: `na` ("no address") should work for any pointer-to-object type. The problem is that `unsigned short` is not a character type, despite the existence of `/Zc:wchar_t-`, so `unsigned short*` doesn't trigger the text visualizer. We explicitly tell the debugger to print our `unsigned short*` like a `wchar_t*` with `su`. You could try `(char16_t*)meow,na` instead and see if that works well.

pps83: > ... The problem is that `unsigned short` is not a character type, despite the existence of `/Zc:wchar_t-`, so `unsigned short*` doesn't trigger the text visualizer.

It's the opposite. Seems like text visualizer actually works with `unsigned short` with `na` when `/Zc:wchar_t-` is used. It doesn't work when `/Zc:wchar_t` is used. Looks like internally the debugger treats `unsigned short*` as `wchar_t*` when `/Zc:wchar_t-` is used.

So, overall, string/string_view visualizers could drop all overloads for `unsinged short` and simply use `na` everywhere. But that means that if some code uses `unsigned short` for strings, then it will start to show numbers

CaseyCarter: > > ... The problem is that `unsigned short` is not a character type, despite the existence of `/Zc:wchar_t-`, so `unsigned short*` doesn't trigger the text visualizer.
> 
> It's the opposite. Seems like text visualizer actually works with `unsigned short` with `na` when `/Zc:wchar_t-` is used. It doesn't work when `/Zc:wchar_t` is used. Looks like internally the debugger treats `unsigned short*` as `wchar_t*` when `/Zc:wchar_t-` is used.
> 
> So, overall, string/string_view visualizers could drop all overloads for `unsinged short` and simply use `na` everywhere. But that means that if some code uses `unsigned short` for strings, then it will start to show numbers for regular builds without disabling native `wchar_t`.

Yes, you mentioned this in the other PR. I stand corrected - I probably shouldn't have conjectured about the behavior of `/Zc:wchar_t-` in the debugger when I haven't had to touch it for years. If it works with `/Zc:wchar_t-` then it's fine. We don't care in the slightest what happens to `basic_string<unsigned short>` without `/Zc:wchar_t-` because people shouldn't be using it in the first place. I retract both of these comments.

stl/debugger/STL.natvis
CaseyCarter: Comment retracted per discussion starting at https://github.com/microsoft/STL/pull/5177#discussion_r1878546856.

CaseyCarter (approved):

stl/debugger/STL.natvis
pps83: I'll double check locally with string/string_view (both using only `na`) that all char types (`char`, `char8_t`, `char16_t`, `wchar_t`, `char32_t`) work identically regardless of `/Zc:wchar_t-` switch. In that case, string/string_view can be greatly simplified.

CaseyCarter: Ah, great, we're on the same page now.

I saw this after my latest reply to the other PR. This slightly confused back-and-forth has me questioning the wisdom of my suggestion to split these changes into two PRs =)

pps83:
This should be final update. Please read this msg with full details: https://github.com/microsoft/STL/pull/5176#issuecomment-2532719516

CaseyCarter (approved):
Wow, that's a lot of cleanup. Very nice!

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: I pushed a comment update and two simplifications for wildcards.

:warning: I'm having trouble getting the VS IDE to pick up the updated natvis (I couldn't get it to display an added "taxicab" node), so I was unable to verify these changes, and I'm totally YOLOing them. Please double-check @CaseyCarter and @pps83 - I apologize if I messed up the PR.

For future reference, after code review begins, please avoid force-pushing changes as that makes it harder to incrementally review what changed.

CaseyCarter (approved):

pps83:
I also updated [CPPDebuggerVisualizers](https://github.com/pps83/CPPDebuggerVisualizers), it appears boost::string didn't even have correct visualizers. What's interesting, it has `InternalBufferChars` [variable to avoid calculations](https://github.com/pps83/CPPDebuggerVisualizers/commit/03f522554c2b191f0eb95d3c1bc03e3c25796f01#diff-fac3013bd55fde1bf9c944664863ed9430930a46a0e7e2cdcc7f44f5658696deR125) in string::capacity for short strings. Perhaps, ms/STL has something like that (or at least it should) and avoid doing this in visualizers:
```
<Intrinsic Name="bufSize" Expression="16 / sizeof($T1) &lt; 1 ? 1 : 16 / sizeof($T1)" />
```

CaseyCarter:
IIRC there was an issue with one of the compilers (clang-cl?) not emitting `static constexpr` class members into the PDB when we last overhauled `std::string` so we replicated the calculation in the visualizer as a workaround. It may not be necessary anymore, but nobody has investigated.

StephanTLavavej:

StephanTLavavej:
Thanks for this dramatic simplification, and congratulations on your second microsoft/STL commit! :cat2: :black_cat: :heart_eyes_cat:

pps83:
> Thanks for this dramatic simplification, and congratulations on your second microsoft/STL commit! 🐈 🐈‍⬛ 😻

idea for another commit: there is a paper about [EAFs](https://github.com/cassioneri/eaf) that allow faster date/time conversion math. It is like 20-30% faster than current algo in `_Days_from_civil`/`_Civil_from_days`

PR #5192, opened 2024-12-16
`<regex>`: Correct characters not matched by special character dot

muellerj2:
This corrects the set of characters the special character dot `.` does not match in a regular expression as specified in the ECMAScript and POSIX standards, and aligns our treatment of `.` with libstdc++ and libc++.

* Adds U+2028 Line Separator and U+2029 Paragraph Separator as characters not matched by `.` in a `wregex` in ECMAScript mode. See [the definition of `.` semantics in Section 22.2.2.7 of ECMAScript 14](https://262.ecma-international.org/14.0/index.html#sec-compileatom), which removes the line terminators from the set of matched characters, and [the list of line terminators in Section 12.3](https://262.ecma-international.org/14.0/index.html#sec-line-terminators). (Note that this links to a newer standard, but the set of unmatched characters has not been changed since ECMAScript 3. Furthermore, the C++ standard does not modify the interpretation of `.`.)
* In all other modes, `.` matches all characters except NUL now. This is in accordance with [Section 9.3.4](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_03_04) and [Section 9.4.4](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_04_04) of the POSIX standard. (I contemplated whether a new line (LF) should not be matched in addition to or instead of NUL in grep or egrep mode, as that is what grep implementations tend to do. The POSIX standard only states that regular expressions cannot match LFs due to the way grep works, but does not explicitly modify the definition of `.` or regular expressions in general, so it is ambiguous on this question.  Since libstdc++ and libc++ only exclude NUL from the set of characters matched by `.` in grep and egrep mode, I decided to align the set of unmatched characters with them.)

Note: Whether NUL should be matched in POSIX regular expressions is the subject of [LWG-3603](https://cplusplus.github.io/LWG/issue3603).


tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -669,6 +669,43 @@ void test_gh_5160() {
     neg_regex.should_search_fail(L"xxxYxx\x2009xxxZxxx"); // U+2009 THIN SPACE
 }
 
+void test_gh_5192() {
+    // GH-5192: Correct characters not matched by special character dot
+    using namespace string_literals;
```
StephanTLavavej, line 674: This is unnecessary - the test begins with `using namespace std;` which makes all of the UDLs available.

(Due to the two layers of `inline namespace`s, `using namespace std::string_literals;`, `using namespace std::literals;`, or `using namespace std;` makes `std::literals::string_literals` available.)

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -669,6 +669,43 @@ void test_gh_5160() {
...
+    // GH-5192: Correct characters not matched by special character dot
+    using namespace string_literals;
+    for (const syntax_option_type option : {
+             regex_constants::basic,
+             regex_constants::extended,
+             regex_constants::awk,
+             regex_constants::grep,
+             regex_constants::egrep,
+         }) {
+        const test_regex regex(&g_regexTester, "^.*", option);
```
StephanTLavavej, line 682: The variable names `regex` and `wregex` below are shadowing the Standard types. Shadowing is intensely confusing and interferes with searching, so we should always avoid it.

(I've stopped complaining when I see `expected`, `count`, etc. but this is a `<regex>` test so we need to avoid shadowing its contents.)

StephanTLavavej (approved):

StephanTLavavej:
Thanks as always for the careful analysis, PR description, and changes! :heart_eyes_cat: I pushed small stylistic changes.

StephanTLavavej:

StephanTLavavej (approved):

StephanTLavavej:
:white_check_mark: :cat: :cat2:
===

PR #5194, opened 2024-12-17
Add support for running only edg tests

jgoshi:
Adding a new target STL-only-edg-tests which will run only edg tests.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -34,4 +34,4 @@ get_property(STL_LIT_TEST_DIRS GLOBAL PROPERTY STL_LIT_TEST_DIRS)
 
 add_custom_target(STL-CI COMMAND ${STL_LIT_COMMAND} -Dnotags=ASAN ${STL_LIT_TEST_DIRS} USES_TERMINAL)
 add_custom_target(STL-ASan-CI COMMAND ${STL_LIT_COMMAND} -Dtags=ASAN ${STL_LIT_TEST_DIRS} USES_TERMINAL)
-add_custom_target(STL-only-edg-tests COMMAND ${STL_LIT_COMMAND} -Dnotags=ASAN -Dedg_only=True --xunit-xml-output test-results.xml ${STL_LIT_TEST_DIRS} USES_TERMINAL)
+add_custom_target(test-only-edg COMMAND ${STL_LIT_COMMAND} -Dnotags=ASAN -Dtest-only-edg=True --xunit-xml-output test-results.xml ${STL_LIT_TEST_DIRS} USES_TERMINAL)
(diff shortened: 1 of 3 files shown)


tests/CMakeLists.txt
```diff
@@ -34,3 +34,4 @@ get_property(STL_LIT_TEST_DIRS GLOBAL PROPERTY STL_LIT_TEST_DIRS)
 
 add_custom_target(STL-CI COMMAND ${STL_LIT_COMMAND} -Dnotags=ASAN ${STL_LIT_TEST_DIRS} USES_TERMINAL)
 add_custom_target(STL-ASan-CI COMMAND ${STL_LIT_COMMAND} -Dtags=ASAN ${STL_LIT_TEST_DIRS} USES_TERMINAL)
+add_custom_target(STL-only-edg-tests COMMAND ${STL_LIT_COMMAND} -Dnotags=ASAN -Dedg_only=True --xunit-xml-output test-results.xml ${STL_LIT_TEST_DIRS} USES_TERMINAL)
```
CaseyCarter, line 37: (No change requested) FWIW the existing targets have `STL-` prefixes only because we decided to make them match the names of the pipelines that use them. You could name this something shorter like `only-edg-tests` or `test-only-edg` if you prefer.

tests/utils/stl/test/tests.py
```diff
@@ -71,6 +71,10 @@ def configureTest(self, litConfig):
             return Result(UNSUPPORTED, "Test does not require any of the features specified in limit_to_features: %s" %
                           msg)
 
+        if 'edg_only' in self.config.available_features:
+            if not 'edg' in self.requires:
+                return Result(UNSUPPORTED, 'We only run /BE tests with the edg_only flag')
```
CaseyCarter: Let's go ahead and combine these conditions. (Also, I'm super pedantic about "only" and I'll use this change as an excuse to fix the pre-existing occurrence, too!)
```suggestion
        if 'edg_only' in self.config.available_features and 'edg' not in self.requires:
            return Result(UNSUPPORTED, 'We run only /BE tests with the edg_only flag')
```

tests/utils/stl/test/tests.py
```diff
@@ -71,6 +71,10 @@ def configureTest(self, litConfig):
             return Result(UNSUPPORTED, "Test does not require any of the features specified in limit_to_features: %s" %
                           msg)
 
+        if 'edg_only' in self.config.available_features:
+            if not 'edg' in self.requires:
+                return Result(UNSUPPORTED, 'We only run /BE tests with the edg_only flag')
+
         if 'edg_drop' in self.config.available_features:
             if not 'edg' in self.requires:
                 return Result(UNSUPPORTED, 'We only run /BE tests with the edg drop')
```
CaseyCarter: Pre-existing:
```suggestion
                return Result(UNSUPPORTED, 'We run only /BE tests with the edg drop')
```

CaseyCarter (approved):
Just a couple of nits with the Python that I'll test and push changes for.

CaseyCarter:

tests/CMakeLists.txt
jgoshi: I changed the target name and flag to be consistent.

CaseyCarter (approved):

StephanTLavavej (approved):

CaseyCarter:
Thanks for working to improve STL IntelliSense support! Let us know if there's anything else we can do make things easier for you.

And congratulations on your first STL commit!

jgoshi:
Will do and thank you for your help!

________________________________

Issue #5204, opened 2024-12-23
`<regex>`: Collating ranges are broken


muellerj2:
When the `regex_constants::collate` flag is set, character ranges fail to match characters according to the locale's collation order.

# Test case
```cpp
#include <iostream>
#include <locale>
#include <regex>

using namespace std;

int main() {
    const locale de("de_DE");

    regex_traits<wchar_t> traits;
    traits.imbue(de);
    wregex re(L"[a-z]", regex_constants::collate);
    re.imbue(de);

    const wchar_t eszett = L'ß';
    const wchar_t s = L's';
    const wchar_t t = L't';
    const wchar_t ss[] = {L's', L's'};
    const wchar_t minus = L'-';
    const auto stransform = traits.transform(&s, &s + 1);
    const auto ttransform = traits.transform(&t, &t + 1);
    const auto eszetttransform = traits.transform(&eszett, &eszett + 1);
    const auto sstransform = traits.transform(begin(ss), end(ss));

    cout << regex_match(L"ß", re) << '\n';

    cout << (stransform <= eszetttransform
                && eszetttransform <= ttransform)
            << ' '
            << (sstransform == eszetttransform)
            << '\n';

    return 0;
}
```
[Godbolt link](https://godbolt.org/z/363ojExdE)

# Expected behavior
The range should match `ß`, since `ß` collates like `ss` and should be sorted between `s` and `t` (and thus between `a` and `z` as well). Hence, the output of this program should be:
```
1
1 1
```

# Actual behavior
The range does not match `ß`, so the actual output is:
```
0
1 1
```
Note that the output of the second line shows that the problem does not lie in the implementation of `regex_traits<wchar_t>::transform`, as it returns sort keys that are equal for `ß` and `ss` and place `ß` between `s` and `t`.

muellerj2:
Related to this, I was initially concerned about the implementation of `regex_traits<_Elem>::translate()`:

https://github.com/microsoft/STL/blob/90820002693fe6eaaec2e55884472c654186207e/stl/inc/regex#L295-L298

This replaces the character by the collation order sort key when the sort key has length 1 and otherwise returns the character itself.

[...]

PR #5209, opened 2024-12-27
`<regex>`: Simplify `regex_traits<_Elem>::translate(_Elem)`

muellerj2:
Simplifies `regex_traits<_Elem>::translate(_Elem)` to just return its only argument. In https://github.com/microsoft/STL/issues/5204#issuecomment-2559846465, I voiced my suspicion that the current implementation essentially does just that in a very complicated and expensive way. I verified this now by running the following program that tested 902 locales available on my machine (no output to stdout produced by the program):

```cpp
#include <iostream>
#include <locale>
#include <regex>
#include <string>
#include <vector>

#include <windows.h>

using namespace std;

BOOL add_locale(LPWSTR str, DWORD flags, LPARAM param) {
    vector<string>& vec = *reinterpret_cast<vector<string>*>(param);
    int size            = WideCharToMultiByte(CP_ACP, 0, str, -1, nullptr, 0, nullptr, nullptr);
    if (size) {
        vector<char> narrow(static_cast<size_t>(size), '\0');
        if (WideCharToMultiByte(CP_ACP, 0, str, -1, narrow.data(), size, nullptr, nullptr)) {
            vec.push_back(narrow.data());
        }
    }
    return TRUE;
}

int main() {
    vector<string> locales;
    locales.push_back("C");
    EnumSystemLocalesEx(&add_locale, 0, reinterpret_cast<LPARAM>(&locales), nullptr);

    for (const string& loc_name : locales) {
        const locale loc(loc_name);
        regex_traits<char> traits;
        regex_traits<wchar_t> wtraits;
        traits.imbue(loc);
        wtraits.imbue(loc);

        for (unsigned int i = 0; i <= 0xff; ++i) {
            try {
                const char x = (char) (unsigned char) i;
                if (traits.translate(x) != x) {
                    cout << loc_name << i << '\n';
                }
            } catch (const length_error&) {
            }
        }

        for (unsigned int i = 0; i <= 0xffff; ++i) {
            try {
                const wchar_t x = (wchar_t) i;
                if (wtraits.translate(x) != x) {
                    cout << loc_name << i << '\n';
                }
            } catch (const length_error&) {
            }
        }
    }
    return 0;
}
```

Even so, this PR still introduces a minor behavior change: The previous implementation can throw `length_error("string too long")` when it is passed a `char` that isn't a valid character in the locale's encoding (e.g., 0x80 in locales using UTF-8 encoding). But I think that the old behavior is undesirable anyway, as it makes the regex engine always fail with an exception in `regex_constants::collate` mode when a locale using UTF-8 encoding is imbued and the regex engine is applied to strings containing non-ASCII characters.


frederick-vs-ja (approved):
Aha, it seems that we took too much time to look into this.
[[re.traits]/4](https://eel.is/c++draft/re.traits#4) simply specifies that an implementation-provided `regex_traits<C>::translate` just returns the argument. See also https://en.cppreference.com/w/cpp/regex/regex_traits/translate.

muellerj2:
I put so much work in this mainly because I was worried about mix-and-match scenarios, not because I consider the old implementation correct. Think of the scenario that the implementations actually commonly produce different results and regex parser and matcher pick up different implementations or some other strange combination. Then this change had the potential to actually degrade the regex engine in such a mix-and-match scenario and lead to regex bugs that are difficult to understand and reproduce.

StephanTLavavej (approved):

StephanTLavavej:
Thanks for the careful analysis! :heart_eyes_cat:

StephanTLavavej:
This has gotta be a perf win.

StephanTLavavej:

StephanTLavavej:
Who knew machine translation was so easy! :robot: :joy_cat: :zany_face:

Issue #992, opened 2020-07-05
<regex> mishandles locale-based character classes outside of the char range


AlexGuteniev:
**Describe the bug**
Regex does not handle non-ASCII characters.

@BillyONeal comments:
> This is a longstanding bug in our regex engine -- when we form negated character classes (like `\S`), we negate the bitmap used for encoding units in the range [0-255], but don't have correct handling for encoding units outside that. We've known about this problem since at least October 5 of 2016, but it's ABI breaking to fix :(.

**Command-line test case**
```
d:\Temp2>type repro.cpp
#include <regex>
#include <iostream>

bool test(std::wstring line, std::wstring query)
{
    std::wregex regex(query);
    std::wsmatch res;
    return std::regex_search(line, res, regex);
}

int main()
{
    std::cout << test(L"xxxYxx\x0078xxxZxxx", L"Y\\S*Z") << std::endl; // 0078 is small Latin X
    std::cout << test(L"xxxYxx\xCF87xxxZxxx", L"Y\\S*Z") << std::endl; // CF87 is small Greek Chi 
}

d:\Temp2>cl /EHsc /W4 /WX .\repro.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.27.29009.1 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.

repro.cpp
Microsoft (R) Incremental Linker Version 14.27.29009.1
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:repro.exe
repro.obj

d:\Temp2>.\repro.exe
1
0
```

**Expected behavior**
Given example should match, the correct output is:
```
1
1
```

**STL version**
```
Microsoft Visual Studio Professional 2019 Preview
Version 16.7.0 Preview 3.1
```

**Additional context**
Original repro:
```
#include <regex>
#include <iostream>

int main()
{
	std::wstring line = L"受注、製造、購買オーダのスケジューリングと自動補充生産機能";
	std::wstring query = L"造\\S*オ";
	std::wregex regex(query);
	std::wsmatch res;
	bool found = std::regex_search(line, res, regex);

	std::cout << found << std::endl;
}
```

This item is also tracked on Developer Community as DevCom-984204 and by Microsoft-internal VSO-273702 / [AB#273702](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/273702).

See also #405

[...]

PR #5214, opened 2024-12-30
`<regex>`: Make negated character classes not match characters not included in the negated character class

muellerj2:
Partially addresses #992 again. While there is no easy way to make negated character class escapes (in square-bracket character classes) match members of the class, we can at least make sure that such escapes don't match non-members.

Drive-by changes:
* Correct the limit check (although I think the correction has no practical consequences).
* Change `_Builder::_Add_named_class` and `_Builder::_Add_elts` to use the `char_class_type` of the traits class instead of `_Regex_traits_base::char_class_type` when adding a character class.
* Remove the default argument from `_Builder::_Add_named_class`.


stl/inc/regex
```diff
@@ -2935,12 +2935,12 @@ void _Builder<_FwdIt, _Elem, _RxTraits>::_Add_elts(
 }
 
 template <class _FwdIt, class _Elem, class _RxTraits>
-void _Builder<_FwdIt, _Elem, _RxTraits>::_Add_named_class(typename _Regex_traits_base::char_class_type _Cl,
```
StephanTLavavej, line 2938: No change requested: Ha, this `typename` was previously spurious.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej (approved):

StephanTLavavej:
Thanks for asymptotically fixing this bug! :joy_cat: :lady_beetle: :chart_with_upwards_trend:

PR #5224, opened 2025-01-06
Revert "Maybe Clang _does_ work with `ALTERNATENAME` (#5098)"

StephanTLavavej:
Although #5098 merged successfully, recently the PR system has been persistently failing in an x86 Clang configuration of `std :: tests/Dev11_1086953_call_once_overhaul:39`. This looks like:

```
name: std.tests/Dev11_1086953_call_once_overhaul
output: OSError(22, 'Operation did not complete successfully because the file contains a virus or potentially unwanted software', None, 225, None) at C:\a\_work\1\s\tests\utils\stl\test\format.py:231
```

This is a false positive (which appears to only repro on our Server 2025 VMs; it doesn't repro on my local Win11 machine), which I suppose we could report to the relevant team at great effort. We could also just skip building this particular test for Clang x86.

However, reverting #5098 entirely appears to work robustly. Since I wasn't eager to mess with this in the first place, I think a revert is appropriate.

This preserves a comment improvement and adjusts other comments.


tests/std/tests/GH_002030_asan_annotate_string/env.lst
```diff
@@ -1,40 +1,57 @@
 # Copyright (c) Microsoft Corporation.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-# This test matrix is the usual test matrix, with all currently unsupported options removed, with
-# some additional /Zc:char8_t coverage, crossed with the ASan flags.
+# This test matrix is the usual test matrix, with all currently unsupported options removed, crossed with the ASan flags.
```
CaseyCarter, line 4: Let's not revert this description improvement.
```suggestion
# This test matrix is the usual test matrix, with all currently unsupported options removed, with
# some additional /Zc:char8_t coverage, crossed with the ASan flags.
```

CaseyCarter (changes requested):

CaseyCarter (approved):

StephanTLavavej:

PR #5226, opened 2025-01-08
VCRuntime fixed default non-throwing `operator new[]` to call throwing `operator new[]`

CaseyCarter:
Internally, the default non-throwing array `operator new` now correctly delegates to throwing array `operator new`. Let's mark the corresponding libcxx test `SKIPPED` instead of `FAIL` so the test suite passes both internally and externally. We can remove the skip altogether once the VCRuntime change ships in preview.

This mirrors MSVC-PR-601619 which fixes DevCom-10373274 "vcruntime nothrow array operator new falls back on the wrong function".

### DIFF
--- a/tests/libcxx/expected_results.txt
+++ b/tests/libcxx/expected_results.txt
@@ -438,8 +438,9 @@ std/input.output/syncstream/syncbuf/syncstream.syncbuf.members/emit.pass.cpp FAI
 # *** VCRUNTIME BUGS ***
 # DevCom-10373274 VSO-1824997 "vcruntime nothrow array operator new falls back on the wrong function"
 # This passes for the :1 (ASan) configuration, surprisingly.
-std/language.support/support.dynamic/new.delete/new.delete.array/new.size_nothrow.replace.indirect.pass.cpp:0 FAIL
-std/language.support/support.dynamic/new.delete/new.delete.array/new.size_nothrow.replace.indirect.pass.cpp:2 FAIL
+# TRANSITION, vs17.14p1
+std/language.support/support.dynamic/new.delete/new.delete.array/new.size_nothrow.replace.indirect.pass.cpp:0 SKIPPED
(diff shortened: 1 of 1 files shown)


tests/libcxx/expected_results.txt
```diff
@@ -438,8 +438,9 @@ std/input.output/syncstream/syncbuf/syncstream.syncbuf.members/emit.pass.cpp FAI
 # *** VCRUNTIME BUGS ***
 # DevCom-10373274 VSO-1824997 "vcruntime nothrow array operator new falls back on the wrong function"
 # This passes for the :1 (ASan) configuration, surprisingly.
-std/language.support/support.dynamic/new.delete/new.delete.array/new.size_nothrow.replace.indirect.pass.cpp:0 FAIL
-std/language.support/support.dynamic/new.delete/new.delete.array/new.size_nothrow.replace.indirect.pass.cpp:2 FAIL
+# TRANSITION, vs17.14p1
+std/language.support/support.dynamic/new.delete/new.delete.array/new.size_nothrow.replace.indirect.pass.cpp:0 SKIPPED
+std/language.support/support.dynamic/new.delete/new.delete.array/new.size_nothrow.replace.indirect.pass.cpp:2 SKIPPED
```
StephanTLavavej, line 443: This is fine, but I question whether it's necessary. The internal harness doesn't deal with expected fails or individual configurations - if any configuration of a test is mentioned, the entire test is skipped. (That's why we can use `:9 SKIPPED` at the end for internal-only skips.)

If you took no action here, then I believe that the internal test harness would pass with your VCRuntime fix, and the GitHub test harness would continue to observe expected FAILs until we pick up the 17.14 Preview 1 toolset update, at which time we'd remove the FAILs.

StephanTLavavej (changes requested):

CaseyCarter:
This change is unnecessary per https://github.com/microsoft/STL/pull/5226#discussion_r1907882704.

PR #5230, opened 2025-01-11
Changed the implementation to simpler code

Kratosix:
Technically the same thing, but it makes the compiler's work somehow easy


StephanTLavavej:
Thanks for opening a PR, but this change is not an improvement. It greatly obfuscates how the character's case is being transformed from uppercase to lowercase (the existing code merely relies on the knowledge that a-z and A-Z are contiguous in ASCII). It might make `stl/src/_tolower.cpp` 3 nanoseconds faster to build, which is a cost that's paid by the MSVC build lab. (Even in a header, this would not be a significant compiler throughput optimization.)

As this doesn't work towards any of our [goals](https://github.com/microsoft/STL#goals), I'm going to close this PR.

PR #5231, opened 2025-01-12
Categorize some failing libcxx tests

muellerj2:
... which are mostly related to `<regex>`.

### DIFF
--- a/tests/libcxx/expected_results.txt
+++ b/tests/libcxx/expected_results.txt
@@ -36,11 +36,19 @@ std/numerics/rand/rand.dist/rand.dist.uni/rand.dist.uni.real/param_ctor.pass.cpp
 # LLVM-113609: [libc++][test] Non-rebindable test_alloc in string.capacity/deallocate_size.pass.cpp
 std/strings/basic.string/string.capacity/deallocate_size.pass.cpp FAIL
 
+# LLVM-116054: [libc++] vprint_[non]unicode(ostream &, string_view, Args&&...) should not pad the output
+std/input.output/iostream.format/output.streams/ostream.formatted/ostream.formatted.print/print.pass.cpp FAIL
+std/input.output/iostream.format/output.streams/ostream.formatted/ostream.formatted.print/vprint_nonunicode.pass.cpp FAIL
+std/input.output/iostream.format/output.streams/ostream.formatted/ostream.formatted.print/vprint_unicode.pass.cpp FAIL
+
 # LLVM-119174: [libcxx][test] Silence nodiscard warnings for std::expected
 std/utilities/expected/expected.expected/monadic/and_then.pass.cpp FAIL
 std/utilities/expected/expected.expected/monadic/or_else.pass.cpp FAIL
 std/utilities/expected/expected.expected/monadic/transform.pass.cpp FAIL
 std/utilities/expected/expected.expected/monadic/transform_error.pass.cpp FAIL
+
+# LLVM-122638: [libc++][test] re.regex.construct/bad_backref.cpp assumes non-standard extension to extended regular expressions
+std/re/re.regex/re.regex.construct/bad_backref.pass.cpp FAIL
 
 # Non-Standard regex behavior.
 # "It seems likely that the test is still non-conforming due to how libc++ handles the 'w' character class."
@@ -151,6 +159,9 @@ std/language.support/support.limits/support.limits.general/ranges.version.compil
 # libc++ is missing various <format> DRs
 std/language.support/support.limits/support.limits.general/format.version.compile.pass.cpp FAIL
 
+# libc++ doesn't implement LWG-2584 (LLVM-99976)
+std/re/re.regex/re.regex.construct/bad_escape.pass.cpp FAIL
+
 # libc++ doesn't implement LWG-3670
 std/ranges/range.factories/range.iota.view/iterator/member_typedefs.compile.pass.cpp FAIL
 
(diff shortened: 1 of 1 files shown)


tests/libcxx/expected_results.txt
```diff
@@ -36,12 +36,20 @@ std/numerics/rand/rand.dist/rand.dist.uni/rand.dist.uni.real/param_ctor.pass.cpp
...
+
 # LLVM-119174: [libcxx][test] Silence nodiscard warnings for std::expected
 std/utilities/expected/expected.expected/monadic/and_then.pass.cpp FAIL
 std/utilities/expected/expected.expected/monadic/or_else.pass.cpp FAIL
 std/utilities/expected/expected.expected/monadic/transform.pass.cpp FAIL
 std/utilities/expected/expected.expected/monadic/transform_error.pass.cpp FAIL
 
+# LLVM-122638: [libc++][test] re.regex.construct/bad_backref.cpp assumes non-standard extension to extended regular expressions
+std/re/re.regex/re.regex.construct/bad_backref.pass.cpp FAIL
+
```
muellerj2, line 52: This assumes that #5167 is merged.

tests/libcxx/expected_results.txt
```diff
@@ -552,6 +565,12 @@ std/utilities/format/format.formatter/format.formatter.locking/enable_nonlocking
 std/containers/sequences/vector.bool/shrink_to_fit.pass.cpp FAIL
 std/containers/sequences/vector/vector.capacity/shrink_to_fit.pass.cpp FAIL
 
+# We disagree on the syntax flags set by the default basic_regex constructor: we set 0, libc++ sets ECMAScript.
+# Relates to LWG-3604.
+std/re/re.regex/re.regex.construct/default.pass.cpp FAIL
+std/re/re.regex/re.regex.nonmemb/re.regex.nmswap/swap.pass.cpp FAIL
+std/re/re.regex/re.regex.swap/swap.pass.cpp FAIL
+
```
muellerj2, line 573: I'm not really sure where to put this one. The underlying issue is that [\[re.regex.construct\]/1](http://eel.is/c++draft/re.regex.construct#1) does not specify the value returned by `flags()` as a postcondition on the default constructor, so MSVC STL returns 0 while libc++ returns `regex_constants::ECMAScript` (which is probably informed by [\[re.synopt\]/1](http://eel.is/c++draft/re.synopt#1)).

[LWG-3604](https://cplusplus.github.io/LWG/issue3604) is related, but discusses the postcondition on the flags imposed by all the other constructors and argues that they are too strict. Here, the problem is kind of the opposite: The standard doesn't state a postcondition on the flags for the default constructor at all.

tests/libcxx/expected_results.txt
```diff
@@ -36,12 +36,20 @@ std/numerics/rand/rand.dist/rand.dist.uni/rand.dist.uni.real/param_ctor.pass.cpp
...
+std/input.output/iostream.format/output.streams/ostream.formatted/ostream.formatted.print/vprint_nonunicode.pass.cpp FAIL
+std/input.output/iostream.format/output.streams/ostream.formatted/ostream.formatted.print/vprint_unicode.pass.cpp FAIL
+
 # LLVM-119174: [libcxx][test] Silence nodiscard warnings for std::expected
 std/utilities/expected/expected.expected/monadic/and_then.pass.cpp FAIL
 std/utilities/expected/expected.expected/monadic/or_else.pass.cpp FAIL
 std/utilities/expected/expected.expected/monadic/transform.pass.cpp FAIL
 std/utilities/expected/expected.expected/monadic/transform_error.pass.cpp FAIL
 
+# LLVM-122638: [libc++][test] re.regex.construct/bad_backref.cpp assumes non-standard extension to extended regular expressions
```
StephanTLavavej, line 50: No change requested: The filename is bad_backref.pass.cpp but I don't care enough to ask for the bug title to be changed and the PR testing to be reset.

tests/libcxx/expected_results.txt
StephanTLavavej: LWG-3604 does discuss this problem at the end:

> Also, the constructors say "Postconditions: flags() returns f." This prevents an implementation from storing f|ECMAScript in flags() if no grammar element is present in f. This seems like an unnecessary restriction, and forces implementations to do extra work to check if the ECMAScript grammar is in use. Arguably, it would even be better to require implementations to set ECMAScript in flags() if no grammar element was set in the flags passed to the constructor. This problem was introduced by LWG [2330](https://cplusplus.github.io/LWG/lwg-defects.html#2330)[(i)](https://cplusplus.github.io/LWG/issue2330).

I agree that there's no obvious place for this one - "LIKELY BOGUS TESTS" seems as good as anything, since if the Standard is vague on what happens, the test shouldn't be expecting something specific. Good enough for me, and a vast improvement over the status quo of "not analyzed".

StephanTLavavej (approved):

StephanTLavavej:
Thanks, I love seeing progress on categorizing test failures!! :heart_eyes_cat:

tests/libcxx/expected_results.txt
muellerj2: I fixed the bug title in the LLVM repository at least.

StephanTLavavej: Thanks! I pushed a commit to update the PR since otherwise it would bother me :joy_cat:

StephanTLavavej (approved):

StephanTLavavej:

PR #5237, opened 2025-01-15
`<thread>`: Avoid overflow in `_To_absolute_time()`

### LINKED ISSUE #5234: <thread>: std::this_thread::sleep_for overflow with high precision durations
# Describe the bug
Using integer durations with greater than nanosecond precision causes `std::this_thread::sleep_for` to not sleep.

See https://godbolt.org/z/s37b31MsM and https://stackoverflow.com/questions/79355000/

# Command-line test case
```
C:\Temp>type repro.cpp
#include <chrono>
#include <print>
#include <thread>

using picoseconds = std::chrono::duration<int64_t, std::ratio<1ll, 1'000'000'000'000ll>>;

int main()
{
    auto t1 = std::chrono::steady_clock::now();
    std::this_thread::sleep_for(picoseconds(1'000'000'000'000ll));
    auto t2 = std::chrono::steady_clock::now();
    std::println("{:%S}", t2 - t1);
}

C:\Temp>cl /EHsc /W4 /WX /std:c++latest .\repro.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.36.32522 for x64

repro.cpp
Microsoft (R) Incremental Linker Version 14.36.32522.0

/out:repro.exe
repro.obj

C:\Temp>.\repro.exe
00.000057300
```

# Expected behavior
The program should sleep for approximately a second but instead returns instantly.

# STL version
MSVC version 19.41.33923

# Additional context
A possible fix for this issue is to change:
```
constexpr auto _Forever = (chrono::steady_clock::time_point::max)();
```
in `_To_absolute_time` to:
```
constexpr auto _Forever = (std::chrono::time_point<std::chrono::steady_clock, decltype(_Abs_time)::duration>::max)();
```

StephanTLavavej:
Fixes #5234.

We were carefully forming `decltype(_Now + _Rel_time) _Abs_time`, the common type of the given `_Rel_time` (which could be coarse or fine) and `chrono::steady_clock::now()` (which is nanoseconds).

In this bug scenario, `_Rel_time` and therefore the common type are picoseconds.

The problem was that we weren't careful about our `_Forever` constant. We were using `(chrono::steady_clock::time_point::max)()`, but then converting it to the common type (implicitly in the subtraction `_Forever - _Rel_time`, and in the assignment `_Abs_time = _Forever`). When the common type is finer-grained, this attempts to multiply the stored value, but it's already the maximum, so we overflow.

The fix is to use the common type for `_Forever`. This stores the maximum rep with the proper period of the common type, so we won't attempt to multiply it further. (When `_Rel_time` is coarse, the common type is nanoseconds, so there's no change. When `_Rel_time` is fine, this means that `_Forever` is effectively "earlier", because `INT64_MAX` picoseconds from the epoch happens earlier than `INT64_MAX` nanoseconds. This is exactly what we want, since we're performing our overflow check and clamping with the common type that we're about to return.)

### DIFF
--- a/stl/inc/thread
+++ b/stl/inc/thread
@@ -169,7 +169,7 @@ _NODISCARD auto _To_absolute_time(const chrono::duration<_Rep, _Period>& _Rel_ti
     const auto _Now                      = chrono::steady_clock::now();
     decltype(_Now + _Rel_time) _Abs_time = _Now; // return common type
     if (_Rel_time > _Zero) {
-        constexpr auto _Forever = (chrono::steady_clock::time_point::max)();
+        constexpr auto _Forever = (decltype(_Abs_time)::max)();
         if (_Abs_time < _Forever - _Rel_time) {
             _Abs_time += _Rel_time;
         } else {


CaseyCarter (approved):

StephanTLavavej:

zacklj89 (approved):

Issue #5236, opened 2025-01-15
`<locale>`: `std::collate<wchar_t>` does not respect collation order when compiled with `/MD(d) /Zc:wchar_t-`


muellerj2:
`std::collate<wchar_t>::compare()` and `std::collate<wchar_t>::transform()` do not actually perform comparisons or transformations with respect to the collation order when the program is compiled with flags `/MD(d) /Zc:wchar_t-`.

## Test case

```cpp
#include <iostream>
#include <locale>

using namespace std;

int main() {
    const wchar_t Ue = L'\u00DC'; // U+00DC LATIN CAPITAL LETTER U WITH DIARESIS
    const wchar_t U = L'U';
    const wchar_t V = L'V';

    // German phonebook order: "U+00DC" collates like string "Ue"
    locale loc("de-DE_phoneb");
    auto &coll = use_facet<collate<wchar_t>>(loc);

    cout << "transform(Ue) != Ue: "
            << (coll.transform(&Ue, &Ue + 1) != L"\u00DC") << '\n';

    cout << "transform(U) <= transform(Ue) <= transform(V): "
            << (coll.transform(&U, &U + 1) <= coll.transform(&Ue, &Ue + 1) &&
                coll.transform(&Ue, &Ue + 1) <= coll.transform(&V, &V + 1)); 
}
```
https://godbolt.org/z/Kx8z43e71

Compiled with `/MD(d) /Zc:wchar_t-`, this prints:
```
transform(Ue) != Ue: 0
transform(U) <= transform(Ue) <= transform(V): 0
```

## Expected result
Because "Ü" should collate like "Ue" in the German phonebook collation order, this should print (as it does with `/MD /Zc:wchar_t` or `/MT /Zc:wchar_t-`):

```
transform(Ue) != Ue: 1
transform(U) <= transform(Ue) <= transform(V): 1
```

PR #5238, opened 2025-01-15
`<regex>`: Implement collating ranges

### LINKED ISSUE #5204: `<regex>`: Collating ranges are broken

muellerj2:
Resolves #5204. Follow-up to #5164 and #5209.

When compiled under `/MD` or `/MDd`, the new test cases specific to collating ranges are skipped:
* if compiled with `/Zc:wchar_t-` because the test will fail due to #5236.
* if there is a mismatch with `_ITERATOR_DEBUG_LEVEL` because the stack will be corrupted when `std::collate<wchar_t>::do_transform()` is called in the linked DLL.

Drive-by changes:
* Mark `_Parser::_Error()` as `[[noreturn]]`.
* In `_Matcher::_Skip()` and `_Matcher::_Do_class()`, `_STD` qualify calls to some free functions and clean up the control flow that translates the current character and converts it to an unsigned value.


stl/inc/regex
```diff
@@ -2917,7 +2917,8 @@ void _Builder<_FwdIt, _Elem, _RxTraits>::_Add_range2(const _Elem _Arg0, const _E
 
         _Node->_Small->_Mark(_Ex0);
     }
-    if (_Ex1 >= _Ex0) {
+
+    if (_Flags & regex_constants::collate || _Ex1 >= _Ex0) {
```
muellerj2, line 2921: The control flow is subtle around here but the function ends up doing the right thing: When the `collate` flag is set, the `_Builder` constructor sets the member variables such that `_Get_tmax()` and `_Get_bmax()` return `0`. This means that the preceding loop is skipped and the inner if condition is false (even when `_Ex1 - _Ex0` in the following line underflows). So in the end, the code just adds the bounds to `_Node->_Ranges`.

tests/std/tests/GH_005204_regex_collating_ranges/test.cpp
```diff
@@ -0,0 +1,333 @@
+// Copyright (c) Microsoft Corporation.
```
CaseyCarter, line 1: This new test needs to be added to `tests/std/tests.lst` so the testing infra picks it up.

tests/std/tests/GH_005204_regex_collating_ranges/test.cpp
```diff
@@ -0,0 +1,333 @@
...
+        }
+    }
+
+    {
+        test_wregex_locale collate_A_to_z_regex(&g_regexTester, L"[A-z]", "de-DE", regex_constants::collate);
+        collate_A_to_z_regex.should_search_fail(L"a");
+        for (wchar_t ascii_upper = L'A'; ascii_upper < 'Z'; ++ascii_upper) {
+            collate_A_to_z_regex.should_search_match(wstring(1, ascii_upper), wstring(1, ascii_upper));
+        }
+        for (wchar_t ascii_lower = L'b'; ascii_lower <= 'z'; ++ascii_lower) {
```
StephanTLavavej, line 311: The comparisons should be against L character literals, although it doesn't make a real difference.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! This makes sense and looks good to me. The comprehensive test coverage is excellent. :heart_eyes_cat:

I pushed a trivial merge with `main` followed by a bunch of nitpicks. The most significant changes were to `regex_with_locale_should_throw()`, to actually use `syntax` and to properly record failure when an exception is thrown with an unexpected code.

StephanTLavavej:

StephanTLavavej:
Thanks for sorting out this bug! :joy_cat: :lady_beetle: :magic_wand:

PR #5240, opened 2025-01-16
Clarify ASan test running instructions

davidmrdavid:
This is an exceedingly _minor_ contribution, so feel free to disregard it. 

When running the ASan'ized tests, I could have used a reminder that the python test script resides within the `.\out\<arch>` folder. The preceeding `cmake` commands should run in the repo's root, so it's easy to forget changing your directory when running the `python <pathToTestScript>/stl-lit.py` command.

This PR adds a reminder that the call to `stl-lit.py` should be performed from `out\x64\` , to make this clearer.

Please feel free to dismiss this, it's a subjective contribution. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,16 +412,19 @@ set PATH=C:\STL\out\x64\out\bin\amd64;%PATH%
 ## Running Tests With Address Sanitizer (ASan)
 
 You don't need any extra steps to run with test code and the code in STL headers instrumented with [ASan][].
-The test matrices include both ASan and non-ASan configurations if you don't pass `-Dtags=ASAN` or `-Dnotags=ASAN` to exclude one or the other.
+The test matrices include both ASan and non-ASan configurations if you don't pass `-Dtags=ASAN` or `-Dnotags=ASAN`
+to exclude one or the other.
 
-However, to instrument the separately-compiled code (the DLL, the satellites, the [Import Library][] - everything that's in `/stl/src`), you need to build the STL with ASan. Change the build steps to add `-DSTL_ASAN_BUILD=ON`:
+However, to instrument the separately-compiled code (the DLL, the satellites, the [Import Library][] - everything that's
+in `/stl/src`), you need to build the STL with ASan. Change the build steps to add `-DSTL_ASAN_BUILD=ON`:
 
 ```
 cmake --preset x64 -DSTL_ASAN_BUILD=ON
 cmake --build --preset x64
 ```
 
-ASan-instrumented STL binaries require that the executable be instrumented as well, so you'll have to skip the non-ASan configurations by passing `-Dtags=ASAN` to `stl-lit.py`:
+ASan-instrumented STL binaries require that the executable be instrumented as well, so you'll have to skip the non-ASan
+configurations by passing `-Dtags=ASAN` to `stl-lit.py`:
 
 ```
 python out\x64\tests\utils\stl-lit\stl-lit.py tests\std\tests\VSO_0000000_vector_algorithms -Dtags=ASAN -v
--- a/README.md
+++ b/README.md
@@ -424,7 +424,7 @@ cmake --build --preset x64
 ASan-instrumented STL binaries require that the executable be instrumented as well, so you'll have to skip the non-ASan configurations by passing `-Dtags=ASAN` to `stl-lit.py`:
 
 ```
-C:\STL\out\x64> python tests\utils\stl-lit\stl-lit.py ..\..\tests\std\tests\VSO_0000000_vector_algorithms -Dtags=ASAN -v
+python out\x64\tests\utils\stl-lit\stl-lit.py tests\std\tests\VSO_0000000_vector_algorithms -Dtags=ASAN -v
 ```
 
 # Benchmarking
--- a/README.md
+++ b/README.md
@@ -412,7 +412,7 @@ set PATH=C:\STL\out\x64\out\bin\amd64;%PATH%
 ## Running Tests With Address Sanitizer (ASan)
 
 You don't need any extra steps to run with test code and the code in STL headers instrumented with [ASan][].
-The test matrices include both ASan and non-ASan configurations.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -414,19 +414,17 @@ set PATH=C:\STL\out\x64\out\bin\amd64;%PATH%
 You don't need any extra steps to run with test code and the code in STL headers instrumented with [ASan][].
 The test matrices include both ASan and non-ASan configurations.
 
-However, to instrument the separately-compiled code (the DLL, the satellites, the [Import Library][] - everything that's
-in `/stl/src`), you need to build the STL with ASan. Change the build steps to add `-DSTL_ASAN_BUILD=ON`:
+However, to instrument the separately-compiled code (the DLL, the satellites, the [Import Library][] - everything that's in `/stl/src`), you need to build the STL with ASan. Change the build steps to add `-DSTL_ASAN_BUILD=ON`:
```
AlexGuteniev, line 417: The lone wrapping was deliberate and should be restored to make code validation pass

README.md
```diff
@@ -414,19 +414,17 @@ set PATH=C:\STL\out\x64\out\bin\amd64;%PATH%
 You don't need any extra steps to run with test code and the code in STL headers instrumented with [ASan][].
 The test matrices include both ASan and non-ASan configurations.
```
AlexGuteniev, line 415: Maybe add "unless `-Dnotags=ASAN` is specified" somewhere here, while we're at it.

README.md
CaseyCarter: We enforce a 120-column maximum line length on most files in the STL. You can build the `validate` target to be yelled at for various violations of rules we enforce on file contents.

README.md
CaseyCarter:
```suggestion
The test matrices include both ASan and non-ASan configurations if you don't pass `-Dtags=ASAN` or `-Dnotags=ASAN` to exclude one or the other.
```
(This probably needs wrapping.)

README.md
```diff
@@ -414,19 +414,17 @@ set PATH=C:\STL\out\x64\out\bin\amd64;%PATH%
...
+However, to instrument the separately-compiled code (the DLL, the satellites, the [Import Library][] - everything that's in `/stl/src`), you need to build the STL with ASan. Change the build steps to add `-DSTL_ASAN_BUILD=ON`:
 
 ```
 cmake --preset x64 -DSTL_ASAN_BUILD=ON
 cmake --build --preset x64
 ```
 
-ASan-instrumented STL binaries require that the executable be instrumented as well, so you'll have to skip the non-ASan
-configurations by passing `-Dtags=ASAN` to `stl-lit.py`:
+ASan-instrumented STL binaries require that the executable be instrumented as well, so you'll have to skip the non-ASan configurations by passing `-Dtags=ASAN` to `stl-lit.py`:
```
CaseyCarter, line 424: This line looks overlong; you should probably revert this bit.

README.md
```diff
@@ -414,19 +414,17 @@ set PATH=C:\STL\out\x64\out\bin\amd64;%PATH%
...
 
 ```
-python tests\utils\stl-lit\stl-lit.py ..\..\tests\std\tests\VSO_0000000_vector_algorithms -Dtags=ASAN -v
+C:\STL\out\x64> python tests\utils\stl-lit\stl-lit.py ..\..\tests\std\tests\VSO_0000000_vector_algorithms -Dtags=ASAN -v
```
CaseyCarter, line 427: The inconsistency with lines 420-421 seems likely to confuse folks. Maybe we should assume this is run from the top of the repo like 420-421 are:
```suggestion
python out\x64\tests\utils\stl-lit\stl-lit.py tests\std\tests\VSO_0000000_vector_algorithms -Dtags=ASAN -v
```

CaseyCarter (changes requested):

README.md
davidmrdavid: Good point, I think that's a better alternative

README.md
davidmrdavid: Oh, interesting. Good to know - I've undone these changes.

README.md
davidmrdavid: Reverted!

CaseyCarter (approved):

davidmrdavid:
naive question: should I go ahead and squash-and-merge is there a procedure to this?

CaseyCarter:
> naive question: should I go ahead and squash-and-merge is there a procedure to this?

We hand-port all changes into the MSVC repo and merge in both places "simultaneously" to minimize divergence between the internal and external copies of the STL. Yes, this is a bit of a PITA. We generally merge a batch of PRs every few days to amortize the annoyance of the port and the cost of the internal CI. The process is documented in https://github.com/microsoft/STL/wiki/Checklist-For-Merging-A-Pull-Request.

TLDR: No, we'll squash and merge it later. Your work is done here once the PR reaches the "Ready to Merge" state.

StephanTLavavej:
> I want to see if STL is bothered by the ASan instructions using the top of the repo as working directory instead of using out/x64 like the non-ASan instructions do.

Yeah, this really bothers me. It's unnecessarily different from the normal test instructions which begin by saying:

> These examples assume that your current directory is `C:\Dev\STL\out\x64`.

I'll push a change.

StephanTLavavej (approved):

CaseyCarter (approved):

StephanTLavavej:
Change pushed (and PR title and description updated).

Sorry to be so nitpicky, but I believe that following a consistent pattern is important. Running the tests from `out\<arch>` has a couple of benefits: (1) if you run a different arch (commonly x86), you only need to alter your `pushd` command, and (2) if you create any log or temp files while investigating tests, they'll be safely contained within the `out` tree instead of polluting the repo root.

davidmrdavid:
no problem at all, it's good for me to absorb these documentation _values_ for future PRs, so I appreciate them being laid out explicitly. Thanks!

StephanTLavavej:

StephanTLavavej:
Thanks for the README clarification! :scroll: :green_heart: :tada:

PR #5241, opened 2025-01-16
Add the ability to opt-out of ASan container annotations on a per-allocator basis

davidmrdavid:
**Context:**

The STL "annotates" the `vector` and `basic_string` containers so that ASan will report `container-overflow` whenever allocated _but un-initialized_ data is accessed.

A simple repro is the [`container overflow`](https://learn.microsoft.com/en-us/cpp/sanitizers/error-container-overflow?view=msvc-170#example-access-reserved-memory-in-a-stdvector) error fired in this sample program (assuming it's `/fsanitize=address`'ed):

```C++
// Compile with: cl /EHsc /fsanitize=address /Zi
#include <vector>

int main() {   
    std::vector<int> v(10);
    v.reserve(20); // we've allocated 20 entries, but only initialized only 10

    // Accessing the 10th entry (0-indexed, naturally) triggers an AV 
    v[10] = 1;

}
```

This is sensible behavior in most cases. 

One case where it does not bode well is when an **arena allocator** is used as the **custom allocator** of the container. Arena allocators often tamper with the entire allocated memory at once (e.g. they commonly deallocate their entire 'arena' at once) which would trigger ASan AVs when the capacity of the container exceeds it's size.

We encountered one such bug in the `msvc` front end.

**This PR:**

This PR introduces the ability for custom allocators to opt-out of `vector` and `basic_string`'s ASan annotations. This is controlled by the newly introduced template variable: `_Disable_ASan_container_annotations_for_allocator<...some allocator type...>`.

**Testing:**
* For the _new_ annotation opt-out feature: a simple test case was added for `basic_string` and `vector` respectively
* For the pre-existing annotation feature: a simple 'death test' test case was added for the aforementioned containers respectively as well. In the case of `basic_string`, this replaces the recently added `test_gh_5251` test (from https://github.com/microsoft/STL/pull/5252) to avoid repetition.

### DIFF
--- a/stl/inc/vector
+++ b/stl/inc/vector
@@ -483,52 +483,50 @@ private:
         _STL_INTERNAL_CHECK(_Old_last_ != nullptr);
         _STL_INTERNAL_CHECK(_New_last_ != nullptr);
 
+        if constexpr (!_Disable_ASan_container_annotations_for_allocator<allocator_type>) {
 #if _HAS_CXX20
-        if (_STD is_constant_evaluated()) {
-            return;
-        }
+            if (_STD is_constant_evaluated()) {
+                return;
+            }
 #endif // _HAS_CXX20
 
-        if (!_Asan_vector_should_annotate) {
-            return;
-        }
-
-        if constexpr (_Disable_ASan_container_annotations_for_allocator<allocator_type>) {
-            return;
-        }
-
-        const void* const _First    = _STD _Unfancy(_First_);
-        const void* const _End      = _STD _Unfancy(_End_);
-        const void* const _Old_last = _STD _Unfancy(_Old_last_);
-        const void* const _New_last = _STD _Unfancy(_New_last_);
-        if constexpr ((_Container_allocation_minimum_asan_alignment<vector>) >= _Asan_granularity) {
-            // old state:
-            //   [_First, _Old_last) valid
-            //   [_Old_last, _End) poison
-            // new state:
-            //   [_First, _New_last) valid
-            //   [_New_last, asan_aligned_after(_End)) poison
-            _CSTD __sanitizer_annotate_contiguous_container(
-                _First, _STD _Get_asan_aligned_after(_End), _Old_last, _New_last);
-        } else {
-            const auto _Aligned = _STD _Get_asan_aligned_first_end(_First, _End);
-            if (_Aligned._First == _Aligned._End) {
-                // The buffer does not end at least one shadow memory section; nothing to do.
+            if (!_Asan_vector_should_annotate) {
                 return;
             }
 
-            const void* const _Old_fixed = _Aligned._Clamp_to_end(_Old_last);
-            const void* const _New_fixed = _Aligned._Clamp_to_end(_New_last);
+            const void* const _First    = _STD _Unfancy(_First_);
+            const void* const _End      = _STD _Unfancy(_End_);
+            const void* const _Old_last = _STD _Unfancy(_Old_last_);
+            const void* const _New_last = _STD _Unfancy(_New_last_);
+            if constexpr ((_Container_allocation_minimum_asan_alignment<vector>) >= _Asan_granularity) {
+                // old state:
+                //   [_First, _Old_last) valid
+                //   [_Old_last, _End) poison
+                // new state:
+                //   [_First, _New_last) valid
+                //   [_New_last, asan_aligned_after(_End)) poison
+                _CSTD __sanitizer_annotate_contiguous_container(
+                    _First, _STD _Get_asan_aligned_after(_End), _Old_last, _New_last);
+            } else {
+                const auto _Aligned = _STD _Get_asan_aligned_first_end(_First, _End);
+                if (_Aligned._First == _Aligned._End) {
+                    // The buffer does not end at least one shadow memory section; nothing to do.
+                    return;
+                }
 
-            // old state:
-            //   [_Aligned._First, _Old_fixed) valid
-            //   [_Old_fixed, _Aligned._End) poison
-            //   [_Aligned._End, _End) valid
-            // new state:
-            //   [_Aligned._First, _New_fixed) valid
-            //   [_New_fixed, _Aligned._End) poison
-            //   [_Aligned._End, _End) valid
-            _CSTD __sanitizer_annotate_contiguous_container(_Aligned._First, _Aligned._End, _Old_fixed, _New_fixed);
+                const void* const _Old_fixed = _Aligned._Clamp_to_end(_Old_last);
+                const void* const _New_fixed = _Aligned._Clamp_to_end(_New_last);
+
+                // old state:
+                //   [_Aligned._First, _Old_fixed) valid
+                //   [_Old_fixed, _Aligned._End) poison
+                //   [_Aligned._End, _End) valid
+                // new state:
+                //   [_Aligned._First, _New_fixed) valid
+                //   [_New_fixed, _Aligned._End) poison
+                //   [_Aligned._End, _End) valid
+                _CSTD __sanitizer_annotate_contiguous_container(_Aligned._First, _Aligned._End, _Old_fixed, _New_fixed);
+            }
         }
     }
 
--- a/stl/inc/vector
+++ b/stl/inc/vector
@@ -493,7 +493,7 @@ private:
             return;
         }
 
-        if (!is_ASan_enabled_for_allocator_v<allocator_type>) {
+        if constexpr (!_Is_ASan_enabled_for_allocator<allocator_type>) {
             return;
         }
 
--- a/stl/inc/vector
+++ b/stl/inc/vector
@@ -493,6 +493,10 @@ private:
             return;
         }
 
+        if (!is_ASan_enabled_for_allocator_v<allocator_type>) {
+            return;
+        }
+
         const void* const _First    = _STD _Unfancy(_First_);
         const void* const _End      = _STD _Unfancy(_End_);
         const void* const _Old_last = _STD _Unfancy(_Old_last_);
--- a/stl/inc/vector
+++ b/stl/inc/vector
@@ -493,7 +493,7 @@ private:
             return;
         }
 
-        if constexpr (!_Is_ASan_enabled_for_allocator<allocator_type>) {
+        if constexpr (_Disable_ASan_container_annotations_for_allocator<allocator_type>) {
             return;
         }
 
--- a/stl/inc/xmemory
+++ b/stl/inc/xmemory
@@ -805,7 +805,7 @@ struct _Simple_types { // wraps types from allocators with simple addressing for
 _INLINE_VAR constexpr size_t _Asan_granularity      = 8;
 _INLINE_VAR constexpr size_t _Asan_granularity_mask = _Asan_granularity - 1;
 
-// Represents whether ASan container annotations are enabled a given allocator.
+// Controls whether ASan `container-overflow` errors are reported for this allocator.
 template <class>
 constexpr bool _Disable_ASan_container_annotations_for_allocator = false;
 
--- a/stl/inc/xmemory
+++ b/stl/inc/xmemory
@@ -805,6 +805,17 @@ struct _Simple_types { // wraps types from allocators with simple addressing for
 _INLINE_VAR constexpr size_t _Asan_granularity      = 8;
 _INLINE_VAR constexpr size_t _Asan_granularity_mask = _Asan_granularity - 1;
 
+// Represents whether ASan is enabled for allocator type 'T'.
+// By default, ASan is enabled.
+template<typename T>
+struct is_ASan_enabled_for_allocator {
+    static constexpr bool value = true;
+};
+
+// Convenience helper to determine if ASan is enabled for a given allocator type 'T'
+template <typename T>
+constexpr bool is_ASan_enabled_for_allocator_v = is_ASan_enabled_for_allocator<T>::value;
+
 struct _Asan_aligned_pointers {
     const void* _First;
     const void* _End;
--- a/stl/inc/xmemory
+++ b/stl/inc/xmemory
@@ -807,7 +807,7 @@ _INLINE_VAR constexpr size_t _Asan_granularity_mask = _Asan_granularity - 1;
 
 // Represents whether ASan container annotations are enabled a given allocator.
 template <class>
-constexpr bool _Is_ASan_enabled_for_allocator = true;
+constexpr bool _Disable_ASan_container_annotations_for_allocator = false;
 
 struct _Asan_aligned_pointers {
     const void* _First;
--- a/stl/inc/xmemory
+++ b/stl/inc/xmemory
@@ -805,7 +805,7 @@ struct _Simple_types { // wraps types from allocators with simple addressing for
 _INLINE_VAR constexpr size_t _Asan_granularity      = 8;
 _INLINE_VAR constexpr size_t _Asan_granularity_mask = _Asan_granularity - 1;
 
-// Represents whether ASan container annotations are enabled for allocator type 'T'.
+// Represents whether ASan container annotations are enabled a given allocator.
 template <class>
 constexpr bool _Is_ASan_enabled_for_allocator = true;
 
--- a/stl/inc/xmemory
+++ b/stl/inc/xmemory
@@ -806,15 +806,8 @@ _INLINE_VAR constexpr size_t _Asan_granularity      = 8;
 _INLINE_VAR constexpr size_t _Asan_granularity_mask = _Asan_granularity - 1;
 
 // Represents whether ASan container annotations are enabled for allocator type 'T'.
-// By default, ASan is enabled.
-template<typename T>
-struct is_ASan_enabled_for_allocator {
-    static constexpr bool value = true;
-};
-
-// Convenience helper to determine if ASan is enabled for a given allocator type 'T'
-template <typename T>
-constexpr bool is_ASan_enabled_for_allocator_v = is_ASan_enabled_for_allocator<T>::value;
+template <class>
+constexpr bool _Is_ASan_enabled_for_allocator = true;
 
 struct _Asan_aligned_pointers {
     const void* _First;
--- a/stl/inc/xmemory
+++ b/stl/inc/xmemory
@@ -805,7 +805,7 @@ struct _Simple_types { // wraps types from allocators with simple addressing for
 _INLINE_VAR constexpr size_t _Asan_granularity      = 8;
 _INLINE_VAR constexpr size_t _Asan_granularity_mask = _Asan_granularity - 1;
 
-// Represents whether ASan is enabled for allocator type 'T'.
+// Represents whether ASan container annotations are enabled for allocator type 'T'.
 // By default, ASan is enabled.
 template<typename T>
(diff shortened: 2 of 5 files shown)


stl/inc/xmemory
```diff
@@ -805,6 +805,17 @@ struct _Simple_types { // wraps types from allocators with simple addressing for
...
+// Represents whether ASan is enabled for allocator type 'T'.
+// By default, ASan is enabled.
+template<typename T>
+struct is_ASan_enabled_for_allocator {
+    static constexpr bool value = true;
+};
+
+// Convenience helper to determine if ASan is enabled for a given allocator type 'T'
+template <typename T>
+constexpr bool is_ASan_enabled_for_allocator_v = is_ASan_enabled_for_allocator<T>::value;
```
CaseyCarter, line 817: * We don't need both a class _and_ variable template. Types are expensive so let's just keep the variable template.
* We conventionally use `class` to introduce template type parameters.
* This variable name is far too pretty! (We have to use reserved identifiers in the STL for all machinery not specified by the Standard. Users are allowed to `#define is_ASan_enabled_for_allocator woof woof woof` before including our headers, and we have to tolerate it. We generally use `_CapitalCamelCase` for template parameters and `_Capitalized_snake_case` for other names.)
* The default is _super_ obvious with a variable template, so we can omit the comment.
* Maybe more controversial, but my personal style (i.e., feel free to ignore this particular comment) I think omitting the names of unused template parameters improves clarity.
```suggestion
template <class>
constexpr bool _Is_ASan_enabled_for_allocator = true;
```
(Be sure to rename other references consistently.)

stl/inc/xmemory
```diff
@@ -805,6 +805,17 @@ struct _Simple_types { // wraps types from allocators with simple addressing for
 _INLINE_VAR constexpr size_t _Asan_granularity      = 8;
 _INLINE_VAR constexpr size_t _Asan_granularity_mask = _Asan_granularity - 1;
 
+// Represents whether ASan is enabled for allocator type 'T'.
```
CaseyCarter, line 808: We can be a bit more precise here:
```suggestion
// Represents whether ASan container annotations are enabled for allocator type 'T'.
```

stl/inc/vector
```diff
@@ -493,6 +493,10 @@ private:
             return;
         }
 
+        if (!is_ASan_enabled_for_allocator_v<allocator_type>) {
```
CaseyCarter, line 496: Since this is testing a constant expression, it should probably be an `if constexpr` containing the entire function body so we don't bother to so much as compile it.

CaseyCarter (changes requested):

stl/inc/vector
davidmrdavid: `constexpr`'ed all the things: https://github.com/microsoft/STL/pull/5241/commits/71c7716a8a53bc36d97bad260f59662c3b1dc725

stl/inc/xmemory
davidmrdavid: Just a quick question on the "types are expensive" comment, @CaseyCarter: when you say this, is that because we want to avoid creating too many identifiers, or is there a compile-time cost, or something of the sort? I buy that we shouldn't have declarations we don't need, but I'm curious if there's something special going on here that I'm overseeing.

CaseyCarter: "Types are expensive" is purely a compile time concern. Compilers have to do a ton of work to instantiate a class template, measurably less work to instantiate a variable template, and even slightly less work than that to substitute the definition of an alias template. We try to prefer the least expensive of those options when metaprogramming.

stl/inc/vector
frederick-vs-ja: I guess it's better to write

```C++
if constexpr (is_ASan_enabled_for_allocator_v<allocator_type>) {
// ASan stuffs
}
```

davidmrdavid: Yeah I thought about that as well. In this case we're inside the `_Apply_annotation` method which, to my understanding, is already ASan-specific. I figure we wouldn't want to indent all of the remaining code to be inside the `if`-statement.

Do let me know if I'm misunderstanding something though :-)

tests/std/tests/GH_002030_asan_annotate_vector/test.cpp
```diff
@@ -274,6 +274,93 @@ struct implicit_allocator : custom_test_allocator<T, Pocma, Stateless> {
...
+    Arena* arena;
+
+    ArenaAllocator(size_t alloc_size, size_t capacity) {
+        assert(sizeof(T) <= alloc_size && "Constructor failed: allocation size is too small for target type.");
+        arena = new Arena(alloc_size, capacity);
+    }
+
+    template <class U>
+    ArenaAllocator(const ArenaAllocator<U>& other) {
+        assert(sizeof(U) <= arena->allocation_size && "Constructor failed: allocation size is too small for target type.");
```
davidmrdavid, line 326: Actually, this should be `sizeof(T)`. Seems I forgot to push that change. Will fix that shortly

tests/std/tests/GH_002030_asan_annotate_vector/test.cpp
```diff
@@ -274,6 +274,93 @@ struct implicit_allocator : custom_test_allocator<T, Pocma, Stateless> {
...
+    template<typename U>
+    bool operator!=(ArenaAllocator<U> const other)
+    {
+        return !(this == other);
+    }
+};
+
+// Opt out of ASan analysis for the ArenaAllocator
+// FIXME: IntelliSense claims '_Is_ASan_enabled_for_allocator is not a templateC/C++(864)',
+//  but it works somehow.
```
CaseyCarter, line 360: This is probably due to IntelliSense picking up the installed STL headers instead of the headers in your STL repo.

tests/std/tests/GH_002030_asan_annotate_vector/test.cpp
```diff
@@ -1002,12 +1089,40 @@ void run_custom_allocator_matrix() {
...
+    // TODO: is it possible to add a 'negative' test case here? One where ASan expectedly fails?
+}
+
 template <class T>
 void run_allocator_matrix() {
     run_tests<allocator<T>>();
     run_custom_allocator_matrix<T, aligned_allocator>();
     run_custom_allocator_matrix<T, explicit_allocator>();
     run_custom_allocator_matrix<T, implicit_allocator>();
+    run_asan_disablement_test();
```
davidmrdavid, line 1125: I suppose I should template this new test with `<T>`

tests/std/tests/GH_002030_asan_annotate_string/test.cpp
```diff
@@ -287,6 +287,94 @@ STATIC_ASSERT(_Container_allocation_minimum_asan_alignment<
                   basic_string<wchar_t, char_traits<wchar_t>, implicit_allocator<wchar_t>>>
               == 2);
 
+// Helper class for `ArenaAllocator`, where data is linearly allocated in a finite buffer.
+class Arena {
```
davidmrdavid, line 291: @CaseyCarter - I have this `Arena` and `ArenaAllocator` classes duplicated in the `basic_string` and `vector` test files. I'd love to refactor it into a shared location so that it can be imported in both tests.

Do you all have some kind of "shared test utils" header in this repo or do we prefer to keep the tests fully self-contained?

tests/std/tests/GH_002030_asan_annotate_string/test.cpp
```diff
@@ -1867,6 +1981,7 @@ void run_allocator_matrix() {
     run_custom_allocator_matrix<CharType, aligned_allocator>();
     run_custom_allocator_matrix<CharType, explicit_allocator>();
     run_custom_allocator_matrix<CharType, implicit_allocator>();
+    run_asan_disablement_test();
```
davidmrdavid, line 1984: TODO: I suppose I should add template-paramterize this new test as well. I'll tackle that tomorrow

tests/std/tests/GH_002030_asan_annotate_vector/test.cpp
davidmrdavid: addressed, resolving

tests/std/tests/GH_002030_asan_annotate_vector/test.cpp
davidmrdavid: addressed, resolving.

tests/std/tests/GH_002030_asan_annotate_string/test.cpp
davidmrdavid: addressed, resolving.

tests/std/tests/GH_002030_asan_annotate_vector/test.cpp
davidmrdavid: Oh, that makes sense. I'll remove this comment in my next commit then.

Just curious - do we have a good way of teling intellisense about this, or do we just 'ignore' the false alarm?

davidmrdavid:
Marking as draft for now to try and simplify the test.

tests/std/tests/GH_002030_asan_annotate_string/test.cpp
StephanTLavavej: `tests/std/include` is where shared test headers go.

tests/std/tests/GH_002030_asan_annotate_vector/test.cpp
StephanTLavavej: We test IntelliSense via the `/BE` command-line option. That should pick up your freshly built STL headers.

As for actual IntelliSense in the IDE, no idea. I would expect that if you've built something with the freshly built STL picked up (remember `set_environment.bat`!), then it would know the correct paths.

davidmrdavid: I'll do some digging and find out, would be good to document :-)

tests/std/tests/GH_002030_asan_annotate_string/test.cpp
davidmrdavid: Thanks. Since we've greatly simplified the test case, I don't think I'll need to refactor any classes into a shared location, but I'll remember this for the future :) . _Resolving_

stl/inc/vector
davidmrdavid: _Resolving due to inactivity. Please re-open if needed :-) _

tests/std/tests/GH_002030_asan_annotate_vector/test.cpp
```diff
@@ -1002,12 +1022,30 @@ void run_custom_allocator_matrix() {
...
+    std::vector<T, implicit_allocator_no_asan_annotations<T>> vec;
+    vec.reserve(100);
+
+    // We access position (50) of the vector, which is within the capacity of the vector
+    // but uninitialized. This would normally trigger an AV because of ASan annotations.
+    // However, the allocator has opted out of ASan analysis through,
+    // `_Is_ASan_enabled_for_allocator`, so this should succeed.
+    vec.data()[50] = T();
+
+    // TODO: is it possible to add a 'negative' test case here? One where ASan expectedly fails?
```
davidmrdavid, line 1039: @StephanTLavavej - I know [from the README](https://github.com/microsoft/STL/blob/9b38fb9ccca90fd8b31fd37d4512b190aeef4fe7/README.md?plain=1#L351-L363) that `stl-lit.py` supports `XFAIL` tests, which would match this "negative test case" I'm trying to write.

Those instructions say to modify the `expected_results` file, which I suppose is [`tests/std/expected_results.txt`](https://github.com/microsoft/STL/blob/9b38fb9ccca90fd8b31fd37d4512b190aeef4fe7/tests/std/expected_results.txt) which to my surprise is empty (despite being populated for [`libcxx`](https://github.com/microsoft/STL/blob/9b38fb9ccca90fd8b31fd37d4512b190aeef4fe7/tests/std/expected_results.txt).

Do we usually not write negative tests again the tests in `std` or am I misunderstanding the purpose of the file? Just asking to understand if I'm misunderstanding how to about writing this negative test. Thanks!

StephanTLavavej: As in #5252, you would need a "death" test. lit's XFAIL doesn't distinguish between "failed to compile" and "failed to run".

davidmrdavid: Forgot to resolve this comment, yes -we had discussed this internally. Thanks, I'll resolve this.

stl/inc/xmemory
```diff
@@ -805,6 +805,10 @@ struct _Simple_types { // wraps types from allocators with simple addressing for
 _INLINE_VAR constexpr size_t _Asan_granularity      = 8;
 _INLINE_VAR constexpr size_t _Asan_granularity_mask = _Asan_granularity - 1;
 
+// Represents whether ASan container annotations are enabled a given allocator.
+template <class>
+constexpr bool _Is_ASan_enabled_for_allocator = true;
```
zacklj89, line 810: Nice!

I may have missed some conversation about it, but I'm wondering if the name should be changed. The feature is disabling annotations based on the _container's_ underlying allocator. 

Since the options for compiler are `_DISABLE_X_ANNOTATION` should we try to follow suite with `Disable` in the name?

davidmrdavid: >Since the options for compiler are _DISABLE_X_ANNOTATION should we try to follow suite with Disable in the name?

Oh I didn't realize the compiler options / flags followed that naming convention? Do you have a link or example where I can see this?

But yeah, don't feel strongly about naming. If it matches the flags, I'm happy to have `Disabled` over `Enabled` in the name. I'll tackle that on my next push.

zacklj89: Yeah, I'm basically commenting on staying in line with [these](https://github.com/microsoft/STL/blob/9b38fb9ccca90fd8b31fd37d4512b190aeef4fe7/stl/inc/__msvc_sanitizer_annotate_container.hpp#L18-L27)

davidmrdavid: Incorporated: https://github.com/microsoft/STL/pull/5241/commits/57e7affd7f3f5d6f32c255c0a7b822c7006f7551

zacklj89 (approved):
Cool 👍 

Is the plan to hold on merge while testing FE allocators?

davidmrdavid:
Thanks, @zacklj89!

> Is the plan to hold on merge while testing the FE?

My current plan is to work on getting the FE to test this by Friday one way or another: either by manually porting this internally on a private branch, or by consuming this 'as normal' by merging this PR and mirror'ing it on ADO's default branch.

So I think it depends on how quickly we can STL's final approval here. One way or another, I'll be working towards enabling this to be validated by FE starting _tomorrow_ :) . 

So I don't think it's strictly necessary to block the PR merge on it - the feature and tests seems up to spec _to me_. But I'll defer to more experienced owners on that decision.

tests/std/tests/GH_002030_asan_annotate_vector/test.cpp
StephanTLavavej: Resolving; comment has been removed.

tests/std/tests/GH_002030_asan_annotate_vector/test.cpp
```diff
@@ -53,6 +55,7 @@ struct non_trivial_can_throw {
     }
 
     non_trivial_can_throw& operator=(const non_trivial_can_throw&) {
+        ++i;
         return *this;
     }
```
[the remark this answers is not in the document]
StephanTLavavej: Unless I am confused, it looks like we should also follow the pattern
```cpp
        if (i == 0) {
            throw i;
        }
```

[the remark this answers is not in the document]
StephanTLavavej: I will validate and push changes.

tests/std/tests/GH_002030_asan_annotate_vector/test.cpp
```diff
@@ -1002,31 +1026,66 @@ void run_custom_allocator_matrix() {
     run_tests<AllocT<T, false_type, false_type>>();
 }
 
+// Test that writing to un-initialized memory in a string triggers ASan container-overflow error.
```
StephanTLavavej, line 1029: Copy-paste error: This is a vector, not a string.

StephanTLavavej (approved):

StephanTLavavej:

tests/std/tests/GH_002030_asan_annotate_string/test.cpp
```diff
@@ -1869,6 +1912,11 @@ void run_allocator_matrix() {
     run_custom_allocator_matrix<CharType, aligned_allocator>();
     run_custom_allocator_matrix<CharType, explicit_allocator>();
     run_custom_allocator_matrix<CharType, implicit_allocator>();
+
+    // To test ASan annotation disablement, we use an ad-hoc allocator type to avoid disrupting other
+    // tests that depend on annotations being enabled. Therefore, unlike the prior tests,
+    // this test is not parametrized by the allocator type.
```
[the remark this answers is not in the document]
davidmrdavid: this typo is news to me! whoa, thanks

tests/std/tests/GH_002030_asan_annotate_string/test.cpp
```diff
@@ -1944,7 +1983,11 @@ int main(int argc, char* argv[]) {
         test_gh_3955();
     });
 #ifdef __SANITIZE_ADDRESS__
-    exec.add_death_tests({test_gh_5251});
+#ifdef __cpp_char8_t
+    exec.add_death_tests({run_asan_container_overflow_death_test<char8_t>});
+#endif // __cpp_char8_t
+    exec.add_death_tests({run_asan_container_overflow_death_test<char16_t>,
+        run_asan_container_overflow_death_test<char32_t>, run_asan_container_overflow_death_test<wchar_t>});
```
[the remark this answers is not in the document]
davidmrdavid: This was actually forced upon me by clang-format, I believe. But I haven't triple-checked.

davidmrdavid:
Thanks for the feedback, I've taken notes on the common feedback items (don't alias, don't add `std::`, remember the null termination _is_ initialized, etc.). Hoping to retain that for next time around.

New diff seems sound to me, thanks. Good to merge _afaict_.

tests/std/tests/GH_002030_asan_annotate_string/test.cpp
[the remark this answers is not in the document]
StephanTLavavej: What happens is that if you don't have a trailing comma, clang-format will format it on long lines. But if you add a trailing comma, clang-format will then format it one-per-line.

In general, clang-format adjusts whitespace only, and will not alter printing characters. (We have configured a couple of exceptions, though - we ask clang-format to remove redundant semicolons, for example.)

tests/std/tests/GH_002030_asan_annotate_string/test.cpp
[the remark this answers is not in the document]
StephanTLavavej: I was like that with "auxiliary" - kept wanting to spell it as "auxilliary" but that's wrong! :joy_cat:

StephanTLavavej:
Thanks for implementing this feature and improving test coverage! :heart_eyes_cat: :white_check_mark: :tada:

PR #5246, opened 2025-01-21
Expand workaround for EDG bug in `P0323R12_expected`

CaseyCarter:
The IntelliSense compiler just disabled compatibility with an outdated MSVC overload resolution bug (the `do_ms_quirk` bit in VS-PR-602338), causing an apparent regression in `P0323R12_expected`. These `static_assert`s that were previously passing - but for the wrong reason - [now fail with VS trunk](https://dev.azure.com/devdiv/DevDiv/_build?definitionId=24811&_a=summary&view=runs) (apologies for this Microsoft-internal pipeline link, it's the new `test-edg-only` pipeline I helped the IntelliSense team put together). It looks a lot like VSO-1601179, so let's just expand the existing conditional to cover the new failures.

Drive-by: Don't bother to implement some member functions that aren't ODR-used.

### DIFF
--- a/tests/std/tests/P0323R12_expected/test.cpp
+++ b/tests/std/tests/P0323R12_expected/test.cpp
@@ -2476,24 +2476,17 @@ static_assert(test_inherited_constructors());
 template <class T, class E>
 struct ambiguating_expected_copy_constructor_caller {
     struct const_lvalue_taker {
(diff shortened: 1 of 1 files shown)


StephanTLavavej (approved):

StephanTLavavej:
I have an active merge batch so I've copied this PR over to #5249.

PR #5255, opened 2025-01-27
Remove usage of `[[likely]]` and `[[unlikely]]`

StephanTLavavej:
@GabrielDosReis informed me that the optimizer team discourages use of the `[[likely]]` and `[[unlikely]]` attributes, and instead encourages PGO/SPGO. (This aligns with what I've always said, "don't help the compiler".) We've accumulated a fair amount of usage of these attributes, *none* of which was motivated by benchmarking. Let's remove this noise from our codebase. If we ever find places where it actually affects performance (presumably for Clang/LLVM, since C2 pays minimal attention to the attribute), we can decide what to do.

* Remove direct usage of `[[likely]]` and `[[unlikely]]`.
* Remove `_LIKELY_X86_X64`, `_LIKELY_ARM_ARM64`, and clang-format suppression.
* Remove `_LIKELY` and `_UNLIKELY`.

### DIFF
--- a/stl/inc/__msvc_chrono.hpp
+++ b/stl/inc/__msvc_chrono.hpp
@@ -653,16 +653,6 @@ namespace chrono {
         using time_point                = _CHRONO time_point<steady_clock>;
         static constexpr bool is_steady = true;
 
-#if defined(_M_ARM) || defined(_M_ARM64) // vvv ARM or ARM64 arch vvv
-#define _LIKELY_ARM_ARM64 _LIKELY
-#define _LIKELY_X86_X64
-#elif defined(_M_IX86) || defined(_M_X64) // ^^^ ARM or ARM64 arch / x86 or x64 arch vvv
-#define _LIKELY_ARM_ARM64
-#define _LIKELY_X86_X64 _LIKELY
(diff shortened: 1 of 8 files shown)


GabrielDosReis (approved):

AlexGuteniev:
> since C2 will reportedly ignore these attributes forever

Well, it uses the attribute for quite long: https://godbolt.org/z/TPabjMos8

GabrielDosReis:
> > since C2 will reportedly ignore these attributes forever
> 
> Well, it uses the attribute for quite long: https://godbolt.org/z/TPabjMos8

And that is basically it.  The compiler team instead recommends its customers use profile-guided technology (like PGO or SPGO) that rely on profiling data.

AlexGuteniev:
> And that is basically it.

That's fine,  I'm not opposed to this PR. or compiler not using `[[unlikely]]` attribute more than that,
I'm only opposed to the misleading statement in the PR description.

StephanTLavavej:
I've updated the PR description, hopefully it's more accurate now.

MahmoudGSaleh (approved):
Looks good to me

StephanTLavavej:

Issue #5245, opened 2025-01-18
`<regex>`: Successful negative lookahead assertions sometimes mistakenly assign matches to capture groups


muellerj2:
When a negative lookahead assertion is successful (i.e., the asserted regular expression does not match), some capture groups might be assigned in the asserted regular expression as if the lookahead regular expression had matched.

## Test case
```cpp
#include <iostream>
#include <regex>

using namespace std;

int main() {
    const regex re("(?!(a)b)..");
    cmatch match;

    cout << "'(?!(a)b)..' matches 'ac': "
         << regex_match("ac", match, re)
         << '\n';
    cout << "capture group 1 matched: "
         << match[1].matched
         << '\n';
}
```
https://godbolt.org/z/EzjGrGP5K

This prints:
```
'(?!(a)b)..' matches 'ac': 1
capture group 1 matched: 1
```

## Expected result
This should print:
```
'(?!(a)b)..' matches 'ac': 1
capture group 1 matched: 0
```

It follows from ECMA-262 3rd ed., Section 15.10.2.8 "Atom", that capture groups in negative lookahead assertions are never matched, whether the asserted regular expression matches or not.

PR #5261, opened 2025-01-31
<format>: Alternate form general floating-point width handling

alexprabhat99:
Fixes: #5011

PR #5269, opened 2025-02-04
<regex>: Successful negative lookahead assertions do not matche to capture groups

alexprabhat99:
Fixes: #5245

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -3580,7 +3580,11 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Match_pat(_Node_base* _Nx) { // c
                     _Tgt_state = _St;
                     _Failed    = true;
                 } else {
-                    _Tgt_state._Cur = _Ch;
+                    if (_Neg) {
+                        _Tgt_state = _St;
+                    } else {
+                        _Tgt_state._Cur = _Ch;
+                    }
                 }
 
                 break;
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -334,10 +334,7 @@ public:
     template <class _Iter>
     char_class_type lookup_classname(_Iter _First, _Iter _Last, bool _Icase = false) const {
         // map [_First, _Last) to character class mask value
-#define _REGEX_CHAR_CLASS_NAME(n, c)                            \
-    {                                                           \
-        n, L##n, static_cast<unsigned int>(_STD size(n) - 1), c \
-    }
+#define _REGEX_CHAR_CLASS_NAME(n, c) {n, L##n, static_cast<unsigned int>(_STD size(n) - 1), c}
         static constexpr _Cl_names _Names[] = {
             // map class names to numeric constants
             _REGEX_CHAR_CLASS_NAME("alnum", _Ch_alnum),
@@ -3610,7 +3607,6 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Match_pat(_Node_base* _Nx) { // c
                 _Node_end_group* _Node = static_cast<_Node_end_group*>(_Nx);
                 _Node_capture* _Node0  = static_cast<_Node_capture*>(_Node->_Back);
                 if (_Cap || _Node0->_Idx != 0) { // update capture data
-                    _Tgt_state._Grp_valid[_Node0->_Idx] = true;
                     _Tgt_state._Grps[_Node0->_Idx]._End = _Tgt_state._Cur;
                 }
                 break;
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -3572,19 +3572,14 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Match_pat(_Node_base* _Nx) { // c
         case _N_neg_assert:
         case _N_assert:
             { // check assert
-                _It _Ch              = _Tgt_state._Cur;
                 bool _Neg            = _Nx->_Kind == _N_neg_assert;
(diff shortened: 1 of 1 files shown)


AlexGuteniev:
I'd recommend running tests locally. 
You can use regex to run all regex tests, this command works for me:
`python tests\utils\stl-lit\stl-lit.py -Dnotags=ASAN ..\..\llvm-project\libcxx\test ..\..\tests\std ..\..\tests\tr1 --filter=(regex^|/re\.)`

stl/inc/regex
```diff
@@ -3610,7 +3607,6 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Match_pat(_Node_base* _Nx) { // c
                 _Node_end_group* _Node = static_cast<_Node_end_group*>(_Nx);
                 _Node_capture* _Node0  = static_cast<_Node_capture*>(_Node->_Back);
                 if (_Cap || _Node0->_Idx != 0) { // update capture data
-                    _Tgt_state._Grp_valid[_Node0->_Idx] = true;
```
muellerj2, line 3613: This deletion does not just correct the matches for negative assertions. With this change, no capture group at all will ever be matched.

Changes to fix #5245 actually have to be applied here:
https://github.com/microsoft/STL/blob/fc15609a0f2ae2a134c34e7c9a13977994f37367/stl/inc/regex#L3575-L3590

Specifically, `_Tgt_state._Cur = _Ch;` insufficiently resets the state after a successful negative assertion. (But it is the appropriate reset for positive assertions.)

muellerj2 (changes requested):

stl/inc/regex
alexprabhat99: I have restored the initial state after a successful negative assertion.

stl/inc/regex
```diff
@@ -3583,7 +3580,11 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Match_pat(_Node_base* _Nx) { // c
                     _Tgt_state = _St;
```
muellerj2, line 3580: Pre-existing: This state reset after a match failure is both unnecessary and pointless, so I think it should be removed. No caller of `_Match_pat()` can assume that the match state hasn't changed after failure, because the function does not fully reset the state after trying to match two or more NFA nodes anyway.

stl/inc/regex
```diff
@@ -3583,7 +3580,11 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Match_pat(_Node_base* _Nx) { // c
                     _Tgt_state = _St;
                     _Failed    = true;
                 } else {
```
muellerj2, line 3582: The opening brace after `else` leads to unnecessary nesting. You can just remove the braces and write `else if` instead.

stl/inc/regex
```diff
@@ -3583,7 +3580,11 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Match_pat(_Node_base* _Nx) { // c
                     _Tgt_state = _St;
                     _Failed    = true;
                 } else {
-                    _Tgt_state._Cur = _Ch;
+                    if (_Neg) {
+                        _Tgt_state = _St;
+                    } else {
+                        _Tgt_state._Cur = _Ch;
```
muellerj2, line 3586: Pre-existing: We can assign `_St._Cur` to `_Tgt_state._Cur` here instead of `_Ch` and eliminate the local variable `_Ch`. This should also decrease the amount of stack this function uses per call a bit, slightly reducing the danger of stack overflow (as in #997).

muellerj2 (commented):
I think a test (like the one in #5245) should be added to verify that the behavior is correct after this PR.

Edit: It would also be nice to add test coverage which verifies that captures are recorded correctly for positive assertions as well.

stl/inc/regex
```diff
@@ -3575,15 +3572,14 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Match_pat(_Node_base* _Nx) { // c
...
                 bool _Neg            = _Nx->_Kind == _N_neg_assert;
                 _Bt_state_t<_It> _St = _Tgt_state;
                 if (_Match_pat(static_cast<_Node_assert*>(_Nx)->_Child) == _Neg) {
                     // restore initial state and indicate failure
                     _Tgt_state = _St;
                     _Failed    = true;
-                } else {
-                    _Tgt_state._Cur = _Ch;
+                } else if (!_Neg) {
+                    _Tgt_state._Cur = _St._Cur;
```
muellerj2, line 3582: This is not equivalent to the prior commit. Now the state is no longer being reset for a successful negative lookahead assertion at all.

(My comment about a superfluous `_Tgt_state = _St;` statement points a different line, namely the one in the failure case. The one removed by the last commit in the successful case isn't superfluous at all.)

This suggests we really need more test coverage for negative assertions if no test catches this.

muellerj2 (changes requested):

StephanTLavavej:
Hi @alexprabhat99, do you have time to address the feedback here, or do you need help to get this ready?

StephanTLavavej:
@muellerj2 There's absolutely no obligation or expectation, but if you know what needs to happen here and would like to pick up this PR, you can create a new one by starting with this one's commit and then adding whatever's necessary. @alexprabhat99 agreed to the CLA in https://github.com/microsoft/STL/pull/5261#issuecomment-2628721418 so this should be perfectly fine, and keeping the original commits will result in co-author credit during merging.

If you're too busy or would prefer not to, that's perfectly fine - I will look into this when I have time.

PR #2110, opened 2021-08-10
Workaround for slow codegen on x86 atomic load

AlexGuteniev:
Compiler emits jumptable or jcc sequence that prevents inlining
of atomic load; separation of order check and barrier condition helps

PR #5270, opened 2025-02-05
Improve debug checks (prelude to STL Hardening)

StephanTLavavej:
# :world_map: Overview

Works towards #5090. For ease of review, this PR contains all of the cleanups and enhancements that I noticed along the way. The following PR will then be able to focus on removing `_CONTAINER_DEBUG_LEVEL` and adding STL Hardening without mixing in worthwhile-but-distracting changes.

Notably, this PR has audited the STL for "Continue on Error" (CoE) compatibility, allowing users to define `_STL_CRT_SECURE_INVALID_PARAMETER` to something that returns. (Throwing an exception is still super ultra forbidden.) Automated test coverage for CoE may follow in a future PR (it would inherently involve UB, so I need to investigate).

This also improves debug codegen for the STL's debug checks, which were emitting unnecessary branches.

# :speech_balloon: Comment cleanups

* Remove CDL preprocessor comments after a single line.
* Remove CDL preprocessor comments after a single `_STL_VERIFY` over multiple lines.

In general, we comment our `#else` and `#endif` directives to clarify where "regions" begin and end. However, for very short regions that aren't densely nested in other preprocessor directives, we're willing to omit the commits as they cause more clutter than clarity.

Here, removing the comments will make it easier to edit the conditions later. (All CDL conditions will be replaced with either the STL Hardening pattern, or IDL != 0.)

# :scroll: Documentation cleanups

* `docs/import_library.md`: Don't mention CDL - it doesn't affect ABI, and will be removed.

# :page_facing_up: Debug message improvements

* Improve debug messages: `"array<T, 0> subscript is invalid"`
  + This identifies that we're dealing with `array<T, 0>`, and that the call is permanently bogus, following the example of `"array<T, 0>::front() is invalid"`.
* Improve debug messages: `"MEOW subscript out of range"`
  + This is consistent with `vector`.
  + `mdspan`'s message was very opaque; starting with `"mdspan subscript out of range"` provides clarity. Additionally, "I" was previously used without being defined; I've replaced it with its definition from /2.
  + `ranges::view_interface` now refers to itself with qualification.
* Improve debug messages: `"MEOW() called on empty WOOF"`
  + `deque` now mentions whether `pop_front()` or `pop_back()` was called.
  + `optional` now mentions whether `operator*()` or `operator->()` was called.
  + `ranges::view_interface` now refers to itself with qualification.
* Improve debug messages: Clarify `span`'s `first<Count>()`, `last<Count>()`, `subspan<Offset, Count>()`.
* Improve debug messages: Clarify `string_view` `remove_prefix()`, `remove_suffix()`.
  + Spell the function names literally for improved searchability. Mention the short version of the class name too. Adjust "longer" to "larger" for style.
* Improve debug messages: `std::expected`.
  + This splits `"expected stores an error, not a value"` into:
    - `"operator->() called on a std::expected that contains an error, not a value"`
    - `"operator*() called on a std::expected that contains an error, not a value"`
  + And changes `"expected stores a value, not an error"` into:
    - `"error() called on a std::expected that contains a value, not an error"`
  + This mentions which member function is being called. Because `"expected"` doesn't really sound like a noun, I've added `"std::"` qualification and made the phrases less terse. And as @sivadeilra suggested, "stores" has connotations of "memory writes", so "contains" is clearer.
* Improve debug messages: Improve grammar and clarity with "basic_format_arg contains an impossible type".
  + In general, I limited the scope of these debug message improvements to those that will be shared with STL Hardening checks. However, I noticed this one in `<format>`'s implementation and couldn't resist fixing it.

# :ballot_box_with_check: Test cleanups

* Test cleanup: Remove CDL definitions from some `<ranges>` tests.
  + These weren't CDL-sensitive at all, and were properly guarded with `IDL != 0`.
* Test cleanup: Drop CDL from `<format>` test, properly test negative dynamic width/precision.
  + `<format>` isn't CDL-sensitive.
  + Negative dynamic width/precision are required to throw `format_error`; we don't use "death tests" for exceptions. For clarity and consistency, add positive and negative test cases for both. This may slightly duplicate existing coverage, but negative dynamic precision wasn't being properly tested AFAICT.
* Test cleanup: Remove VSO-847348 workarounds.
  + VSO-847348 "c1xx ICEs deducing array size inside a temploid" was fixed at some point during the last 6 years (verified with prod/fe x86chk).
* Test cleanup: Use `if constexpr` instead of tag dispatch and plain `if`.

# :magic_wand: Code cleanups

* Code cleanups: Add `const`, unwrap "strengthened".
* Code cleanup: Conditional `_STL_REPORT_ERROR` => `_STL_VERIFY` or `_STL_ASSERT`
  + In `<memory_resource>`, we always call `_Prepare_oversized`, but the `_STL_REPORT_ERROR` was guarded by `#ifdef _DEBUG`. Convert this to `_STL_ASSERT`, which is consistent with the `_STL_ASSERT` that appears immediately below.
  + In `<numeric>` `gcd()`, we're using [De Morgan's laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws), the STL maintainer's best friend.
  + In `<numeric>` `lcm()`, extract `const bool _Overflow` so the `_STL_VERIFY` has no side effects.
* Code cleanup: `_STL_VERIFY(false)` => `_STL_REPORT_ERROR`
  + In `<mdspan>`, add `return 1;` in `stride()` to support upcoming changes for continue-on-error.
  + In `<regex>`, unconditionally `return false;` for the same reason.
* Code cleanup: Consistently order `case` before `default`.
* Code cleanup: Fuse `_Check_alignment` into `atomic_ref` ctor.
* Code cleanup: Collapse `_ATOMIC_REF_CHECK_ALIGNMENT` into `_STL_ASSERT`.
  + This doesn't change behavior (it remains a debug-only check), it just removes an escape hatch that was never used.
* Code cleanup: Avoid duplicating `_Off` and `_Count`. Add `const` to `_Moved`.
* Code cleanup: Avoid duplicating `_STD copy()` call.
* Code cleanup: Avoid more duplication.
  + We can always extract `_Newsize` for clarity.
* Code cleanup: IDL == 2 implies debug, so `_STL_ASSERT` should be `_STL_VERIFY`.
  + This is especially noticeable when `_STL_VERIFY` already follows.
* Code cleanup: Extract `_Can_memcpy` for clarity, simplifying following CoE fix.
* Code cleanup: Use `_Check_MEOW_memory_order` to avoid some `_FALLTHROUGH` in `<atomic>`.
  + For `_ATOMIC_LOAD_ARM64` and `_ATOMIC_POST_LOAD_BARRIER_AS_NEEDED`, use `_Check_load_memory_order` and arrange the control flow to preserve the fallthrough behavior.
  + Because `_Check_load_memory_order` takes a real enum, we need to compare against the enumerators and avoid `static_cast<unsigned int>(_Order)`.
  + For `_ATOMIC_STORE_PREFIX`, expand this **absolutely hideous macro** :vomiting_face: at each point of usage, then perform a similar transformation. As a bonus, we get to drop a bunch of early `return;`s.

# :rocket: Enhancements

* Enhancement (for CoE): Adjust `uninitialized_meow`'s logic.
  + We'll continue right into an infinite loop, but at least we won't fall off the end of a non-void function.
* Enhancement (for CoE): After `_INVALID_MEMORY_ORDER`, consistently fallthrough to `seq_cst`.
  + We were already doing this in other places.
* Enhancement: `_STL_ASSERT` => `_STL_VERIFY` for checks within IDL >= 1.
  + We don't really care about IDL exactly equal to 1, but these should clearly be checked in IDL=1 release, consistent with the following checks. Note that we already had an `_STL_VERIFY` below.
* Enhancement: Replace `_STL_ASSERT` with `_STL_INTERNAL_STATIC_ASSERT`.
  + These conditions are already guaranteed by `duration` and `ratio`. They don't need to be checked for users, especially at runtime.
* Enhancement: Remove IDL=2 null asserts from `auto_ptr` deref/arrow.
  + It's silly to assert that `_Myptr` is non-null when we're about to dereference it (or return it for imminent dereferencing). We purged this category of asserts from the STL aeons ago, but must have missed `auto_ptr`. The idea is that null dereferences are just as obvious in the debugger. Note that this makes the code consistent with `unique_ptr` and `shared_ptr`.
* Enhancement: `_Span_iterator::operator-=` was unconditionally checking.
  + First, a cleanup: `_Verify_offset` does something for IDL >= 1 only, so we can guard the callsite in `operator+=`.
  + In `operator-=`, the pathological "integer overflow" check was being performed unconditionally. Guard it for IDL >= 1, and the following `_Verify_offset` call.
* Enhancement: Improve `_STL_REPORT_ERROR`, `_STL_VERIFY` codegen.
  + See: https://godbolt.org/z/b95sjPPGo
  + And: https://github.com/microsoft/STL/blob/fc15609a0f2ae2a134c34e7c9a13977994f37367/stl/inc/atomic#L103-L105
  + `do { ... } while (false)` has debug codegen costs. We don't need this because we ALWAYS use braces for control flow.
  + `if (cond) { } else { ... }` also has debug codegen costs. We control the conditions, and the Standard's specification for boolean-testable has improved, so we shouldn't need this.
  + We can drop `-Wassume` ("assumption is ignored because it contains (potential) side-effects") suppression for Clang in `_STL_VERIFY`. Clang doesn't support `/analyze`, but `VSO_0938757_attribute_order` is special - it runs Clang with `/D_PREFAST_` to verify that we're ordering standard attributes before non-standard attributes. Because this is the only test that's activating `_Analysis_assume_` for Clang, compiling it with `-Wno-assume` is better than modifying the product code.

### DIFF
--- a/docs/import_library.md
+++ b/docs/import_library.md
@@ -41,7 +41,7 @@ The caveats of this technique are:
   + This limitation is subtle (not readily apparent from the source code) and critical.
     If shared global state is necessary, our only option while preserving bincompat is adding a satellite DLL.
 * Due to having just two flavors of the import library (debug and release),
-  we cannot use anything that depends on `_CONTAINER_DEBUG_LEVEL` or `_ITERATOR_DEBUG_LEVEL`.
+  we cannot use anything that depends on `_ITERATOR_DEBUG_LEVEL`.
 
 For these reasons, especially the last one, we need to strictly control what is used by the import library.
 In particular, `basic_string` must not be used there.
--- a/stl/inc/__msvc_ranges_tuple_formatter.hpp
+++ b/stl/inc/__msvc_ranges_tuple_formatter.hpp
@@ -324,7 +324,7 @@ public:
         case _Basic_format_arg_type::_Custom_type:
             return _STD forward<_Visitor>(_Vis)(_Custom_state);
         default:
-            _STL_VERIFY(false, "basic_format_arg is in an impossible state");
+            _STL_REPORT_ERROR("basic_format_arg is in an impossible state");
             int _Dummy{};
             return _STD forward<_Visitor>(_Vis)(_Dummy);
         }
--- a/stl/inc/__msvc_ranges_tuple_formatter.hpp
+++ b/stl/inc/__msvc_ranges_tuple_formatter.hpp
@@ -324,7 +324,7 @@ public:
         case _Basic_format_arg_type::_Custom_type:
             return _STD forward<_Visitor>(_Vis)(_Custom_state);
         default:
-            _STL_VERIFY(false, "basic_format_arg is in impossible state");
+            _STL_VERIFY(false, "basic_format_arg is in an impossible state");
             int _Dummy{};
             return _STD forward<_Visitor>(_Vis)(_Dummy);
         }
--- a/stl/inc/__msvc_ranges_tuple_formatter.hpp
+++ b/stl/inc/__msvc_ranges_tuple_formatter.hpp
@@ -324,7 +324,7 @@ public:
         case _Basic_format_arg_type::_Custom_type:
             return _STD forward<_Visitor>(_Vis)(_Custom_state);
         default:
-            _STL_REPORT_ERROR("basic_format_arg is in an impossible state");
+            _STL_REPORT_ERROR("basic_format_arg contains an impossible type");
             int _Dummy{};
             return _STD forward<_Visitor>(_Vis)(_Dummy);
         }
--- a/stl/inc/array
+++ b/stl/inc/array
@@ -708,7 +708,7 @@ public:
 
     _NODISCARD reference operator[](size_type) noexcept /* strengthened */ {
 #if _CONTAINER_DEBUG_LEVEL > 0
-        _STL_REPORT_ERROR("array subscript out of range");
+        _STL_REPORT_ERROR("array<T, 0> subscript invalid");
 #endif
 
         return *data();
@@ -716,7 +716,7 @@ public:
 
     _NODISCARD const_reference operator[](size_type) const noexcept /* strengthened */ {
 #if _CONTAINER_DEBUG_LEVEL > 0
-        _STL_REPORT_ERROR("array subscript out of range");
+        _STL_REPORT_ERROR("array<T, 0> subscript invalid");
 #endif
 
         return *data();
--- a/stl/inc/array
+++ b/stl/inc/array
@@ -533,7 +533,7 @@ public:
     _NODISCARD _CONSTEXPR17 reference operator[](_In_range_(<, _Size) size_type _Pos) noexcept /* strengthened */ {
 #if _CONTAINER_DEBUG_LEVEL > 0
         _STL_VERIFY(_Pos < _Size, "array subscript out of range");
-#endif // _CONTAINER_DEBUG_LEVEL > 0
+#endif
 
         return _Elems[_Pos];
     }
@@ -542,7 +542,7 @@ public:
     /* strengthened */ {
 #if _CONTAINER_DEBUG_LEVEL > 0
         _STL_VERIFY(_Pos < _Size, "array subscript out of range");
-#endif // _CONTAINER_DEBUG_LEVEL > 0
+#endif
 
         return _Elems[_Pos];
     }
@@ -709,7 +709,7 @@ public:
     _NODISCARD reference operator[](size_type) noexcept /* strengthened */ {
 #if _CONTAINER_DEBUG_LEVEL > 0
         _STL_REPORT_ERROR("array subscript out of range");
-#endif // _CONTAINER_DEBUG_LEVEL > 0
+#endif
 
         return *data();
     }
@@ -717,7 +717,7 @@ public:
     _NODISCARD const_reference operator[](size_type) const noexcept /* strengthened */ {
 #if _CONTAINER_DEBUG_LEVEL > 0
         _STL_REPORT_ERROR("array subscript out of range");
-#endif // _CONTAINER_DEBUG_LEVEL > 0
+#endif
 
         return *data();
     }
@@ -725,7 +725,7 @@ public:
     _NODISCARD reference front() noexcept /* strengthened */ {
 #if _CONTAINER_DEBUG_LEVEL > 0
         _STL_REPORT_ERROR("array<T, 0>::front() invalid");
-#endif // _CONTAINER_DEBUG_LEVEL > 0
+#endif
 
         return *data();
     }
@@ -733,7 +733,7 @@ public:
     _NODISCARD const_reference front() const noexcept /* strengthened */ {
 #if _CONTAINER_DEBUG_LEVEL > 0
         _STL_REPORT_ERROR("array<T, 0>::front() invalid");
-#endif // _CONTAINER_DEBUG_LEVEL > 0
+#endif
 
         return *data();
     }
@@ -741,7 +741,7 @@ public:
     _NODISCARD reference back() noexcept /* strengthened */ {
 #if _CONTAINER_DEBUG_LEVEL > 0
         _STL_REPORT_ERROR("array<T, 0>::back() invalid");
-#endif // _CONTAINER_DEBUG_LEVEL > 0
+#endif
 
         return *data();
     }
@@ -749,7 +749,7 @@ public:
     _NODISCARD const_reference back() const noexcept /* strengthened */ {
 #if _CONTAINER_DEBUG_LEVEL > 0
         _STL_REPORT_ERROR("array<T, 0>::back() invalid");
-#endif // _CONTAINER_DEBUG_LEVEL > 0
+#endif
 
         return *data();
     }
--- a/stl/inc/array
+++ b/stl/inc/array
@@ -708,7 +708,7 @@ public:
 
     _NODISCARD reference operator[](size_type) noexcept /* strengthened */ {
 #if _CONTAINER_DEBUG_LEVEL > 0
-        _STL_REPORT_ERROR("array<T, 0> subscript invalid");
+        _STL_REPORT_ERROR("array<T, 0> subscript is invalid");
 #endif
 
         return *data();
@@ -716,7 +716,7 @@ public:
 
     _NODISCARD const_reference operator[](size_type) const noexcept /* strengthened */ {
 #if _CONTAINER_DEBUG_LEVEL > 0
-        _STL_REPORT_ERROR("array<T, 0> subscript invalid");
+        _STL_REPORT_ERROR("array<T, 0> subscript is invalid");
 #endif
 
         return *data();
@@ -724,7 +724,7 @@ public:
 
     _NODISCARD reference front() noexcept /* strengthened */ {
 #if _CONTAINER_DEBUG_LEVEL > 0
-        _STL_REPORT_ERROR("array<T, 0>::front() invalid");
+        _STL_REPORT_ERROR("array<T, 0>::front() is invalid");
 #endif
 
         return *data();
@@ -732,7 +732,7 @@ public:
 
     _NODISCARD const_reference front() const noexcept /* strengthened */ {
 #if _CONTAINER_DEBUG_LEVEL > 0
-        _STL_REPORT_ERROR("array<T, 0>::front() invalid");
+        _STL_REPORT_ERROR("array<T, 0>::front() is invalid");
 #endif
 
         return *data();
@@ -740,7 +740,7 @@ public:
 
     _NODISCARD reference back() noexcept /* strengthened */ {
 #if _CONTAINER_DEBUG_LEVEL > 0
-        _STL_REPORT_ERROR("array<T, 0>::back() invalid");
+        _STL_REPORT_ERROR("array<T, 0>::back() is invalid");
 #endif
 
         return *data();
@@ -748,7 +748,7 @@ public:
 
     _NODISCARD const_reference back() const noexcept /* strengthened */ {
 #if _CONTAINER_DEBUG_LEVEL > 0
-        _STL_REPORT_ERROR("array<T, 0>::back() invalid");
+        _STL_REPORT_ERROR("array<T, 0>::back() is invalid");
 #endif
 
         return *data();
--- a/stl/inc/atomic
+++ b/stl/inc/atomic
@@ -134,54 +134,21 @@ extern "C" inline void _Check_memory_order(const unsigned int _Order) noexcept {
 #if _STD_ATOMIC_USE_ARM64_LDAR_STLR == 1
 
 #define _ATOMIC_LOAD_ARM64(_Result, _Width, _Ptr, _Order_var) \
-    switch (_Order_var) {                                     \
-    case _Atomic_memory_order_relaxed:                        \
+    _Check_load_memory_order(_Order_var);                     \
+    if (_Order_var == memory_order_relaxed) {                 \
         _Result = __iso_volatile_load##_Width(_Ptr);          \
-        break;                                                \
-    case _Atomic_memory_order_release:                        \
-    case _Atomic_memory_order_acq_rel:                        \
-    default:                                                  \
-        _INVALID_MEMORY_ORDER;                                \
-        _FALLTHROUGH;                                         \
-    case _Atomic_memory_order_consume:                        \
-    case _Atomic_memory_order_acquire:                        \
-    case _Atomic_memory_order_seq_cst:                        \
+    } else {                                                  \
         _Result = __LOAD_ACQUIRE_ARM64(_Width, _Ptr);         \
         _Compiler_barrier();                                  \
-        break;                                                \
     }
 
 #endif // _STD_ATOMIC_USE_ARM64_LDAR_STLR == 1
 
 #define _ATOMIC_POST_LOAD_BARRIER_AS_NEEDED(_Order_var) \
-    switch (_Order_var) {                               \
-    case _Atomic_memory_order_relaxed:                  \
-        break;                                          \
-    case _Atomic_memory_order_release:                  \
-    case _Atomic_memory_order_acq_rel:                  \
-    default:                                            \
-        _INVALID_MEMORY_ORDER;                          \
-        _FALLTHROUGH;                                   \
-    case _Atomic_memory_order_consume:                  \
-    case _Atomic_memory_order_acquire:                  \
-    case _Atomic_memory_order_seq_cst:                  \
+    _Check_load_memory_order(_Order_var);               \
+    if (_Order_var != memory_order_relaxed) {           \
         _Compiler_or_memory_barrier();                  \
-        break;                                          \
     }
-
-#define _ATOMIC_STORE_PREFIX(_Width, _Ptr, _Desired)      \
-    case _Atomic_memory_order_relaxed:                    \
-        __iso_volatile_store##_Width((_Ptr), (_Desired)); \
-        return;                                           \
-    case _Atomic_memory_order_release:                    \
-        __STORE_RELEASE(_Width, _Ptr, _Desired);          \
-        return;                                           \
-    case _Atomic_memory_order_consume:                    \
-    case _Atomic_memory_order_acquire:                    \
-    case _Atomic_memory_order_acq_rel:                    \
-    default:                                              \
-        _INVALID_MEMORY_ORDER;                            \
-        _FALLTHROUGH;
 
 #define _ATOMIC_STORE_SEQ_CST_ARM(_Width, _Ptr, _Desired) \
     _Memory_barrier();                                    \
@@ -730,11 +697,15 @@ struct _Atomic_storage<_Ty, 1> { // lock-free using 1-byte intrinsics
     void store(const _TVal _Value, const memory_order _Order) noexcept { // store with given memory order
         const auto _Mem      = _STD _Atomic_address_as<char>(_Storage);
         const char _As_bytes = _STD _Atomic_reinterpret_as<char>(_Value);
-        switch (static_cast<unsigned int>(_Order)) {
-            _ATOMIC_STORE_PREFIX(8, _Mem, _As_bytes)
-        case _Atomic_memory_order_seq_cst:
+
+        _Check_store_memory_order(_Order);
+
+        if (_Order == memory_order_relaxed) {
+            __iso_volatile_store8(_Mem, _As_bytes);
+        } else if (_Order == memory_order_release) {
+            __STORE_RELEASE(8, _Mem, _As_bytes);
+        } else {
             store(_Value);
-            return;
         }
     }
 
@@ -743,10 +714,10 @@ struct _Atomic_storage<_Ty, 1> { // lock-free using 1-byte intrinsics
         const auto _Mem = _STD _Atomic_address_as<char>(_Storage);
         char _As_bytes;
 #if _STD_ATOMIC_USE_ARM64_LDAR_STLR == 1
-        _ATOMIC_LOAD_ARM64(_As_bytes, 8, _Mem, static_cast<unsigned int>(_Order))
+        _ATOMIC_LOAD_ARM64(_As_bytes, 8, _Mem, _Order)
 #else // ^^^ _STD_ATOMIC_USE_ARM64_LDAR_STLR == 1 / _STD_ATOMIC_USE_ARM64_LDAR_STLR == 0 vvv
         _As_bytes = __iso_volatile_load8(_Mem);
-        _ATOMIC_POST_LOAD_BARRIER_AS_NEEDED(static_cast<unsigned int>(_Order))
+        _ATOMIC_POST_LOAD_BARRIER_AS_NEEDED(_Order)
 #endif // ^^^ _STD_ATOMIC_USE_ARM64_LDAR_STLR == 0 ^^^
         return reinterpret_cast<_TVal&>(_As_bytes);
     }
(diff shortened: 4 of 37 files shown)


stl/inc/atomic
```diff
@@ -138,18 +138,17 @@ extern "C" inline void _Check_memory_order(const unsigned int _Order) noexcept {
     case _Atomic_memory_order_relaxed:                        \
         _Result = __iso_volatile_load##_Width(_Ptr);          \
         break;                                                \
+    case _Atomic_memory_order_release:                        \
+    case _Atomic_memory_order_acq_rel:                        \
+    default:                                                  \
+        _INVALID_MEMORY_ORDER;                                \
+        _FALLTHROUGH;                                         \
```
AlexGuteniev, line 145: Can be structured as:
```C++
if (_Order_var == _Atomic_memory_order_release || _Order_var == _Atomic_memory_order_acq_rel) {
     _INVALID_MEMORY_ORDER;
}

if (_Order_var == _Atomic_memory_order_relaxed) {
     _Result = __iso_volatile_load##_Width(_Ptr);
} else {
     _Result = __LOAD_ACQUIRE_ARM64(_Width, _Ptr);         \
     _Compiler_barrier();    
}
```

This avoids falling through, and likely to improve codegen in some cases, as #2110.
Occurs below.

StephanTLavavej: I am moderately persuaded by fallthrough being somewhat gross, weakly persuaded by possible codegen improvements (I don't have good intuition or ability to measure that), but strongly persuaded by `_Check_load_memory_order` already having centralized the ability to distinguish good kitties :smile_cat: from bad kitties :smirk_cat:. I'll audit the header, thanks.

StephanTLavavej: I got rid of as many `_FALLTHROUGH`s as I felt was reasonable here (the remaining ones seemed justified). Completely eliminating the abominable `_ATOMIC_STORE_PREFIX` made me very happy, thanks.

stl/inc/expected
```diff
@@ -621,39 +621,39 @@ public:
     // [expected.object.obs]
     _NODISCARD constexpr const _Ty* operator->() const noexcept {
 #if _CONTAINER_DEBUG_LEVEL > 0
-        _STL_VERIFY(_Has_value, "expected stores an error, not a value");
-#endif // _CONTAINER_DEBUG_LEVEL > 0
+        _STL_VERIFY(_Has_value, "operator->() called on a std::expected that stores an error, not a value");
```
sivadeilra, line 624: Minor suggestion: "contains" instead of "stores".  "Store" to me usually implies a memory write.  "Contains" is more familiar, especially since this is a container.

stl/inc/list
```diff
@@ -143,7 +143,7 @@ public:
     _NODISCARD reference operator*() const noexcept {
 #if _ITERATOR_DEBUG_LEVEL == 2
         const auto _Mycont = static_cast<const _Mylist*>(this->_Getcont());
-        _STL_ASSERT(_Mycont, "cannot dereference value-initialized list iterator");
+        _STL_VERIFY(_Mycont, "cannot dereference value-initialized list iterator");
```
sivadeilra, line 146: suggest `_Mycont != nullptr` instead of just `_Mycont`

stl/inc/span
```diff
@@ -432,7 +437,7 @@ public:
 #if _CONTAINER_DEBUG_LEVEL > 0
         _STL_VERIFY(_Mysize <= dynamic_extent / sizeof(element_type),
```
sivadeilra, line 438: Is this check redundant with similar checks in the constructor?  Is it even possible to construct a span that would violate this?

sivadeilra (approved):

docs/import_library.md
```diff
@@ -41,7 +41,7 @@ The caveats of this technique are:
   + This limitation is subtle (not readily apparent from the source code) and critical.
     If shared global state is necessary, our only option while preserving bincompat is adding a satellite DLL.
 * Due to having just two flavors of the import library (debug and release),
-  we cannot use anything that depends on `_CONTAINER_DEBUG_LEVEL` or `_ITERATOR_DEBUG_LEVEL`.
```
dmachaj, line 44: Out of curiosity, is the removal of `_CONTAINER_DEBUG_LEVEL` possibly a source-breaking change?  If anyone was assuming it is defined without explicitly ifdef checking it.  (This seems like a bad practice and probably ok to break).

stl/inc/bitset
```diff
@@ -132,7 +132,7 @@ private:
     static constexpr unsigned long long _Mask = (1ULL << (_Need_mask ? _Bits : 0)) - 1ULL;
 
 public:
-    _NODISCARD constexpr bool operator[](size_t _Pos) const noexcept /* strengthened */ {
+    _NODISCARD constexpr bool operator[](const size_t _Pos) const noexcept /* strengthened */ {
```
dmachaj, line 135: This is pass by value.  Is the `const` going to change behavior in any way?

stl/inc/memory
```diff
@@ -1042,18 +1049,10 @@ public:
     }
 
     _NODISCARD _Ty& operator*() const noexcept {
-#if _ITERATOR_DEBUG_LEVEL == 2
```
dmachaj, line 1045: I don't think this change made it into the PR description.  It seems unexpected to be removing a safety check.

stl/inc/xtree
```diff
@@ -178,7 +178,7 @@ public:
     _NODISCARD reference operator*() const noexcept {
 #if _ITERATOR_DEBUG_LEVEL == 2
         const auto _Mycont = static_cast<const _Mytree*>(this->_Getcont());
-        _STL_ASSERT(_Mycont, "cannot dereference value-initialized map/set iterator");
+        _STL_VERIFY(_Mycont, "cannot dereference value-initialized map/set iterator");
```
dmachaj, line 181: The other adjusted error messages make me wonder if this should be something like `operator*() called on a value-initialized iterator`.

stl/inc/list
StephanTLavavej: Thank you, but our stylistic conventions both allow and often prefer directly testing pointers for non-nullness. (It's shorter, avoids the negation in `!=`, and reads naturally; an expression like `if (ptr && ptr->whatever)` reads as "if we have `ptr` and its `whatever` is true...")

`_STL_VERIFY` is well-known to us as testing its condition, so I view this context as equivalent to an `if` statement.

(But conversely, our stylistic conventions do not allow directly testing integers for non-zeroness, except in special circumstances.)

stl/inc/span
StephanTLavavej: It is difficult, but possible, to violate this precondition. We have a "death test" for this check here:

https://github.com/microsoft/STL/blob/fc15609a0f2ae2a134c34e7c9a13977994f37367/tests/std/tests/P0122R7_span_death/test.cpp#L263-L266

We should probably move this check into every constructor. I've already written the hardening changes and this will be tangled up with them (the overflow precondition shouldn't be part of hardening, but it should interact with the `if constexpr` for static vs. dynamic extents), so instead of altering this prelude PR, I've recorded a todo and will perform this enhancement in the hardening PR as a special exception. Thanks!

stl/inc/memory
StephanTLavavej: It's in the PR description:

> * Enhancement: Remove IDL=2 null asserts from `auto_ptr` deref/arrow.
>   + It's silly to assert that `_Myptr` is non-null when we're about to dereference it (or return it for imminent dereferencing). We purged this category of asserts from the STL aeons ago, but must have missed `auto_ptr`. The idea is that null dereferences are just as obvious in the debugger. Note that this makes the code consistent with `unique_ptr` and `shared_ptr`.

It is not a safety check:

* Null dereferences will terminate program execution in all modes. Thou shalt not follow the null pointer, for chaos and madness await thee at its end.
* This is an IDL=2 check, which is debug mode only. (IDL=2 cannot be enabled in release mode.)

As an aside, I removed `auto_ptr` in the C++17 Standard: WG21-N4190. (There's an escape hatch to restore it in our implementation, though.)

dmachaj: Ah, didn't look closely enough.  Yeah `auto_ptr` is very deprecated at this point.

docs/import_library.md
StephanTLavavej: In theory, yes. We currently always define the CDL macro (if not already defined), and after my following PR we'll never define the macro. If someone was inspecting the CDL macro, this could affect them.

A few notes:

* CDL was intended to be a control macro, not an observation macro. Users were intended to define CDL to opt-in, but only the STL would be checking the macro. Aside from CDL's virtually-zero usage, I can't imagine why users would try to observe the macro.
* The off-by-default [warning C4668](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4668?view=msvc-170) will complain "'symbol' is not defined as a preprocessor macro, replacing with '0' for 'directives'". Although this is Standard-mandated behavior, this warning is quite useful for detecting typos, and the STL enables it throughout its test suites so that we can remain clean with respect to it. (i.e. We never say `#if MACRO` for a potentially-undefined macro.) This warning could be used to find occurrences of `#if CDL` when it's no longer defined by the STL.
* In the following PR, I will be explicitly rejecting any definitions of CDL from users, so they'll be informed of the change if they were attempting to opt-in.

stl/inc/bitset
StephanTLavavej: No, this change is completely unobservable to callers. WG21-N5001 \[dcl.fct\]/3:

> After producing the list of parameter types, any top-level *cv-qualifier*s modifying a parameter type are deleted when forming the function type.

Top-level `const` on value parameters is near-immediately ignored for function declarations. For function definitions, it affects the implementation, but in both cases it doesn't affect callers and doesn't affect the function type. (So the function's mangled name is unchanged, and this is ABI-invisible.)

We routinely add top-level `const` to STL functions and this has never caused trouble.

stl/inc/xtree
StephanTLavavej: Yeah, that would be consistent.

Unlike the continue-on-error audit, where I went through every single `_STL_VERIFY` / `_STL_ASSERT` / etc. in the STL, for the debug messages I audited only those that are shared by checks that will be part of STL Hardening. (I realize that I didn't mention the scope of this cleanup in the PR description.)

Since I'll be adjusting other messages, I can do a quick review and clean some of these up. There are ~700 such checks in the STL so I don't want to promise a complete cleanup here. :joy_cat:

stl/inc/expected
StephanTLavavej: That's reasonable, I'll make that change. `expected` is not formally a container, but it plays one on TV. :tv:

stl/inc/xtree
dmachaj: Yeah that seems like a lot of checks.  I commented because this line was modified so it seemed like fair game to ask about.

stl/inc/bitset
dmachaj: I agree that this won't cause any trouble.  I was moreso asking if there's any purpose/benefit.  It prevents you from reusing the input parameter as another local variable but that's rarely something anyone tries to do.

stl/inc/xtree
StephanTLavavej: After searching for `_STL_VERIFY|_STL_ASSERT`, there are a ton of messages that could be rephrased. I'll record a personal todo but I'm going to rule that this is out of scope for this PR. I'll adjust the PR description to note the scope here.

stl/inc/bitset
jovibor: @dmachaj You can consider this as yet another function-member-variable. It just happened to be declared in the function argument list. But the rules of `const/non const` absolutely the same as for variables within the function.
This approach is used throughout the STL.

StephanTLavavej: > It prevents you from reusing the input parameter as another local variable but that's rarely something anyone tries to do.

It's low risk, but it's also cheap to prevent. There are two other reasons that we love marking value parameters and local variables as `const` whenever possible:

* It allows us to avoid worrying about whether the state will later change.
* When consistently followed (we aren't perfect about this but we try, especially in new code), the *lack* of `const` begins to become a signal that the parameter/variable's value will be altered later, so we should pay special attention to it.

stl/inc/expected
StephanTLavavej: Done, thanks!

stl/inc/span
StephanTLavavej: I've implemented and tested this change - it's a nice improvement.

It wasn't as tangled up with hardening as I feared - I found that it was better to perform in `span`'s base class which has different cases for static vs. dynamic extents. However, it's still slightly tangled up with CDL removal and test changes. Although I could backport it to the prelude PR if necessary, rebasing the changes would be annoying enough that I still plan to leave it in the next PR.

Thanks again!

stl/inc/atomic
davidmrdavid: What's the reasoning behind 'falling through' here, instead of just breaking (like before)?

stl/inc/chrono
```diff
@@ -1608,8 +1608,12 @@ namespace chrono {
             // Returns the number of fractional digits of _Num / _Den in the range [0, 18].
             // If it can't be represented, 6 is returned.
             // Example: _Fractional_width(1, 8) would return 3 for 0.125.
+
+            _STL_INTERNAL_STATIC_ASSERT(_Duration::period::num > 0); // N5001 [time.duration.general]/3
+            _STL_INTERNAL_STATIC_ASSERT(_Duration::period::den > 0); // N5001 [ratio.ratio]/1, /2.2
+
             auto _Den = _Duration::period::den;
-            _STL_ASSERT(_Duration::period::num > 0 && _Den > 0, "Numerator and denominator can't be less than 1.");
+
```
davidmrdavid, line 1616: nit - possibly unintended newline?

stl/inc/yvals.h
```diff
@@ -181,35 +181,15 @@ _STL_DISABLE_CLANG_WARNINGS
 #endif // ^^^ !defined(_DEBUG) ^^^
 #endif // !defined(_STL_CRT_SECURE_INVALID_PARAMETER)
 
-#define _STL_REPORT_ERROR(mesg)                  \
-    do {                                         \
-        _RPTF0(_CRT_ASSERT, mesg);               \
-        _STL_CRT_SECURE_INVALID_PARAMETER(mesg); \
-    } while (false)
```
davidmrdavid, line 188: I'm a bit lost as to why we had a `do { .. } while (false)` in the first place (it will never loop, after all). Either I'm missing something basic, or there's some wizardry at play. I'll reach out internally.

davidmrdavid (commented):
I've completed a first pass. Thanks _a lot_, again, for the pristine commit history, it made this a lot easier. I've left some comments, most are non-blockers and others reveal bits I don't fully grok yet. I'll look to resolve these by tomorrow so I can unblock you sooner rather than later. 

_Excited for hardening!_ 💎 💎 💎

stl/inc/array
```diff
@@ -708,15 +708,15 @@ public:
 
     _NODISCARD reference operator[](size_type) noexcept /* strengthened */ {
 #if _CONTAINER_DEBUG_LEVEL > 0
-        _STL_REPORT_ERROR("array subscript out of range");
+        _STL_REPORT_ERROR("array<T, 0> subscript invalid");
```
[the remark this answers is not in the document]
StephanTLavavej: > (1) I'm personally a fan of error messages that suggest a course of action / a possible fix.
> For instance, I wonder if the error should read "array<T, 0> subscript is invalid. To fix this, provide a non-zero size for the array"

I agree that it's desirable to suggest a course of action when we can do so with high confidence. (For example, some of our deprecation messages do this.)

But in this case, I don't think that "provide a non-zero size for the array" is a high-confidence suggestion. That's saying "actually, you should have stored some elements here", but how can we possibly know that in the library? It's more likely that they're iterating through their array in an undisciplined way (e.g. in a way that would have been wrong for a non-zero-size array too), or that they should have guarded their subscript with an index check or `!arr.empty()`, etc.

For this reason I would prefer to suggest nothing here.

> (2) I feel this error message should include a verb :) . So I like: "array<T, 0> subscript is invalid".

That's definitely reasonable. We're often very terse in comments, but user-visible messages should make more of an effort. I'll do another audit.

stl/inc/span
```diff
@@ -351,11 +351,11 @@ public:
     template <size_t _Count>
     _NODISCARD constexpr auto first() const noexcept /* strengthened */ {
         if constexpr (_Extent != dynamic_extent) {
-            static_assert(_Count <= _Extent, "Count out of range in span::first()");
+            static_assert(_Count <= _Extent, "Count out of range in span::first<Count>()");
```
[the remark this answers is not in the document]
StephanTLavavej: It's mostly stylistic. The main reason to make this change is to help point the user at the overload they should be looking at, since we've got a compile-time and a run-time overload here.

stl/inc/__msvc_ranges_tuple_formatter.hpp
```diff
@@ -324,7 +324,7 @@ class basic_format_arg {
         case _Basic_format_arg_type::_Custom_type:
             return _STD forward<_Visitor>(_Vis)(_Custom_state);
         default:
-            _STL_VERIFY(false, "basic_format_arg is in impossible state");
+            _STL_VERIFY(false, "basic_format_arg is in an impossible state");
```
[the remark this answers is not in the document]
StephanTLavavej: This probably ought to be an `_STL_INTERNAL_CHECK`, actually, since (according to my understanding) only an STL logic error could lead to this.

I don't think there's more information to provide here, certainly not anything that would be actionable to a user. I could rephrase it to "contains an impossible type", I suppose.

stl/inc/atomic
```diff
@@ -2328,8 +2328,7 @@ public:
 
     explicit atomic_ref(_Ty& _Value) noexcept /* strengthened */ : _Base(_Value) {
         if constexpr (is_always_lock_free) {
-            _ATOMIC_REF_CHECK_ALIGNMENT(
-                (reinterpret_cast<uintptr_t>(_STD addressof(_Value)) & (required_alignment - 1)) == 0,
+            _STL_ASSERT((reinterpret_cast<uintptr_t>(_STD addressof(_Value)) & (required_alignment - 1)) == 0,
```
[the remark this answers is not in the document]
StephanTLavavej: That would unconditionally check, including in release mode.

Perhaps this would make sense as a special exception. Presumably, `atomic_ref` construction is infrequent, and if the alignment precondition isn't met, catastrophic things *could* happen later. There are also some sneaky cases where natural code won't achieve the desired alignment (x86's stack is 4-aligned, and `double` only has a "suggested" 8-alignment in the compiler, despite the fact that its `alignof` is reported as 8).

Currently this cleanup is preserving the existing behavior, so I would prefer to make no further change in this PR. I've recorded a personal todo to consider unconditional checking here.

stl/inc/memory
```diff
@@ -97,20 +97,26 @@ namespace ranges {
...
-                    _STL_ASSERT(false, "Tried to uninitialized_copy two ranges with unreachable sentinels");
                 }
             } else {
+                if constexpr (_Can_memcpy) {
+                    // We were eligible for the memcpy optimization above, except for both sentinels being unreachable.
+                    // The following classic code is doomed, because no exceptions will end the infinite loop.
+                    // Following our usual pattern, let's emit a debug assertion, then run the loop anyways.
+                    _STL_ASSERT(false, "Tried to uninitialized_copy two ranges with unreachable sentinels");
+                }
+
```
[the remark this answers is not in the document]
StephanTLavavej: > (1) should we say Tried to std::uninitialized_copy() ... instead of just 'Tried to uninitialized_copy'. Qualifying 'uninitialized_copy' helps me quickly identify we're talking about a particular function call going wrong.

I can add parentheses. (We usually do this when mentioning functions.) Many of our debug messages omit `std::` for "well-known" names (with the exception that `std::` qualifying names like `string` that are also common English words can really help improve clarity). I personally consider `uninitialized_copy()` to be a well-known Standard Library function - but my brain has also been irradiated by the Standard for decades. If it was unclear to you, let's qualify it.

I will also update the `uninitialized_move` message.

> (2) I'll admit - I don't think I fully grok how exactly we get into this case, and the implications. Would love to sync internally.

Sure, ask me next week during office hours, I'll be happy to walk you through the control flow and what we're doing here (it's quite elaborate).

stl/inc/atomic
StephanTLavavej: Previously, this executed `_Result = __iso_volatile_load##_Width(_Ptr);` (same as the `_Atomic_memory_order_relaxed` case, which is the weakest) before diagnosing `_INVALID_MEMORY_ORDER` followed by `break;`.

That was a less safe, more risky approach. After diagnosing a bogus memory order, performing the *weakest* operation could lead to the program observing unexpected results. Instead, if we perform the strongest operation, that of full sequential consistency, we're basically saying "look, you gave us a bogus memory order, we told you about the precondition violation, and now we're going to do the slowest but safest operation". (Atomic memory order is a massive subject that I could explain at some length, but the half-minute summary is that full sequential consistency is the ordering that programmers usually expect, when considering all possible interleavings of execution. Weaker-than-SC memory orders require very careful reasoning.)

stl/inc/chrono
StephanTLavavej: I was being clever by introducing a newline to avoid clang-format aligning the `=`, but these are all modifiable variables and grouping them together does make sense. I'll remove it.

stl/inc/yvals.h
StephanTLavavej: Yep, that's it exactly. I mentioned this in the PR description (very tersely):

> We don't need this because we ALWAYS use braces for control flow.

I think the `do { ... } while (false)` might actually have been introduced during an earlier era where our formatting was radically different and did contain unbraced control flow. Fortunately, this is one of the syntax things that I feel the strongest about (it led to a major bug in the real world!), and we converted over the entire codebase before open-sourcing.

stl/inc/atomic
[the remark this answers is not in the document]
AlexGuteniev: I don't think `atomix_ref` constructor is neccessarily infrequent. For exsmple for a loop.over an array of values it will be created on every iteration.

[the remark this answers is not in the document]
StephanTLavavej: That's a good reason to keep the debug-only status quo then, thanks.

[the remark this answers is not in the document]
AlexGuteniev: In ideal world I would expect this to vectorize, because why not, but otherwise I'd expect this loop not to contain extra steps. And this is a typical usage of `atomic_ref`, according to its proposal.
```c++
void clear_each_integer_atomically(int* data, std::size_t size) {
    for (std::size_t i = 0; i != size; ++i) {
       std::atomic_ref<int> r(data[i]);
       r.store(0, std::memory_order_relaxed);
    }
}
```

stl/inc/memory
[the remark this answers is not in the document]
davidmrdavid: >Sure, ask me next week during office hours, I'll be happy to walk you through the control flow and what we're doing here (it's quite elaborate).

Sounds great. After this, I should be ready to approve.

davidmrdavid (approved):

StephanTLavavej:

Issue #5282, opened 2025-02-14
CUDA's installer hangs for VS 2022 17.13 and later


StephanTLavavej:
Known issue: https://forums.developer.nvidia.com/t/stuck-installing-nsight-on-cuda-12-8-0-571-96-windows/323732/2

I encountered this when updating the toolset (from Preview 2) to VS 2022 17.13 Preview 3 in #5247 but didn't understand the cause then.

Unfortunately, my "solution" of installing just 'nvcc_12.4' didn't work, but I didn't realize it. `nvcc` was no longer being found on the path, causing the test harness to skip `GH_000639_nvcc_include_all`, and I didn't think to look for this.

I can't figure out how `nvcc` is normally added to the developer command prompt path (i.e. if there's a particular subpackage that will do it, or only the full installation). As a workaround, I'm going to try to manually add it to the path. I also need to install the CUDA runtime libraries.

This is a tracking issue to remember what's going on here. I'm not sure if this will be fixed on the CUDA side or the VS side. Either way, my workaround should be removed at that time, and if CUDA is the one that's fixed, we should strongly consider increasing our minimum required CUDA version.

PR #5284, opened 2025-02-15
Toolset update: VS 2022 17.14 Preview 1

StephanTLavavej:
# :scroll: Changelog

- Code cleanups:
  * Updated dependencies.
    + Updated build compiler to VS 2022 17.14 Preview 1.
    + Updated Google Benchmark to 1.9.1.
    + Updated Python to 3.13.2.

# :gear: Commits

* VS 2022 17.14 Preview 1.
* Remove workaround for DevCom-10373274 "vcruntime nothrow array operator new falls back on the wrong function".
* Remove workaround for DevCom-10808176 "MSVC doesn't properly destroy a loop variable in constant evaluation".
  + However, `mdspan/index_operator.pass.cpp` is still failing due to a likely MSVC constexpr bug, "note: failure was caused by a read of a variable outside its lifetime".

# :phone: `static operator()` Commits

All of MSVC/Clang/EDG now support C++23 WG21-P1169R4 `static operator()`, so let's update the codebase:

* Remove `#ifdef __cpp_static_call_operator` from `P1169R4_static_call_operator`.
  + This uses `usual_latest_matrix.lst`.
* Guard `std::function` CTAD for static call operators with `_HAS_CXX23`.
  + There's a question of whether we should attempt to expand this CTAD downlevel, to match how static call operators are supported downlevel. It would probably work, even for CUDA (which I believe currently lacks support for static call operators, but also doesn't appear to reject the SFINAE syntax here on sight). However, `std::function` CTAD is a "nice to have" feature, and we aren't obligated to support it downlevel. Restricting it to C++23 is less risky, and completely avoids any question of disrupting CUDA (as CUDA 12.4 supports only up to C++20).
* Suppress MSVC Future Technology warnings C5280 and C5281.
  + Clang uses the same warning option with different messages; record both.
* Use static call operators downlevel, except for CUDA.
  + In C++14 mode, as of VS 2022 17.14 Preview 1 and Clang 19.1.1:
    - MSVC does NOT define the feature-test macro, but accepts the code (with a C5280 warning).
    - Clang DOES define the feature-test macro, and accepts the code (with a `-Wc++23-extensions` warning).
    - EDG does NOT define the feature-test macro, but accepts the code silently.
  + For Intel's compiler, which we don't support but which we do avoid gratuitously breaking, my experiments on Compiler Explorer indicate that they have Clang's behavior.
  + CUDA 12.4 was released in March 2024, and I suspect that their EDG version doesn't support this yet.
  + Additionally, split the macro into `_STATIC_LAMBDA` to work around VSO-2383148 "C++23 static lambdas aren't accepted by the old lambda processor". This means that they aren't accepted in C++14/17, without enabling `/permissive-` (strict mode) or `/Zc:lambda` (new processor). This also means that they aren't accepted in C++20/23 when `/Zc:lambda-` (old processor) is used. (Note: C++20/23 with `/permissive` (permissive mode) still uses the new processor, and therefore works.) Because none of this is Standard-mandated (it's a "nice to have" minor performance improvement), we should just avoid static lambdas with MSVC while the bug remains active.
    - This will be fixed in 17.14 Preview 3 by MSVC-PR-611685.
* Enable libcxx tests for MSVC.
  + However, `range_adaptor_closure.pass.cpp `is blocked by a compiler bug affecting ranges.

# :gear: More Commits

* Drop the `wmemchr()` optimization for `unreachable_sentinel_t`.
  + Fixes `GH_002431_byte_range_find_with_unreachable_sentinel`, which regressed either with VS 2022 17.14 Preview 1
  or the newer Win11 SDK 26100 that I had unintentionally installed briefly; I didn't care enough to find out what was the cause. (Our README currently recommends installing the older Win11 SDK 22621 because that's what the MS-internal build currently uses.)
* Google Benchmark 1.9.1.
  + No MS-internal changes needed.
* 2025-datacenter-azure-edition.
  + The extra crispy flavor is said to be more efficient than the original recipe. It doesn't seem to make a difference for us, but it seems ideal to use.
* Python 3.13.2.
* Send `create-1es-hosted-pool.ps1` output to a log file.
  + I'm updating the `.gitignore` to ignore `*.log` anywhere in the tree, not just the root. (We don't have any such files in the repo.)
  + This captures output (via appending to ` >> $LogFile`) that I was previously piping to ` | Out-Null` to reduce script noise. I've had some trouble with sporadic failures and this should make it easier to diagnose in the future.
* Fix the CUDA installation.
  + In #5247, I encountered a CUDA installer hang, #5282. While working around it, I didn't notice that `nvcc` was no longer being added to the path, so the test harness was just skipping that test. And I didn't notice that we need to install `cudart` for headers that it wants to include.
  + I don't know how to get the installer to update the path when installing selected subpackages (I tried the "VS integration" one and it wasn't sufficient, and trying different subsets is extremely time-consuming). I'm just going to manually update the path.
  + I've verified that all 6 configurations of `GH_000639_nvcc_include_all` are now passing instead of being skipped.
* New pool.
* Teach `print_failures.py` to understand XPASS in XML.
  + This is parsing:
    ```xml
    <testcase classname="libc++.std/utilities/template_bitset/bitset_members" name="left_shift_eq.pass.cpp" time="8.86">
        <failure><![CDATA[]]></failure>
    </testcase>
    ```
  + It handles XPASS in JSON just fine.
* Fix STL-ASan-CI failures by upgrading `bitset.members` tests from FAIL to SKIPPED.
  + These tests are stressing `constexpr` step limits. One started passing for ASan, which makes little sense. All of the tests in this directory are similar, so skip 'em.
* Expand `memchr`/`wmemchr` comments, fix `__std_find_trivial_unsized_2`.

:white_check_mark: [STL-ASan-CI passed.](https://dev.azure.com/vclibs/STL/_build/results?buildId=18252&view=results)

# :infinity: Commits Forever

* Work around the MSVC-internal toolset being 17.12 and not supporting static call operators.
* Add an MSVC-internal skip for a test with `ADDITIONAL_COMPILE_FLAGS(cl-style-warnings)`.
  + This test was otherwise re-enabled after VSO-2319111 was fixed.
* Cite CWG-1656 with our usual style, improve wrapping.
  + Thanks @frederick-vs-ja!
* Code review feedback: clarify `memchr`/`wmemchr` comments.
  + Thanks @davidmrdavid!

### DIFF
--- a/stl/inc/xutility
+++ b/stl/inc/xutility
@@ -6216,7 +6216,9 @@ namespace ranges {
                         _Count = SIZE_MAX;
                     }
 
-                    // C23 7.32.4.6.9 "The wmemchr generic function"/2 doesn't work with unreachable_sentinel_t.
+                    // C23 7.27.5.2 "The memchr generic function"/2 says "The implementation shall behave as if
+                    // it reads the characters sequentially and stops as soon as a matching character is found."
+                    // C23 7.32.4.6.9 "The wmemchr generic function"/2 lacks such wording, so we don't use wmemchr().
                     _Result = static_cast<_Ptr_t>(_CSTD memchr(_First_ptr, static_cast<unsigned char>(_Val), _Count));
 
                     if constexpr (_Is_sized) {
--- a/stl/inc/xutility
+++ b/stl/inc/xutility
@@ -6218,7 +6218,8 @@ namespace ranges {
 
                     // C23 7.27.5.2 "The memchr generic function"/2 says "The implementation shall behave as if
                     // it reads the characters sequentially and stops as soon as a matching character is found."
-                    // C23 7.32.4.6.9 "The wmemchr generic function"/2 lacks such wording, so we don't use wmemchr().
+                    // C23 7.32.4.6.9 "The wmemchr generic function"/2 lacks such wording,
+                    // so we don't use wmemchr(), avoiding issues with unreachable_sentinel_t.
                     _Result = static_cast<_Ptr_t>(_CSTD memchr(_First_ptr, static_cast<unsigned char>(_Val), _Count));
 
                     if constexpr (_Is_sized) {
--- a/stl/inc/xutility
+++ b/stl/inc/xutility
@@ -6179,12 +6179,12 @@ namespace ranges {
     template <input_iterator _It, sentinel_for<_It> _Se, class _Ty, class _Pj = identity>
         requires indirect_binary_predicate<ranges::equal_to, projected<_It, _Pj>, const _Ty*>
     _NODISCARD constexpr _It _Find_unchecked(_It _First, const _Se _Last, const _Ty& _Val, _Pj _Proj = {}) {
-        constexpr bool _Elements_are_1_or_2_bytes = sizeof(_Iter_value_t<_It>) <= 2;
-        constexpr bool _Is_sized                  = sized_sentinel_for<_Se, _It>;
+        constexpr bool _Elements_are_1_byte = sizeof(_Iter_value_t<_It>) == 1;
+        constexpr bool _Is_sized            = sized_sentinel_for<_Se, _It>;
 
         if constexpr (_Vector_alg_in_find_is_safe<_It, _Ty>
-                      && (_Elements_are_1_or_2_bytes ? _Is_sized || same_as<_Se, unreachable_sentinel_t>
-                                                     : _Is_sized && _USE_STD_VECTOR_ALGORITHMS)
+                      && (_Elements_are_1_byte ? _Is_sized || same_as<_Se, unreachable_sentinel_t>
+                                               : _Is_sized && _USE_STD_VECTOR_ALGORITHMS)
                       && same_as<_Pj, identity>) {
             if (!_STD is_constant_evaluated()) {
                 if (!_STD _Could_compare_equal_to_value_type<_It>(_Val)) {
@@ -6202,13 +6202,13 @@ namespace ranges {
                 _Ptr_t _Result;
 #if _USE_STD_VECTOR_ALGORITHMS
                 if constexpr (_Is_sized) {
-                    // When _Is_sized && _Elements_are_1_or_2_bytes, prefer this over memchr()/wmemchr() for performance
+                    // When _Is_sized && _Elements_are_1_byte, prefer this over memchr() for performance
                     const auto _Last_ptr = _First_ptr + (_Last - _First);
                     _Result              = _STD _Find_vectorized(_First_ptr, _Last_ptr, _Val);
                 } else
 #endif // ^^^ _USE_STD_VECTOR_ALGORITHMS ^^^
                 {
-                    _STL_INTERNAL_STATIC_ASSERT(_Elements_are_1_or_2_bytes);
+                    _STL_INTERNAL_STATIC_ASSERT(_Elements_are_1_byte);
                     size_t _Count;
                     if constexpr (_Is_sized) {
                         _Count = static_cast<size_t>(_Last - _First);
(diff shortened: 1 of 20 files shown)


stl/inc/xutility
```diff
@@ -6202,28 +6202,22 @@ namespace ranges {
...
 
-                    if constexpr (sizeof(_Iter_value_t<_It>) == 1) {
-                        _Result =
-                            static_cast<_Ptr_t>(_CSTD memchr(_First_ptr, static_cast<unsigned char>(_Val), _Count));
-                    } else {
-                        _STL_INTERNAL_STATIC_ASSERT(sizeof(_Iter_value_t<_It>) == 2);
-                        _Result = reinterpret_cast<_Ptr_t>(const_cast<wchar_t*>(_CSTD wmemchr(
-                            reinterpret_cast<const wchar_t*>(_First_ptr), static_cast<wchar_t>(_Val), _Count)));
-                    }
+                    // C23 7.32.4.6.9 "The wmemchr generic function"/2 doesn't work with unreachable_sentinel_t.
```
AlexGuteniev, line 6219: This seems too indirect to me. Maybe link `memchr` behavior quote instead?
```c
// C23 7.27.5.2 "The memchr generic function"/2 as if reads the characters sequentially.
// wmemchr does not have to do that

Comment (AlexGuteniev) [reply stl/inc/xutility:6219]: The same incorrect assumption about `wmemchr` is also made in `vector_algorithms.cpp`, it is in a compat path for no longer existent vectorization.

Comment (StephanTLavavej) [reply stl/inc/xutility:6219]: I can update the comment and the `vector_algorithms.cpp` usage is super easy to fix - thanks, I totally missed that!

**Review by davidmrdavid — COMMENTED** [review]
Thanks for this work! 🦖 
Left some nits. After reaching consensus, I should be ready to approve.

File: stl/inc/xutility [hunk 2]
```diff
@@ -6216,7 +6216,9 @@ namespace ranges {
                         _Count = SIZE_MAX;
                     }
 
-                    // C23 7.32.4.6.9 "The wmemchr generic function"/2 doesn't work with unreachable_sentinel_t.
+                    // C23 7.27.5.2 "The memchr generic function"/2 says "The implementation shall behave as if
+                    // it reads the characters sequentially and stops as soon as a matching character is found."
+                    // C23 7.32.4.6.9 "The wmemchr generic function"/2 lacks such wording, so we don't use wmemchr().
```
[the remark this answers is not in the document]
StephanTLavavej: Excellent suggestion, thanks. I think there's a bit of ambiguity in that phrasing - is it:

* { we don't use wmemchr() } to { avoid issues with unreachable_sentinel_t }
* we don't { use wmemchr() to avoid issues with unreachable_sentinel_t }

I'll go with "we don't use wmemchr(), avoiding issues with unreachable_sentinel_t".

stl/src/vector_algorithms.cpp
```diff
@@ -4481,12 +4481,15 @@ extern "C" {
...
 const void* __stdcall __std_find_trivial_unsized_1(const void* const _First, const uint8_t _Val) noexcept {
+    // C23 7.27.5.2 "The memchr generic function"/2 says "The implementation shall behave as if
+    // it reads the characters sequentially and stops as soon as a matching character is found."
     return memchr(_First, _Val, SIZE_MAX);
 }
 
 // TRANSITION, ABI: preserved for binary compatibility
 const void* __stdcall __std_find_trivial_unsized_2(const void* const _First, const uint16_t _Val) noexcept {
-    return wmemchr(static_cast<const wchar_t*>(_First), static_cast<wchar_t>(_Val), SIZE_MAX);
+    // C23 7.32.4.6.9 "The wmemchr generic function"/2 lacks such wording, so we don't use wmemchr().
```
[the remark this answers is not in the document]
StephanTLavavej: It's a few lines above, but you have a good point that we usually don't expect comments to have *cross-function* context, and during maintenance functions can be reordered/modified/etc. I'll update.

StephanTLavavej:
Thanks! I've pushed a conflict-free merge with `main`, fixed issues found by a trial mirror to the MSVC-internal repo, and addressed CR feedback (all commits described at the bottom of the PR description).

StephanTLavavej:

davidmrdavid (approved):

PR #1544, opened 2020-12-18
Always use `if constexpr`, require CUDA 10.1 Update 2

StephanTLavavej:
Works towards #189.

For ease of reviewing, this is organized into separate commits (which will be squashed).

First, `yvals_core.h` now requires CUDA 10.1 Update 2. Note that VS 2019 was first supported by CUDA 10.1 RTW, so only it and CUDA 10.1 Update 1 are being blocked here. There were no objections when I discussed this planned policy change in the past few months.

Then, this replaces `_CONSTEXPR_IF` with `constexpr`. I looked for warning pragmas in the immediate vicinity and didn't see any. (In some cases, the conditions had been extracted into `constexpr bool`s, both to improve clarity and to avoid warnings; I left those alone.)

Next, this drops all of the `_HAS_IF_CONSTEXPR` logic, resulting in 2000+ lines of removed code. In a few cases, this results in clang-format unwrapping of `} else {`, but nothing major.

This also drops the definition of `_HAS_IF_CONSTEXPR` itself - as usual, we are assuming that users have wisely avoided taking dependencies on undocumented and unsupported STL-internal macros.

Finally, this removes `/D_HAS_IF_CONSTEXPR=0` test coverage, which was the primary way we avoided regressing these duplicate codepaths (with the old CUDA 9.2 coverage being a secondary way). These configurations were near-identical duplicates of adjacent configurations, so we weren't relying on them for unique coverage otherwise. (In a few cases, the adjacent configurations had accumulated some options that weren't applied to the `/D_HAS_IF_CONSTEXPR=0` duplicates, but this is not a concerning issue - the absence of such options is still being tested by all of the other configurations.)

As a result, this will speed up testing by some amount.

This PR does **not** attempt to replace any occurrences of tag dispatch with `if constexpr`; that can be saved for future cleanups. Also, this will probably cause merge conflicts with in-flight PRs, but they should be simple to resolve, and removing these duplicate codepaths will reduce our maintenance burden. It's time!

PR #1342, opened 2020-10-05
Replace __pragma with _Pragma

CdTCzech:
This change was already merged as #662, but later reverted in #675.
Resolves #659

PR #772, opened 2020-05-01
Fix CUDA warnings/errors, _Noinit shadowing, other cleanups

StephanTLavavej:
`<random>`
This was defining a doubly-nested `struct _Noinit`, shadowing the STL-wide machinery that we should have been using:
https://github.com/microsoft/STL/blob/f0f0b55d57e2c9b9da5206b7dbdd71f17e7a0ee3/stl/inc/yvals.h#L311-L313

`<istream>`
Rename `bool _Noinit` to `_Skip_init`, avoiding more shadowing.

`<yvals_core.h>`
CUDA 9.2 was activating `__has_cpp_attribute(msvc::known_semantics)` yet emitting `warning: attribute namespace "msvc" is unrecognized`. I've added a TRANSITION workaround.

`<memory>`
We were unconditionally using `__declspec(empty_bases)`, causing CUDA 9.2 to emit `error: unknown attribute "empty_bases"`. @CaseyCarter noticed that this had no effect (only one base can possibly be empty here), so we can simply remove `__declspec(empty_bases)`. (These control blocks are new to VS 2019 16.7, so this is extra safe.)

`<string>`
CUDA 9.2 was emitting a valid `warning: pointless comparison of unsigned integer with zero`. Because it doesn't support `if constexpr`, I've added a separate function, marked TRANSITION. Note for the perceptive: I left `// can hold -2^63 and 2^64 - 1, plus NUL` unchanged in the original function so we can restore it easily.

`<xlocmon>`
CUDA 9.2 noticed bogus code here - we `static_cast<size_t>` the `int` return value of `sprintf_s`, so `if (_Count < 0)` was always false!

`<xutility>`
CUDA 9.2 correctly complained about unreachable code here. I've adjusted the preprocessor/control flow to be simpler. This is a permanent improvement, not TRANSITION.

`<xstring>`
Unwrap two comments.

`<xlocmon>`
`<xlocnum>`
Avoid `&_Src[sizeof(_Src)]` and `&_Buf[_Count]`. (The former technically performs a forbidden dereferencing; the latter can too, depending on the value of `_Count`.) I considered using `_Src, _Src + sizeof(_Src)`, but figured the throughput/debug codegen of `begin/end` was acceptable in iostreams.

PR #5335, opened 2025-03-12
Toolset update: VS 2022 17.14 Preview 2, Clang 19.1.5

### LINKED ISSUE #5074: libcxx: Sporadic crash in `thread.thread.destr/dtor.pass.cpp`
#5073 encountered a sporadic crash in [`libcxx/test/std/thread/thread.threads/thread.thread.class/thread.thread.destr/dtor.pass.cpp`](https://github.com/llvm/llvm-project/blob/69ead949d08ff0bb8cbbf4f7143aaa6687830f6b/libcxx/test/std/thread/thread.threads/thread.thread.class/thread.thread.destr/dtor.pass.cpp).

[Test logs](https://dev.azure.com/vclibs/STL/_build/results?buildId=17863&view=results), x64:

```
Build steps:
Command: "C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Tools\MSVC\14.42.34430\bin\HostX64\x64\cl.EXE" "D:\a\_work\1\s\llvm-project\libcxx\test\std\thread\thread.threads\thread.thread.class\thread.thread.destr\dtor.pass.cpp" "-ID:\build\out\inc" "-ID:\a\_work\1\s\llvm-project\libcxx\test\support" "/nologo" "/Od" "/W4" "/w14061" "/w14242" "/w14265" "/w14582" "/w14583" "/w14587" "/w14588" "/w14749" "/w14841" "/w14842" "/w15038" "/w15214" "/w15215" "/w15216" "/w15217" "/w15262" "/sdl" "/WX" "/D_ENABLE_STL_INTERNAL_CHECK" "/bigobj" "/EHsc" "/MTd" "/std:c++latest" "/permissive-" "/utf-8" "/FImsvc_stdlib_force_include.h" "/wd4643" "/D_STL_CALL_ABORT_INSTEAD_OF_INVALID_PARAMETER" "/analyze:autolog-" "/Zc:preprocessor" "/wd6262" "-FeD:\build\tests\libcxx\std\thread\thread.threads\thread.thread.class\thread.thread.destr\Output\dtor.pass.cpp.dir\0\dtor.pass.exe" "-link" "-LIBPATH:D:\build\out\lib\amd64" "-LIBPATH:C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Tools\MSVC\14.42.34430\lib\x64" "/MANIFEST:EMBED"
Exit Code: 0 (0x0)
Standard Output:
--
dtor.pass.cpp
--

Intellisense response file steps:
Test setup steps:
Test steps:
Test step failed unexpectedly.
Command: "D:\build\tests\libcxx\std\thread\thread.threads\thread.thread.class\thread.thread.destr\Output\dtor.pass.cpp.dir\0\dtor.pass.exe"
Exit Code: 3221225477 (0xC0000005)
```

### LINKED ISSUE #5282: CUDA's installer hangs for VS 2022 17.13 and later

StephanTLavavej:
Resolves #5074. Resolves #5282. Resolves DevCom-10863472 VSO-2411436.

# :scroll: Changelog
- Fixed bugs:
  * Fixed compiler errors when using `<format>` in a CUDA project, by adding a compiler bug workaround.
- Code cleanups:
  * Updated dependencies.
    + Updated build compiler to VS 2022 17.14 Preview 2 (now required).
    + Updated Clang to 19.1.5.

# :gear: Commits

* Node.js 23.9.0.
  + Increasing our minimum required Node version for the Status Chart isn't necessary at this time, but I like to keep this updated on a semi-regular basis.
* Require VS 2022 17.14.
  + The MSVC-internal toolset has *finally* been updated, allowing us to require 17.14.
* Remove workaround for CUDA 9.2's lack of `if constexpr`.
  + I added this workaround in #772, merged 2020-05-01. We began using `if constexpr` unconditionally (requiring CUDA 10.1 Update 2) in #1544, merged 2021-01-05. This workaround could have been removed then. At least I noticed this 4 years later! I'm going back to always calling `_Integral_to_string()`, now enhanced with `if constexpr`.
* Remove workaround for VSO-612785 "'expect_error' assertion when testing a constexpr 'hidden friend'".
  + This was fixed for VS by EDG 5.0 on 2018-12-15. In #759 merged on 2020-04-30, I had to preserve the workaround because it was still needed for CUDA 9.2 (released 2018-05). Fortunately I commented it, although very poorly. It's been 5 years and we now require CUDA 12.4 (released 2024-03), so we can remove this workaround.
* Remove workaround for VSO-1132704 "Bogus C3615 when implicitly-constexpr defaulted SMF calls non-trivial SMF in base class".
* Remove workaround for VSO-1284799 "constexpr ICE: binding a reference to a temporary emits Assertion failed: Nerrors > 0".
* Remove workaround for VSO-1664293 "/clr C++20 chk assertion failed: rhs.is_lvalue() in constexpr.cpp" in `Dev11_0535636_functional_overhaul`.
  + Update citation to VSO-2417635 "/clr C++20 chk assertion failed: rhs.is_lvalue() in constexpr.cpp, again" in `P0466R5_layout_compatibility_and_pointer_interconvertibility_traits`.
* Remove workaround for VSO-2343282 "C1XX assertion 'If lookup found RDSymbol before it should find something this time, too' affecting STL atomic_ref test".
* Remove workaround now that the internal toolset supports static call operators.
* Remove workaround for DevCom-10841757 (CUDA's installer hangs for VS 2022 17.13 and later).
  + Goodbye #5282!
  + :white_check_mark: Verified that all 6 configurations of `GH_000639_nvcc_include_all` are passing instead of being skipped.
* Skip a sporadic libcxx failure.
  + Goodbye #5074!
* Work around VSO-2253317 "ICE when calling std::to_underlying()".
  + This is a very obscure bug that I encountered while modernizing the MSVC codebase to C++23. (It involves multiplying a `float` by the result of `to_underlying()`.) As a compiler fix hasn't arrived in a timely manner, temporarily removing our "MSVC intrinsic" annotation is easy.
* create-1es-hosted-pool.ps1: Opt-in early to disable default outbound access, [before it's retired on 2025-09-30](https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/default-outbound-access).
  + I verified that this takes effect (i.e. with no other changes, this prevents outbound Internet access, so the script can't download anything and totally fails).
* create-1es-hosted-pool.ps1: Create a NAT gateway.
  + This is the most preferred method for explicit outbound connectivity, [according to Azure docs](https://learn.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-connections#scenarios).
* create-1es-hosted-pool.ps1: Consistently indent wrapped options.
* create-1es-hosted-pool.ps1: Add more progress bar messages.
* New pool.
* VS 2022 17.14 Preview 2.
* Add workaround to fix DevCom-10863472 VSO-2411436 "CUDA (12.8) host code compilation fails with std::format and VS2022".
  + And add test coverage, verified in a VM.
* Remove CUDA workaround for `__has_cpp_attribute(msvc::x)`.
  + This was added by #772, merged 2020-05-01, for CUDA 9.2 (released May 2018). I don't observe this warning with CUDA 12.4 (released March 2024).
* Remove workarounds for CUDA/Intel, always use Standard `_Pragma`.
  + These were added by #1342, merged 2021-05-20, for VSO-1329304 "EDG doesn't understand Standard _Pragma". We immediately fixed that locally and reported it upstream to EDG. In that release, VS 2019 16.10, we had just begun requiring CUDA 10.1 Update 2 (released Aug 2019). We now require CUDA 12.4 (released March 2024), and I observe that it understands Standard `_Pragma` just fine.
  + I've verified that Intel accepts Standard `_Pragma` on Windows, see below.
* Remove all `__INTEL_COMPILER` workarounds.
  + Intel deprecated and removed their old compiler (icc), which we no longer attempt to avoid breaking in any way. (See DevCom-10200300.)
  + Their new compiler (icx) is LLVM-based and doesn't define `__INTEL_COMPILER` anymore, so all of our accumulated workarounds don't affect it and can be removed. Instead, icx uses `__INTEL_LLVM_COMPILER` to specifically distinguish itself. It also identifies itself as a `__clang__`, activating all of our Clang codepaths. For example, it accepts Clang-style diagnostic pragmas. See icx usage example below.
  + We still don't officially support or test icx. However, following our unofficial policy of avoiding *gratuitous* breakage should be easier now. (We were treating it as a Clang without realizing it, and it worked!)
* Clarify comment: "TRANSITION, fixed in CUDA 12.5"
  + Added by #4974.
* Add `retryCountOnTaskFailure: 4` to checkout-self.yml.
  + See: https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-checkout?view=azure-pipelines
  + This matches what we do in checkout-submodule.yml: https://github.com/microsoft/STL/blob/0d8f517ae3828284fed594741b847db940167a59/azure-devops/checkout-submodule.yml#L12-L15
  + Retrying *tests* is an abomination. However, it's fine to retry checkouts, to be resilient to network gremlins.
* Clang 19.1.5.
  + VS performed a minor version update. It doesn't affect clang-format output, or any of our product/test code, but we should still update our expected clang-format version to avoid emitting warnings during CMake configuration.

:white_check_mark: [STL-ASan-CI passed.](https://dev.azure.com/vclibs/STL/_build/results?buildId=18317&view=results)

# :computer: icx Usage Example

```
C:\Temp>type meow.cpp
```
```cpp
#include <cstdio>
#include <vector>

[[deprecated]] int square(int x) { return x * x; }

int main() {
#ifdef __INTEL_COMPILER
    std::puts("__INTEL_COMPILER is defined.");
#else
    std::puts("__INTEL_COMPILER is NOT defined!");
#endif

#ifdef __INTEL_LLVM_COMPILER
    std::puts("__INTEL_LLVM_COMPILER is defined.");
#endif

#ifdef __clang__
    std::puts("__clang__ is defined.");
#endif

#ifdef SILENCE_DEPRECATED_WARNING
    _Pragma("clang diagnostic push")
    _Pragma("clang diagnostic ignored \"-Wdeprecated-declarations\"")
    return square(0);
    _Pragma("clang diagnostic pop")
#else
    return square(0);
#endif
}
```
```
C:\Temp>icx --version
Intel(R) oneAPI DPC++/C++ Compiler for applications running on Intel(R) 64, Version 2025.0.4 Build 20241205
Copyright (C) 1985-2024 Intel Corporation. All rights reserved.

Intel(R) oneAPI DPC++/C++ Compiler 2025.0.4 (2025.0.4.20241205)
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Program Files (x86)\Intel\oneAPI\compiler\2025.0\bin\compiler
Configuration file: C:\Program Files (x86)\Intel\oneAPI\compiler\2025.0\bin\compiler\..\icx.cfg

C:\Temp>icx /EHsc /nologo /W4 /DSILENCE_DEPRECATED_WARNING meow.cpp && meow
__INTEL_COMPILER is NOT defined!
__INTEL_LLVM_COMPILER is defined.
__clang__ is defined.

C:\Temp>icx /EHsc /nologo /W4 meow.cpp
C:\Temp\meow.cpp(27,12): warning: 'square' is deprecated [-Wdeprecated-declarations]
27 |     return square(0);
    |            ^
C:\Temp\meow.cpp(4,3): note: 'square' has been explicitly marked deprecated here
    4 | [[deprecated]] int square(int x) { return x * x; }
    |   ^
1 warning generated.
```

### DIFF
--- a/stl/inc/yvals_core.h
+++ b/stl/inc/yvals_core.h
@@ -923,8 +923,8 @@ _EMIT_STL_ERROR(STL1002, "Unexpected compiler version, expected CUDA 12.4 or new
 _EMIT_STL_ERROR(STL1000, "Unexpected compiler version, expected Clang 19.0.0 or newer.");
 #endif // ^^^ old Clang ^^^
 #elif defined(_MSC_VER)
-#if _MSC_VER < 1942 // Coarse-grained, not inspecting _MSC_FULL_VER
-_EMIT_STL_ERROR(STL1001, "Unexpected compiler version, expected MSVC 19.42 or newer.");
+#if _MSC_VER < 1944 // Coarse-grained, not inspecting _MSC_FULL_VER
+_EMIT_STL_ERROR(STL1001, "Unexpected compiler version, expected MSVC 19.44 or newer.");
 #endif // ^^^ old MSVC ^^^
 #else // vvv other compilers vvv
 // not attempting to detect other compilers
--- a/stl/inc/yvals_core.h
+++ b/stl/inc/yvals_core.h
@@ -698,8 +698,6 @@
 
 #ifndef __has_cpp_attribute
 #define _HAS_MSVC_ATTRIBUTE(x) 0
-#elif defined(__CUDACC__) // TRANSITION, CUDA - warning: attribute namespace "msvc" is unrecognized
-#define _HAS_MSVC_ATTRIBUTE(x) 0
 #else
 #define _HAS_MSVC_ATTRIBUTE(x) __has_cpp_attribute(msvc::x)
 #endif
--- a/stl/inc/yvals_core.h
+++ b/stl/inc/yvals_core.h
@@ -526,14 +526,7 @@
 #define _STL_STRINGIZE_(S) #S
 #define _STL_STRINGIZE(S)  _STL_STRINGIZE_(S)
 
-// Note that _STL_PRAGMA is load-bearing;
-// it still needs to exist even once CUDA and ICC support _Pragma.
-#if defined(__CUDACC__) || defined(__INTEL_COMPILER)
-#define _STL_PRAGMA(PRAGMA) __pragma(PRAGMA)
-#else
-#define _STL_PRAGMA(PRAGMA) _Pragma(#PRAGMA)
-#endif
-
+#define _STL_PRAGMA(PRAGMA)          _Pragma(#PRAGMA)
 #define _STL_PRAGMA_MESSAGE(MESSAGE) _STL_PRAGMA(message(MESSAGE))
 #define _EMIT_STL_MESSAGE(MESSAGE)   _STL_PRAGMA_MESSAGE(__FILE__ "(" _STL_STRINGIZE(__LINE__) "): " MESSAGE)
 
@@ -883,26 +876,20 @@
 #define _STL_DISABLE_DEPRECATED_WARNING \
     _Pragma("clang diagnostic push")    \
     _Pragma("clang diagnostic ignored \"-Wdeprecated-declarations\"")
-#elif defined(__CUDACC__) || defined(__INTEL_COMPILER)
-#define _STL_DISABLE_DEPRECATED_WARNING \
-    __pragma(warning(push))             \
-    __pragma(warning(disable : 4996)) // was declared deprecated
-#else // vvv MSVC vvv
+#else // ^^^ defined(__clang__) / !defined(__clang__) vvv
 #define _STL_DISABLE_DEPRECATED_WARNING \
     _Pragma("warning(push)")            \
     _Pragma("warning(disable : 4996)") // was declared deprecated
-#endif // ^^^ MSVC ^^^
+#endif // ^^^ !defined(__clang__) ^^^
 #endif // _STL_DISABLE_DEPRECATED_WARNING
 // clang-format on
 
 #ifndef _STL_RESTORE_DEPRECATED_WARNING
 #ifdef __clang__
 #define _STL_RESTORE_DEPRECATED_WARNING _Pragma("clang diagnostic pop")
-#elif defined(__CUDACC__) || defined(__INTEL_COMPILER)
-#define _STL_RESTORE_DEPRECATED_WARNING __pragma(warning(pop))
-#else // vvv MSVC vvv
+#else // ^^^ defined(__clang__) / !defined(__clang__) vvv
 #define _STL_RESTORE_DEPRECATED_WARNING _Pragma("warning(pop)")
-#endif // ^^^ MSVC ^^^
+#endif // ^^^ !defined(__clang__) ^^^
 #endif // !defined(_STL_RESTORE_DEPRECATED_WARNING)
 
 #define _CPPLIB_VER       650
--- a/stl/inc/yvals_core.h
+++ b/stl/inc/yvals_core.h
@@ -2024,10 +2024,6 @@ compiler option, or define _ALLOW_RTCc_IN_STL to suppress this error.
 #define _STATIC_CALL_OPERATOR static
 #define _CONST_CALL_OPERATOR
 #define _STATIC_LAMBDA static
-#elif _MSC_VER < 1944 // TRANSITION, internal toolset needs to be updated to VS 2022 17.14 Preview 1
-#define _STATIC_CALL_OPERATOR
-#define _CONST_CALL_OPERATOR const
-#define _STATIC_LAMBDA
 #else // TRANSITION, VSO-2383148, fixed in VS 2022 17.14 Preview 3
 #define _STATIC_CALL_OPERATOR static
 #define _CONST_CALL_OPERATOR
--- a/tests/std/tests/P0896R4_views_take/test.cpp
+++ b/tests/std/tests/P0896R4_views_take/test.cpp
@@ -561,22 +561,17 @@ constexpr void move_only_test() {
 }
 
 constexpr void output_range_test() {
-#if !defined(__clang__) && !defined(__EDG__) // TRANSITION, VSO-1132704
-    if (!is_constant_evaluated())
-#endif // ^^^ workaround ^^^
-    {
-        using R = test::range<output_iterator_tag, int, test::Sized::no, test::CanDifference::no, test::Common::no,
-            test::CanCompare::no, test::ProxyRef::yes, test::CanView::yes, test::Copyability::move_only>;
-        int some_writable_ints[] = {0, 1, 2, 3};
-        static_assert(same_as<decltype(views::take(R{some_writable_ints}, 99999)), ranges::take_view<R>>);
+    using R = test::range<output_iterator_tag, int, test::Sized::no, test::CanDifference::no, test::Common::no,
+        test::CanCompare::no, test::ProxyRef::yes, test::CanView::yes, test::Copyability::move_only>;
+    int some_writable_ints[] = {0, 1, 2, 3};
+    static_assert(same_as<decltype(views::take(R{some_writable_ints}, 99999)), ranges::take_view<R>>);
 
-        // How do I implement "Fill up to n elements in {output range} with {value}"?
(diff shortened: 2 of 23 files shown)


StephanTLavavej:

tests/std/tests/P0896R4_views_take/test.cpp
```diff
@@ -561,22 +561,17 @@ constexpr void move_only_test() {
...
-        static_assert(same_as<decltype(views::take(R{some_writable_ints}, 99999)), ranges::take_view<R>>);
+    using R = test::range<output_iterator_tag, int, test::Sized::no, test::CanDifference::no, test::Common::no,
+        test::CanCompare::no, test::ProxyRef::yes, test::CanView::yes, test::Copyability::move_only>;
+    int some_writable_ints[] = {0, 1, 2, 3};
+    static_assert(same_as<decltype(views::take(R{some_writable_ints}, 99999)), ranges::take_view<R>>);
 
-        // How do I implement "Fill up to n elements in {output range} with {value}"?
-        ranges::fill(R{some_writable_ints} | views::take(99999), 42);
-        assert(ranges::equal(some_writable_ints, initializer_list<int>{42, 42, 42, 42}));
+    // How do I implement "Fill up to n elements in {output range} with {value}"?
```
davidmrdavid, line 569: out of scope - what's the context behind this comment? Isn't the line below doing precisely this (filling up to `99999` elements with `42`)?

stl/inc/yvals_core.h
```diff
@@ -526,14 +526,7 @@
...
 
-// Note that _STL_PRAGMA is load-bearing;
-// it still needs to exist even once CUDA and ICC support _Pragma.
-#if defined(__CUDACC__) || defined(__INTEL_COMPILER)
-#define _STL_PRAGMA(PRAGMA) __pragma(PRAGMA)
-#else
-#define _STL_PRAGMA(PRAGMA) _Pragma(#PRAGMA)
-#endif
-
+#define _STL_PRAGMA(PRAGMA)          _Pragma(#PRAGMA)
```
davidmrdavid, line 529: nit - why the large spacing? Shouldn't it be:

```suggestion
#define _STL_PRAGMA(PRAGMA) _Pragma(#PRAGMA)
```

davidmrdavid (approved):
Thanks :star:!

No blockers, just two questions:

tests/std/tests/P0896R4_views_take/test.cpp
StephanTLavavej: I'm not sure - this goes back to the original PR #1253 on 2020-09-04. I think it's a rhetorical question, to be immediately answered by the code below, but it's not our usual style.

stl/inc/yvals_core.h
StephanTLavavej: This is the effect of [`AlignConsecutiveMacros: Consecutive`](https://releases.llvm.org/19.1.0/tools/clang/docs/ClangFormatStyleOptions.html#alignconsecutivemacros):

https://github.com/microsoft/STL/blob/0d8f517ae3828284fed594741b847db940167a59/.clang-format#L53

PR #2498, opened 2022-01-21
constexpr all the generate_canonical parameters

MattStephanson:
...up to 64 bits of entropy. Leave the original implementation for
more bits and naughty generators (I'm looking at you, tr1) whose
min and max functions aren't static.

Fixes #1964. Alternative to #2452.

A slightly modified version of @CaseyCarter's release (`/O2 /fp:strict` for me) codegen test:

```cpp
#include <random>

template <size_t kMax>
struct Engine {
    using result_type = size_t;

    static constexpr result_type min() noexcept {
        return 0;
    }
    static constexpr result_type max() noexcept {
        return kMax;
    }

    result_type operator()() noexcept;
};

template <class R, class G>
R meow(G& g) {
    return std::generate_canonical < R, ~std::size_t{ 0 } > (g);
}

template double meow(Engine<10'000>&);
template double meow(Engine<~size_t{0}>&);
```

<details>
<summary>asm diff</summary>

```diff
diff --git a/before.asm b/after.asm
index 5a4e672e..a5d6f74f 100644
--- a/before.asm
+++ b/after.asm
@@ -12,102 +12,43 @@ PUBLIC	??$meow@NU?$Engine@$0CHBA@@@@@YANAEAU?$Engine@$0CHBA@@@@Z ; meow<double,E
 PUBLIC	??$meow@NU?$Engine@$0?0@@@@YANAEAU?$Engine@$0?0@@@Z ; meow<double,Engine<-1> >
 PUBLIC	??$generate_canonical@N$0?0U?$Engine@$0CHBA@@@@std@@YANAEAU?$Engine@$0CHBA@@@@Z ; std::generate_canonical<double,-1,Engine<10000> >
 PUBLIC	??$generate_canonical@N$0?0U?$Engine@$0?0@@@std@@YANAEAU?$Engine@$0?0@@@Z ; std::generate_canonical<double,-1,Engine<-1> >
-PUBLIC	?min@?$Engine@$0CHBA@@@SA_KXZ			; Engine<10000>::min
-PUBLIC	?max@?$Engine@$0CHBA@@@SA_KXZ			; Engine<10000>::max
-PUBLIC	?min@?$Engine@$0?0@@SA_KXZ			; Engine<-1>::min
-PUBLIC	?max@?$Engine@$0?0@@SA_KXZ			; Engine<-1>::max
 PUBLIC	__real@3ff0000000000000
-PUBLIC	__real@404a800000000000
-PUBLIC	__real@40c3880000000000
[...]

PR #5336, opened 2025-03-12
Various cleanups

StephanTLavavej:
Low-risk cleanups that I've been accumulating.

# :gear: Commits
* Comment: Fix alignment for `_Is_word_table` in `<regex>`.
* Comment: Drop extra space in `P0088R3_variant_msvc`.
* Uglify `Enable` => `_Enable` in `<variant>`.
* Use `if constexpr` for `is_floating_point_v` in `<chrono>`.
* Use `if constexpr` for `is_base_of_v` in `P0220R1_sample`.
* Comments: `concept available` => `_HAS_CXX20` in `<functional>`.
* Add braces to `_Idl_dist_add` control flow in `<algorithm>`.
* Replace away the `_NRAND` macro in `<random>`.
  + Before #2498 on 2022-02-07, the macro was `#define _NRAND(eng, resty) (_STD generate_canonical<resty, static_cast<size_t>(-1)>(eng))` which avoided significant verbosity, but now it serves no purpose.
  + Replaced `_NRAND\((\w+), (\w+)\)` => `_Nrand_impl<$2>($1)` because surrounding parens were never necessary.
* Add `const` to `unsigned int timeout` in `primitives.hpp`.
* ~~Change `_Has_static_min_max` from `struct` to `constexpr bool` in `<random>`.~~
  + Reverted, reported VSO-2417491 "C1XX `/permissive-` (strict mode) mishandles a variable template detecting `static constexpr` member functions", commented. (This isn't worth maintaining separate "workaround" / "no workaround" paths.)
* Drop unnecessary parens in `P0355R7_calendars_and_time_zones_dates`.
* Avoid unnecessary parens and `+= bool` in `VSO_0000000_vector_algorithms`.
  + Directly adding a `bool` is weird, so we conventionally avoid that.
* Drop unnecessary `typename` in `<format>`.
* Extract an alias template `_P_arg_type_t` in `<future>`.
* `_mm_mfence` is SSE2; update comment after #4741.
* Change unnecessary `inline constexpr` to `constexpr` for a function.
* `while (true)` => `for (;;)`
  + We traditionally used "for ever" to avoid "conditional expression is constant" warnings. While the compiler doesn't warn about `true` anymore, our usage remains consistent. (I don't feel super strongly about this one though.)
* Mark `_HAS_STD_BYTE` as `// TRANSITION, OS-14273702`.
  + Internal OS-14273702 has been active since 2017-10-17, but one can hope.
* `_Name.length()` => `_Name.size()` for `string`.
  + We're very particular about the synonym that we call.
* Fix comments: `%[E0Q#]` => `%[EOQ#]`
* Use braces to construct a temporary `bitset`.
  + When possible, we prefer braces to construct temporaries, since this avoids looking like a function call.
* util.py: `executeCommandVerbose` is dead.
* `_AllocNode` => `_Alnode` for consistency.
  + `_Erase_tree_and_orphan_guard` calls `_Erase_tree_and_orphan`, which already uses this naming convention: https://github.com/microsoft/STL/blob/0d8f517ae3828284fed594741b847db940167a59/stl/inc/xtree#L756-L757
  + As I always say, templates don't have a "type system" (especially without concepts), so naming is super important. This is an example of how avoiding unnecessary name changes between template layers is a good idea.

### DIFF
--- a/stl/src/print.cpp
+++ b/stl/src/print.cpp
@@ -273,7 +273,7 @@ extern "C" {
     _Allocated_string _Allocated_str{};
     _Transcode_result _Transcoded_str{};
 
-    while (true) {
+    for (;;) {
         _Curr_str_segment = _Get_next_utf8_string_segment(_Remaining_str, _Remaining_str_size);
         _Transcoded_str   = _Transcode_utf8_string(_Allocated_str, _Curr_str_segment);
 
--- a/stl/inc/algorithm
+++ b/stl/inc/algorithm
@@ -7477,11 +7477,13 @@ _NODISCARD constexpr auto _Idl_dist_add(_Diff1 _Lhs, _Diff2 _Rhs) {
         return ranges::_Distance_unbounded{};
     } else
 #endif // _HAS_CXX20
+    {
         if constexpr (is_same_v<_Diff1, _Distance_unknown> || is_same_v<_Diff2, _Distance_unknown>) {
             return _Distance_unknown{};
         } else {
             return _Lhs + _Rhs;
         }
+    }
 }
 
 _EXPORT_STD template <class _InIt1, class _InIt2, class _OutIt, class _Pr>
--- a/stl/inc/atomic
+++ b/stl/inc/atomic
@@ -203,7 +203,7 @@ extern "C" inline void _Atomic_thread_fence(const unsigned int _Order) noexcept 
     if (_Order == _Atomic_memory_order_seq_cst) {
         volatile long _Guard; // Not initialized to avoid an unnecessary operation; the value does not matter
 
-        // _mm_mfence could have been used, but it is not supported on older x86 CPUs and is slower on some recent CPUs.
+        // _mm_mfence could have been used, but it is slower on some CPUs.
         // The memory fence provided by interlocked operations has some exceptions, but this is fine:
         // std::atomic_thread_fence works with respect to other atomics only; it may not be a full fence for all ops.
         (void) _InterlockedIncrement(&_Guard);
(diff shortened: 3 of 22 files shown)


StephanTLavavej:

stl/src/print.cpp
```diff
@@ -273,7 +273,7 @@ extern "C" {
     _Allocated_string _Allocated_str{};
     _Transcode_result _Transcoded_str{};
 
-    while (true) {
+    for (;;) {
```
davidmrdavid, line 276: nit - is this just stylistic preference? I've always found `for (;;)` loops to be a bit strange and less clear than the obvious `while(true)`, but I won't insist / trust your intuition on what's idiomatic.

davidmrdavid (approved):
:shipit: LGTM, just a question below.

stl/src/print.cpp
StephanTLavavej: It was previously more than stylistic preference, but now it's just for consistency. As I explained in the PR description:

> We traditionally used "for ever" to avoid "conditional expression is constant" warnings. While the compiler doesn't warn about `true` anymore, our usage remains consistent. (I don't feel super strongly about this one though.)

What changed a few years ago is that the compiler specifically got smarter about conditional expressions that are literally and directly `true` and `false`. I think it'll also accept individual compile-time constants, but it still warns about more complicated expressions.

AlexGuteniev: I actually like the scariness of `for (;;)` loop.
It says, like, be wary, there must be a break or a return somewhere in the middle.

PR #5362, opened 2025-03-25
`<regex>`: Properly parse dollar anchors in basic and grep mode

muellerj2:
This PR provides the symmetric change to #5165 for the dollar sign in basic and grep mode. Similar to the handling of the caret anchor, this PR recognizes a dollar sign as an anchor in a basic regular expression if it is at the end of an alternative. It does so by doing a bit of lookahead.

Drive-by: Remove a duplicated test that was accidentally added in #5165.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -3947,15 +3947,17 @@ void _Parser<_FwdIt, _Elem, _RxTraits>::_Trans() { // map character to meta-char
                     ++_Next;
                 }
 
-                const bool _Has_escaped_capture_groups = (_L_flags & _L_nex_grp) == 0;
+                // Only the basic and grep grammars set _L_anch_rstr, so _L_alt_pipe and _L_nex_grp must be unset.
+                // Therefore, we don't need to handle "dollar followed by pipe '|' for alternation"
+                // or "dollar followed by non-escaped right parenthesis ')' closing a group" below.
+                _STL_INTERNAL_CHECK((_L_flags & (_L_alt_pipe | _L_nex_grp)) == 0);
 
                 const _Elem _Ch = *_Next;
                 const bool _Is_end_of_alternative =
-                    ((_L_flags & _L_alt_pipe) && _Ch == _Meta_bar) // dollar followed by pipe '|' for alternation
-                    || ((_L_flags & _L_alt_nl) && _Ch == _Meta_nl
+                    ((_L_flags & _L_alt_nl) && _Ch == _Meta_nl
                         && _Disj_count == 0) // dollar followed by newline '\n' for alternation
-                    || (_Escaped == _Has_escaped_capture_groups && _Ch == _Meta_rpar
-                        && _Disj_count != 0); // dollar followed by right parenthesis ')' closing a group
+                    || (_Escaped && _Ch == _Meta_rpar
+                        && _Disj_count != 0); // dollar followed by (escaped) right parenthesis ')' closing a group
 
                 if (!_Is_end_of_alternative) {
                     _Mchar = _Meta_chr;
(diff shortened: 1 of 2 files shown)


stl/inc/regex
```diff
@@ -3941,9 +3940,23 @@ void _Parser<_FwdIt, _Elem, _RxTraits>::_Trans() { // map character to meta-char
...
+                if (_Escaped) {
+                    ++_Next;
+                }
+
+                const _Elem _Ch = *_Next;
+                const bool _Is_end_of_alternative =
+                    ((_L_flags & _L_alt_pipe) && _Ch == _Meta_bar) // dollar followed by pipe |
+                    || ((_L_flags & _L_alt_nl) && _Ch == _Meta_nl && _Disj_count == 0) // dollar followed by pipe \n
+                    || ((_Escaped || (_L_flags & _L_nex_grp)) && _Ch == _Meta_rpar
+                        && _Disj_count != 0); // dollar followed by parenthesis closing a group
```
muellerj2, line 3955: Strictly speaking, `_L_flags & _L_alt_pipe` and `_L_flags & _L_nex_grp` are always 0 here, so the first disjunction could be removed completely and `_L_flags & _L_nex_grp` could be omitted in the third.

I still opted for this expression, though, because it clearly handles all possible cases of "end of an alternative", independent of how the flags for alternatives correlate in practice with the flag `_L_anch_rstr`.

StephanTLavavej:
Thanks! :smile_cat: I pushed changes to fix product code logic (even though it "can't happen") and various copy-paste relics in the test - please double-check.

StephanTLavavej (approved):

stl/inc/regex
muellerj2: An OR is actually fine here because `_Is_esc()` never returns true for a closing parenthesis if `(_L_flags & _L_nex_grep)` is set:

https://github.com/microsoft/STL/blob/2d01415877fa785543b89edbaf7758dcf9430106/stl/inc/regex#L3878-L3882

So it can't happen that  `(_L_flags & _L_nex_grp)` and `_Escaped` are both true when `_Ch == _Meta_rpar`. The remaining three cases behave like an OR.

Well, I guess I shouldn't have used this logical equivalence because it is too subtle.

StephanTLavavej: Aha, thanks - I didn't look at what `_Is_esc` was actually doing.

Hopefully my logic change is still reasonable.

muellerj2: It's fine. I wrote this down as `_Escaped == !(_L_flags & _L_nex_grp)` at first as well, but then I thought that `x == !(y)` looks terrible and feels like an invitation for someone to change it to `x != y` (which is wrong), so I rewrote it as a logical or. But your solution to move `!(y)` out of the expression works as well.

That said, the logic works slightly differently than you assumed: Because of `_Is_esc()`'s implementation, `_Escaped == _Has_escaped_capture_groups` will always be true when the `_L_nex_grp` flag is set. But since `_Escaped` is always false, the backslash before an escaped parenthesis is never skipped, so the comparison succeeds at `_Ch == _Meta_rpar` for a non-escaped parenthesis, also fails there for an escaped parenthesis (because `_Ch` will then be `_Meta_esc`).

StephanTLavavej: Oh, thanks for explaining. That's more confusing than I thought. Now I'm thinking we should just have an internal assert that this isn't possible with the set of grammars available.

StephanTLavavej (approved):

stl/inc/regex
StephanTLavavej: Ok, I pushed another commit, which I hope captures your intention to describe all of the possible cases, but simplifies the logic to only handle the cases that can actually occur.

StephanTLavavej:

StephanTLavavej:
:heart_eyes_cat: :dollar: :money_mouth_face:
===

PR #5364, opened 2025-03-25
`<regex>`: Allow initial `]` to start character ranges in POSIX regular expressions

muellerj2:
Fixes a regex miscompilation noticed in #5165.

The POSIX standard says that the right bracket ] is not special at the beginning of a bracketed character class (https://pubs.opengroup.org/onlinepubs/009696699/basedefs/xbd_chap09.html#tag_09_03_05). This means that an initial ] behaves like any other character, so it may also be the start of a character range.

### DIFF
--- a/tests/std/tests/VSO_0000000_regex_use/test.cpp
+++ b/tests/std/tests/VSO_0000000_regex_use/test.cpp
@@ -1171,6 +1171,23 @@ void test_gh_5253() {
     g_regexTester.should_not_match("a", "()*");
 }
 
+void test_gh_5364() {
+    // GH-5364 `<regex>`: Allow initial ] to start character ranges in basic regular expressions
+    for (syntax_option_type option : {basic, extended, grep, egrep}) {
+        g_regexTester.should_match("]", "[]-_]", option);
+        g_regexTester.should_match("^", "[]-_]", option);
+        g_regexTester.should_match("_", "[]-_]", option);
+        g_regexTester.should_not_match("-", "[]-_]", option);
+
+        g_regexTester.should_match("]", "[]a]", option);
+        g_regexTester.should_match("a", "[]a]", option);
+        g_regexTester.should_not_match("a]", "[]a]", option);
+        g_regexTester.should_not_match("]a", "[]a]", option);
+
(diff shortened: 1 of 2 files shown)


tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -1171,6 +1171,23 @@ void test_gh_5253() {
     g_regexTester.should_not_match("a", "()*");
 }
 
+void test_gh_5364() {
+    // GH-5364 `<regex>`: Allow initial ] to start character ranges in basic regular expressions
```
StephanTLavavej, line 1175: This says "basic", but the PR title says "POSIX".

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -1171,6 +1171,23 @@ void test_gh_5253() {
...
+    for (syntax_option_type option : {basic, extended, grep, egrep}) {
+        g_regexTester.should_match("]", "[]-_]", option);
+        g_regexTester.should_match("^", "[]-_]", option);
+        g_regexTester.should_match("_", "[]-_]", option);
+        g_regexTester.should_not_match("-", "[]-_]", option);
+
+        g_regexTester.should_match("]", "[]a]", option);
+        g_regexTester.should_match("a", "[]a]", option);
+        g_regexTester.should_not_match("a]", "[]a]", option);
+        g_regexTester.should_not_match("]a", "[]a]", option);
```
StephanTLavavej, line 1185: We should also test that it doesn't match a single character like `_` (which is in the range of `]` to `a`, but here they are being used individually).

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: I pushed changes to significantly expand the test coverage.

StephanTLavavej:

StephanTLavavej:
Resolved adjacent-add conflict in `VSO_0000000_regex_use`.

StephanTLavavej (approved):

StephanTLavavej:
:carrot: :dollar: :star:
===

Issue #5359, opened 2025-03-25
`<bit>`: Could `has_single_bit()` be faster?


pps83:
Seems like it should convert to `popcount(x) == 1` if it's available on the target arch, but looking at the code, doesn't seem like compiler would do it:
```c++
_NODISCARD constexpr bool has_single_bit(const _Ty _Val) noexcept {
    return _Val != 0 && (_Val & (_Val - 1)) == 0;
}
```

https://github.com/microsoft/STL/blob/f2a2933dd65d9e8d3fa698a97b6074f7ef00e1fd/stl/inc/bit#L87

AlexGuteniev:
I'm not sure.

For the current implementation I'd expect something as following (assuming x64 arch, as it is primary optimization target):
```
or   reg1, reg1
jz   label1
lea  reg2, [reg1-1]
test reg1, reg2
label1:
```
And for `popcount` as following;
```
popcnt reg2, reg1
cmp    reg2, 1
```

Which of these is better would highly depend on the context. I'd expect the current branchy function to win for predictable input, and the `popcnt` branchless function towin for unpredictable input.

In sight of this uncertainty, I would avoid making any changes and introducing unnecessary complexity.

AlexGuteniev:
Oh, it also seems CPU vendor specific. AMDs have cheaper `popcnt`, could be the winner most of the time there.

Alcaro:
According to GCC codegen, popcnt is the winner if available, yes.

If not, the best answer is `(v ^ (v-1)) > v-1`.

https://godbolt.org/z/d61bxW4r1

StephanTLavavej:
Ideally, our compilers would do the right thing here, so we may need to report optimization bug(s) to C1XX and/or Clang.

pps83:
I just edited that file in github.dev and made that PR directly from the browser. Crazy, full blown code editor in browser :)

![Image](https://github.com/user-attachments/assets/0cbb9270-8328-477d-b2a3-20aa708da999)

AlexGuteniev:
@pps83 are you interested in `popcnt` PR? Otherwise I can do it.

pps83:
> [@pps83](https://github.com/pps83) are you interested in `popcnt` PR? Otherwise I can do it.

I can do it, but I'll need some assistance to be able to do it quickly.
I just looked into it and here's how it can be done.

[...]

Issue #4496, opened 2024-03-21
benchmark: it is built with `/Ob1`, so vector algorithm dispatcher is noticeable


AlexGuteniev:
Noticed while working #4495 . When I decided to use sized `if constexpr` dispatch, instead of using the same version for all element sizes, I observed significant perf degradation for small element sizes. A part of it is due to not inlining the dispatcher.

The benchmark is built with  `/Ob1`. Looks like it is implied due to CMake `RelWithDebugInfo` configuration, as opposed to `Release`.

What are our takeaways?

----

I see the following options:
* Mark vector algorithms dispatchers `inline`, consider making other STL functions `inline`.
   * This helps other projects with  `RelWithDebugInfo` to inline STL, though it would obfuscate the debugger
* Override the option in the benchmark
* Make the benchmark `Release` by default, instead of `RelWithDebugInfo` 
  * I wouldn't like that. `RelWithDebugInfo` is convenient for profiling
* Accept the cost of dispatching as a penalty for vector algorithms that use dispatching
  * I don't think this is fair
* Use specializations instead of `if constexpr`
  *  Throughput?
* Manually inline the dispatch like for `__std_reverse_copy_trivially_copyable...`
  * Copypasta

StephanTLavavej:
We talked about this at the weekly maintainer meeting and @barcharcraz had a suggestion - build as `Release`, but then add compiler options (e.g. `/Zi`) to generate debug info.

https://github.com/microsoft/STL/blob/8e2d724cc1072b4052b14d8c5f81a830b8f1d8cb/benchmarks/CMakeLists.txt#L62-L65

Charlie was additionally concerned that we shouldn't be trying to *force* the build type here - it would be better to provide a default that can be overridden when building the benchmarks, if a contributor temporarily wants something different. We don't immediately know the proper incantations for that.

PR #5367, opened 2025-03-27
`<bit>`: Improve `has_single_bit()` codegen

pps83:
code avoids branches, provides 20% speed up on x64 targets.

**x64 results**
```
---------------------------------------
Benchmark                          Time
---------------------------------------
Rand16_has_single_bit            284 ns
Rand16_has_single_bit_new        232 ns
Rand32_has_single_bit            306 ns
Rand32_has_single_bit_new        246 ns
Rand64_has_single_bit            289 ns
Rand64_has_single_bit_new        233 ns
```

partially addresses #5359


pps83:
seems like code is better for x86/x64/arm/arm64 : https://godbolt.org/z/fjshqMf4d

the only potential drawback: old code would check for _Val = 0 and jump. New code has no jumps (except on arm)

stl/inc/bit
```diff
@@ -84,7 +84,7 @@ _NODISCARD constexpr int countl_zero(_Ty _Val) noexcept;
 
 _EXPORT_STD template <_Standard_unsigned_integral _Ty>
 _NODISCARD constexpr bool has_single_bit(const _Ty _Val) noexcept {
-    return _Val != 0 && (_Val & (_Val - 1)) == 0;
+    return (_Val ^ (_Val - 1)) > _Val - 1;
```
StephanTLavavej, line 87: This looks good, and `P0556R3_bit_integral_power_of_two_operations` provides sufficient correctness testing. However, can you add a benchmark to demonstrate that this makes an observable difference in performance?

StephanTLavavej (changes requested):

stl/inc/bit
StephanTLavavej: Additionally, if this fully resolves the issue, please update your PR description to use the [magic words](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) "Fixes #nnnn" instead of merely "Addresses #nnnn".

AlexGuteniev: For benchmark I suggest:
1. Both reading result into a variable 
```c++
bool r = has_single_bit(...)
DoNotOptimize(r);
```
and directly into branch
```c++
if (has_single_bit(...)) {
  DoNotOptimize(v1);
}  else {
  DoNotOptimize(v2);
}
```
2. Both testing on constant nonzero input (but still variable with optimization barrier), and randomized input with some zeros (a loop over an array of 10s or maybe 100s elements

This would give combinations, but this will cover different usage contexts

AlexGuteniev:
> (except on arm)

Note that only 64-bit arm is an important optimization target.
It is referred to as ARM64 here and AArch64 somewhere else.

32-bit arm is abandoned as optimization target.

stl/inc/bit
pps83: > Additionally, if this fully resolves the issue, please update your PR description to use the [magic words](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) "Fixes #nnnn" instead of merely "Addresses #nnnn".

I don't think it resolves. It's improves default codegen, but should use popcount for avx2 target.

pps83:
> 32-bit arm is abandoned as optimization target.

Imo for mobiles it should still be relevant. More relevant that x86 for x64 CPUs. In any case, 32bit arm was the same

stl/inc/bit
pps83: https://github.com/pps83/has_single_bit_test
in short, 20-30% improvement for x64 target (even when all zeros are tested). Old code appears to be better for 16/32bit on x86 (and the same perf when testing 64 bit uints on x86).

pps83: ### x86 results
```
--------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations
--------------------------------------------------------------------
Rand16_has_single_bit_old        106 ns          107 ns      7466667
Rand16_has_single_bit_new        130 ns          131 ns      5600000
Rand32_has_single_bit_old       99.0 ns          100 ns      7466667
Rand32_has_single_bit_new        149 ns          151 ns      4977778
Rand64_has_single_bit_old        208 ns          209 ns      3446154
Rand64_has_single_bit_new        194 ns          193 ns      3733333

Zero16_has_single_bit_old        104 ns          105 ns      7466667
Zero16_has_single_bit_new        146 ns          146 ns      4072727
Zero32_has_single_bit_old        108 ns          110 ns      6400000
Zero32_has_single_bit_new        169 ns          169 ns      4072727
Zero64_has_single_bit_old        134 ns          134 ns      5600000
Zero64_has_single_bit_new        247 ns          246 ns      2986667
```

### x64 results
```
--------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations
--------------------------------------------------------------------
Rand16_has_single_bit_old        284 ns          283 ns      2488889
Rand16_has_single_bit_new        232 ns          230 ns      2986667
Rand32_has_single_bit_old        306 ns          311 ns      2357895
Rand32_has_single_bit_new        246 ns          246 ns      2800000
Rand64_has_single_bit_old        289 ns          285 ns      2357895
Rand64_has_single_bit_new        233 ns          235 ns      2986667

Zero16_has_single_bit_old        278 ns          279 ns      2635294
Zero16_has_single_bit_new        238 ns          241 ns      2986667
Zero32_has_single_bit_old        322 ns          321 ns      1947826
Zero32_has_single_bit_new        247 ns          243 ns      2635294
Zero64_has_single_bit_old        320 ns          325 ns      2357895
Zero64_has_single_bit_new        244 ns          244 ns      3200000
```

AlexGuteniev: Good results!

We have benchmarks as a part of this repo, and usually expand them in new performance PR as needed. 
We benchmark the STL implementation directly, and for "before" results we get the before version.
See any other currently open performance PR as an example, and also this Wiki page::
https://github.com/microsoft/STL/wiki/Benchmarking-the-STL

benchmarks/src/has_single_bit.cpp
```diff
@@ -0,0 +1,51 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <assert.h>
+#include <benchmark/benchmark.h>
+#include <bit>
+
+#include <utility.hpp>
+
+template <typename T>
+static void has_single_bit_if(benchmark::State& state) {
```
AlexGuteniev, line 11: `static` linkage were dropped in existing benchmarks in #5370 along with unnamed namespace, because they serve no purpose for a single TU program. Let's not introduce them here. Also occurs once below.

benchmarks/src/has_single_bit.cpp
```diff
@@ -0,0 +1,51 @@
...
+    const auto random_v = random_vector<T>(8);
+    for (auto _ : state) {
+        benchmark::DoNotOptimize(random_v);
+        int count_true = 0, count_false = 0;
```
AlexGuteniev, line 15: We conventionally avoid multiple variables in a single declaration, even for the non-problematic cases.

benchmarks/src/has_single_bit.cpp
```diff
@@ -0,0 +1,51 @@
...
+        for (const auto& x : random_v) {
+            if (std::has_single_bit(x)) {
+                benchmark::DoNotOptimize(count_true++);
```
AlexGuteniev, line 18: I suspect that the signed counters might overflow for long benchmark run (and the benchmark run may be longer than usual using command-line `--benchamrk_min_time` argument). The signed overflow is Undefined Behavior. 

This can be easily avoided by using `unsinged int` instead of `int`. Unsigned overflow is defined.

benchmarks/src/has_single_bit.cpp
pps83: done.

benchmarks/src/has_single_bit.cpp
pps83: done.

stl/inc/bit
pps83: > Old code appears to be better for 16/32bit on x86 (and the same perf when testing 64 bit uints on x86).

I removed testing e elements per iteration. x86 results got improved

AlexGuteniev:
i5-1235U

x64

P-core

Benchmark                    |  Before  |    After  
-----------------------------|----------|-----------
has_single_bit_if<uint16_t>  | 13,5 ns  |  12,4 ns  
has_single_bit_if<uint32_t>  | 14,4 ns  |  11,1 ns  
has_single_bit_if<uint64_t>  | 13,8 ns  |  13,1 ns  
has_single_bit<uint16_t>     | 5,96 ns  |  4,53 ns  
has_single_bit<uint32_t>     | 5,53 ns  |  5,03 ns  
has_single_bit<uint64_t>     | 5,22 ns  |  5,02 ns  

E-core

Benchmark                    |  Before  |   After  
-----------------------------|----------|-----------
has_single_bit_if<uint16_t>  | 19,7 ns  |  17.7 ns  
has_single_bit_if<uint32_t>  | 19,7 ns  |  17.7 ns  
has_single_bit_if<uint64_t>  | 19,7 ns  |  19.9 ns  
has_single_bit<uint16_t>     | 12,5 ns  |  9.35 ns  
has_single_bit<uint32_t>     | 12,0 ns  |  9.33 ns  
has_single_bit<uint64_t>     | 13,2 ns  |  9.37 ns

pps83:
Can you please try this local change also:

```c++
_NODISCARD constexpr bool has_single_bit(const _Ty _Val) noexcept {
#if 1 // _POPCNT_INTRINSICS_ALWAYS_AVAILABLE
#if _HAS_CXX20
    if (_STD is_constant_evaluated())
        return 1 == _Popcount_fallback(_Val);
#endif // _HAS_CXX20
    return 1 == _Unchecked_popcount(_Val);
#else
    return (_Val ^ (_Val - 1)) > _Val - 1;
#endif
}
```

AlexGuteniev:
Turns out this particular benchmark is very much affected by #4496 and `popcnt` version is slower. 

That's why I've hidden the previous results, and obtained the updated results by changing also this line to say `Release` instead of `RelWithDebInfo`

https://github.com/microsoft/STL/blob/main/benchmarks/CMakeLists.txt#L62

i5-1235U

x64

P-core

Benchmark                    |  Before  |    After  | popcnt
-----------------------------|----------|-----------|---------
has_single_bit_if<uint16_t>  | 11.5 ns  |  10.7 ns  | 6.68 ns 
has_single_bit_if<uint32_t>  | 11.3 ns  |  11.3 ns  | 10.6 ns 
has_single_bit_if<uint64_t>  | 6.98 ns  |  11.7 ns  | 6.57 ns 
has_single_bit<uint16_t>     | 6.36 ns  |  4.51 ns  | 3.51 ns 
has_single_bit<uint32_t>     | 6.08 ns  |  4.71 ns  | 3.52 ns 
has_single_bit<uint64_t>     | 5.27 ns  |  4.71 ns  | 3.53 ns 

E-core

Benchmark                    |  Before  |   After   | popcnt  
-----------------------------|----------|-----------|---------
has_single_bit_if<uint16_t>  | 16.2 ns  |  14.1 ns  | 14.1 ns 
has_single_bit_if<uint32_t>  | 16.7 ns  |  14.0 ns  | 14.0 ns 
has_single_bit_if<uint64_t>  | 16.2 ns  |  16.2 ns  | 14.0 ns 
has_single_bit<uint16_t>     | 12.2 ns  |  9.07 ns  | 7.24 ns 
has_single_bit<uint32_t>     | 11.9 ns  |  9.07 ns  | 7.24 ns 
has_single_bit<uint64_t>     | 11.6 ns  |  9.07 ns  | 7.25 ns 

Apparently `popcnt` is a good idea. Even for the "if" case.
----

#4496 wasn't a severe obstacle for most previous optimization, but at such micro level it is really disrupting, we'll need to get into that.

AlexGuteniev:
> Can you please try this local change also:

Before you go with this in the PR:
 * We don't compare like `1 == function()` in new code, although there are some relic precedents. In new code, we use natural order `function() == 1`
 * Always control flow braces `{}` even around single statement in `if`
 * You don't need to check `#if _HAS_CXX20`. It is C++20 header, and the check is already there at the very beginning for the whole content

pps83:
> Before you go with this in the PR...

This PR doesn't have popcnt change. The popcnt change doesn't use dispatcher, it's the unchecked version that does the `popcnt` opcode directly and it should only be used when building for AVX+. The default (version with this PR) is better than using dispatcher imo.

I can add popcnt it as a separate PR, i think it's better  that way.

StephanTLavavej:
For the record, ARM32 performance is indeed utterly irrelevant. Windows ripped out their ARM32 build, and will even be dropping support for ARM32 binaries running on ARM64 machines, see https://learn.microsoft.com/en-us/windows/arm/arm32-to-arm64 .

ARM32 still needs to behave correctly (until DevDiv/MSVC drops support for targeting it, which I am constantly asking about), but we don't want to waste any more time on it.

stl/inc/bit
pps83: @StephanTLavavej pr should be ok to merge

StephanTLavavej: Thanks, I'll review this soon!

benchmarks/src/has_single_bit.cpp
```diff
@@ -0,0 +1,49 @@
...
+        for (const auto& x : random_v) {
+            r += std::has_single_bit(x);
+        }
+        benchmark::DoNotOptimize(r);
+    }
+}
+
+BENCHMARK(has_single_bit_if<uint16_t>);
+BENCHMARK(has_single_bit_if<uint32_t>);
+BENCHMARK(has_single_bit_if<uint64_t>);
```
StephanTLavavej, line 43: We should benchmark `uint8_t` even if we don't expect it to change here.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! I pushed some changes to the benchmark for consistency with repo conventions.

StephanTLavavej:

benchmarks/src/has_single_bit.cpp
```diff
@@ -0,0 +1,52 @@
...
+                benchmark::DoNotOptimize(++count_true);
+            } else {
+                benchmark::DoNotOptimize(++count_false);
+            }
+        }
+    }
+}
+
+template <typename T>
+void has_single_bit(benchmark::State& state) {
```
StephanTLavavej, line 30: This shadows the library function, which we conventionally avoid. It happened to work with google-benchmark 1.9.1, but conflicts with google-benchmark 1.9.2 in #5370.

StephanTLavavej (approved):

StephanTLavavej:
:zero: :zero: :one:
===

PR #5375, opened 2025-03-28
`<regex>`: Fix word bounds `\b` and `\B` when applied to the empty string

### LINKED ISSUE #5371: `<regex>`: \b and \B are backwards on empty strings
# Describe the bug
\b and \B are backwards on empty strings

# Command-line test case
https://godbolt.org/z/9TnqKsvbe
```
#include <regex>
#include <stdio.h>

int main() {
    try {
        std::string s{""};
        std::regex r{"\\b"};
        std::smatch m;
        bool result = std::regex_search(s, m, r);
        printf("regex_search: %d\n", result);
        for (unsigned i = 0; i < m.size(); ++i) {
            printf("m[%d]", i);
            if (m[i].matched) {
                printf(".str(): \"%s\"\n", m[i].str().c_str());
            } else {
                puts(".matched: false");
            }
        }
    } catch (const std::exception& e) {
        printf("Exception: %s\n", e.what());
    }
}
```

# Expected behavior
Same as libstdc++, libc++ and Firefox: \B should match empty string, \b should not

# STL version
Whatever's on Godbolt (I couldn't check if it's fixed on latest, but I did check for duplicate issues/PRs and found nothing)

# Additional context
Yes

muellerj2:
Fixes #5371.

The [ECMAScript standard](https://262.ecma-international.org/#sec-compileassertion) effectively states that `\b` matches a position `i` in the string iff `IsWordChar(i-1) xor IsWordChar(i)` is true. It also defines that for positions `i = -1` and `i = length(input)`, `IsWordChar(i)` always returns `false`. Thus, `\b` does not match anything in the empty string.

Since `\b` does not match the empty string, the flags `match_not_bow` and `match_not_eow` don't matter in this case.

### DIFF
--- a/tests/std/tests/VSO_0000000_regex_use/test.cpp
+++ b/tests/std/tests/VSO_0000000_regex_use/test.cpp
@@ -499,7 +499,7 @@ void test_VSO_225160_match_eol_flag() {
 
 void test_VSO_226914_word_boundaries() {
     const test_regex emptyAnchor(&g_regexTester, R"(\b)");
-    emptyAnchor.should_search_match("", "");
+    emptyAnchor.should_search_fail("");
     emptyAnchor.should_search_fail("", match_not_bow);
     emptyAnchor.should_search_fail("", match_not_eow);
     emptyAnchor.should_search_fail("", match_not_bow | match_not_eow);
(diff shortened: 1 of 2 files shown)


tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -1171,6 +1171,12 @@ void test_gh_5253() {
     g_regexTester.should_not_match("a", "()*");
 }
 
+void test_gh_5371() {
+    // GH-5371 <regex>: \b and \B are backwards on empty strings
+    g_regexTester.should_not_match("", "\\b");
+    g_regexTester.should_match("", "\\B");
```
StephanTLavavej, line 1177: Let's use raw string literals for clarity.

StephanTLavavej:
Thanks for the fix, test coverage, and clear writeup with citations! I pushed a trivial change to use raw string literals.

StephanTLavavej:

StephanTLavavej:
Resolved adjacent-add conflict in `VSO_0000000_regex_use`.

StephanTLavavej (approved):

PR #5376, opened 2025-03-28
`<regex>`: Make back-references to unmatched capture groups not match anything in POSIX basic regexes

### LINKED ISSUE #5374: `<regex>`: Back-references to unmatched capture groups should not match in POSIX basic regular expressions
# Describe the bug

According to [Section 9.3.6 of the POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_03_06), back-references to unmatched capture groups ("subexpressions") should not match. However, MSVC STL's implementation currently matches such back-references to the empty string.

# Test case
```
#include <iostream>
#include <regex>
#include <string>

using namespace std;

int main() {
        string pattern = R"(\(.\)*\1)";
        regex re{pattern, regex_constants::basic};
        cout << "regex '"<< pattern
             << "' matches the empty string: "
             << regex_match("", re);
}
```
https://godbolt.org/z/83nxo39qj

# Expected behavior
The regular expression should not match.

# Additional context
Note that the equivalent ECMAScript regex `(.)*\1` does match the empty string, since the standard says that back-references to unmatched capture groups match the empty string.

muellerj2:
Fixes #5374.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -3643,6 +3643,9 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Match_pat(_Node_base* _Nx) { // c
 
         case _N_back:
             { // check back reference
+                _STL_INTERNAL_CHECK(
+                    (_Sflags & (regex_constants::extended | regex_constants::egrep | regex_constants::awk))
+                    == 0); // these grammars don't have backreferences
                 _Node_back* _Node = static_cast<_Node_back*>(_Nx);
                 if (_Tgt_state._Grp_valid[_Node->_Idx]) { // check for match
                     _It _Res0 = _Tgt_state._Cur;
@@ -3654,9 +3657,7 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Match_pat(_Node_base* _Nx) { // c
                     } else {
                         _Tgt_state._Cur = _Res0;
                     }
-                } else if (_Sflags
-                           & (regex_constants::basic | regex_constants::extended | regex_constants::grep
-                               | regex_constants::egrep | regex_constants::awk)) {
+                } else if (_Sflags & (regex_constants::basic | regex_constants::grep)) {
                     _Failed = true;
                 }
                 break;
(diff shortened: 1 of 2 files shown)


stl/inc/regex
```diff
@@ -3654,6 +3654,10 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Match_pat(_Node_base* _Nx) { // c
                     } else {
                         _Tgt_state._Cur = _Res0;
                     }
+                } else if (_Sflags
+                           & (regex_constants::basic | regex_constants::extended | regex_constants::grep
+                               | regex_constants::egrep | regex_constants::awk)) {
```
StephanTLavavej, line 3659: I find it confusing to check for `extended`, `egrep`, and `awk` here, because they don't support backreferences. It would be better to have an internal check for these "can't happen" cases.

StephanTLavavej (approved):

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -1171,6 +1171,17 @@ void test_gh_5253() {
...
+void test_gh_5374() {
+    // GH-5374: <regex>: Back-references to unmatched capture groups
+    // should not match in POSIX basic regular expressions
+    for (syntax_option_type option : {basic, grep}) {
+        g_regexTester.should_not_match("", R"(\(.\)*\1)", option);
+        g_regexTester.should_match("", R"(\(.*\)\1)", option);
+        g_regexTester.should_not_match("bc", R"(\(a\)*b\1c)", option);
+        g_regexTester.should_match("bc", R"(\(a*\)b\1c)", option);
+    }
+}
```
StephanTLavavej, line 1183: Hmm. The fact that you didn't have to edit any existing tests tells me that we didn't have any coverage for any of the grammars. Which means that we should add a test for ECMAScript, to verify that it does match. I will verify and push changes.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej (approved):

StephanTLavavej:
Resolved adjacent-add conflicts in `VSO_0000000_regex_use`.

PR #5378, opened 2025-03-28
Prepare ranges test for MSVC fix

cdacamar:
The compiler recently addressed an [external bug](https://developercommunity.visualstudio.com/t/conditional-operator-gives-wrong-result-/10665330) and along with that is a better system for handling user-authored casts in `decltype` expressions.

The fix above does not completely fix the compiler behavior mentioned in the workaround, it only makes the compiler more consistently return the type determined in cast expressions.

### DIFF
--- a/tests/std/tests/P0896R4_ranges_iterator_machinery/test.cpp
+++ b/tests/std/tests/P0896R4_ranges_iterator_machinery/test.cpp
@@ -1076,7 +1076,11 @@ namespace iterator_cust_move_test {
 #if defined(__clang__) || defined(__EDG__) // TRANSITION, VSO-1008447
     static_assert(same_as<iter_rvalue_reference_t<int(int)>, int (&)(int)>);
 #else // ^^^ no workaround / workaround vvv
+#ifdef _MSVC_INTERNAL_TESTING // TRANSITION, assertion will fire once VSO-2066340 ships.
+    static_assert(same_as<iter_rvalue_reference_t<int(int)>, int (&&)(int)>);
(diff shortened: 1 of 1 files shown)


tests/std/tests/P0896R4_ranges_iterator_machinery/test.cpp
```diff
@@ -1076,7 +1076,11 @@ namespace iterator_cust_move_test {
 #if defined(__clang__) || defined(__EDG__) // TRANSITION, VSO-1008447
     static_assert(same_as<iter_rvalue_reference_t<int(int)>, int (&)(int)>);
 #else // ^^^ no workaround / workaround vvv
+#ifdef _MSVC_INTERNAL_TESTING // TRANSITION, assertion will fire once VSO-2066340 ships.
+    static_assert(same_as<iter_rvalue_reference_t<int(int)>, int (&&)(int)>);
+#else // ^^^ defined(_MSVC_INTERNAL_TESTING) / !defined(_MSVC_INTERNAL_TESTING) vvv
     static_assert(same_as<iter_rvalue_reference_t<int(int)>, int (*)(int)>);
+#endif // ^^^ !defined(_MSVC_INTERNAL_TESTING)
```
StephanTLavavej, line 1083: No change requested: This comment is missing `^^^` at the end but it's fine, I'll be removing this preprocessor logic pretty soon anyways.

StephanTLavavej (approved):

StephanTLavavej:
Cameron is mirroring this in MSVC-PR-622777.

StephanTLavavej:
Thanks for keeping the STL and compiler working together! :smile_cat: :tada: :rocket:

Issue #5379, opened 2025-03-29
`<regex>`: Backslashes in character classes are sometimes not matched in basic regular expressions


muellerj2:
The basic regular expressions `[\(]`, `[\)]`, `[\{]` and `[\}]` do not match the backslash, even though all of them should. The lexer internally recognizes and handles these "escaped" parentheses and braces as the opening and closing parentheses of capture groups or braces of quantifiers, even though there are no capture groups or quantifiers within character classes.

Noticed while writing test coverage for closely related #5244.

PR #5380, opened 2025-03-30
`<regex>`: Revise parsing of escape sequences

### LINKED ISSUE #992: <regex> mishandles locale-based character classes outside of the char range

### LINKED ISSUE #5244: `<regex>`: Some escape sequences are mishandled
There are a number of escape sequences that the parser mistakenly accepts or miscompiles.

## ECMAScript
* Backreferences with leading zero digits (e.g., `\01` for capture group 1) should be rejected. [ECMA-262 3rd ed., Section 15.10.2.11 "DecimalEscape"]
* `\00` and more zero digits should be rejected and not be interpreted as an escape for NUL. Only `\0` is a valid escape sequence for NUL. [ECMA-262 3rd ed., Section 15.10.2.11 "DecimalEscape"]
* When a custom traits implementation defines a new character class "z", `[\z]` matches the characters in this class and not the character z. (Meanwhile, `\z` without brackets matches the character z and not the characters in the class "z".) [ECMA-262 3rd ed., Sections 15.10.1 "Patterns" and 15.10.2.12 "CharacterClassEscape"]
* `[\b]` should match U+0008 BACKSPACE, not b. [ECMA-262 3rd ed., Section 15.10.2.19 "ClassEscape"]

## awk

See [Section "Regular expressions"](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/awk.html#tag_20_06_13_04) in the awk specification.

* Octal escape sequences are not parsed correctly in square-bracket character class definitions. (E.g., `[\040]` should match U+0020 SPACE.)
* Similarly, `[\"]` and `[\/]` match backslashes as well even though they shouldn't.
* While the awk specification says that using unspecified escape sequences results in undefined behavior, I think we should reject them. (I believe we should handle this differently from ECMAScript mode, where unrecognized escape sequences just yield the escaped character.)

muellerj2:
Fixes #5244. There is still one remaining issue related to escape sequences in basic regular expressions. I created #5379 to track this.

## Changes

* ECMAScript: `[\b]` now matches "backspace" (see [the semantics of the production in the recent ECMAScript standard, unchanged since the 3rd edition](https://262.ecma-international.org/#sec-patterns-static-semantics-character-value)). I repurposed and  renamed one of the `_L_flags` to enable this oddity in the parser for ECMAScript. This flag is only set for ECMAScript and had previously been used to represent that the backslash can escape itself in ECMAScript -- but this is superfluous because the flag `_L_ident_ECMA` implies the same. 
* ECMAScript: `[\z]` and `[\Z]` no longer match a custom character class "z" provided by a custom regex traits class. Instead, they will match the characters "z" and "Z", as the escapes also do outside of squared character classes. (This change will also make it easier to solve #992 in the future.)
* ECMAScript: The standard states that a decimal escape starting with digit `\0` represents the literal NUL. The standard also demands that such an escape must not be followed by another decimal digit ([which is still specified this way in recent ECMAScript standards](https://262.ecma-international.org/#prod-CharacterEscape)), so we now throw an `error_escape` in this case. As a consequence of this change, a backreference escape can no longer be written with leading zeros (see also [the most recent ECMAScript standard](https://262.ecma-international.org/#sec-regular-expressions)).
* basic, grep: The escape sequence `\|` is no longer interpreted to mean the character `|`. Instead, it is rejected, since its meaning is undefined according to the POSIX standard version referenced by the C++ standard ([Section 9.3.2](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_03_02)). Users can write `|` instead.
  * The current edition of the POSIX standard makes the meaning of `\|`implementation-defined (in a not particularly good way, I think): It states that `\|` could either be interpreted as the character `|` or an actual pipe ([Section 9.3.2 in the recent standard](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap09.html#tag_09_03_02)). The standard then continues that a future version may require `\|` to denote a pipe.  
  In any case, the interpretation of `\|` as the character `|`, which was implemented before this PR, is not the one recommended by the current version of the standard.
* basic, grep: The regular expression `]` is accepted now, as [the POSIX standard requires because it does not declare `]` to be a special character](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_03_02)). (I noticed this issue when I added `[\]]` as a test case.)
  * Unfortunately, there is no truly suitable `_L_flags` for this, so I check one of the flags that are set for basic regular expressions but not for any of the other supported grammars. Setting `_L_paren_bal` for basic regular expressions would make the parser accept wrong usages of `\}` and `\)` as well.
* awk: As required by the POSIX standard ([see here](https://pubs.opengroup.org/onlinepubs/009604399/utilities/awk.html#tag_04_06_13_04)), octal escape sequences and all awk-specific character escapes are now recognized within squared character classes (before this PR, only the escapes \a, \b, \f, \n, \r, \t and \v were recognized). On the other hand, unknown escape sequences are now rejected in character classes, because their meaning is ambiguous:
  * The underlying POSIX extended regular expressions don't treat \ as an escape character in character classes, so `[\a]` matches the backslash and a. This was how these escape sequences were interpreted before this PR.
  * The POSIX standard leaves the interpretation of unknown sequences undefined in the awk specification.
  * Many awk implementations treat such unknown sequences as identity escapes (as `std::regex` does in ECMAScript mode, too), but others stick closer to the underlying extended regular expressions semantics and treat them as backslash + character.

## `\}` and `\]` in extended regular expressions
According to the POSIX standard referenced by the C++ standard, the escape sequence `\}` is undefined in extended regular expressions. However, [more recent POSIX versions do define it](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap09.html#tag_09_04_02) and it also feels unbalanced to reject `\}` while `\{` is accepted, so I opted to keep supporting the escape sequence `\}`. 
* The parser supports the escape sequence `\[` but not `\]`. More recent POSIX versions define `\]` as well in basic and extended regular expressions. Should we add support for `\]`?


stl/inc/regex
```diff
@@ -4406,22 +4407,24 @@ bool _Parser<_FwdIt, _Elem, _RxTraits>::_CharacterEscape() { // check for valid
...
-            }
+    if ((_L_flags & _L_bzr_chr) && _Traits.value(_Char, 10) == 0) { // handle \0
+        _Next();
+        if (_Traits.value(_Char, 10) != -1) {
+            _Error(regex_constants::error_escape);
+        }
+        _Nfa._Add_char2(_Elem{});
+    } else if ((_L_flags & _L_bckr)
+               && _DecimalDigits2(regex_constants::error_backref,
+                   (_L_flags & _L_lim_bckr) ? _Bre_max_backref_digits : INT_MAX)) { // check for valid back reference
```
muellerj2, line 4418: This is processing the first character as a decimal digit twice, but I didn't find an elegant way to adjust the control flow such that this is avoided.

muellerj2: I adjusted the logic now to avoid this double-processing of the first digit in the latest commit, but I'm not sure if I like this change. If you prefer the code before these changes, feel free to undo the commit.

StephanTLavavej:
I am reviewing this now - thanks as always for the extremely detailed PR description with citations, it is super appreciated for these complex changes. I edited it to fix a typo that made a citation harder to understand - please meow if I messed it up instead.

> The parser supports the escape sequence `\[` but not `\]`. More recent POSIX versions define `\]` as well in basic and extended regular expressions. Should we add support for `\]`?

Yes, I would be in favor of that, in a followup PR - it seems analogous to implementing a C++ Defect Report, and I suspect 99.9% of users would already expect that behavior.

tests/std/tests/GH_005244_regex_escape_sequences/test.cpp
```diff
@@ -0,0 +1,882 @@
...
+    g_regexTester.should_not_match("\\", R"([\111])", awk);
+    g_regexTester.should_match("\111", R"([\1111])", awk);
+    g_regexTester.should_match("1", R"([\1111])", awk);
+    g_regexTester.should_not_match("\\", R"([\1111])", awk);
+    g_regexTester.should_match("A", R"([\101])", awk);
+    g_regexTester.should_not_match("\\", R"([\101])", awk);
+    g_regexTester.should_not_match("1", R"([\101])", awk);
+    g_regexTester.should_not_match("0", R"([\101])", awk);
+    g_regexTester.should_not_match("\1", R"([\101])", awk);
+    g_regexTester.should_not_match("\0", R"([\101])", awk);
```
StephanTLavavej, line 757: Plain `"\0"` will result in an empty `std::string` being constructed for the function call. We need `"\0"s`.

tests/std/tests/GH_005244_regex_escape_sequences/test.cpp
```diff
@@ -0,0 +1,882 @@
...
+    check_classescape_noescape("+", option);
+    check_classescape_noescape("?", option);
+    check_classescape_noescape("|", option);
+
+    // TRANSITION, GH-5379
+    if (option & (extended | egrep | awk)) {
+        check_classescape_noescape("(", option);
+        check_classescape_noescape(")", option);
+        check_classescape_noescape("{", option);
+    }
```
StephanTLavavej, line 634: We're in `test_gh_5244_classescape_posix_not_awk`, so we don't need to test for awk. And in fact, this is forming `string pattern = "[\\" + c + "]";`, which awk will throw exceptions for (as tested in the awk-specific function).

stl/inc/regex
StephanTLavavej: I like the new logic, thanks!

StephanTLavavej (approved):

StephanTLavavej:
Thanks for this amazing overhaul and extreme attention to detail! :heart_eyes_cat: I pushed one significant change to the product code and a few to the test code - please meow if I messed anything up.

StephanTLavavej:

StephanTLavavej:
Resolved adjacent-edit conflicts with #5364 in `<regex>` where:

* it changed to `_ClassAtom(bool)` and this PR changed to `_ClassEscape3()`
* it changed to `_ClassAtom(const bool _Initial)` and this PR expanded "check for valid escape sequence"

StephanTLavavej (approved):

StephanTLavavej:
:arrow_left: :runner: :tada:
===

PR #5381, opened 2025-03-31
Added a missing case for `__std_is_file_not_found()`.

tylerbrawl:
I noticed some rather peculiar behavior regarding the Win32 API `GetFileAttributesExW()`. Specifically, I found that upon calling this function with an invalid path (in my case, a remote URI) in two separate applications, both would fail as expected. What is interesting is that - for some inexplicable reason - the error code returned by `GetLastError()` differed between these two applications. Specifically, it was always either one of the following:

  * `ERROR_INVALID_NAME`: The file name, directory name, or volume label syntax is incorrect.

  * `ERROR_BAD_PATHNAME`: The specified path is invalid.

When this occurred, I was making use of the throwing `std::filesystem::exists()` overload. So, you could imagine my surprise when this function threw an exception on one of the applications and returned `false` as expected on the other. I have not spent any real time investigating the cause for this difference in behavior myself. In any case, though, it is clear that the `__std_is_file_not_found()` function should have returned `true` when given an `__std_win_error` value corresponding to either of the two error codes.

This commit thus modifies the `__std_is_file_not_found()` function to take into consideration an additional Win32 error code: `ERROR_BAD_PATHNAME`. Prior to this commit, this value did not have a unique value assigned to it in the `__std_win_error` `enum class`. Thus, a new value by the name of `__std_win_error::_Bad_pathname` was also added in this commit.

### DIFF
--- a/stl/inc/xfilesystem_abi.h
+++ b/stl/inc/xfilesystem_abi.h
@@ -39,6 +39,7 @@ enum class __std_win_error : unsigned long {
     _Insufficient_buffer       = 122, // #define ERROR_INSUFFICIENT_BUFFER        122L
     _Invalid_name              = 123, // #define ERROR_INVALID_NAME               123L
     _Directory_not_empty       = 145, // #define ERROR_DIR_NOT_EMPTY              145L
+    _Bad_pathname              = 161, // #define ERROR_BAD_PATHNAME               161L
     _Already_exists            = 183, // #define ERROR_ALREADY_EXISTS             183L
     _Filename_exceeds_range    = 206, // #define ERROR_FILENAME_EXCED_RANGE       206L
     _Directory_name_is_invalid = 267, // #define ERROR_DIRECTORY                  267L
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
Thanks for the info. Do you have a moment to provide the invalid path ("remote URI") that was causing the error?

StephanTLavavej:
Aha, I believe you've fixed DevCom-10825862 VSO-2345522 "`std::filesystem::status` throws on bad paths when the current directory is a network path" which matches your description, but also describes the necessary environment.

With VS 2022 17.14 Preview 2, I repro:

```
C:\Temp>type meow.cpp
```
```cpp
#include <filesystem>
#include <print>
using namespace std;

int main() {
    try {
        println("_MSVC_STL_UPDATE: {}", _MSVC_STL_UPDATE);
        const bool b = filesystem::exists(R"(https://example.com/bogus/path)");
        println("{}", b);
    } catch (const filesystem::filesystem_error& e) {
        println("{}", e.what());
    }
}
```
```
meow.cpp
_MSVC_STL_UPDATE: 202501
false

C:\Temp>pushd \\localhost\C$\Temp

Y:\Temp>meow
_MSVC_STL_UPDATE: 202501
exists: The specified path is invalid.: "https://example.com/bogus/path"
```

I observe that your fix is effective:

```
meow.cpp
_MSVC_STL_UPDATE: 202503
false

C:\Temp>pushd \\localhost\C$\Temp

X:\Temp>meow
_MSVC_STL_UPDATE: 202503
false
```

Awesome! :heart_eyes_cat:

StephanTLavavej (approved):

StephanTLavavej:

PR #5390, opened 2025-04-02
`STL.natvis`: Add alternative type name for `time_point`

Nerixyz:
`clang-cl` will produce 
```
std::chrono::time_point<std::chrono::system_clock,std::chrono::duration<long long,std::ratio<1,10000000>>>
```
as the type name for `std::chrono::system_clock::time_point` whereas MSVC will produce 
```
std::chrono::time_point<std::chrono::system_clock,std::chrono::duration<__int64,std::ratio<1,10000000>>>
```
Notice: `long long` vs `__int64` (https://github.com/llvm/llvm-project/issues/134156).

As a workaround, I added the type from `clang-cl` as an `<AlternativeType>`.


StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for helping to support Clang as a first-class citizen! :tada: :cat: :cat2:

PR #5398, opened 2025-04-10
`<barrier>`: Fix constructor's exception specification

JMazurkiewicz:
Currently, `barrier`'s constructor is unconditionally `noexcept`, but according to [`[thread.barrier.class]/11`](http://eel.is/c++draft/thread.barrier.class#11) this is incorrect. This PR fixes the problem by changing exception specification from `noexcept(true)` to `noexcept(is_nothrow_move_constructible_v<_Completion_function>)`.

### DIFF
--- a/stl/inc/barrier
+++ b/stl/inc/barrier
@@ -77,8 +77,8 @@ public:
 
     using arrival_token = _Arrival_token<_Completion_function>;
 
-    constexpr explicit barrier(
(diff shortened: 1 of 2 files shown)


StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for noticing and fixing this bug! :lady_beetle: :detective: :hammer_and_wrench:

PR #5399, opened 2025-04-10
`<regex>`: Permit escaping of closing brackets in POSIX regexes

muellerj2:
Short follow-up to #5380 to support escapes for closing brackets in POSIX regexes in accordance with the most recent POSIX standard version: https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap09.html#tag_09_03_02 for basic and https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap09.html#tag_09_04_02 for extended regexes.

Because of the (expected) test failures caused by this change, I noticed that the tests in `gh_5244_atomescape_posix_common` are called twice for `extended` and `egrep` from `test_gh_5244_atomescape_extended_or_egrep`: Once via `gh_5244_atomescape_posix_not_awk` and once via `gh_5244_atomescape_extended_egrep_awk`. That's once more than necessary, so let's get rid of this double-testing. I opted to remove the call of `gh_5244_atomescape_posix_common` in `gh_5244_atomescape_extended_egrep_awk`, but this means a call has to be added to `gh_5244_atomescape_awk` so that these common POSIX tests are still exercised in awk mode.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -4348,6 +4348,7 @@ bool _Parser<_FwdIt, _Elem, _RxTraits>::_IsIdentityEscape(bool _In_character_cla
         return true;
     case _Meta_dot:
     case _Meta_lsq:
+    case _Meta_rsq:
     case _Meta_star:
     case _Meta_caret:
     case _Meta_dlr:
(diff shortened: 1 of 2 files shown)


StephanTLavavej (approved):

StephanTLavavej:

PR #5400, opened 2025-04-11
Add `std::fill` and `std::fill_n` benchmarks

barcharcraz:
I wrote these benchmarks while fixing some issues with memset loop recognition in the compiler, but I think they are more broadly useful.


benchmarks/src/fill.cpp
```diff
@@ -0,0 +1,127 @@
+#include <benchmark/benchmark.h>
+#include <cstring>
+#include <memory>
+#include <ranges>
+#include <vector>
+namespace {
+    template <typename Contained, Contained Value>
+    void handwritten_loop(benchmark::State& state) {
+        const size_t r0 = static_cast<size_t>(state.range(0));
+        std::unique_ptr<Contained[]> buffer(new Contained[r0]);
```
AlexGuteniev, line 10: Any advantage of using `new` and not `make_unique` ?

benchmarks/src/fill.cpp
```diff
@@ -0,0 +1,127 @@
+#include <benchmark/benchmark.h>
+#include <cstring>
+#include <memory>
+#include <ranges>
+#include <vector>
```
AlexGuteniev, line 5: +newline

benchmarks/src/fill.cpp
AlexGuteniev: You might want to use `vector` with `highly_aligned_allocator` / `not_highly_aligned_allocator` to make allocation alignment more predictable. You can have `.data()` for pointer. Currently it can be 16 bytes (`malloc` alignment), or 32 (accidentally) or more (there's a chance)

barcharcraz: No particular reason. I think I'd need `make_unique_for_overwrite` I can switch over.

I used unique_ptr just to avoid double initializing things, even though that's not really a big deal. I didn't know about those allocators.

I guess it's probably best to switch between highly/not-highly aligned for different variants, instead of just 2x-ing the number of tests.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :smile_cat: I pushed significant changes - please meow if you have concerns.

barcharcraz:
nah, lgtm.
However:

I don't really like pulling the allocators out into a template parameter, as I don't think it's worth multiplying the test matrix for them (although having some benchmarks use one and some the other probably was bad).

AlexGuteniev:
> I don't really like pulling the allocators out into a template parameter, as I don't think it's worth multiplying the test matrix for them (although having some benchmarks use one and some the other probably was bad).

If there's little difference in performance between allocators, I'd just use the worse one (`not_highly_aligned_allocator`), to make AVX2 or AVX-512 vectorization, if they exist here, of less advantage.

StephanTLavavej:
Sounds good to me since I saw no perf difference. I'll validate and push.

barcharcraz:
lgtm

StephanTLavavej:

StephanTLavavej:
Thanks for filling out our benchmark directory! :stopwatch: :chart_with_upwards_trend: :joy_cat:

Issue #1528, opened 2020-12-11
<regex>: regex_match gets caught in recursive loop until stack overflow occurs


freezerdev:
**Describe the bug**
I have found a case where calling regex_match can result in a recursive loop until a stack overflow occurs.  Problem occurs when run as x64 debug.  When run as x64 release an exception occurs.  And when run as x86 debug or release, program works as expected.

**Command-line test case**
```
#include <iostream>
#include <regex>

int main(void)
{
	auto flags = std::regex_constants::ECMAScript | std::regex::icase;

	try
	{
		std::wregex rgx(L"^http[s]?\\:\\/\\/([^.]+\\.)*github\\.com\\/.*$", flags);
		bool bResult = std::regex_match(L"https://www.google.com/search?client=firefox-b-1-d&q=tea%20box&tbs=lf:1,lf_ui:10&tbm=lcl&sxsrf=ALeKk03GLkevbpot_6JmlWw6_IJKeKBKbw:1607540768824&rflfq=1&num=10&rldimm=11474290690396232776&lqi=Cgd0ZWEgYm94Gd9r-yQGeFTPSKvez_DmgICACFoWCgd0ZWEgYm94EAAQASIHdGVhIGJveIABAQ&phdesc=-P7Y2CJJHVU&ved=2ahUKEwjA9PKDzMHtAhVTMH0KHfnkAUIQvS4wBHoECAQQOQ&rlst=f#rlfi%3Dhd%3A%3Bsi%3A11474290690396232776%2Cl%2CCgd0ZWEgYm94Gd9r-yQGeFTPSKvez_DmgICACFoWCgd0ZWEgYm94EAAQASIHdGVhIGJveIABAQ%2Cy%2C-P7Y2CJJHVU%3Bmv%3A%5B%5B35.638039299999996%2C-117.9412485%5D%2C%5B34.634577799999995%2C-120.8545755%5D%5D%3Btbs%3Alrf%3A%211m4%211u3%212m2%213m1%211e1%211m4%211u2%212m2%212m1%211e1%212m1%211e2%212m1%211e3%213sIAE%2Clf%3A1%2Clf_ui%3A10", rgx);

		if(bResult)
			std::cout << "Match" << std::endl;
		else
			std::cout << "Does not match" << std::endl;
	}
	catch(const std::regex_error &e)
	{
		std::cout << "Exception" << std::endl;
	}

	return 0;
}
```

**Expected behavior**
[...]

PR #5405, opened 2025-04-13
`<regex>`: Simplify matching of `_N_if` NFA nodes in leftmost-longest mode

muellerj2:
Since we settled on some reasonable semantics for leftmost-longest matching in #5218, I think we should remove the code for some other (abandoned) attempt to implement the leftmost-longest rule in `_Matcher::_Do_if`: An attempt to set `Tgt_state` to the leftmost-longest match found under this `_N_if` node.

This is neither necessary (in leftmost-longest mode, we take the final result from `_Res`, not `_Tgt_state`) nor sufficient (it assigns one of the longest matches to `_Tgt_state`, but if there are several of the same length it doesn't necessarily pick the correct one among these matches).

On the other hand, this saves 96 bytes of stack space per call to `_Do_if` in debug mode, slightly alleviating the stack overflow issues (#997, #1528).

Potentially leaving `_Tgt_state` in a garbage state is fine here, because none of the functions further up on the callstack rely on its value:
* `_Match_pat` in the `_N_if` case: Will just return immediately to its caller without changing `_Tgt_state`.
* `_Do_rep0`: Can't be a caller of `_Match_pat` because a repetition containing an `_N_if` node anywhere is not simple.
* `_Do_assert`/`_Do_neg_assert`: Can't be the callers of `_Match_pat` because there are no lookahead assertions in the POSIX grammars that demand application of the leftmost-longest rule. (But even if there were lookahead assertions -- notwithstanding their currently unknown semantics -- it would be fine in the sense that the matcher wouldn't crash: `_Do_assert` would reset the position pointer to a savepoint, while `_Do_neg_assert` would fail, resulting in `_Match_pat` returning immediately to its caller so that the remaining analysis here applies. Even so, there is the issue that `_Do_assert` would not reset the capture groups in `_Tgt_state` -- but we don't know what to set them to either as long as we don't know the semantics of such assertions. Nevertheless, all "valid" capture groups would still point to legal ranges in the input, so even the matcher with this PR's change wouldn't crash. This means we wouldn't have to worry about a newer parser emiting assertion nodes because running them with the old matcher would at worst produce wrong results. To get correct semantics, an updated parser *and* matcher are required, but this would also be the case if we didn't do this PR's change.)
* Another `_Do_if`: Will either reset `_Tgt_state` to some savepoint before calling `_Match_pat` or immediately return to its caller and leave `_Tgt_state` as-is.
* `_Do_rep`: Will either reset `_Tgt_state` to some savepoint before doing the next `_Match_pat` call or return to its caller while leaving `_Tgt_state` as-is.
* `_Match_pat` in the `_N_rep` or `_N_end_rep` cases: Will just return immediately to its caller without changing `_Tgt_state`.
* `_Match`: Will evaluate `_Res` in leftmost-longest mode, not `_Tgt_state`. (`_Match` only evaluted `_Res` before #5218 as well, so this change also doesn't pose a problem if old and new functions are mixed.)

So in all cases, either `Tgt_state` isn't evaluated anymore or it is reset to some savepoint before it is used again.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -3190,8 +3190,6 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Do_if(_Node_if* _Node) { // apply
     }
 
     // see if there is a longer match
-    _Tgt_state_t<_It> _Final = _Tgt_state;
-    auto _Final_len          = _STD distance(_St._Cur, _Tgt_state._Cur);
     for (;;) { // process one branch of if
         _Node = _Node->_Child;
         if (!_Node) {
@@ -3199,17 +3197,8 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Do_if(_Node_if* _Node) { // apply
         }
 
         _Tgt_state = _St;
-        if (_Match_pat(_Node->_Next)) { // record match if it is longer
-            const auto _Len = _STD distance(_St._Cur, _Tgt_state._Cur);
-            if (_Final_len < _Len) { // memorize longest so far
-                _Final     = _Tgt_state;
-                _Final_len = _Len;
-            }
-        }
+        (void) _Match_pat(_Node->_Next);
     }
-
-    // set the input end to the longest match
-    _Tgt_state = _Final;
     return true;
 }
 


StephanTLavavej (approved):

StephanTLavavej:
Thanks for the exceptional analysis and writeup - I wouldn't have noticed that this code was dead! :heart_eyes_cat:

Anything to mitigate stack consumption is also greatly appreciated.

StephanTLavavej:

StephanTLavavej:
Thanks again for this simplification! :broom: :sparkles: :cat:

PR #5407, opened 2025-04-14
`<optional>`: Extend compiler bug workaround to MSVC

StephanTLavavej:
#5369 revealed VSO-2444237 "\[RWC\]\[prod/fe\]\[std:c++latest\]\[Regression\] LLVM failed with Assertion failed: `ParseTree::Utility::SigStack::sigStack.size() == 1`".

@xiangfan-ms tracked this down to the introduction of `_Derived_from_optional` and reduced it to a library-free repro. The compiler is having trouble with the lambda's templated function call operator.

To unblock our Real World Code test suite, we should extend the workaround we're currently using for EDG to also cover MSVC.

### DIFF
--- a/stl/inc/optional
+++ b/stl/inc/optional
@@ -952,7 +952,7 @@ _NODISCARD constexpr bool operator>=(const _Ty1& _Left, const optional<_Ty2>& _R
 }
 
 #if _HAS_CXX20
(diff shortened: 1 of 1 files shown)


davidmrdavid (approved):

StephanTLavavej:

PR #5412, opened 2025-04-17
`<random>`: Fast `discard()` for `minstd_rand` and `minstd_rand0` via fast exponentiation

mirounga:
discard() minstd_rand and minstd_rand0 now takes logarithmic time
due to fast exponentiation and avoiding constant divisions.


stl/inc/random
```diff
@@ -617,12 +617,37 @@ public:
...
+            auto _Temp = static_cast<unsigned long long>(_Prev);
+            auto _Mul  = static_cast<unsigned long long>(_Ax);
+
+            for (;;) {
+                if (_Nskip & 1) {
+                    _Temp = _Temp * _Mul;
+                    _Temp = (_Temp >> 31) + (_Temp & _Mx);
+                    _Temp = _Temp < _Mx ? _Temp : _Temp - _Mx;
+                }
+                if (_Nskip >>= 1) {
```
StephanTLavavej, line 632: No change requested: We generally avoid directly testing integers for non-zero-ness (except when working with bitmasks), and in modern code we generally avoid mixing side effects into conditions (although our older code did that quite often).

However, in this case, I am okay with the style as-is. Perhaps it's the near-symmetry of the `_Temp` and `_Mul` logic, but I feel that mechanically following our usual conventions would decrease clarity.

stl/inc/random
```diff
@@ -617,12 +617,37 @@ public:
 
     void discard(unsigned long long _Nskip) noexcept /* strengthened */ {
         // discard _Nskip elements
-        auto _Temp = _Prev;
-        for (; 0 < _Nskip; --_Nskip) {
-            _Temp = _Next_linear_congruential_value<_Uint, _Ax, _Cx, _Mx>(_Temp);
-        }
+        if constexpr (_Cx == 0 && _Mx == 2147483647) {
+            // for minstd_rand and minstd_rand0 we can improve performance by
```
StephanTLavavej, line 621: Let's update the benchmark for this.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: I cleaned up a mistakenly added file, added a benchmark, and got the following hilarious results on my 5950X:

Benchmark                             |    Before | After   | Speedup
--------------------------------------|----------:|--------:|--------
`BM_discard<std::minstd_rand>/0`      |   1.68 ns | 1.69 ns | 0.99
`BM_discard<std::minstd_rand>/1`      |   3.32 ns | 2.93 ns | 1.13
`BM_discard<std::minstd_rand>/8`      |   14.8 ns | 5.99 ns | 2.47
`BM_discard<std::minstd_rand>/64`     |    107 ns | 9.52 ns | 11.24
`BM_discard<std::minstd_rand>/512`    |    843 ns | 16.6 ns | 50.78
`BM_discard<std::minstd_rand>/4096`   |   6805 ns | 19.0 ns | 358.16
`BM_discard<std::minstd_rand>/32768`  |  53885 ns | 24.6 ns | 2190.45
`BM_discard<std::minstd_rand>/262144` | 430735 ns | 29.1 ns | 14801.89

Skipping an exponential amount of work is pretty awesome.

StephanTLavavej:

PR #5419, opened 2025-04-19
Fix [flat.map.modifiers] bugs found by libcxx tests

seishun:
* Add missing constraints:
  * `is_constructible_v<pair<key_type, mapped_type>, Args...>` is `true`
* Always call `insert` in `insert`


StephanTLavavej (approved):
Wow, the Standardese for the `emplace()` constraint is very subtly hidden - good catch!

PR #5420, opened 2025-04-19
Vectorize `is_sorted_until`

AlexGuteniev:
# 🗺️ Overview

Another vectorziation, apparently one of the last relatively low hanging fruits 🍎.
Uses offset load like in `remove`, `unique` or `adjacent_find`, and comparisons from `minmax` family.

# ⚖️ Less and greater

Sorting in both ascending and descending order makes equal sense, although the standard picks `less` as the default predicate. We support both `less` and `greater` here. This is done for the first time. It wasn't done for `minmax` family in the past, because for `minmax` it is very squirrelly to reverse the predicate 🐿️.

To reduce the number of functions, `less` and `greater` distinguished by a `bool` parameter, which turns to a couple of offsets for the loads, making difference which of the loads are actually with a negative offset.

The runtime cost of applying extra offset on each iteration is nonzero, but still I expect it to be small.

Fun observation: one could attempt to pass `less_equal` or `greater_equal` too, and expect the algorithm to be checking for strictly asceding/desceding values. But it is UB according to the sorted algorithms requirements, and there's `_DEBUG_LT_PRED` to alert about that 😈. The vectorization will not affect this case anyhow, as it looks for the specific predicates.

# 🐞 Floating bugs

We support floating point types here too.

I don't expect anything bad this time. This algorithm is position-based, and we only use the intrinsics which do correct math.

One caveat I see is signaling values🚨. As this is early return algorithm, for some data it may be expected not to signal when handling elements one-by-one, and it would start signaling when working with vectors. But we already do not enable floating vector algorithms for `/fp:except`, so it should be fine:
https://github.com/microsoft/STL/blob/5762e6bcaf7f5f8b5dba0a9aabf0acbd0e335e80/stl/inc/xutility#L55-L59

If anything goes wrong, the escape hatch is still there to help.

# ➕ Unsigned values

Like in `minmax_element`, we need the `*_gt` intrinsics that only exist for signed values, so we have to do sign correction.

Unlike in `minmax_element`, we apply the correction at compile-time, having different signed and unsigned functions. The whole algorithm is faster, so the correction overhead is more noticeable.

Can be changed if the benchmark results do not show the signed/unsigned difference persuasive enough. Like we have ~33.5 for `int8_t` and ~39.5 for `iint8_t`, and in runtime correction applying I expect ~39.5 for both, but with less machine code generated, and with simpler dispatch in the header.

# ⏱️ Benchmark results

Benchmark                                  |  Before   |  After   | Speedup
-------------------------------------------|-----------|----------|-------
bm<std::int8_t, AlgType::Std>/3000/1800    |  434 ns   | 33.7 ns  | 12.88
bm<std::int8_t, AlgType::Rng>/3000/1800    |  434 ns   | 33.4 ns  | 12.99
bm<std::int16_t, AlgType::Std>/3000/1800   |  440 ns   | 67.8 ns  |  6.49
bm<std::int16_t, AlgType::Rng>/3000/1800   |  432 ns   | 66.5 ns  |  6.50
bm<std::int32_t, AlgType::Std>/3000/1800   |  432 ns   |  129 ns  |  3.35
bm<std::int32_t, AlgType::Rng>/3000/1800   |  431 ns   |  130 ns  |  3.32
bm<std::int64_t, AlgType::Std>/3000/1800   |  437 ns   |  238 ns  |  1.84
bm<std::int64_t, AlgType::Rng>/3000/1800   |  440 ns   |  241 ns  |  1.83
bm<std::uint8_t, AlgType::Std>/3000/1800   |  437 ns   | 39.4 ns  | 11.09
bm<std::uint8_t, AlgType::Rng>/3000/1800   |  440 ns   | 39.5 ns  | 11.14
bm<std::uint16_t, AlgType::Std>/3000/1800  |  438 ns   | 79.0 ns  |  5.54
bm<std::uint16_t, AlgType::Rng>/3000/1800  |  433 ns   | 80.9 ns  |  5.35
bm<std::uint32_t, AlgType::Std>/3000/1800  |  432 ns   |  146 ns  |  2.96
bm<std::uint32_t, AlgType::Rng>/3000/1800  |  429 ns   |  147 ns  |  2.92
bm<std::uint64_t, AlgType::Std>/3000/1800  |  500 ns   |  279 ns  |  1.79
bm<std::uint64_t, AlgType::Rng>/3000/1800  |  446 ns   |  280 ns  |  1.59
bm<float, AlgType::Std>/3000/1800          |  658 ns   |  112 ns  |  5.88
bm<float, AlgType::Rng>/3000/1800          |  670 ns   |  100 ns  |  6.70
bm<double, AlgType::Std>/3000/1800         |  653 ns   |  179 ns  |  3.65
bm<double, AlgType::Rng>/3000/1800         |  657 ns   |  177 ns  |  3.71


benchmarks/src/is_sorted_until.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include <vector>
+
+#include "skewed_allocator.hpp"
+#include "utility.hpp"
+
+enum class AlgType { Std, Rng };
+
+template <class T, AlgType Alg>
+void bm(benchmark::State& state) {
+    const std::size_t size     = static_cast<size_t>(state.range(0));
```
StephanTLavavej, line 17: Inconsistent qualification of `std::size_t`.

benchmarks/src/is_sorted_until.cpp
```diff
@@ -0,0 +1,71 @@
...
+    const std::size_t sort_pos = static_cast<size_t>(state.range(1));
+
+    std::vector<T, not_highly_aligned_allocator<T>> v;
+    if constexpr (std::is_integral_v<T>) {
```
StephanTLavavej, line 21: Need `<type_traits>`.

benchmarks/src/is_sorted_until.cpp
```diff
@@ -0,0 +1,71 @@
...
+        if constexpr (Alg == AlgType::Std) {
+            benchmark::DoNotOptimize(std::is_sorted_until(v.begin(), v.end()));
+        } else {
+            benchmark::DoNotOptimize(std::ranges::is_sorted_until(v));
+        }
+    }
+}
+
+void common_args(auto bm) {
+    bm->ArgPair(3000, 1800);
```
StephanTLavavej, line 45: This is pre-existing in other tests, but the parameter `bm` shadows the function `bm`. As the function name shows up in the output, I think that's the one that should indicate what it's testing.

benchmarks/src/is_sorted_until.cpp
```diff
@@ -0,0 +1,71 @@
...
+}
+
+BENCHMARK(bm<std::int8_t, AlgType::Std>)->Apply(common_args);
```
StephanTLavavej, line 48: Need `<cstdint>`.

stl/inc/algorithm
```diff
@@ -74,6 +74,17 @@ __declspec(noalias) _Min_max_8u __stdcall __std_minmax_8u(const void* _First, co
 __declspec(noalias) _Min_max_f __stdcall __std_minmax_f(const void* _First, const void* _Last) noexcept;
 __declspec(noalias) _Min_max_d __stdcall __std_minmax_d(const void* _First, const void* _Last) noexcept;
 
+const void* __stdcall __std_is_sorted_until_1i(const void* _First, const void* _Last, const bool _Greater) noexcept;
```
StephanTLavavej, line 77: These declarations shouldn't have top-level const on `bool _Greater`.

stl/inc/algorithm
```diff
@@ -192,6 +203,49 @@ auto _Minmax_vectorized(_Ty* const _First, _Ty* const _Last) noexcept {
     }
 }
 
+template <class _Ty>
+_Ty* _Is_sorted_until(_Ty* const _First, _Ty* const _Last, const bool _Greater) noexcept {
```
StephanTLavavej, line 207: This should follow the `_Meow_vectorized` naming convention.

stl/inc/algorithm
```diff
@@ -192,6 +203,49 @@ auto _Minmax_vectorized(_Ty* const _First, _Ty* const _Last) noexcept {
...
+template <class _Ty>
+_Ty* _Is_sorted_until(_Ty* const _First, _Ty* const _Last, const bool _Greater) noexcept {
+    constexpr bool _Signed = is_signed_v<_Ty>;
+
+    if constexpr (is_pointer_v<_Ty>) {
+#ifdef _WIN64
+        return const_cast<_Ty*>(static_cast<const _Ty*>(::__std_is_sorted_until_8u(_First, _Last, _Greater)));
+#else
+        return const_cast<_Ty*>(static_cast<const _Ty*>(::__std_is_sorted_until_4u(_First, _Last, _Greater)));
+#endif
```
StephanTLavavej, line 215: I am torn - on the one hand, this is clear. On the other hand, I don't believe you need the pointer cases at all, as they should be exactly absorbed by the integer cases. The critical bit is that `is_signed_v` is false for non-arithmetic types. A comment mentioning that would be sufficient. Do you think that's too weird?

stl/inc/algorithm
```diff
@@ -8040,6 +8094,25 @@ namespace ranges {
...
+
+                    if constexpr (is_pointer_v<_It>) {
+                        return _Result;
+                    } else {
+                        return _First + static_cast<iter_difference_t<_It>>(_Result - _First_ptr);
+                    }
+                }
+            }
+        }
+#endif
```
StephanTLavavej, line 8114: Needs comment.

stl/inc/algorithm
```diff
@@ -8040,6 +8094,25 @@ namespace ranges {
             return _First;
         }
 
+#if _USE_STD_VECTOR_ALGORITHMS
+        if constexpr (_Is_min_max_iterators_safe<_It> && sized_sentinel_for<_Se, _It> && is_same_v<_Pj, identity>) {
+            constexpr bool _Is_greater = _Is_predicate_greater<_It, _Pr>;
+            if constexpr (_Is_greater || _Is_predicate_less<_It, _Pr>) {
+                if (!_STD _Is_constant_evaluated()) {
```
StephanTLavavej, line 8101: C++20 should be able to use pretty `is_constant_evaluated`.

stl/inc/algorithm
```diff
@@ -10908,6 +10981,28 @@ _NODISCARD _CONSTEXPR20 _FwdIt is_sorted_until(const _FwdIt _First, _FwdIt _Last
...
+                } else {
+                    _UFirst += static_cast<_Iter_diff_t<decltype(_UFirst)>>(_Result - _First_ptr);
+                }
+
+                _STD _Seek_wrapped(_Last, _UFirst);
+                return _Last;
+            }
+        }
+    }
+#endif
```
StephanTLavavej, line 11004: Needs comment.

stl/src/vector_algorithms.cpp
```diff
@@ -2249,6 +2250,110 @@ namespace {
...
     }
+
+
+    template <class _Traits, class _Ty>
+    const void* __std_is_sorted_until_impl(const void* _First, const void* const _Last, const bool _Greater) noexcept {
+        const ptrdiff_t _Left_off  = 0 - static_cast<ptrdiff_t>(_Greater);
+        const ptrdiff_t _Right_off = static_cast<ptrdiff_t>(_Greater) - 1;
+
+        if constexpr (_Traits::_Vectorized) {
+#ifndef _M_ARM64EC
```
StephanTLavavej, line 2261: This `if constexpr` is empty for ARM64EC - it seems that the guards should be moved to encompass it entirely.

tests/std/include/test_is_sorted_until_support.hpp
```diff
@@ -0,0 +1,43 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#pragma once
+
+#include <algorithm>
+#include <cassert>
+#include <cmath>
+#include <cstddef>
+#include <type_traits>
+#include <utility>
```
StephanTLavavej, line 11: `<cmath>`, `<cstddef>`, `<type_traits>`, and `<utility>` all appear to be unused.

tests/std/include/test_is_sorted_until_support.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include <vector>
+
+#if _HAS_CXX20
+#include <ranges>
+#endif
```
StephanTLavavej, line 16: Perhaps surprisingly, `<algorithm>` is sufficient for `ranges::is_sorted_until`.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -601,6 +602,34 @@ void test_min_max_element_special_cases() {
...
+    input.reserve(dataCount);
+
+    test_case_is_sorted_until(input, less<>{});
+    test_case_is_sorted_until(input, greater<>{});
+
+    for (size_t attempts = 0; attempts < dataCount; ++attempts) {
+        original_input.push_back(static_cast<T>(dis(gen)));
+        input = original_input;
+
+        auto it = input.begin() + uniform_int_distribution<ptrdiff_t>{0, static_cast<ptrdiff_t>(input.size() - 1)}(gen);
```
StephanTLavavej, line 623: This would be clearer if the distribution were extracted into a named variable. It's worth a line.

tests/std/tests/VSO_0000000_vector_algorithms_floats/test.cpp
```diff
@@ -91,14 +91,42 @@ void test_min_max_element_floating_zero_predef() {
...
+void test_is_sorted_until_floating_with_values(mt19937_64& gen, const vector<T>& input_of_input) {
+    uniform_int_distribution<size_t> idx_dis(0, input_of_input.size() - 1);
+
+    vector<T> original_input;
+    vector<T> input;
+    original_input.reserve(dataCount);
+    input.reserve(dataCount);
+
+    test_case_is_sorted_until(input, less<>{});
+    test_case_is_sorted_until(input, greater<>{});
```
StephanTLavavej, line 109: Should include `<functional>`.

tests/std/tests/VSO_0000000_vector_algorithms_floats/test.cpp
```diff
@@ -91,14 +91,42 @@ void test_min_max_element_floating_zero_predef() {
...
+
+    for (size_t attempts = 0; attempts < dataCount; ++attempts) {
+        original_input.push_back(input_of_input[idx_dis(gen)]);
+        input = original_input;
+
+        auto it = input.begin() + uniform_int_distribution<ptrdiff_t>{0, static_cast<ptrdiff_t>(input.size() - 1)}(gen);
```
StephanTLavavej, line 115: Should extract distribution to named variable.

tests/std/tests/VSO_0000000_vector_algorithms_floats/test.cpp
```diff
@@ -91,14 +91,42 @@ void test_min_max_element_floating_zero_predef() {
...
+        sort(input.begin(), it, less<>{});
```
StephanTLavavej, line 116: Should include `<algorithm>`.

StephanTLavavej (changes requested):

stl/inc/algorithm
AlexGuteniev: ### 🔥🐶☕🔥 

We already have this relying on pointers being not signed in `_Minmax_element_vectorized`  wihtout even a comment

AlexGuteniev: But are pointers indeed unsigned? 🤔 
Yes they are, even the explicitly signed ones: https://godbolt.org/z/jnWPd3on4

stl/src/vector_algorithms.cpp
AlexGuteniev: It is better to `static_assert(false)` here, like in two other occurrences.

tests/std/include/test_is_sorted_until_support.hpp
```diff
@@ -0,0 +1,35 @@
...
+    return last;
+}
+
+template <class T, class Comp>
+void test_case_is_sorted_until(const std::vector<T>& input, Comp comp) {
+    auto expected = last_known_good_is_sorted_until(input.begin(), input.end(), comp);
+    auto actual   = std::is_sorted_until(input.begin(), input.end(), comp);
+    assert(expected == actual);
+#if _HAS_CXX20
+    auto actual_r = std::ranges::is_sorted_until(input.begin(), input.end(), comp);
```
StephanTLavavej, line 32: We usually pass ranges instead of iterator-sentinel pairs.

StephanTLavavej (approved):

AlexGuteniev:
> 5950X results

Interesting thing about floating types in your results. In mine, they were consistently slower, but I suspected "code alignment gremlins", rather than slower instructions itself. The instructions have a bit longer encoding, making stepping into that a bit easier. Your result seem to confirm that, having one of floating results consistent with integer results.

AlexGuteniev:
> Your result seem to confirm that, having one of floating results consistent with integer results

And this happens not because of different CPU, but because a different overload of `ranges::is_sorted` was used.

AlexGuteniev:
> > Your result seem to confirm that, having one of floating results consistent with integer results
> 
> And this happens not because of different CPU, but because a different overload of `ranges::is_sorted` was used.

No, it has changed in tests, not benchmark. Then probably merging re-shuffled things.

StephanTLavavej:

StephanTLavavej:
Thanks for sorting this out! :joy_cat: :heart_eyes_cat: :1234:

PR #5425, opened 2025-04-22
Reduce vector algorithms test run time with ASan

AlexGuteniev:
Optimize LKG `search` and `find_end`.
Avoid calling `equal` in a loop.

Similar to #4654


tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -271,10 +271,19 @@ auto last_known_good_find_first_of(FwdItH h_first, FwdItH h_last, FwdItN n_first
 
 template <class RanItH, class RanItN>
 auto last_known_good_search(RanItH h_first, RanItH h_last, RanItN n_first, RanItN n_last) {
-    const auto n_len = n_last - n_first;
+    const ptrdiff_t n_len = n_last - n_first;
 
     for (; h_last - h_first >= n_len; ++h_first) {
-        if (equal(h_first, h_first + n_len, n_first, n_last)) {
+        bool is_eq = true;
```
StephanTLavavej, line 277: This shadows `std::is_eq` from `<compare>`. https://en.cppreference.com/w/cpp/utility/compare/named_comparison_functions

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: This makes a big difference - on my 5950X, with an ASan build of the STL and ASan tests, I measured a reduction from 152.44 seconds to 96.21 seconds, for a speedup of 1.58x.

I removed the ASan label because this isn't necessarily ASan-specific though.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
AlexGuteniev: I made it `is_eq` in attempt to avoid shadowing, thinking that abbreviation is less likely to shadow.
There's `std::pmr::memory_resource::is_equal`, but probably it does not count, as it is a member.

StephanTLavavej:

Issue #5418, opened 2025-04-19
`vector_algorithms.cpp`: split into multiple translation units or otherwise enhance file navigation?


AlexGuteniev:
`vector_algorithms.cpp` grew to about 6000 lines. That's about a half of `<algoirthms>` size. Out of separately compiled files it stands out, with only `xcharconv_ryu_tables.cpp` exceeding it.

The navigation became somewhat inconvenient, especially with a lot of unnamed namespaces, each containing its own functions set:

![Image](https://github.com/user-attachments/assets/57c187e1-07e6-437c-99f3-baa0e5872f33)

Do we want to do something about it?

We can:
 * Do nothing
 * Split `vector_algorithms.cpp` into multiple files
 * Fuse all unnamed namespaces into one
 * Do something diffrerent

jovibor:
> * Fuse all unnamed namespaces into one

Or create new namespace for all these methods, for convenience. Something like `std::__vecalgos`, if it's possible.

StephanTLavavej:
I would prefer to avoid splitting up the file - that interferes with history, makes the STL take longer to build, and makes it somewhat harder to see commonality.

However, since we're gathering so many different implementations, I think it would be an improvement to create more grouping by introducing separate namespaces for each family of support functions, within the unnamed namespace (which continues to make it clear that these are never used by anything outside).

Basically, I don't think that large files are inherently problematic, but "who uses what" is more important to keep clear.

AlexGuteniev:
> However, since we're gathering so many different implementations, I think it would be an improvement to create more grouping by introducing separate namespaces for each family of support functions, within the unnamed namespace (which continues to make it clear that these are never used by anything outside).

Sounds good! The only issue I see is increased indentation. Will we accept it or resist it somehow?

StephanTLavavej:
I wouldn't worry about it.

AlexGuteniev:
Ok, it is actionable now. 

It would be a good idea to do when no other PRs changing `vector_algorithms.cpp`. 
But possibly can combine with other minor cleanups, like missed `const` or inconsistently present `_CSTD`.

AlexGuteniev:
`__std_bitset_from_string` /  `__std_bitset_to_string` captured in that screenshot too looking like a good example of what should be done.

I would rename them though to `_Bitset_from_string` / `_Bitset_to_string` to make them shorter, and add other namespaces like this.

AlexGuteniev:
[...]

PR #5429, opened 2025-04-24
`vector_algorithms.cpp`: cleanup

AlexGuteniev:
Fix accumulated inconsistencies in `vector_algorithms.cpp`, separate commit each:
 * More `const`, lately I've been trying to add `const` everywhere, but there is missing one in older code
 * Use the same type for a variable as in the other similar branch. Unsigned type is selected, as we do bitwise thing with it
 * Drop `_CSTD`, the C run time things used unqualified in this file in some other places
 * `_Avx2_tail_mask_32` now takes bytes count. It is odd from the perspective of the function itself, but almost every call site has bytes, and even for the ones that don't, the lack of multiplication by 4 and multiplication by 2 instead of by 8 look odd.

Deliberately skipped namespace as per #5418 for now. These look more controversial, so I'd do just clean stuff first.

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -1948,7 +1948,7 @@ namespace {
                             _Cur_max_val = _H_max_val;
                             const auto _Eq_mask =
                                 _Traits::_Cmp_eq(_H_max, _Cur_vals_max); // Mask of all elems eq to max
-                            int _Mask = _Traits::_Mask(_Traits::_Mask_cast(_Eq_mask));
+                            unsigned long _Mask = _Traits::_Mask(_Traits::_Mask_cast(_Eq_mask));
 
                             unsigned long _H_pos;
                             if constexpr (_Mode == _Mode_both) {
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -4160,7 +4160,7 @@ namespace {
             constexpr size_t _Length_el = 32 / sizeof(_Ty);
 
             const __m256i _Last2val = _mm256_maskload_epi32(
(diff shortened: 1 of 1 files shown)


AlexGuteniev:
Added another commit with `noexcept` lambdas. 
This is still consistency. `noexcept` isn't useful there.

stl/src/vector_algorithms.cpp
```diff
@@ -42,11 +42,12 @@ namespace {
         }
     };
 
-    __m256i _Avx2_tail_mask_32(const size_t _Count_in_dwords) noexcept {
+    __m256i _Avx2_tail_mask_32(const size_t _Count_in_bytes) noexcept {
         // _Count_in_dwords must be within [0, 8].
```
StephanTLavavej, line 46: This comment is bogus now.

stl/src/vector_algorithms.cpp
```diff
@@ -4125,7 +4160,7 @@ namespace {
             constexpr size_t _Length_el = 32 / sizeof(_Ty);
 
             const __m256i _Last2val = _mm256_maskload_epi32(
-                reinterpret_cast<const int*>(_Stop2), _Avx2_tail_mask_32(_Last2_length_el * (sizeof(_Ty) / 4)));
+                reinterpret_cast<const int*>(_Stop2), _Avx2_tail_mask_32(_Last2_length_el * (sizeof(_Ty))));
```
StephanTLavavej, line 4163: Extra parens now.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:broom: :rocket: :smile_cat:
===

PR #5431, opened 2025-04-24
`<locale>`: Improve `std::collate::do_transform()`'s handling of wrongly encoded input

### LINKED ISSUE #5210: `<locale>`: `std::collate<_Elem>::do_transform()` should behave appropriately when `_LStrxfrm()` fails
This is mainly an annoyance I noticed while preparing #5209, but there are two related problems here:

### `std::collate<char>`
When `_Strxfrm()` fails, it returns `-1` (`SIZE_MAX`) as an error code.`std::collate<char>::do_transform()` passes this return value to `basic_string<char>::resize()`, which throws a `length_error("string too long")`.

This is misleading, since the problem isn't that the result can't be represented in memory, but that there is no result at all because no sort key could be generated by `_Strxfrm()`. We should check whether the return value of `_Strxfrm()` indicates an error and then handle this appropriately (either by throwing an appropriate exception or returning a substitute key).

### `std::collate<wchar_t>`

When `_Wcsxfrm()` fails, it returns `INT_MAX` as an error code. `std::collate<char>::do_transform()` passes this return value to `basic_string<char>::resize()`, which is likely to succeed on x64. If so, it calls `_Wcsxfrm()` again, which returns `INT_MAX`, but because the string size equals `INT_MAX`, this is considered successful and the contents of the string are returned. However, the contents of the returned string are not guaranteed, so could be garbage.

### Additional remarks

While `_Strxfrm()` returns `-1` (`SIZE_MAX`) on error, the comment above the function actually claims that it returns `INT_MAX` in this case: https://github.com/microsoft/STL/blob/90820002693fe6eaaec2e55884472c654186207e/stl/src/xstrxfrm.cpp#L53

I think the implementations of `_Strxfrm()` and `_Wcsxfrm()` should be aligned. They essentially do the same thing and pass off the actual work to `LCMapStringA/W`, but their internal logic is different for no discernible reason.

muellerj2:
Resolves #5210.

Changes:

* This fixes the comments in xstrxform.cpp and xwcsxfrm.cpp: The functions return `SIZE_MAX` on error, not `INT_MAX` like their UCRT counterparts.
* In `_Wcsxfrm()`, the return value is now correctly set to `static_cast<size_t>(-1)` for one error case, bringing it in line with the other error case. Previously, it was set to `INT_MAX`, which is potentially a valid length for a sort key.
  - While this is clearly a bug in the code, I couldn't actually find any way to exert this error path. (I was a bit disappointed to find that my fix can't actually be observed to fix anything, but on the other hand, this means that we don't have to worry about any potential compatibility impact.)
* I changed `collate::do_transform()` to return an empty string in the error case instead of throwing a confusing `length_error("string too large")`.
  - The standard remains completely silent on what `collate::do_transform()` should do in case of error.
  -  The empty string can only really be a valid return value for the empty string, so there isn't any potential for confusion whether the return value represents an actual sort key or an error.
  - Alternatively, we could consciously choose that the function throws an exception. I would prefer this exception to be more appropriate than `length_error`. Maybe `runtime_error`?
    * This would mean that the regex implementation would have to defend against this exception, because `regex_match` and `regex_search` really shouldn't throw just because the searched input string isn't valid UTF-8.

### DIFF
--- a/tests/std/tests/GH_005236_collate_facet/test.cpp
+++ b/tests/std/tests/GH_005236_collate_facet/test.cpp
@@ -23,7 +23,7 @@ using namespace std;
 void test_gh_5210() {
 #ifndef SKIP_COLLATE_TRANSFORM_TESTS
     {
-        locale utf8_locale("en_US.utf8");
+        locale utf8_locale("en-US.UTF-8");
         const auto& coll = use_facet<collate<char>>(utf8_locale);
 
         const string test = "this i\xA0s a very brok\x80n utf-8\xC8string";
@@ -31,7 +31,7 @@ void test_gh_5210() {
     }
 
     {
-        locale en_us_locale("en_US");
+        locale en_us_locale("en-US");
         const auto& coll = use_facet<collate<char>>(en_us_locale);
 
         {
@@ -49,7 +49,7 @@ void test_gh_5210() {
     }
 
     {
-        locale en_us_locale("en_US");
+        locale en_us_locale("en-US");
         const auto& coll = use_facet<collate<wchar_t>>(en_us_locale);
 
         {
@@ -67,7 +67,7 @@ void test_gh_5210() {
     }
 
     {
-        locale de_DE_phone_locale("de_DE_phoneb");
+        locale de_DE_phone_locale("de-DE_phoneb");
         const auto& coll = use_facet<collate<wchar_t>>(de_DE_phone_locale);
 
         {


tests/std/tests/GH_005236_collate_facet/test.cpp
```diff
@@ -19,6 +19,83 @@
 
 using namespace std;
 
+// GH-5210 "std::collate<_Elem>::do_transform() should behave appropriately when _LStrxfrm() fails"
+void test_gh_5210() {
+#ifndef SKIP_COLLATE_TRANSFORM_TESTS
+    {
+        locale utf8_locale("en_US.utf8");
```
StephanTLavavej, line 26: We conventionally use locale names with dashes ("en-US.UTF-8" is used elsewhere). It used to make a difference on old OSes, which didn't understand underscores. I don't *think* we're using those systems in our internal harness anymore, but consistency seems worthwhile to keep. Occurs below (e.g. this test has pre-existing "de-DE_phoneb" but this PR is adding "de_DE_phoneb").

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: An empty string for error seems fine.

StephanTLavavej:

StephanTLavavej:
Thanks for noticing and fixing these long-standing mistakes! :eagle: :eye: :heart_eyes_cat:

Issue #4858, opened 2024-07-24
Support for Win7 / Server 2008 R2 is DOOMED


StephanTLavavej:
In VS 18.0 Preview 1 (whenever that happens, and whatever its official branding is), we will permanently drop support for targeting Windows 7 and Server 2008 R2, as originally explained in #4742. This time, we've got ultraboss approval (internal "experience" VSO-2164004).

# :gear: Code Changes
* [ ] Revert #4857.
* [ ] Revert MSVC-PR-566956.
  + This will re-apply the commit "NON-GitHub: Link `src/vctools/PDB` binaries against `synchronization.lib`."
* [ ] Block the VCRedist executable from being installed on Win7.
  + Thanks to Curtis Smith for explaining that we need to update `src/SetupPackages/wix/Runtime/Bundle/vcredist.wxs` to increase the `bal:Condition` from `VersionNT >= v6.1` to `VersionNT >= v6.2`, and update comments/messages from "Windows 7" to "Windows 8".
  + While Win8.0 client is no longer supported by Microsoft, Server 2012 classic is receiving (paid) extended security updates until Oct 2026, so this is close enough. (Analysis from #4751.)

# :bookmark_tabs: VS/MSVC Documentation
* [ ] Document this in the Dev18 release notes.
  + Suggested wording: "Visual C++ no longer supports targeting Windows 7 and Windows Server 2008 R2, which have stopped receiving security updates."
* [ ] Document this on the https://aka.ms/vcredist page.
* [ ] Write a [C++ Team Blog](https://devblogs.microsoft.com/cppblog/) post.

# :scroll: STL Changelog
- Improved performance:
  * Improved the performance of `system_clock::now()` and `atomic::wait()` by directly calling APIs that were added in Windows 8.
    + The STL no longer supports targeting Windows 7 and Server 2008 R2.

riverar:
The current redistributable URLs contain the VC version in the URL (`https://aka.ms/vs/17/release/vc_redist.x64.exe`). Will these continue to provide a frozen-in-time bundle for affected targets? Or will that unexpectedly start offering `/vs/18/` content? Perhaps the creation of a `/vs/latest` short-link is in order?

StephanTLavavej:
[...]

PR #5432, opened 2025-04-24
Win8 baseline: Drop support for Win7 / Server 2008 R2

StephanTLavavej:
# :world_map: Overview
* This is a clean revert of #4857.
* I will need to add several MSVC-internal changes when mirroring.
* Works towards #4858.
* This is for VS 18.0 Preview 1 and will not be backported.
  + For the rest of its lifecycle, VS 2022 17.14 will support Win7.

# :scroll: Changelog
  * Improved the performance of `system_clock::now()` and `atomic::wait()` by directly calling APIs that were added in Windows 8.


StephanTLavavej:

zacklj89 (approved):

PR #5433, opened 2025-04-24
Win8 baseline: Improve STL Hardening codegen size

StephanTLavavej:
# :world_map: Overview
* Will be merged after #5432.
* Followup to #5274.
* For MSVC, this uses [`__fastfail`](https://learn.microsoft.com/en-us/cpp/intrinsics/fastfail?view=msvc-170):
  + "Support for the native fast fail mechanism began in Windows 8."
  + I am manually declaring the intrinsic because it's currently declared by the large `<intrin.h>`. We generally try to avoid duplicating intrinsic declarations (due to calling convention madness) but in this one case I think it's fine. Because `<yvals.h>` is very central and special, moving the declaration into `<intrin0.h>` might still be overly impactful.
  + I also need to hardcode `FAST_FAIL_INVALID_ARG` because there's no way we can drag in `<winnt.h>`.
* For Clang, this uses [`__builtin_verbose_trap`](https://clang.llvm.org/docs/LanguageExtensions.html#builtin-verbose-trap):
  + This is the same mechanism used by [libc++ Hardening](https://libcxx.llvm.org/Hardening.html#hardening-assertion-failure).
  + It's cool in that it emits a single instruction, but when debugging info is enabled, it records the reason without bloating the codegen itself. Conveniently, we're already passing messages down to this layer.
* With STL Hardening enabled, I measured `v[idx]` on x64 with `/O2`:
  + This reduced codegen size from 56 bytes to 29 bytes by avoiding the `_invoke_watson()` function call.
  + Clang codegen was similarly improved.
* In `<ranges>`, we need to avoid a Clang error: "argument to `__builtin_verbose_trap` must be a pointer to a constant string".
* We need to work around VSO-2457624 "`/clr /std:c++20` silent bad codegen with the `__fastfail` intrinsic".

# :scroll: Changelog
  * Improved the performance of STL Hardening by using the MSVC [`__fastfail`](https://learn.microsoft.com/en-us/cpp/intrinsics/fastfail?view=msvc-170) intrinsic that was added in Windows 8, and the Clang [`__builtin_verbose_trap`](https://clang.llvm.org/docs/LanguageExtensions.html#builtin-verbose-trap) intrinsic.

### DIFF
--- a/stl/inc/yvals.h
+++ b/stl/inc/yvals.h
@@ -242,12 +242,16 @@ _EMIT_STL_ERROR(STL1008, "_STL_CALL_ABORT_INSTEAD_OF_INVALID_PARAMETER has been 
 //     a non-void function, etc.), but it will not attempt to replace undefined behavior with implementation-defined
 //     behavior. (For example, we will not transform `pop_back()` of an empty `vector` to be a no-op.)
 #ifndef _MSVC_STL_DOOM_FUNCTION
-#ifdef _MSVC_STL_USE_ABORT_AS_DOOM_FUNCTION
+#ifdef _MSVC_STL_USE_ABORT_AS_DOOM_FUNCTION // The user wants to use abort():
 #define _MSVC_STL_DOOM_FUNCTION(mesg) _CSTD abort()
-#else // ^^^ defined(_MSVC_STL_USE_ABORT_AS_DOOM_FUNCTION) / !defined(_MSVC_STL_USE_ABORT_AS_DOOM_FUNCTION) vvv
-// TRANSITION, GH-4858: after dropping Win7 support, we can directly call __fastfail(FAST_FAIL_INVALID_ARG).
+#elif defined(__clang__) // Use the Clang intrinsic:
+#define _MSVC_STL_DOOM_FUNCTION(mesg) __builtin_verbose_trap("MSVC STL error", mesg)
+#elif defined(_M_CEE_PURE) // Use the classic function because /clr:pure lacks the MSVC __fastfail intrinsic:
 #define _MSVC_STL_DOOM_FUNCTION(mesg) ::_invoke_watson(nullptr, nullptr, nullptr, 0, 0)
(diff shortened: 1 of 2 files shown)


StephanTLavavej:

stl/inc/yvals.h
```diff
@@ -242,12 +242,16 @@ _EMIT_STL_ERROR(STL1008, "_STL_CALL_ABORT_INSTEAD_OF_INVALID_PARAMETER has been
 //     a non-void function, etc.), but it will not attempt to replace undefined behavior with implementation-defined
 //     behavior. (For example, we will not transform `pop_back()` of an empty `vector` to be a no-op.)
 #ifndef _MSVC_STL_DOOM_FUNCTION
-#ifdef _MSVC_STL_USE_ABORT_AS_DOOM_FUNCTION
+#ifdef _MSVC_STL_USE_ABORT_AS_DOOM_FUNCTION // The user wants to use abort():
 #define _MSVC_STL_DOOM_FUNCTION(mesg) _CSTD abort()
-#else // ^^^ defined(_MSVC_STL_USE_ABORT_AS_DOOM_FUNCTION) / !defined(_MSVC_STL_USE_ABORT_AS_DOOM_FUNCTION) vvv
-// TRANSITION, GH-4858: after dropping Win7 support, we can directly call __fastfail(FAST_FAIL_INVALID_ARG).
+#elif defined(__clang__) // Use the Clang intrinsic:
```
AlexGuteniev, line 247: Wanted to ask for `#elifdef` but then recalled about the variety of targets, so perhaps no `#elifdef` in headers for now

StephanTLavavej: Yeah, we have to barely tolerate non-compiler tools like rc.exe, Qt MoC, etc. I noticed your use of elifdef in `stl/src` and didn't complain because that's MSVC-only, but in headers I don't want to try.

zacklj89 (approved):

PR #5434, opened 2025-04-25
Win8 baseline: Simplify `<filesystem>` API calls

StephanTLavavej:
# :world_map: Overview
* Will be merged after #5432.
* These changes are somewhat daring, hence the isolated PR. Our `<filesystem>` implementation (and older `<experimental/filesystem>`) contains several codepaths of the form "if we're a UWP app, call a modern API, otherwise call a legacy API". This is not only twice as much code to maintain, but we don't have automated test coverage for UWP apps, which greatly increases risk. Now that our baseline OS is Win8, we can send everything through the modern APIs.
* The change I'm the least certain about is in Standard `filesystem.cpp`, where `_Get_file_id_by_handle()` has the unusual pattern of calling modern `GetFileInformationByHandleEx()` and then falling back to legacy `GetFileInformationByHandle()`. See the commit notes below for why I suspect this is unnecessary. This code was moved around by #2718 but the original logic dates back to the initial GitHub commit.
* By front-loading this change at the beginning of the Dev18 Preview 1 cycle, we'll have the most time available to gather and react to feedback (including from our Real World Code test suite). If issues are encountered, we can selectively or entirely revert these changes, and we can add comments explaining why the codepaths are necessary.

# :gear: Commits
* Filesystem: Unconditionally call `CopyFile2` instead of `CopyFileW`.
  + [`CopyFile2`][] is "Windows 8 \[desktop apps | UWP apps\]".
  + [`CopyFileW`][] is "Windows XP \[desktop apps | UWP apps\]".
  + We were inconsistently guarding this with `defined(_ONECORE)` versus `defined(_CRT_APP)`.
* Filesystem: Unconditionally call `CreateFile2` instead of `CreateFileW`.
  + [`CreateFile2`][] is "Windows 8 \[desktop apps | UWP apps\]".
  + [`CreateFileW`][] is "Windows XP \[desktop apps only\]".
* Filesystem: Unconditionally call `GetFileInformationByHandleEx` instead of `GetFileInformationByHandle`.
  + [`GetFileInformationByHandleEx`][] is "Windows Vista \[desktop apps | UWP apps\]".
  + [`GetFileInformationByHandle`][] is "Windows XP \[desktop apps only\]".
  + The documentation around `FileIdInfo`/[`FILE_ID_INFO`][] is confusing. It claims "Minimum supported client: None supported" and "Minimum supported server: Windows Server 2012 \[desktop apps only\]", but in modern `filesystem.cpp` it is our only codepath when `defined(_CRT_APP)`. That suggests that it works on Client and works for UWP apps, so I think it might actually be "Windows 8 \[desktop apps | UWP apps\]". This PR certainly appears to work on my client Windows 11 machine.

[`CopyFile2`]: https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-copyfile2
[`CopyFileW`]: https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-copyfilew
[`CreateFile2`]: https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfile2
[`CreateFileW`]: https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew
[`GetFileInformationByHandleEx`]: https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getfileinformationbyhandleex
[`GetFileInformationByHandle`]: https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfileinformationbyhandle
[`FILE_ID_INFO`]: https://learn.microsoft.com/en-us/windows/win32/api/winbase/ns-winbase-file_id_info

# :scroll: Changelog
  * Simplified the `<filesystem>` implementation by unconditionally calling APIs that were added in Windows 8.

### DIFF
--- a/stl/src/filesys.cpp
+++ b/stl/src/filesys.cpp
@@ -223,21 +223,12 @@ _FS_DLL uintmax_t __CLRCALL_PURE_OR_CDECL _Hard_links(const wchar_t* _Fname) noe
         return static_cast<uintmax_t>(-1);
     }
 
-#ifdef _CRT_APP
     FILE_STANDARD_INFO _Info = {0};
 
     // get file info
     const auto _Ok = GetFileInformationByHandleEx(_Handle, FileStandardInfo, &_Info, sizeof(_Info));
     CloseHandle(_Handle);
     return _Ok ? _Info.NumberOfLinks : static_cast<uintmax_t>(-1);
-#else // ^^^ defined(_CRT_APP) / !defined(_CRT_APP) vvv
-    BY_HANDLE_FILE_INFORMATION _Info = {0};
-
-    // get file info
-    const auto _Ok = GetFileInformationByHandle(_Handle, &_Info);
-    CloseHandle(_Handle);
-    return _Ok ? _Info.nNumberOfLinks : static_cast<uintmax_t>(-1);
-#endif // ^^^ !defined(_CRT_APP) ^^^
 }
 
 _FS_DLL uintmax_t __CLRCALL_PURE_OR_CDECL _File_size(const wchar_t* _Fname) noexcept { // get file size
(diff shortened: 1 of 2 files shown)


StephanTLavavej:

zacklj89 (approved):
Love the 🟥

PR #5438, opened 2025-04-25
`<regex>`: Make `wregex` properly match the combination of `\w` and another named character class

### LINKED ISSUE #5243: `<regex>`: `wregex` with regular expression `[\w\s]` fails to match some spaces
The regular expression `[\w\s]` fails to match whitespace characters with code points > 255.

## Test case
```cpp
#include <iostream>
#include <regex>

using namespace std;

int main() {
    const wregex re1(LR"([\s])");
    const wregex re2(LR"([\w\s])");
    cout << R"(U+0020 SPACE is matched by "[\s]": )" <<  regex_match(L" ", re1) << '\n';
    cout << R"(U+0020 SPACE is matched by "[\w\s]": )" <<  regex_match(L" ", re2) << '\n';
    cout << R"(U+2028 LINE SEPARATOR is matched by "[\s]": )" <<  regex_match(L"\u2028", re1) << '\n';
    cout << R"(U+2028 LINE SEPARATOR is matched by "[\w\s]": )" <<  regex_match(L"\u2028", re2) << '\n';
}
```
https://godbolt.org/z/oEdTs3Th4

This prints:
```
U+0020 SPACE is matched by "[\s]": 1
U+0020 SPACE is matched by "[\w\s]": 1
U+2028 LINE SEPARATOR is matched by "[\s]": 1
U+2028 LINE SEPARATOR is matched by "[\w\s]": 0
```

## Expected result
This should print:
```
U+0020 SPACE is matched by "[\s]": 1
U+0020 SPACE is matched by "[\w\s]": 1
U+2028 LINE SEPARATOR is matched by "[\s]": 1
U+2028 LINE SEPARATOR is matched by "[\w\s]": 1
```

## Additional remarks
The underlying cause is #5242. But while I consider fixing #5242 ABI-breaking, I think this issue can be fixed without breaking ABI.

muellerj2:
Fixes #5243 by repurposing another unused flag bit when the combined character class of "w" and another named class has to be represented.

After this PR, a different value is assigned to the `_Classes` field in the NFA node when we run into the buggy scenario, but the code takes care that the same value is assigned in all other cases. This means that the old matcher may only match different characters in the buggy scenario for code points >= 0x100: It will now match the members of the other named character classes rather than the members of the character class "w".

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -1202,16 +1202,17 @@ _INLINE_VAR constexpr unsigned int _Bmp_size  = (_Bmp_max + _Bmp_chrs - 1U) / _B
 _INLINE_VAR constexpr unsigned int _ARRAY_THRESHOLD = 4U;
 
 enum _Node_flags : int { // flags for nfa nodes with special properties
-    _Fl_none            = 0x000,
-    _Fl_negate          = 0x001,
-    _Fl_greedy          = 0x002,
-    _Fl_longest         = 0x008, // TRANSITION, ABI: 0x004 is unused; the parser previously marked some nodes with it
-    _Fl_class_negated_w = 0x100,
-    _Fl_class_negated_s = 0x200,
-    _Fl_class_negated_d = 0x400,
-    _Fl_begin_needs_w   = 0x100,
-    _Fl_begin_needs_s   = 0x200,
-    _Fl_begin_needs_d   = 0x400
+    _Fl_none              = 0x000,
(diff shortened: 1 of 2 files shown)


StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for figuring out how to fix this bug while preserving ABI! :heart_eyes_cat: :green_heart: :cat2:

PR #5439, opened 2025-04-25
Test coverage for `search_n` with forward iterators

### LINKED ISSUE #933: <algorithm>: randomized test coverage for std::search_n and ranges::search_n

AlexGuteniev:
Resolves #933

The iterator was copied and adapted from 'P0202R3_constexpr_algorithm_and_exchange'.

Not doing fancy stuff from example test in #933: no pretty print failure and no command-line seed. These should be done consistently for all vector algorithms, if we want that. They would be convenient to have, but not strictly necessary.


tests/std/tests/VSO_0000000_vector_algorithms_search_n/test.cpp
```diff
@@ -113,6 +159,10 @@ void test_vector_algorithms(mt19937_64& gen) {
     test_search_n<unsigned int>(gen);
     test_search_n<long long>(gen);
     test_search_n<unsigned long long>(gen);
+
+    // Test only one case with forward iteators.  It is different complex enough code path, hence worth testing,
```
StephanTLavavej, line 163: Typo.

tests/std/tests/VSO_0000000_vector_algorithms_search_n/test.cpp
```diff
@@ -12,8 +12,49 @@
...
+        return *ptr;
+    }
+    constexpr pointer operator->() const {
+        return ptr;
+    }
+    constexpr forward_iter_adaptor& operator++() {
+        ++ptr;
+        return *this;
+    }
+    forward_iter_adaptor operator++(int) {
```
StephanTLavavej, line 43: Although it isn't being exercised, `constexpr` is missing here. (This may be a relic of the old `constexpr` rule about single statements.)

StephanTLavavej (approved):

tests/std/tests/VSO_0000000_vector_algorithms_search_n/test.cpp
AlexGuteniev: This was deleted in P0202R3_constexpr_algorithm_and_exchange test it was copied from.
The iterator with deleted post-increment was not accepted in ranges algorithm, this made me implement it.

StephanTLavavej:

StephanTLavavej:
Thanks for exercising more codepaths! :weight_lifting_man: :heart_eyes_cat: :rocket:

PR #5440, opened 2025-04-26
Test coverage for `search_n` with impossible value

AlexGuteniev:
Thanks for adding some impossible coverage to vector algorithms, @StephanTLavavej! (#5355)

But you seem to miss the one you've then-recently reviewed 😹 (#5352)

I had to damage the sequence in the test array a bit, because for `search_n` coverage to properly cover the right branch, we need to search for more than one consecutive element, otherwise we get special _n=1_ case.


StephanTLavavej:
Thanks for correcting my oversight! :heart_eyes_cat:

StephanTLavavej:

StephanTLavavej:
:heart_eyes_cat: :detective: :mag_right:
===

PR #5443, opened 2025-04-26
Benchmark: use `not_highly_aligned_allocator` in more places

### LINKED ISSUE #5035: Provide an intentional (mis)alignment that corresponds to typical usage in benchmarks for plain arrays
> > Results on my 5950X:
> 
> A big difference in "After" columnbetween `std_swap<9000, uint8_t>` and `std_swap_ranges<uint8_t>/9000` made me curious. It is big for me too.
> 
> Looks like the main reason is vector over-alignment. adding `alignas(64)` to stack array makes significant improvement there.
> 
> I don't think the benchmark should be modified to have that though.

If we know that alignment has a significant effect on the results, shouldn't we control it instead of letting it vary and skew the benchmark? Note that "control" doesn't mean "pick an unrealistic value that skews the results" but either pick a value that is representative or average over a set of values that are representative.

_Originally posted by @CaseyCarter in https://github.com/microsoft/STL/issues/4991#issuecomment-2433020551_

AlexGuteniev:
Resolves #5035

To avoid trying both aligned and unaligned allocators, try just unaligned. This makes sure we're checking the worst case, where vectorization would be of less  benefit.

Only change the container for potentially-vectorized algorithm benchmark. For, like, random it does not make sense. Also `vector<bool>` is vectorized, if we consider GPR-based vectorization as still vectorization, but this will not be sensitive to the alignment.

Some vector algorithms that are potentially sensitive to the alignment are in fact sensitive to alignment, some are not or almost not. The ones that are sensitive are simplest searches, or data movement. Consider `adjacent_find` as a good example of the sensitive one, here's how the results became worse:

Benchmark                            |  Before    |  After  
-------------------------------------|------------|------------
bm<AlgType::Std, int8_t>/2525/1142   |   19.9 ns  |   22.1 ns  
bm<AlgType::Std, int16_t>/2525/1142  |   33.2 ns  |   51.5 ns  
bm<AlgType::Std, int32_t>/2525/1142  |   75.5 ns  |   89.1 ns  
bm<AlgType::Std, int64_t>/2525/1142  |    139 ns  |    163 ns  
bm<AlgType::Rng, int8_t>/2525/1142   |   16.7 ns  |   20.1 ns  
bm<AlgType::Rng, int16_t>/2525/1142  |   33.0 ns  |   50.5 ns  
bm<AlgType::Rng, int32_t>/2525/1142  |   76.9 ns  |   89.0 ns  
bm<AlgType::Rng, int64_t>/2525/1142  |    141 ns  |    163 ns  

`adjacent_find` does two AVX loads of the same input data at a time, with aligned allocator just one of the loads is unaligned, with unaligned allocator both of them are unaligned. So it stresses the processor ability to deal with unaligned data.

Skipped also `bitset` benchmarks. They are harder to unalign, since they use some stack containers with deduced types. I'm confident that `bitset` from/to string conversion are examples of algorithm that are not very sensitive to the alignment.

Some of recently added benchmark are not changed in this PR, because they already use not highly aligned allocator.

Also I expected `replace_copy` as one of the sensitive, but auto-vectorization is broken at all in latest Preview 🐛.
Created DevCom-10895463

**Drive-by**: proper optimization barriers in replace family benchmarks.

### DIFF
--- a/benchmarks/src/adjacent_difference.cpp
+++ b/benchmarks/src/adjacent_difference.cpp
@@ -11,6 +11,8 @@
 #include <type_traits>
 #include <vector>
 
+#include "skewed_allocator.hpp"
+
 using namespace std;
 
 template <class T>
@@ -19,8 +21,8 @@ void bm(benchmark::State& state) {
 
     const size_t size = static_cast<size_t>(state.range(0));
 
-    vector<T> input(size);
-    vector<T> output(size);
+    vector<T, not_highly_aligned_allocator<T>> input(size);
+    vector<T, not_highly_aligned_allocator<T>> output(size);
 
     if constexpr (is_floating_point_v<T>) {
         normal_distribution<T> dis(0, 100000.0);
--- a/benchmarks/src/adjacent_find.cpp
+++ b/benchmarks/src/adjacent_find.cpp
@@ -8,6 +8,8 @@
 #include <cstdlib>
 #include <vector>
 
+#include "skewed_allocator.hpp"
+
 using namespace std;
 
 enum class AlgType { Std, Rng };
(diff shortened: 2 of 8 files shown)


StephanTLavavej (approved):

StephanTLavavej:
Thanks for improving the consistency of our benchmarks! :chart_with_upwards_trend: :chart_with_downwards_trend: :bar_chart:

PR #5450, opened 2025-04-29
`vector_algorithms.cpp`: introduce namespaces

### LINKED ISSUE #5418: `vector_algorithms.cpp`: split into multiple translation units or otherwise enhance file navigation?

AlexGuteniev:
Resolves #5418

# Why namespaces

The goals were:

* Being able to locate functions using IDE navigation
* Group related stuff to make it clearer what uses what

Both goals are achieved with partial success:

* IDE navigation is way better, but still clobbered with now-empty unnamed namespaces. DevCom-10889095 is about that.
* Grouping works, but some algorithms still use `_Find_trats_` or `find` itself, now it happens across namespaces

# Changes

* Move functions so that the related are grouped and unrelated are ungrouped; each group contains its trait structs, implementation functions, and export functions
* Separate traits from `_Find_traits_` where makes sense
* Rename existing namespaces and functions to match usual _Naming_style.
* Add new namespaces
* Rename some identifiers to shorter, as they now, in particular `_Predicate`
* Move comments, change comment, introduce variables to resist silly wrapping
* Remove some `using namespace` to make it more clear what is used from `_Bitmap_details`
* Exclude some traits from `_ARM64_EC` and replace by `void` or completely remove

Changes are incremental to enable per-commit review. 
(It may be complex to review the whole change, as something is both moved and modified)

# Algorithms and namespaces

Namespaces:
 * **_Bitset_from_string**, **_Bitset_to_string**: contain corresponding `bitset` member functions implementations. These were fine, just renamed.
 * **_Sorting**: `minmax_element`, `minmax`, `is_sorted_until`. These have obvious commonalities, and even referred by the Standard with this name.
 * **_Removing**: `remove`, `unique`, `remove_copy`, `unique_copy`: They do the same shuffle and share tables for it.
 * **_Mismatch**: only `mismatch` itself. Originally was among find-like algorithm due to sharing traits. Decided to move out as it does not resemble find, and re-implmenet similar traits separately, which only need some functions from find traits. The use of traits can be completely avoided for this algorithm, but that's another story.
 * **_Count**: only `count` itself. As originally implemented, it was seen as find-like algorithm, however later it was re-implemented with a dedicated approach. The traits are its own, but still inherit from `_Find_traits_`, as all functions from them are needed.
 * **_Find_meow_of**: `find_first_of` and `basic_string` members that match against a set of characters. Pre-existing namespace, but reorganized a bit, in particular, some `using namespace` removed.
 * **_Find_seq**: `search`, `find_end`, and `basic_string` members members that match subsequence. These are distinct enough from `_Find` in that they use SSE4.2 like `_Find_meow_of`. They still use `_Find_traits_`, but only for fallbacks. Although if there would ever be 32 and 64 bit implementation, it will likely use `_Find_traits_` fully.
 * **_Find**: `find`, `find_last`, `adjacent_find`, `search_n`, and `basic_string` members members that match one character, including single-chararcter `find_meow_not`. They have also much of commonality. `search_n` stands out a bit, but it still uses the same traits, and generally matches a string against a single character, so that is probably fine to keep it here.
 * **_Reversing**: `reverse`, `reverse_copy`. The namespace only contains common fallback

The algorithms that don't have traits or common functions don't need namespaces:
 * `swap_ranges`
 * `replace`

Though I'm sure it is now better with more structure, I'm not sure in particular namespaces usage.

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -5325,103 +5325,154 @@ const void* __stdcall __std_find_end_2(
 } // extern "C"
 
 namespace {
-    template <class _Traits, class _Ty>
-    __declspec(noalias) size_t __stdcall __std_mismatch_impl(
-        const void* const _First1, const void* const _First2, const size_t _Count) noexcept {
-        size_t _Result = 0;
+    namespace _Mismatch {
+#ifdef _M_ARM64EC
+        using _Traits_1 = void;
+        using _Traits_2 = void;
+        using _Traits_4 = void;
+        using _Traits_8 = void;
+#else // ^^^ defined(_M_ARM64EC) / !defined(_M_ARM64EC) vvv
+        struct _Traits_1 {
+            static __m256i _Cmp_avx(const __m256i _Lhs, const __m256i _Rhs) noexcept {
+                return _mm256_cmpeq_epi8(_Lhs, _Rhs);
+            }
+
+            static __m128i _Cmp_sse(const __m128i _Lhs, const __m128i _Rhs) noexcept {
+                return _mm_cmpeq_epi8(_Lhs, _Rhs);
+            }
+        };
+
+        struct _Traits_2 {
+            static __m256i _Cmp_avx(const __m256i _Lhs, const __m256i _Rhs) noexcept {
+                return _mm256_cmpeq_epi16(_Lhs, _Rhs);
+            }
+
+            static __m128i _Cmp_sse(const __m128i _Lhs, const __m128i _Rhs) noexcept {
+                return _mm_cmpeq_epi16(_Lhs, _Rhs);
+            }
+        };
+
+        struct _Traits_4 {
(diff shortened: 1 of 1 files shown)


StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for this major cleanup! :broom: :sparkles: :magic_wand:

PR #5454, opened 2025-05-01
`<flat_map>`: move invariants preserving

AlexGuteniev:
There are some situations to care about:
 * The comparator should always be copied, not ever moved, since moved out comparator may not satisfy invariants as it may not be able to compare at all (consider moved-out `std::function`)
 * The keys and values size should always be equal, even for moved-out container when exception is thrown. The easiest way to accomplish that is to clear both in a scope guard
 * For assignment, need to care about the target object as well

The PR addresses these situations as follows:
 * Turn `_Flat_map_base(_Derived&& _Other)` to actual move constructor
 * Add copy and move assignment operators
 * In `noexcept()` of move constructor and move-with-allocator constructor ask for both noexcept-copy (as we need it) and noexcept-move (as per spec, there is a test for that)
 * Rely on `extract()` to keep container size invariant (pre-existing)
 * Move `extract()` target object, the result is rvalue enough
 * For move-with-allocator constructor add a special implementation of `extract()`
 * Always copy comparator in mentioned constructors
 * Always copy assign comparator in move-assignment
 * Use guards in assignments to preserve target invariants on exception

The following were not addressed for now:
 * Add own coverage for that. Specifically, we apparently have no coverage for assignment target invariants
 * Look if anything is wrong with `<flat_set>` in this regard

### DIFF
--- a/stl/inc/flat_map
+++ b/stl/inc/flat_map
@@ -442,17 +442,13 @@ public:
         : _Flat_map_base(_Tag, _Ilist, key_compare(), _Alloc) {}
 
     // Copy constructors
-    _Flat_map_base(const _Flat_map_base& _Other)
-        : _Key_compare(static_cast<const _Flat_map_base&>(_Other)._Key_compare),
-          _Data(static_cast<const _Flat_map_base&>(_Other)._Data) {}
+    _Flat_map_base(const _Flat_map_base& _Other) : _Key_compare(_Other._Key_compare), _Data(_Other._Data) {}
 
     template <_Usable_allocator_for<key_container_type, mapped_container_type> _Allocator>
-    _Flat_map_base(const _Derived& _Other, const _Allocator& _Alloc)
-        : _Key_compare(static_cast<const _Flat_map_base&>(_Other)._Key_compare),
-          _Data{.keys = _STD make_obj_using_allocator<key_container_type>(
-                    _Alloc, static_cast<const _Flat_map_base&>(_Other)._Data.keys),
-              .values = _STD make_obj_using_allocator<mapped_container_type>(
-                  _Alloc, static_cast<const _Flat_map_base&>(_Other)._Data.values)} {}
+    _Flat_map_base(const _Flat_map_base& _Other, const _Allocator& _Alloc)
+        : _Key_compare(_Other._Key_compare),
+          _Data{.keys = _STD make_obj_using_allocator<key_container_type>(_Alloc, _Other._Data.keys),
+              .values = _STD make_obj_using_allocator<mapped_container_type>(_Alloc, _Other._Data.values)} {}
 
     // Move constructors
     _Flat_map_base(_Flat_map_base&& _Other)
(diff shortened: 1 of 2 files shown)


stl/inc/flat_map
```diff
@@ -442,7 +442,7 @@ public:
         : _Flat_map_base(_Tag, _Ilist, key_compare(), _Alloc) {}
 
     // Copy constructors
-    _Flat_map_base(const _Derived& _Other)
+    _Flat_map_base(const _Flat_map_base& _Other)
         : _Key_compare(static_cast<const _Flat_map_base&>(_Other)._Key_compare),
           _Data(static_cast<const _Flat_map_base&>(_Other)._Data) {}
```
StephanTLavavej, line 447: The `static_cast`s here are no longer necessary.

stl/inc/flat_map
```diff
@@ -442,7 +442,7 @@ public:
...
```
StephanTLavavej, line 448: Below, I think we should change `_Flat_map_base(const _Derived& _Other, const _Allocator& _Alloc)` for consistency. Ditto for move-alloc.

StephanTLavavej (approved):

StephanTLavavej (approved):

StephanTLavavej:
Resolved a trivial adjacent-remove conflict with #5455 in `tests/libcxx/expected_results.txt`. :heart::heart:

stl/inc/flat_map
```diff
@@ -443,34 +443,48 @@ public:
...
                  && is_nothrow_move_constructible_v<mapped_container_type>)
-        : _Key_compare(_STD move(static_cast<_Flat_map_base&>(_Other)._Key_compare)),
-          _Data(_STD move(static_cast<_Flat_map_base&>(_Other).extract())) {}
+        : _Key_compare(_Other._Key_compare), _Data(_STD move(_Other).extract()) {}
 
     template <_Usable_allocator_for<key_container_type, mapped_container_type> _Allocator>
-    _Flat_map_base(_Derived&& _Other, const _Allocator& _Alloc)
-        noexcept(is_nothrow_move_constructible_v<key_compare> && is_nothrow_move_constructible_v<key_container_type>
+    _Flat_map_base(_Flat_map_base&& _Other, const _Allocator& _Alloc)
+        noexcept(is_nothrow_copy_constructible_v<key_compare> && is_nothrow_move_constructible_v<key_compare>
```
AlexGuteniev, line 463: This exception specification is wrong, ditto for `_Extract_using_allocator`.
Allocator may throw (typically `bad_alloc`).

StephanTLavavej: Good catch, let's fix it in a followup!

StephanTLavavej:
Thanks for clearing out lots of the known failures! :heart_eyes_cat: :white_check_mark: :100:

stl/inc/flat_map
AlexGuteniev: #5460

Issue #174, opened 2019-10-15
<regex>: Should regex_match() use the match_results allocator?


StephanTLavavej:
Reported by https://stackoverflow.com/questions/58399158/use-custom-allocator-for-stdregex-match .

Calling `regex_match()` or `regex_search()` with a `match_results` containing a custom allocator ends up constructing ordinary `vector` containers in `_Matcher`, specifically:

https://github.com/microsoft/STL/blob/6aa3b02fde6f885581279ff28d7a6b4f3e38fcb7/stl/inc/regex#L1650-L1652

The Standard is silent on this topic, but libc++ uses the `match_results` allocator here.

It may be possible to fix this while preserving ABI, by renaming `_Matcher` (which is a large but internal helper class).

Note: `<regex>` mostly ignores the existence of allocators; see LWG-1396 "`regex` should support allocators" and LWG-2216 "`regex_replace(basic_string)` allocator handling" for examples.

DJLink:
Hi,

Ok so apologies if this probably has nothing to do with this, but I'm posting because I feel it could somehow be related.

I'm working on a project that has global new /free operator overloads. Recently regex_match() was added, and now this causes a crash on exit. After a bit of searching, and from what I could understand, regex_match internally we will now use our allocator via global new etc. However if seems that facets template somehow only deletes some things much later one? Now it tries to delete but custom allocator is already gone and crash.

So the question really is, would allowing a custom allocator to match_results somehow work for this, in this case actually forcing a std::malloc use so it can safely free whenever it chooses to? or any other way to do this. Again sorry if this is not related.

Thanks.

muellerj2:
I have a draft fix for #5365, but the layout of `_Loop_vals_t` must be changed for it to be reasonably efficient. This means I will have to rename `_Matcher` (and `_Loop_vals_t` obviously) to preserve ABI. Should I include this issue's proposed change at this opportunity as well?

StephanTLavavej:
@muellerj2 I am somewhat uncomfortable from a risk perspective with the idea of chaining together two significant changes into a single mega-fix, especially when the Standard doesn't speak clearly about this one. Could you make your change pave the way for this potential change, without actually combining them? I'm thinking you could do something like template your `_Matcher2` (or whatever name) on an allocator type, but always use it with `std::allocator` in your PR. That would then make it possible for a later change to use it with a different allocator type, which would be ABI-safe because different specializations don't conflict.

PR #5456, opened 2025-05-01
`<regex>`: Correctly reset capture groups for ECMAScript grammar

### LINKED ISSUE #5365: `<regex>`: Implementation divergence for capture group behavior
```
C:\Temp>type meow.cpp
```
```cpp
#include <cstddef>
#include <exception>
#include <print>
#include <string>

#ifdef USE_BOOST
#include <boost/regex.hpp>
namespace std_or_boost = boost;
constexpr auto library = "boost";
#else
#include <regex>
namespace std_or_boost = std;
constexpr auto library = "std";
#endif

using std::exception, std::print, std::println, std::size_t, std::string;

int main() {
    try {
        println("Library: {}", library);
        const string s{"acbd"};
        const std_or_boost::regex r{"(?:(a)|(b)|(c)|(d))+"};
        std_or_boost::smatch m;
        const bool result{std_or_boost::regex_search(s, m, r)};
        println("regex_search: {}", result);
        for (size_t i = 0; i < m.size(); ++i) {
            print("m[{}]", i);
            if (m[i].matched) {
                println(R"(.str(): "{}")", m[i].str());
            } else {
                println(".matched: false");
            }
        }
        println("Exception: {}", e.what());
    }
}
```

VS 2022 17.14 Preview 2 prints:

```
meow.cpp
Library: std
regex_search: true
m[0].str(): "acbd"
m[1].str(): "a"
m[2].str(): "b"
m[3].matched: false
m[4].str(): "d"
```

microsoft/STL `main` prints the exact same thing (as of https://github.com/microsoft/STL/commit/f2a2933dd65d9e8d3fa698a97b6074f7ef00e1fd with @muellerj2's amazing #5218 merged), so we haven't regressed or improved. #5218 *did* fix several other long-standing bugs in our internal database, so I was surprised to see that this one remained.

And we have implementation divergence! See: https://godbolt.org/z/cjz8PWaf7

libstdc++ 14.2 and Boost 1.87.0 agree, differing only in their Library output:

```
Library: std
regex_search: true
m[0].str(): "acbd"
m[1].str(): "a"
m[2].str(): "b"
m[3].str(): "c"
m[4].str(): "d"
```

```
Library: boost
regex_search: true
m[0].str(): "acbd"
m[1].str(): "a"
m[2].str(): "b"
m[3].str(): "c"
m[4].str(): "d"
```

But libc++ 20.1 says:

```
Library: std
regex_search: true
m[0].str(): "acbd"
m[1].matched: false
m[2].matched: false
m[3].matched: false
m[4].str(): "d"
```

Originally reported as VSO-110491 / AB#110491 (in 2014 or earlier via the now-defunct Microsoft Connect). The original user expected libstdc++/Boost's behavior.

muellerj2:
Fixes #5365 and does some preparatory work for #174.

ECMAScript requires that all capture groups inside a loop are reset to "unmatched" at the start of each repetition. To do this correctly, we have to know the first capture group that appears inside the loop. We actually do not have to know the last one to reset the capture groups correctly: If a loop is repeated, then all capture groups between the first capture group inside the loop and the last one in the regex appear inside this loop or are already unmatched. (Knowing the last capture group inside the loop would avoid some unnecessary work, but given that the number of capture groups is usually low in practice it's not clear that this additional work is worth it when traded against more NFA traversal at the time of matching. This would require some benchmarking.)

To find the first capture group, the matcher now traverses the NFA until it encounters a capture group or has visited all of the nodes inside a loop. For POSIX regexes, the traversal is skipped; instead the first capture group is set to the number of capture groups in the whole expression, which means that no capture groups will be reset when processing loops as required by POSIX semantics.

To avoid this NFA traversal for each repetition, the result of this traversal is now cached in the loop state by extending the type `_Loop_vals_t`. But this means the type `_Loop_vals_t` and `_Matcher` have to be renamed for binary compatibility. At this opportunity, I also make some minor cleanups.

I split this PR into several commits to make reviewing easier:

* Rename `_Matcher` to `_Matcher2` and add a template parameter for an allocator to prepare for #174 as suggested by STL. For now, I'm passing `void` as the allocator argument because it's obviously not an allocator, so we don't have to worry about ABI when allocator support is actually added.
* Fix the type of the `_Ncap` member variable: It should be `unsigned int`, not `int`.
* Remove the member function `_Get_ncap()`: It has become superfluous because the type of `_Ncap` has been corrected.
* Remove the matcher's duplicate member variable `_First` in favor of `_Begin` (to match `_End`).
* Remove the unnecessary local variable `_Cur_iter` in `_Do_rep` to minimally reduce stack consumption.
* Actual fix of #5365:
  - Extend `_Loop_vals_t` and rename it to `_Loop_vals_v2_t`.
  - Add function `_Find_first_inner_capture_group` that recursively traverses the NFA to find the first capture group. The recursive traversal is protected against stack overflow like all other recursive calls in the matcher.
  - Split original `_Do_rep` into `_Do_rep` (doing the main work for any repetition) and  `_Do_rep_first` (doing the preparation before the first repetition).
  - Update `_Do_rep` to reset capture groups.
  - Update `_Do_rep0` to reset capture groups in one for loop. (They already are reset correctly in the other for loop.)
  - Implement `_Do_rep_first` to perform the NFA traversal if necessary and update the loop state accordingly.


stl/inc/regex
```diff
@@ -3231,13 +3231,20 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Do_if(_Node_if* _Node) { // apply
...
     _Tgt_state_t<_It> _St = _Tgt_state;
 
     for (; _Ix < _Node->_Min; ++_Ix) { // do minimum number of reps
+        // GH-5365: We have to reset the capture groups from the second iteration on.
+        // We can avoid the reset for the first iteration
+        // because we know that a simple repetition was not encountered before.
+        if (_Ix > 0) {
+            _Tgt_state._Grp_valid = _St._Grp_valid;
+        }
+
```
muellerj2, line 3247: I think this part of the fix can actually not be observed yet, because it's currently impossible to have a capture group within a loop that is marked simple. But this is a side effect of something I would rather consider a bug: The generation of single-branch `_N_if` nodes, whose existence causes these loops to be marked non-simple.

Even so, we should include this change so that this remains fixed even when we change the parser to no longer emit these unnecessary nodes (or to skip them in the categorization of "simpleness" of these loops).

stl/inc/regex
```diff
@@ -3313,8 +3314,10 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Do_rep(_Node_rep* _Node, bool _Gr
             _Matched0 = _Match_pat(_Node->_End_rep->_Next); // empty, try tail
         } else { // try another required match
             _Psav->_Loop_idx  = _Init_idx + 1;
-            _Psav->_Loop_iter = _STD addressof(_Cur_iter);
-            _Matched0         = _Match_pat(_Node->_Next);
+            _Psav->_Loop_iter = _STD addressof(_St._Cur);
+            _STD fill(_Tgt_state._Grp_valid.begin() + static_cast<ptrdiff_t>(_Psav->_Group_first),
+                _Tgt_state._Grp_valid.end(), false);
```
muellerj2, line 3319: I think we could actually skip all `fill` calls in this function when `_Init_idx` is zero: If so, a repetition of this loop has not been performed before (so the capture groups cannot ever have matched before) or this is the first repetition of this loop when repeating an outer loop (so the capture groups have been reset when the outer loop was repeated).

However, I'm not sure if this makes a noticeable difference for performance or is even a win in practice, because the number of capture groups is usually quite small. This would need some measurement.

stl/inc/regex
```diff
@@ -3313,8 +3314,10 @@ bool _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Do_rep(_Node_rep* _Node, bool _Gr
...
         } else { // try another required match
             _Psav->_Loop_idx  = _Init_idx + 1;
-            _Psav->_Loop_iter = _STD addressof(_Cur_iter);
-            _Matched0         = _Match_pat(_Node->_Next);
+            _Psav->_Loop_iter = _STD addressof(_St._Cur);
+            _STD fill(_Tgt_state._Grp_valid.begin() + static_cast<ptrdiff_t>(_Psav->_Group_first),
+                _Tgt_state._Grp_valid.end(), false);
+            _Matched0 = _Match_pat(_Node->_Next);
         }
     } else if (_Longest) { // longest, try any number of repetitions
```
muellerj2, line 3322: There are no `fill` calls when `Longest` is true because this case is POSIX-only.

StephanTLavavej (approved):

StephanTLavavej:
Thanks, this is great! :heart_eyes_cat: I pushed a conflict-free merge with `main` and minor stylistic cleanups.

StephanTLavavej:

StephanTLavavej:
Thanks for resetting all of these bugs! :rewind: :lady_beetle: :joy_cat:

PR #5457, opened 2025-05-01
`<regex>`: Fix quadratic complexity in `regex_search` when regex starts with `?` quantifier or several alternatives

### LINKED ISSUE #5452: `<regex>`: Nonlinear slowdown with increasing string length
Here, this regex is searching for UUIDs optionally wrapped in braces with `\{?` and `\}?` (no attempt is made to ensure pairing). There are no UUIDs in the remaining string (2.5 million characters in the original user repro DevCom-10895059, internal VSO-2458266 / AB#2458266 ). The optional braces cause matching to nonlinearly slow down with increasing length of the non-matching string.

Repro with VS 2022 17.14 Preview 5 x64 and microsoft/STL `main` as of https://github.com/microsoft/STL/commit/8e9f4ee42faec6931a09bc6c1ef018cef712e52a.

```
C:\Temp>type meow.cpp
```
```cpp
#include <cassert>
#include <chrono>
#include <cstddef>
#include <iostream>
#include <regex>
using namespace std;
using namespace std::chrono;

int main() {
#ifdef AVOID_OUTER_BRACES
    const regex UuidRegEx{R"([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12})"};
#else
    const regex UuidRegEx{R"(\{?[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}\}?)"};
#endif

    for (size_t n = 1'000; n < 3'000'000; n *= 2) {
        const string allContent(n, 'x');
        const auto start  = steady_clock::now();
        const bool result = regex_search(allContent, UuidRegEx);
        const auto finish = steady_clock::now();
        const auto ms     = duration_cast<milliseconds>(finish - start).count();
        assert(!result);
        cout << "n=" << n << "; ms=" << ms << "; speed n/ms=";
        if (ms == 0) {
            cout << "instant\n";
        } else {
            cout << n * 1.0 / ms << "\n";
        }
    }
}
```
```
C:\Temp>cl /EHsc /nologo /W4 /std:c++latest /MT /O2 meow.cpp && meow
meow.cpp
n=1000; ms=1; speed n/ms=1000
n=2000; ms=8; speed n/ms=250
n=4000; ms=28; speed n/ms=142.857
n=8000; ms=107; speed n/ms=74.7664
n=16000; ms=415; speed n/ms=38.5542
n=32000; ms=1629; speed n/ms=19.644
n=64000; ms=6543; speed n/ms=9.78145
n=128000; ms=25819; speed n/ms=4.95759
^C
C:\Temp>cl /EHsc /nologo /W4 /std:c++latest /MT /O2 /DAVOID_OUTER_BRACES meow.cpp && meow
meow.cpp
n=1000; ms=0; speed n/ms=instant
n=2000; ms=0; speed n/ms=instant
n=4000; ms=1; speed n/ms=4000
n=8000; ms=1; speed n/ms=8000
n=16000; ms=3; speed n/ms=5333.33
n=32000; ms=7; speed n/ms=4571.43
n=64000; ms=14; speed n/ms=4571.43
n=128000; ms=30; speed n/ms=4266.67
n=256000; ms=60; speed n/ms=4266.67
n=512000; ms=122; speed n/ms=4196.72
n=1024000; ms=244; speed n/ms=4196.72
n=2048000; ms=472; speed n/ms=4338.98
```

muellerj2:
Fixes #5452. This is the minimal fix in `_Matcher::_Skip` that I mentioned in the issue: It avoids worst-case running time quadratic in the size of the searched string by giving up on `_N_if` nodes with two or more branches.

Unfortunately, we can't just give up whenever an `_N_if` is encountered because the parser currently likes to generate such nodes with a single branch. In particular, there is one such node at the beginning of each generated NFA. This means that giving up on such nodes would render `_Skip` pointless for all regular expressions.

I confirmed locally that this change is sufficient to fix the quadratic running time that #5452's test case exposes (though the quantifiers still almost double the running time.)

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -3937,7 +3937,7 @@ _BidIt _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Skip(_BidIt _First_arg, _BidIt 
                 // Thus, we only continue if this node has a single branch only.
                 _Node_if* _Node = static_cast<_Node_if*>(_Nx);
 
-                if (_Node->_Child != nullptr) {
+                if (_Node->_Child) {
                     return _First_arg;
                 }
                 break;
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -3931,23 +3931,22 @@ _BidIt _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Skip(_BidIt _First_arg, _BidIt 
             break;
 
         case _N_if:
-            { // check for soonest string match
+            {
+                // GH-5452: If this node has two or more branches,
+                // examining all these branches has quadratic worst-case complexity.
+                // Thus, we only continue if this node has a single branch only.
                 _Node_if* _Node = static_cast<_Node_if*>(_Nx);
 
-                for (; _First_arg != _Last && _Node; _Node = _Node->_Child) {
-                    _Last = _Skip(_First_arg, _Last, _Node->_Next);
+                if (_Node->_Child != nullptr) {
+                    return _First_arg;
                 }
-
-                return _Last;
+                break;
             }
 
         case _N_begin:
             break;
 
         case _N_end:
-            _Nx = nullptr;
-            break;
-
         case _N_none:
         case _N_wbound:
         case _N_dot:


stl/inc/regex
```diff
@@ -3931,23 +3931,22 @@ _BidIt _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Skip(_BidIt _First_arg, _BidIt
...
-            { // check for soonest string match
+            {
+                // GH-5452: If this node has two or more branches,
+                // examining all these branches has quadratic worst-case complexity.
+                // Thus, we only continue if this node has a single branch only.
                 _Node_if* _Node = static_cast<_Node_if*>(_Nx);
 
-                for (; _First_arg != _Last && _Node; _Node = _Node->_Child) {
-                    _Last = _Skip(_First_arg, _Last, _Node->_Next);
+                if (_Node->_Child != nullptr) {
```
AlexGuteniev, line 3940: I'm not sure if `<regex>` have different convention here too, but overall this repo prefers just `if (pointer)` check, not `if (pointer != nullptr)`.
(Conversely, we avoid checking non-`bool` integers directly)

muellerj2: > (Conversely, we avoid checking non-bool integers directly)

`<regex>` currently doesn't. (Although I did make the comparisons explicit for `char_class_type` because that one doesn't have to be an integer, but could be some user-provided bitmask type.)

stl/inc/regex
```diff
@@ -3931,23 +3931,22 @@ _BidIt _Matcher<_BidIt, _Elem, _RxTraits, _It>::_Skip(_BidIt _First_arg, _BidIt
...
+                break;
             }
 
         case _N_begin:
             break;
 
         case _N_end:
-            _Nx = nullptr;
-            break;
-
```
StephanTLavavej, line 3950: Is this a cleanup? I didn't see it mentioned in the PR description.

StephanTLavavej (approved):

StephanTLavavej:
Locally verified too. Given the very clear comment, and the difficulty of verifying this in automated testing without burning a bunch of CPU time, I'm okay with not having test coverage for this.

stl/inc/regex
muellerj2: Yeah, it is. It shortcuts this a bit. (I don't know why this was its own case in the first place.)

StephanTLavavej:

StephanTLavavej:
O(Thanks<sup>2</sup>) :joy_cat: :rocket: :tada:

PR #5460, opened 2025-05-02
`<flat_map>`: drop bogus exception specifications

AlexGuteniev:
Remove these exception specifications:
* move-with-allocator constructor throws because allocator throws
* ditto the extracted machinery `_Extract_using_allocator`
* ditto copy assignment

Bugs introduced in #5454, before that there was only one of them, and it was in a constructor that was never called.

I'm not sure if move constructor and move assignment exception specifications are right. That's vector-like thing we have in `flat_map`, not necessarily `vector`, what if it might not have `noexcept` moves.


StephanTLavavej (approved):

StephanTLavavej:
Thanks for noticing and fixing this!

PR #5471, opened 2025-05-04
Add casts to vector algorithms entry points

### LINKED ISSUE #5421: `<algorithm>`, `<xutility>`: vector algorithms may break integer class type difference iterators
I was surprised that I needed some casts in #5420 to pass #2885 tests.

Need to see if any other vector algorithms need similar casts.

AlexGuteniev:
Fixes #5421


tests/std/tests/GH_005421_vector_algorithms_integer_class_type_iterator/test.cpp
```diff
@@ -0,0 +1,320 @@
...
+    decltype(auto) operator-=(const difference_type diff) {
+        ptr -= static_cast<underlying_diff_type>(diff);
+        return *this;
+    }
+
+    decltype(auto) operator[](const difference_type diff) const {
+        return ptr[static_cast<underlying_diff_type>(diff)];
+    }
+
+    template <class T>
```
AlexGuteniev, line 95: Should I use abbreviated templates here? It is the "type is so unimportant, doesn't even deserve a name" case.

frederick-vs-ja: There is possibly a blocking issue, see DevCom-10735214. But I wonder whether we can reuse `test::redifference_iterator`.

AlexGuteniev: Oh, I found there's difference range, but didn't realize a single iterator from it can also be reused. Will try.

AlexGuteniev: Yes. This way we lose protection from accidental from pointer construction, but I think it is not that important.

tests/std/tests/GH_005421_vector_algorithms_integer_class_type_iterator/test.cpp
```diff
@@ -0,0 +1,221 @@
...
+#include <cassert>
+#include <compare>
+#include <iterator>
+
+#include "range_algorithm_support.hpp"
+
+using namespace std;
+
+template <class T>
+using picky_contiguous_iterator = test::redifference_iterator<_Signed128, T*>;
```
AlexGuteniev, line 15: In a previous version of this PR, custom iterator was implemented here. 
If there's need to make this tests into pre-C++20 modes, it can be restored as a starting point for that.

Although I do not want to implement this test for pre-C++20 modes.

frederick-vs-ja: I think pre-C++20 iterators are not allowed to use non-integer types as their difference types, so this test seems invalid for pre-C++20 modes.

AlexGuteniev: Didn't know that. But then I think with an unusual difference type (like `short`) there would be the same casts needed.

tests/std/tests/GH_005421_vector_algorithms_integer_class_type_iterator/test.cpp
StephanTLavavej, line 1: Need to update `test.lst`.

stl/inc/algorithm
```diff
@@ -720,7 +720,7 @@ _NODISCARD _CONSTEXPR20 _FwdIt adjacent_find(const _FwdIt _First, _FwdIt _Last,
                 if constexpr (is_pointer_v<decltype(_ULast)>) {
                     _ULast = _Result;
                 } else {
-                    _ULast = _UFirst + (_Result - _First_ptr);
+                    _ULast = _UFirst + static_cast<_Iter_diff_t<_FwdIt>>(_Result - _First_ptr);
```
StephanTLavavej, line 723: This is adding to an unwrapped iterator, but using the difference type of the wrapped iterator. This is inconsistent with changes below.

tests/std/tests/GH_005421_vector_algorithms_integer_class_type_iterator/test.cpp
```diff
@@ -0,0 +1,223 @@
...
+    {
+        int arr_src[arr_size];
+        picky_contiguous_iterator arr_src_begin(begin(arr_src));
+        picky_contiguous_iterator arr_src_end(end(arr_src));
+
+        int arr_dest[arr_size] = {};
+        picky_contiguous_iterator arr_dest_begin(begin(arr_dest));
+        picky_contiguous_iterator arr_dest_end(end(arr_dest));
+
+        int remove_expected[] = {200, 210, 220, 240, 270, 280, 290, 300, 310, 320, 340, 370, 380, 390};
```
StephanTLavavej, line 155: For these "sub-tests", can you add braces when possible to scope the additional variables like `remove_expected` to a smaller region? This helps defend against copy-paste errors.

StephanTLavavej, line 155: Can the expected arrays be const?

tests/std/tests/GH_005421_vector_algorithms_integer_class_type_iterator/test.cpp
```diff
@@ -0,0 +1,223 @@
...
+        assert(equal(arr_src_begin, arr_src_end, begin(reverse_expected), end(reverse_expected)));
+
+        ranges::copy(arr_begin, arr_end, arr_src_begin);
+        ranges::reverse_copy(arr_src_begin, arr_src_end, arr_dest_begin);
+        assert(ranges::equal(arr_dest_begin, arr_dest_end, begin(reverse_expected), end(reverse_expected)));
+        ranges::reverse(arr_src_begin, arr_src_end);
+        assert(ranges::equal(arr_src_begin, arr_src_end, begin(reverse_expected), end(reverse_expected)));
+
+        // Out of replace family, only replace for 32 and 64 bytes elements is manually vectorized,
+        // the _copy version is auto vectorized
```
StephanTLavavej, line 200: I recommend explicitly writing `replace_copy` here, which helps with global codebase searching.

tests/std/tests/GH_005421_vector_algorithms_integer_class_type_iterator/test.cpp
StephanTLavavej: Having the test be usual_20_matrix seems fine, and I agree that this should additionally help defend against narrow difference types in older modes without explicit coverage of that.

StephanTLavavej (changes requested):

AlexGuteniev:
I was thinking whether the runtime test is needed at all, or compile-time would have been fine.

Looks like the runtime test also checks the correctness of the unwrapped-to-non-pointer branches, which the main tests don't do, since most of them only use `vector<T>`.

tests/std/tests/GH_005421_vector_algorithms_integer_class_type_iterator/test.cpp
AlexGuteniev: > narrow difference types in older modes 

Just realized that this seems impossible. There's no `contiguous_iterator` in older modes, so it should be pointers with custom `iterator_traits`. But then it should be pointers to user defined types. Which we can't vectorize because don't know if they compare bitwise.

tests/std/tests/GH_005421_vector_algorithms_integer_class_type_iterator/test.cpp
```diff
@@ -0,0 +1,231 @@
...
+            ranges::copy(arr_begin, arr_end, arr_src_begin);
+            ranges::replace(arr_src_begin, arr_src_end, 250, 333);
+            assert(ranges::equal(arr_src_begin, arr_src_end, begin(replace_expected), end(replace_expected)));
+        }
+        {
+            const int swap_ranges_expected[] = {
+                300, 310, 320, 250, 340, 250, 250, 370, 380, 390, 200, 210, 220, 250, 240, 250, 250, 270, 280, 290};
+
+            copy(arr_begin, arr_end, arr_src_begin);
+            swap_ranges(arr_src_begin, arr_src_begin + _Signed128{10}, arr_src_begin + _Signed128{10});
```
StephanTLavavej, line 222: This occurs 4 times, so extracting a mid iterator would improve clarity.

StephanTLavavej:
Thanks! I fixed minor nitpicks and pushed a moderate simplification to the test. :heart_eyes_cat:

tests/std/tests/GH_005421_vector_algorithms_integer_class_type_iterator/test.cpp
```diff
@@ -0,0 +1,231 @@
...
+        assert(find_first_of(arr_begin, arr_end, needle_begin, needle_end) == arr_begin + _Signed128{10});
+        assert(ranges::find_first_of(arr_begin, arr_end, needle_begin, needle_end) == arr_begin + _Signed128{10});
+    }
+
+    assert(adjacent_find(arr_begin, arr_end) == arr_begin + _Signed128{5});
+    assert(ranges::adjacent_find(arr_begin, arr_end) == arr_begin + _Signed128{5});
+
+    {
+        // We're short of 32 and 64 bit elements 'search' and 'find_end' for now
+        short short_arr[arr_size];
```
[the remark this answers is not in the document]
AlexGuteniev: Now since you fixed that, the PR completely lacks any puns (the other one was in a defined here iterator, which I've replaced with `#include`  one).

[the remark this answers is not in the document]
AlexGuteniev: I'm going to remove this anyway with future `search_4_8` PR though

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for adding these important `static_cat`s! :zap: :cat: :joy_cat:

PR #5477, opened 2025-05-07
`<mdspan>`: Fix `layout_stride::mapping<E>::is_exhaustive()`

JMazurkiewicz:
Currently, our implementation of `layout_stride::mapping<E>::is_exhaustive()` gives incorrect result for some "unusual" extents. This PR attempts to fix it by adding a new way to compute an answer to the problem described in [[mdspan.layout.stride.obs]/5.2](http://eel.is/c++draft/mdspan.layout.stride.obs#5.2) and extra return paths for simple cases:

* When `E::rank() == 1`, `is_exhaustive()` returns `true` if and only if `stride(0) == 1`.
* When `E::rank() == 2`, `is_exhaustive()` returns `true` if and only if one of the strides is equal to `1` and the other stride is equal to extent corresponding to the previous stride.
* When two or more extents are equal to `0`, then `is_exhaustive()` always returns `false`.
* When there are no extents equal to `0` or `1` (these are "unusual" extents) this functions uses current method of comparing `required_span_size()` to `fwd-prod-of-extents(rank)` (extracted to separate function named `_Is_exhaustive_common_case`).
* Otherwise, `_Is_exhaustive_special_case` computes correct answer by finding a permutation of `(stride, extent)` pairs described in [[mdspan.layout.stride.obs]/5.2](http://eel.is/c++draft/mdspan.layout.stride.obs#obs-5.2). If such permutation does not exist this function returns `false`.
  * Sometimes, when one or more extents are equal to `1`, mapping might be exhaustive, but this function (because of the way it is described) should return `false`. For example:
  ```console
  extents = [4, 1]
  strides = [1, 5]
  required_span_size = 1 + (4 - 1) * 1 + (1 - 1) * 5 = 4
  fwd-prod-of-extents = 4 * 1 = 4 // equal to required_span_size, `_Is_exhaustive_common_case` would return true
  m(0, 0) = 0
  m(1, 0) = 1
  m(2, 0) = 2
  m(3, 0) = 3
  ```
  This mapping is clearly exhaustive according to [[mdspan.layout.reqmts]/16](http://eel.is/c++draft/mdspan.layout.reqmts#16), but not per [[mdspan.layout.stride.obs]/5.2](http://eel.is/c++draft/mdspan.layout.stride.obs#5.2).

`libc++` test fails now due to bogus test - mapping tested [here](https://github.com/llvm/llvm-project/blob/a061998a14a0ff16b633d6cf48c250d50c6acad2/libcxx/test/std/containers/views/mdspan/layout_stride/is_exhaustive_corner_case.pass.cpp#L40) is clearly not exhaustive (at least one of the strides should be equal to `1`). This also indicates that their implementation is incorrect.

### DIFF
--- a/tests/libcxx/expected_results.txt
+++ b/tests/libcxx/expected_results.txt
@@ -177,6 +177,9 @@ std/ranges/range.adaptors/range.join/range.join.iterator/arrow.pass.cpp FAIL
 # If any feature-test macro test is failing, this consolidated test will also fail.
 std/language.support/support.limits/support.limits.general/version.version.compile.pass.cpp FAIL
 
+# libc++ incorrectly implements `layout_stride::mapping<E>::is_exhaustive()`
+std/containers/views/mdspan/layout_stride/is_exhaustive_corner_case.pass.cpp FAIL
+
 
 # *** INTERACTIONS WITH MSVC THAT UPSTREAM LIKELY WON'T FIX ***
 # These tests set an allocator with a max_size() too small to default construct an unordered container
@@ -918,9 +921,6 @@ std/time/time.syn/formatter.month_day_last.pass.cpp FAIL
 
 # Our monotonic_buffer_resource takes "user" space for metadata, which it probably should not do.
 std/utilities/utility/mem.res/mem.res.monotonic.buffer/mem.res.monotonic.buffer.mem/allocate_with_initial_size.pass.cpp FAIL
-
-# Likely STL bug in layout_stride::mapping::is_exhaustive().
-std/containers/views/mdspan/layout_stride/is_exhaustive_corner_case.pass.cpp FAIL
 
 
 # *** NOT YET ANALYZED ***
--- a/tests/std/tests/P0009R18_mdspan_layout_stride/test.cpp
+++ b/tests/std/tests/P0009R18_mdspan_layout_stride/test.cpp
@@ -400,22 +400,85 @@ constexpr void check_required_span_size() {
 }
 
 constexpr void check_is_exhaustive() {
-    { // Check exhaustive mappings (all possibilities)
-        using E = extents<int, 2, 3, 5>;
-        assert((layout_stride::mapping<E>{E{}, array{1, 2, 6}}.is_exhaustive()));
-        assert((layout_stride::mapping<E>{E{}, array{1, 10, 2}}.is_exhaustive()));
-        assert((layout_stride::mapping<E>{E{}, array{3, 1, 6}}.is_exhaustive()));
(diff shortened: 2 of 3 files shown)


tests/libcxx/expected_results.txt
```diff
@@ -177,6 +177,9 @@ std/ranges/range.adaptors/range.join/range.join.iterator/arrow.pass.cpp FAIL
 # If any feature-test macro test is failing, this consolidated test will also fail.
 std/language.support/support.limits/support.limits.general/version.version.compile.pass.cpp FAIL
 
+# libc++ incorrectly implements `layout_stride::mapping<E>::is_exhaustive()`
+std/containers/views/mdspan/layout_stride/is_exhaustive_corner_case.pass.cpp FAIL
```
StephanTLavavej, line 181: No change requested: This is being added to the "issues reported/known upstream" section, but I didn't see an LLVM bug about it. I reported it to their Discord.

tests/std/tests/P0009R18_mdspan_layout_stride/test.cpp
```diff
@@ -400,22 +400,85 @@ constexpr void check_required_span_size() {
...
+    check(extents<int, dynamic_extent, 0, 7>{2}, array{1, 14, 2}, true);
+    check(extents<int, 2, dynamic_extent, 7>{0}, array{14, 28, 1}, false);
+    check(extents<int, 2, dynamic_extent, dynamic_extent>{0, 7}, array{1, 2, 2}, false);
+    check(dextents<int, 3>{2, 0, 7}, array{2, 28, 4}, false);
+    check(extents<int, 5, 0, 0>{}, array{3, 1, 1}, false);
+    check(extents<int, 5, dynamic_extent, 0>{0}, array{1, 5, 1}, false);
+    check(dextents<int, 3>{5, 0, 0}, array{2, 1, 10}, false);
+    check(extents<int, 0, 0, 0>{}, array{1, 1, 1}, false);
+    check(extents<int, 0, 1, 1>{}, array{1, 1, 1}, true);
+    check(std::extents<int, 0, 3>{}, std::array{6, 2}, false);
```
StephanTLavavej, line 459: Unnecessary qualification. But more significantly, this doesn't follow "rank() is greater than 2". I'll group it with the other `check(extents<int, 0, 3>{}`.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: I pushed minor changes, please meow if I messed something up.

StephanTLavavej:

StephanTLavavej:
Thanks for investigating this libcxx skip! :hammer_and_wrench: :detective: :heart_eyes_cat:

PR #5485, opened 2025-05-10
`vector_algorithms.cpp`: further cleanup

AlexGuteniev:
## search_n

Generalize the first step for every element size, making the whole approach more readable and use fewer lines of code.

## mismatch

Reduce element size variation. We can always compare bytewise, the division by element size would truncate the unneeded remainder length.

The 8 byte element instruction is a bit of odd ball 🎱. The SSE version has a bit higher CPU requirement (SSE4.2, not SSE2), it is a bit longer, and a bit less efficient. Still I don't think it is a performance change: I think the compiler would emit memory-operand version on AVX2 path, which does not differ in performance from the 1-byte equivalent.

## minmax_element

Just more tails closer to the function where they are used. Traits are used in all functions here, tails are only for `minmax_element`

## bitset from string

Move AVX2 exit intrinsic inside the implementation function to make the compiler emit a tail call. A perf change, though a very small one.

## bitset to string

Ditto. But this time it didn't help emitting tail call. Still done for consistency.


StephanTLavavej:

StephanTLavavej:
:broom: :broom: :broom:
===

PR #5487, opened 2025-05-10
`<regex>`: Cache bitmasks of negated character classes during matching

muellerj2:
Follow-up to #5403: Since the matcher class just got renamed, we can add some member variables to cache the bitmasks for negated character classes when such negated character classes appear in the regex. This way, we can avoid the class lookup whenever we encounter such character classes during matching.

I also removed the duplicated class-matching logic in `_Skip`: It was annoying to keep `_Skip` and `_Do_class` in sync, and the logic in `_Skip` is by far not as well-tested as `_Do_class`.

To fuse the duplicated logic, I changed `_Do_class` to no longer modify the matcher's current position. Instead, it now takes a position starting from which the class should match and returns the position *just after* the sequence actually matched by the class. (Note that this can be more than one character because of collating elements.) If the character class doesn't match, `_Do_class` just returns the starting position.

The current matcher position is now updated in `_Match_pat` after calling `_Do_class`, while `_Skip` just checks whether `_Do_class` returns the starting position or not.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -1657,6 +1657,17 @@ public:
           _Longest((_Re->_Flags & _Fl_longest) && !(_Mf & regex_constants::match_any)), _Traits(_Tr) {
         _Loop_vals.resize(_Re->_Loops);
         _Adl_verify_range(_Pfirst, _Plast);
+        if (_Re->_Flags & _Fl_begin_needs_w) {
+            _Char_class_w = _Lookup_char_class(static_cast<_Elem>('W'));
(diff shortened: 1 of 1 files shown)


StephanTLavavej (approved):

StephanTLavavej:
This is great, thank you! :heart_eyes_cat:

StephanTLavavej:

StephanTLavavej:
:dollar: :zero: :one: :diving_mask:
===

PR #5491, opened 2025-05-11
`stl/CMakeLists.txt`: Link against `softintrin.lib` when targeting ARM64/ARM64EC

cpplearner:
See @StephanTLavavej's comment in https://github.com/microsoft/STL/issues/5459#issuecomment-2867947300:

> I had to push another commit to fix the ARM64EC internal build. The use of `popcount()` in the STL's DLL revealed that (unlike VCRuntime) we weren't linking against `softintrin.lib`, which is necessary to be able to use the `__popcnt64` intrinsic. (Most people don't have to do anything special for this, but because the STL *is* a defaultlib, we link with `/nodefaultlib` and so we have to do everything manually.)

We need to apply the same fix to `CMakeLists.txt`.


stl/CMakeLists.txt
```diff
@@ -507,6 +507,9 @@ function(add_stl_dlls D_SUFFIX REL_OR_DBG)
 
     add_library(msvcp${D_SUFFIX} SHARED)
     target_link_libraries(msvcp${D_SUFFIX} PRIVATE msvcp${D_SUFFIX}_eha_objects msvcp${D_SUFFIX}_objects msvcp${D_SUFFIX}_init_objects "${TOOLSET_LIB}/vcruntime${D_SUFFIX}.lib" "${TOOLSET_LIB}/msvcrt${D_SUFFIX}.lib" "ucrt${D_SUFFIX}.lib")
+    if(VCLIBS_TARGET_ARCHITECTURE MATCHES "^arm64(ec)?$")
```
AlexGuteniev, line 510: Are you sure ARM64 non EC needs that?
It somehow linked and built tests before the change.

cpplearner: I'm not sure. `msvcp.settings.targets` seems to do this, but I don't understand why.

https://github.com/microsoft/STL/blob/da7307d68dbf39a805372e91a740393cbf06595c/stl/msbuild/stl_base/msvcp.settings.targets#L74

cpplearner:
fused into #5492

Issue #2310, opened 2021-10-30
Make ARM64EC built on GitHub


AlexGuteniev:
So that we see issues that caused urgent #2306 change

StephanTLavavej:
It's worth noting that such coverage is also missing internally (e.g. the ARM64EC compile-time errors in #2306 were discovered by a team outside MSVC), so adding such coverage would be very valuable.

PR #5492, opened 2025-05-11
Add arm64EC pipeline

cpplearner:
This PR attempts to build STL targeting ARM64EC as part of the CI process. Note that neither benchmarks nor tests are built.

I am completely novice to Azure Pipelines, and I hope that I've done everything correctly.

Towards #2310


azure-pipelines.yml
```diff
@@ -97,7 +97,6 @@ stages:
           hostArch: x64
           targetArch: arm64
           targetPlatform: arm64ec
-          analyzeBuild: true
```
cpplearner: Note: It turns out that `/analyze` is problematic for ARM64EC. Both `<xmmintrin.h>` (included by `<intrin.h>`) and `<winnt.h>` (included by `<Windows.h>`) provide declarations of `_mm_setcsr` when targeting ARM64EC, but only the declarations in `<winnt.h>` have SAL annotation. This leads to warning C28301 when `/analyze` is enabled and the headers are included in a certain order, and the warning is treated as an error because we use `/WX`.

<details><summary>Min repro</summary>

```
D:\test>type test-include.cpp
```
```c++
#include <intrin.h>
#include <Windows.h>
```
```
D:\test>cl /analyze /arm64EC /Zs test-include.cpp
```
```
Microsoft (R) C/C++ Optimizing Compiler Version 19.44.35128 for ARM64

test-include.cpp
C:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\um\winnt.h(3292) : warning C28301: No annotations for first declaration of '_InterlockedCompareExchange128'. See c:\program files\microsoft visual studio\2022\preview\vc\tools\msvc\14.44.35128\include\intrin0.inl.h(185).
C:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\um\winnt.h(3711) : warning C28301: No annotations for first declaration of '_mm_setcsr'. See c:\program files\microsoft visual studio\2022\preview\vc\tools\msvc\14.44.35128\include\xmmintrin.h(370).
C:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\um\winnt.h(6611) : warning C28301: No annotations for first declaration of '_mm_setcsr'. See c:\program files\microsoft visual studio\2022\preview\vc\tools\msvc\14.44.35128\include\xmmintrin.h(370). Note: There are several prototypes for this function. This warning compares the first with instance number 3.
```
</details>

stl/CMakeLists.txt
```diff
@@ -507,6 +507,9 @@ function(add_stl_dlls D_SUFFIX REL_OR_DBG)
 
     add_library(msvcp${D_SUFFIX} SHARED)
     target_link_libraries(msvcp${D_SUFFIX} PRIVATE msvcp${D_SUFFIX}_eha_objects msvcp${D_SUFFIX}_objects msvcp${D_SUFFIX}_init_objects "${TOOLSET_LIB}/vcruntime${D_SUFFIX}.lib" "${TOOLSET_LIB}/msvcrt${D_SUFFIX}.lib" "ucrt${D_SUFFIX}.lib")
+    if(VCLIBS_TARGET_ARCHITECTURE MATCHES "^arm64(ec)?$")
```
AlexGuteniev, line 510: I repeat in this PR the question if we really need `arm64`. 
This is mostly question to @StephanTLavavej.
I think we could drop non-EC from MSBuild project instead.

StephanTLavavej: I blindly copied what VCRuntime was doing. I'll see if this is necessary.

azure-pipelines.yml
StephanTLavavej: I think I can fix xmmintrin.h and intrin0.inl.h (can't touch winnt.h but it doesn't seem to be the offender here).

stl/CMakeLists.txt
StephanTLavavej: I just checked by modifying all occurrences to be ARM64EC-only. The "ARM64 and ARM64EC" pattern is necessary for VCRuntime which otherwise breaks with:

```
setjmpex.obj : error LNK2001: unresolved external symbol _mm_getcsr (EC Symbol)
D:\msvc\binaries\arm64chk\bin\arm64\vcruntime140.dll : fatal error LNK1120: 1 unresolved externals
```

I don't really understand what's going on here or if the STL's usage is different, but since there's no real cost to linking an unnecessary library, I'd prefer to follow the same incantation.

azure-pipelines.yml
StephanTLavavej: Argh, I can't silence this on the VCRuntime side.

StephanTLavavej: I will silence this with a bigger hammer.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for this significant infrastructure improvement, I'm really looking forward to not having to manually catch ARM64EC omissions! :heart_eyes_cat: :smile_cat: :smiley_cat:

PR #5495, opened 2025-05-11
Use `_Verify_ranges_do_not_overlap` in more places

### LINKED ISSUE #5472: `<algorithm>`: use `_Verify_ranges_do_not_overlap` function in more places
During #5471 I've noticed interesting function `_Verify_ranges_do_not_overlap`. This one is used in `std::swap_ranges` to check [alg.swap]/2 precondition:
> The two ranges [first1, last1) and [first2, last2) do not overlap

When IDL = 2, and this precondition is not satisfied, it traps with `"ranges should not overlap each other"`.

We can use it in more places. Specifically, `meow_copy` / `meow_copy_if`. 

Sometimes overlapping won't hurt, but for example with `remove_cpy` and `unique_copy`  (#5355) the output may be very surprising if there's an overlap, and the input range leads a bit; this would resemble similar `memcpy` UB manifestation.

Note that the function has plain error message, not smartass Standard citing. I doubt that it is worth citing the Standard, as we need to propagate different messages to a potentially-user-defined trap. We've been propagating only literals there so far, and propagating literals from all potential `_Verify_ranges_do_not_overlap` call sites would mean macroizing these call sites, which seems to be much of abomination for this purpose. Still, if it is desired, `_Verify_ranges_do_not_overlap` can be turn into `bool` returning, and put into a macro that traps with citing message.

AlexGuteniev:
Resolves #5472.

Only used the function in the places where it is a perfect match. So no `ranges::swap_ranges` with two sentinels, and no `std::copy` etc with weaker precondition.

I was very surprised to learn that `copy_if` / `remove_copy` / `remove_copy_if` / `unique_copy` require non-overlap [as follows](https://eel.is/c++draft/alg.modifying.operations#alg.unique-8.1.sentence-1):
> The ranges [first, last) and [result, result + (last - first)) do not overlap.

This implies that the destination range may fit the same element number as the source range contains. I thought only fitting the actually returned number of elements is enough.

Overall, now as I've done this, I'm not even sure if it is at all useful.


AlexGuteniev:
libc++ tests fail specifically due to this:

> This implies that the destination range may fit the same element number as the source range contains. I thought only fitting the actually returned number of elements is enough.

They have:
```c++
      std::array<S, 4> in = {{{4, 2}, {1, 3}, {3, 4}, {3, 5}}};
      std::array<S, 2> out;
      auto ret = std::ranges::copy_if(in.begin(), in.end(), out.begin(), [](int i) { return i == 3; }, &S::val);
```

https://github.com/llvm/llvm-project/blob/79210feb2993ff9a79ef11f8a7016a527d4fcf22/libcxx/test/std/algorithms/alg.modifying.operations/alg.copy/ranges.copy_if.pass.cpp#L160-L162

AlexGuteniev:
LLVM-139464 created.

Now that copycats are not enforced to have non-overlap, this looks even less useful.

AlexGuteniev:
> I was very surprised to learn that `copy_if` / `remove_copy` / `remove_copy_if` / `unique_copy` require non-overlap [as follows](https://eel.is/c++draft/alg.modifying.operations#alg.unique-8.1.sentence-1):

I think that standard shoould have a weaker precondition here. Just dest not in [first, last)

stl/inc/algorithm
```diff
@@ -4534,6 +4535,8 @@ namespace ranges {
             _STL_INTERNAL_STATIC_ASSERT(indirectly_copyable<_It, _Out>);
             _STL_INTERNAL_STATIC_ASSERT(indirect_binary_predicate<equal_to, projected<_It, _Pj>, const _Ty1*>);
 
+            _STD _Verify_ranges_do_not_overlap(_First, _Last, _Output);
```
StephanTLavavej, line 4538: This check is being added to `_Replace_copy_unchecked()`, which doesn't feel right. I think that the checks should be added outside of any unchecked helpers.

stl/inc/algorithm
```diff
@@ -4603,6 +4607,7 @@ namespace ranges {
         _STATIC_CALL_OPERATOR constexpr replace_copy_if_result<_It, _Out> operator()(
             _It _First, _Se _Last, _Out _Output, _Pr _Pred, const _Ty& _Newval, _Pj _Proj = {}) _CONST_CALL_OPERATOR {
             _STD _Adl_verify_range(_First, _Last);
+            _STD _Verify_ranges_do_not_overlap(_First, _Last, _Output);
```
StephanTLavavej, line 4610: This is checking (first, last, output) but the (range, output) overload below isn't being checked.

tests/std/tests/GH_005472_do_not_overlap/test.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+int main(int argc, char* argv[]) {
+#if _ITERATOR_DEBUG_LEVEL == 2
+    std_testing::death_test_executive exec;
+
+    exec.add_death_tests({
+        +[] { swap_ranges(arr, arr + 2, arr + 1); },
+        +[] { reverse_copy(arr, arr + 2, arr + 1); },
+        +[] { replace_copy(arr, arr + 2, arr + 1, 1, 2); },
+        +[] { replace_copy_if(arr, arr + 2, arr + 1, [](auto) { return false; }, 2); },
```
StephanTLavavej, line 20: Suggestion: Taking `(int)` would be simpler than taking `(auto)`.

StephanTLavavej (changes requested):

stl/inc/algorithm
AlexGuteniev: It is both centralized for all overloads and already unwrapped iterator.

AlexGuteniev: Though the check would unwrap to pointers on its own

stl/inc/algorithm
AlexGuteniev: +lack of coverage, need to check every overload

stl/inc/algorithm
AlexGuteniev: After looking into this, I'm asking for confirmation if `_unchecked` is really bad idea. We would need to extract iterators from the places like here and introduce variables, this would make the change more massive:
https://github.com/microsoft/STL/blob/da7307d68dbf39a805372e91a740393cbf06595c/stl/inc/algorithm#L4621-L4627

As an alternative, "unchecked' functions can be renamed to "common", like this one:
https://github.com/microsoft/STL/blob/da7307d68dbf39a805372e91a740393cbf06595c/stl/inc/algorithm#L5703-L5708

StephanTLavavej: I think I would be OK with renaming "unchecked" to "common". That would be a reminder to potential callers that some checking might be performed. We just want to avoid the situation where one function performs checks and then calls something that performs redundant checks.

(If the centralized function is private-static then that substantially mitigates my concerns too.)

tests/std/tests/GH_005472_do_not_overlap/test.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+    exec.add_death_tests({
+        +[] { swap_ranges(arr, arr + 2, arr + 1); },
+        +[] { reverse_copy(arr, arr + 2, arr + 1); },
+        +[] { replace_copy(arr, arr + 2, arr + 1, 1, 2); },
+        +[] { replace_copy_if(arr, arr + 2, arr + 1, [](int) { return false; }, 2); },
+        +[] { rotate_copy(arr, arr + 1, arr + 2, arr + 1); },
+#if _HAS_CXX20
+        +[] { ranges::reverse_copy(arr, arr + 2, arr + 1); },
+        +[] { ranges::reverse_copy(ranges::subrange(arr, arr + 2), arr + 1); },
```
StephanTLavavej, line 24: Actually need `<ranges>` for `ranges::subrange`.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:mountain: :x: :mountain_snow:
===

AlexGuteniev:
> > I was very surprised to learn that `copy_if` / `remove_copy` / `remove_copy_if` / `unique_copy` require non-overlap [as follows](https://eel.is/c++draft/alg.modifying.operations#alg.unique-8.1.sentence-1):
> 
> I think that standard shoould have a weaker precondition here. Just dest not in [first, last)

For the record, this is now LWG-4262

PR #5496, opened 2025-05-12
`stl/CMakeLists.txt`: Remove redundant `Synchronization.lib`

StephanTLavavej:
Followup to #5432, which linked against `Synchronization.lib` for all architectures.

This will be paired with a change to MSVC-internal `src/vctools/crt/crt_build.targets` (validated internally).


zacklj89 (approved):

StephanTLavavej:

PR #5508, opened 2025-05-15
`<flat_set>`: some libc++ tests look bogus

AlexGuteniev:
`MinSequenceContainer` in the test suite looks underimplemented


StephanTLavavej (approved):

StephanTLavavej:
Thanks! Standard citations with paragraph numbers should include the WP number (because paragraphs are frequently renumbered) but as this is a feature branch, we don't need to reset testing. Hopefully the tests will just be fixed upstream soon.

PR #5509, opened 2025-05-15
`<regex>`: Improve search performance for regexes with initial `+` quantifiers

muellerj2:
This is a small change that greatly speeds up searches for regexes like `a+` that start with some letter/string/character class followed by a `+` quantifier (or any other quantifier requiring at least one repetition). Because this loop must be matched at least once, we can enter the repeated subpattern and look for the first position a letter/string/character class in the subpattern can match.

While working about this, I noticed that I didn't think the implementation of `text_regex::should_search_match_capture_groups()` through well enough: I designed it to use relative coordinates for expected submatches, but this isn't so helpful when one wants to ensure that the whole match is in a particular position. So I changed the implementation to use absolute coordinates from the start of the matched string. Luckily, no test seems to have relied on the previous behavior, meaning all of them just matched the start of the input string anyway.

## Benchmark

Running on my machine:

| Benchmark                   | Before       | After         | Speedup |
| --------------------------- | ------------ | ------------- | ------- |
bm_lorem_search/"bibe"/2      |     38504 ns |      39237 ns |    0.98 |
bm_lorem_search/"bibe"/3      |     76730 ns |      76730 ns |    1.00 |
bm_lorem_search/"bibe"/4      |    153460 ns |     153460 ns |    1.00 |
bm_lorem_search/"(bibe)+"/2   |   4814680 ns |      92076 ns |   52.29 | 
bm_lorem_search/"(bibe)+"/3   |   9521484 ns |     204041 ns |   46.66 |
bm_lorem_search/"(bibe)+"/4   |  18158784 ns |     401088 ns |   45.27 |
bm_lorem_search/"(?:bibe)+"/2 |   4743304 ns |      97656 ns |   48.57 |
bm_lorem_search/"(?:bibe)+"/3 |   9521484 ns |     192540 ns |   49.45 |
bm_lorem_search/"(?:bibe)+"/4 |  19531250 ns |     384976 ns |   50.73 |


StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
I resolved a trivial adjacent-add conflict with #5494 in `VSO_0000000_regex_use`.

StephanTLavavej (approved):

StephanTLavavej:
:heavy_plus_sign: :rocket: :stopwatch:
===

PR #5514, opened 2025-05-16
`<flat_set>`: insert range invariant fix

AlexGuteniev:
On exception have to restore invariants. As usual, the only easy way is to clear it.


StephanTLavavej (approved):

StephanTLavavej:
:green_heart:
===

PR #5517, opened 2025-05-17
Build benchmarks for ARM64EC pipeline

cpplearner:
In #5492, I did not build the benchmarks because I didn't know how to get google-benchmark built for ARM64EC. Now I seem to find the solution.

To create static libraries for ARM64EC, the `/machine:arm64ec` option needs to be passed to `lib.exe`. [CMake doc](https://cmake.org/cmake/help/latest/command/add_link_options.html) mentions that the [`STATIC_LIBRARY_OPTIONS`](https://cmake.org/cmake/help/latest/prop_tgt/STATIC_LIBRARY_OPTIONS.html) target property holds the options for `lib.exe`. This means `benchmarks/CMakeLists.txt` needs to set this property for each relevant target, including those defined by google-benchmark. But it turns out that [`CMAKE_STATIC_LINKER_FLAGS`](https://cmake.org/cmake/help/latest/variable/CMAKE_STATIC_LINKER_FLAGS.html) will also be passed to `lib.exe`.

Setting `CMAKE_STATIC_LINKER_FLAGS` is simpler, and it affects every target, including those added in the future (either by us or by google-benchmark), which is desirable in this specific case. So this PR chooses to set `CMAKE_STATIC_LINKER_FLAGS`.

(Note: it seems that `CMAKE_STATIC_LINKER_FLAGS` defaults to `/machine:arm64`, so I decided to override it instead of appending `/machine:arm64ec` to it, but appending might be more future-proof.)

(Note 2: just in case you are interested: `stl/CMakeLists.txt` handles this by adding `${VCLIBS_EXPLICIT_MACHINE}` to the `STATIC_LIBRARY_OPTIONS` target property for every relevant target, where `${VCLIBS_EXPLICIT_MACHINE}` is `/machine:arm64ec` for ARM64EC and empty otherwise.)


StephanTLavavej (approved):

StephanTLavavej:
Thanks for figuring out these incantations! :mechanical_arm: :six: :four:

StephanTLavavej:
All shall love ARM64EC and despair! :elf_woman: :ring: :supervillain_woman:

PR #5518, opened 2025-05-17
`<regex>`: Avoid allocating during matching when regex defines no captures

muellerj2:
There is no need to allocate the capture group vectors in the matcher state when the only capture group is 0, because capture group 0 is just the whole match and we already know the bounds of the whole match (start is the position where `_Matcher::_Match` was called, end is the value of `_Cur` in the matcher state).

I can't come up with any additional test cases that should be added for this change and I think the existing test coverage should already catch it if I made a mistake here.

Benchmark on my machine:
| Benchmark                     | Before       | After         | Speedup |
| ----------------------------- | ------------ | ------------- | ------- |
| bm_lorem_search/"bibe"/2      |     37667 ns |      29157 ns |    1.29 |
| bm_lorem_search/"bibe"/3      |     78474 ns |      64174 ns |    1.22 |
| bm_lorem_search/"bibe"/4      |    147524 ns |     114397 ns |    1.29 |
| bm_lorem_search/"(bibe)"/2    |     67188 ns |      50000 ns |    1.34 |
| bm_lorem_search/"(bibe)"/3    |    146484 ns |     144385 ns |    1.01 |
| bm_lorem_search/"(bibe)"/4    |    311440 ns |     291561 ns |    1.07 |
| bm_lorem_search/"(bibe)+"/2   |     89979 ns |      92072 ns |    0.98 | 
| bm_lorem_search/"(bibe)+"/3   |    184168 ns |     180303 ns |    1.02 |
| bm_lorem_search/"(bibe)+"/4   |    376989 ns |     352926 ns |    1.07 |
| bm_lorem_search/"(?:bibe)+"/2 |     94164 ns |      59375 ns |    1.59 |
| bm_lorem_search/"(?:bibe)+"/3 |    191812 ns |     109863 ns |    1.75 |
| bm_lorem_search/"(?:bibe)+"/4 |    359869 ns |     240626 ns |    1.50 |

### DIFF
--- a/benchmarks/src/regex_search.cpp
+++ b/benchmarks/src/regex_search.cpp
@@ -32,6 +32,7 @@ void bm_lorem_search(benchmark::State& state, const char* pattern) {
 }
 
 BENCHMARK_CAPTURE(bm_lorem_search, "bibe", "bibe")->Arg(2)->Arg(3)->Arg(4);
+BENCHMARK_CAPTURE(bm_lorem_search, "(bibe)", "(bibe)")->Arg(2)->Arg(3)->Arg(4);
 BENCHMARK_CAPTURE(bm_lorem_search, "(bibe)+", "(bibe)+")->Arg(2)->Arg(3)->Arg(4);
 BENCHMARK_CAPTURE(bm_lorem_search, "(?:bibe)+", "(?:bibe)+")->Arg(2)->Arg(3)->Arg(4);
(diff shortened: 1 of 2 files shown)


StephanTLavavej:

StephanTLavavej:
:rocket: :mag_right: :heart_eyes_cat:
===

PR #5521, opened 2025-05-18
Build tests for ARM64EC pipeline

### LINKED ISSUE #2310: Make ARM64EC built on GitHub

cpplearner:
Closes #2310 

- 7 libc++ tests and 51 std tests include `<Windows.h>` (possibly via a centralized header, e.g. `test_death.hpp`), and produce warning C28301: `C:\Program Files (x86)\Windows Kits\10\include\10.0.22621.0\um\winnt.h(3209) : warning C28301: No annotations for first declaration of '_InterlockedCompareExchange128'. See c:\program files\microsoft visual studio\2022\preview\vc\tools\msvc\14.44.35207\include\intrin0.inl.h(185).`
  - I've suppressed the warning by adding the `/wd28301` compiler option. It might also be possible to instead add a pragma to each affected file.
- Windows SDK 10.0.22621.0 uses `#pragma intrinsic(fabsf)` in `<corecrt_math.h>` for ARM64EC, which seems to cause problems in several tests.
  - `GH_002206_unreserved_names` needs to avoid defining `intrinsic` as a macro.
  - 3 modules tests fail with `error LNK2019: unresolved external symbol fabsf referenced in function #fabsf$exit_thunk (EC Symbol)`.
    - 2 of them pass locally with Windows SDK 10.0.26100.0, but `VSO_1775715_user_defined_modules` is still failing with `tests\std\tests\VSO_1775715_user_defined_modules\user.ixx(24): error C2678: binary '==': no operator found which takes a left-hand operand of type 'const std::string' (or there is no acceptable conversion)`.
- Windows SDK 10.0.26100.0 instead uses `#pragma function(fabsf)` in `<corecrt_math.h>`, which seems to fix the abovementioned problems, but breaks Clang 19. This is LLVM-116256 and fixed in Clang 20.
   - I didn't check whether all tests pass with Clang 19 and Windows SDK 10.0.22621.0.
- `GH_001103_countl_zero_correctness` needs to update its preprocessor condition to exclude ARM64EC. I've copied the condition from `<__msvc_bit_utils.hpp>`.

### DIFF
--- a/tests/std/tests/VSO_1775715_user_defined_modules/test.cpp
+++ b/tests/std/tests/VSO_1775715_user_defined_modules/test.cpp
@@ -3,8 +3,9 @@
 
 // Note: To properly test the fix for VSO-1775715, don't include any headers here.
 
-// TRANSITION, this test fails for ARM64EC when built with Windows SDK 10.0.22621.0.
+// TRANSITION, Windows SDK 10.0.22621.0 causes this test to fail for ARM64EC with:
 // "error LNK2019: unresolved external symbol fabsf referenced in function #fabsf$exit_thunk (EC Symbol)"
+// It passes when built with Windows SDK 10.0.26100.0.
 // UNSUPPORTED: arm64ec
 
 import User;
--- a/tests/std/tests/VSO_1775715_user_defined_modules/test.cpp
+++ b/tests/std/tests/VSO_1775715_user_defined_modules/test.cpp
@@ -5,7 +5,9 @@
 
 // TRANSITION, Windows SDK 10.0.22621.0 causes this test to fail for ARM64EC with:
 // "error LNK2019: unresolved external symbol fabsf referenced in function #fabsf$exit_thunk (EC Symbol)"
-// It passes when built with Windows SDK 10.0.26100.0.
+// Windows SDK 10.0.26100.0 will avoid that error, but we'll need to investigate why user.ixx emits:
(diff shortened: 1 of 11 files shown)


StephanTLavavej (approved):

StephanTLavavej (approved):

tests/std/tests/VSO_1775715_user_defined_modules/test.cpp
```diff
@@ -3,6 +3,13 @@
 
 // Note: To properly test the fix for VSO-1775715, don't include any headers here.
 
+// TRANSITION, Windows SDK 10.0.22621.0 causes this test to fail for ARM64EC with:
+// "error LNK2019: unresolved external symbol fabsf referenced in function #fabsf$exit_thunk (EC Symbol)"
+// Windows SDK 10.0.26100.0 will avoid that error, but we'll need to investigate why user.ixx emits:
+// "error C2678: binary '==': no operator found which takes a left-hand operand of type 'const std::string'
+// (or there is no acceptable conversion)"
```
cpplearner, line 10: This error C2678 turns out to be unaffected by architecture or Windows SDK version, and affected by the execution charset: error C2678 manifests if and only if the execution charset is not self-synchronizing (e.g. cp932, cp936). With cp1252 or UTF-8, the test compiles successfully.

PR #5516, opened 2025-05-17
`<flat_set>`: more invariants fixes

AlexGuteniev:
Flat adapter rule: when in doubt, clear the underlying collection.

PR #5522, opened 2025-05-18
Revert incorrect `<flat_set>` change, and update test failure reason

AlexGuteniev:
* Reverts #5516, as supporting non-standard requirement is non-goal
    - #5514 still looks valid
 * Link these test failures to LLVM-140448
 * Also link previously reported test failures to LLVM-140287

### DIFF
--- a/stl/inc/flat_set
+++ b/stl/inc/flat_set
@@ -486,12 +486,9 @@ private:
                 return pair{_Where, false};
             }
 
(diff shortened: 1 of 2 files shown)


StephanTLavavej (approved):

StephanTLavavej:
Thanks for continuing to look into this! :cat2:

PR #5523, opened 2025-05-18
`<regex>`: Properly parse backslashes in character classes of basic regexes

### LINKED ISSUE #5379: `<regex>`: Backslashes in character classes are sometimes not matched in basic regular expressions

muellerj2:
Fixes #5379.

This renames the parser to add a new member variable describing the lexer mode: Default or inside character class. This allows the lexer to correctly process a backslash when parsing a character class/bracket expression.

I also tried to do this without renaming the parser, but this would mean we would have to pass the lexer mode (in or outside a character class) as an argument to all the functions processing escapes in any way, which is a bit of a pain. By renaming the parser, we need the least changes to the logic itself.

Since the parser is renamed, this PR is also doing a number of minor cleanups to the parser and builder (which is also renamed to do these cleanups).

The PR is split into several commits to simplify reviewing:

* Rename `_Parser` to `_Parser2`.
* Since we have renamed the parser, we can strip any version numbers from member functions.
* Clean up the parse flags, which we can do now because there is no longer any chance of mix-and-matching the parser constructor and the parser member function `_Compile`. Specifically:
  * `_L_paren_bal` is assigned its own bit.
  * The `_L_grp_esc` flag is added to the awk flags so that the workaround in `_ClassAtom` can be removed.
  * I also extended the `_Lang_flags` enum and the `_L_flags` member variable to `unsigned long long` so that we can add more flags more easily in the future. (This already adds `_L_dsh_rstr` to signify that the dash `-` cannot appear as the starting point of a character range in BREs and EREs, but doesn't perform the parser changes to support it yet.)
* Remove the unused member `_Begin` from the parser.
* Slightly reorder the parser member variables to reduce padding  a bit. (`_Char` is usually a `char` or `wchar_t`, so it [plus the single-byte `_Mode` member variable added in the last commit] can usually fit into the four bytes the compiler must add after `_Mchar`.
* Rename `_Builder` to `_Builder2`.
* Strip version numbers from member functions of the builder.
* Remove obsolete members `_Bmax` and `_Tmax` from the builder.
* Actually fix #5379 essentially by making `_Is_esc()` always return false when not in default (read: outside-bracketed-character-class) mode. Note that it matters how we change the lexer mode in `_Parser2::_Alternative()`: `_Next()` and `_Expect()` process the first token inside or outside the square brackets, so we must change the mode before calling these functions. The tests check that we didn't get this wrong.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -1838,11 +1838,11 @@ private:
     vector<bool> _Finished_grps;
     _Builder<_FwdIt, _Elem, _RxTraits> _Nfa;
     const _RxTraits& _Traits;
+    unsigned long long _L_flags;
     regex_constants::syntax_option_type _Flags;
     int _Val;
-    _Elem _Char;
     _Meta_type _Mchar;
-    unsigned long long _L_flags;
+    _Elem _Char;
 };
 
 enum _Lang_flags2 : unsigned long long { // describe language properties
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -1620,8 +1620,6 @@ private:
     _Node_base* _Current;
     regex_constants::syntax_option_type _Flags;
     const _RxTraits& _Traits;
-    const int _Bmax; // TRANSITION, ABI: preserved for binary compatibility
-    const int _Tmax; // TRANSITION, ABI: preserved for binary compatibility
 
 public:
     _Builder2& operator=(const _Builder2&) = delete;
@@ -2838,9 +2836,7 @@ _EXPORT_STD using wsregex_token_iterator = regex_token_iterator<wstring::const_i
 
 template <class _FwdIt, class _Elem, class _RxTraits>
 _Builder2<_FwdIt, _Elem, _RxTraits>::_Builder2(const _RxTraits& _Tr, regex_constants::syntax_option_type _Fx)
-    : _Root(new _Root_node), _Current(_Root), _Flags(_Fx), _Traits(_Tr),
-      _Bmax(static_cast<int>(_Fx & regex_constants::collate ? 0U : _Bmp_max)),
-      _Tmax(static_cast<int>(_Fx & regex_constants::collate ? 0U : _ARRAY_THRESHOLD)) {}
+    : _Root(new _Root_node), _Current(_Root), _Flags(_Fx), _Traits(_Tr) {}
 
 template <class _FwdIt, class _Elem, class _RxTraits>
 void _Builder2<_FwdIt, _Elem, _RxTraits>::_Setlong() { // set flag
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -1782,6 +1782,8 @@ enum _Prs_ret { // indicate class element type
     _Prs_set
 };
 
+enum class _Lex_mode : unsigned char { _Default, _Character_class };
+
 template <class _FwdIt, class _Elem, class _RxTraits>
 class _Parser2 { // parse a regular expression
 public:
@@ -1840,6 +1842,7 @@ private:
     regex_constants::syntax_option_type _Flags;
     int _Val;
     _Meta_type _Mchar;
+    _Lex_mode _Mode;
     _Elem _Char;
 };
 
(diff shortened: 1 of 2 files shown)


stl/inc/regex
```diff
@@ -1832,53 +1832,54 @@ private:
...
-    _L_brk_rstr   = 0x20000000, // ']' not special when first character in set
-    _L_mtch_long  = 0x40000000, // find longest match (ERE, BRE)
+enum _Lang_flags2 : unsigned long long { // describe language properties
+    _L_ext_rep    = 0x000000001ULL, // + and ? repetitions
+    _L_alt_pipe   = 0x000000002ULL, // uses '|' for alternation
+    _L_alt_nl     = 0x000000004ULL, // uses '\n' for alternation (grep, egrep)
+    _L_nex_grp    = 0x000000008ULL, // has non-escaped capture groups
+    _L_nex_rep    = 0x000000010ULL, // has non-escaped repeats
+    _L_nc_grp     = 0x000000020ULL, // has non-capture groups (?:xxx)
+    _L_asrt_gen   = 0x000000040ULL, // has generalized assertions (?=xxx), (?!xxx)
```
StephanTLavavej, line 1856: I recorded a personal todo saying that `_L_nc_grp` and `_L_asrt_gen`: https://github.com/microsoft/STL/blob/5feb814078916f519851c30028581b41941122d4/stl/inc/regex#L1777-L1778

are set in `_ECMA_flags` and nowhere else: https://github.com/microsoft/STL/blob/5feb814078916f519851c30028581b41941122d4/stl/inc/regex#L4700

But only `_L_nc_grp` is ever inspected to control non-capture groups, negative lookahead assertions, and positive lookahead assertions: https://github.com/microsoft/STL/blob/5feb814078916f519851c30028581b41941122d4/stl/inc/regex#L4296-L4313
 
`_L_asrt_gen` is therefore a relic.

If this analysis is still accurate, now would be a good time to purge this flag.

muellerj2: The idea was probably that `_L_nc_grp` should control the `if` in line 4300 and `L_asrt_grp` the following `else if`s. We could remove `_L_asrt_grp` or make it control what it is supposed to control, I don't think either decision has much impact on the (clarity of the) logic. But if we remove it, we should perhaps change its name such that the controlled features are better described (but I don't have a good name for it now).

There is also `_L_no_nl` that is set for grep/egrep and supposed to control that newlines in the regex are rejected, but this behavior hasn't been implemented.

StephanTLavavej: I would prefer to remove `_L_asrt_gen` because the logic is intertwined (we inspect the question mark first, followed by colon/exclamation/equal).

> we should perhaps change _L_nc_grp's name such that the controlled features are better described (but I don't have a good name for it now).

I suggest `_L_nc_asrt` with the comment `// has non-capture groups (?:xxx) and generalized assertions (?=xxx), (?!xxx)` because I am a lazy kitty.

We can rename it later without breaking bincompat.

> There is also _L_no_nl that is set for grep/egrep and supposed to control that newlines in the regex are rejected, but this behavior hasn't been implemented.

Oh, good catch, should we strike that flag too?

muellerj2: > Oh, good catch, should we strike that flag too?

I do plan to implement the behavior later. But I guess we could also fuse it with `_L_alt_nl` and name the combination something like `_L_grep_nl`.

(Background: A list of regexes separated by newlines can be passed to grep, and grep will search by trying all of them for each line. This means that newlines function like alternations, but only if the expressions between newlines complete can be interpreted as complete regexes. If the newline appears within parentheses or in a bracket expression, it's a syntax error because splitting the string at newlines will result in mismatched parentheses or brackets in the substrings. `_L_alt_nl` is supposed to mark that newlines outside parentheses/brackets represent alternations, while `_L_no_nl` describes that newlines within parentheses/brackets should result in a syntax error.)

muellerj2: If we remove these two flags, we could use `unsigned int` as the underlying type for the flags.

StephanTLavavej:
Reviewing now, I'll push changes soon. I updated the PR description from saying "`_L_paren_bal` is assigned its own bit." to say `_L_brk_bal` instead; please meow if I was somehow confused.

StephanTLavavej (approved):

StephanTLavavej:
Thanks for the infinite bugfixes in infinite combinations, as the Vulcans say! :vulcan_salute: :lady_beetle: :hammer_and_wrench:

PR #5524, opened 2025-05-18
`<regex>`: Restrict control letters in escapes to alphabetic ASCII characters

muellerj2:
Towards #995. ECMAScript 15.10.1 defines the production `ControlLetter` such that only ASCII alphabetic characters are allowed. Such restricted character sets are often extended to corresponding POSIX character classes (to include Unicode characters) by Boost.Regex or `<regex>` implementations. But such an extension doesn't make much sense in this specific case, because these `\c` + letter escapes are supposed to emulate what pressing Ctrl + ASCII character does in the terminal. (It would actually make more sense to support "\c[" and similar, but the ECMAScript standard doesn't say so.)

Alternative: If non-ASCII characters should remain supported, we could query the bitmask of the character class "alpha" from the traits class.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -4707,7 +4707,11 @@ bool _Parser<_FwdIt, _Elem, _RxTraits>::_CharacterEscape(bool _In_character_clas
         _Next();
     } else if (_Char == _Esc_ctrl && (_L_flags & _L_esc_ctrl)) { // handle control escape sequence
         _Next();
-        if (!_Traits.isctype(_Char, _RxTraits::_Ch_alpha)) {
+
+        using _Uelem = typename _RxTraits::_Uelem;
+        _Uelem _UCh  = static_cast<_Uelem>(_Char);
+        if ((static_cast<_Uelem>('a') > _UCh || static_cast<_Uelem>('z') < _UCh)
+            && (static_cast<_Uelem>('A') > _UCh || static_cast<_Uelem>('Z') < _UCh)) {
             _Error(regex_constants::error_escape);
         }
 
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -4710,8 +4710,8 @@ bool _Parser<_FwdIt, _Elem, _RxTraits>::_CharacterEscape(bool _In_character_clas
 
         using _Uelem = typename _RxTraits::_Uelem;
         _Uelem _UCh  = static_cast<_Uelem>(_Char);
-        if ((static_cast<_Uelem>('a') > _UCh || static_cast<_Uelem>('z') < _UCh)
-            && (static_cast<_Uelem>('A') > _UCh || static_cast<_Uelem>('Z') < _UCh)) {
+        if (!((static_cast<_Uelem>('a') <= _UCh && _UCh <= static_cast<_Uelem>('z'))
+                || (static_cast<_Uelem>('A') <= _UCh && _UCh <= static_cast<_Uelem>('Z')))) {
             _Error(regex_constants::error_escape);
         }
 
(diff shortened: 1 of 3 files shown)


tests/std/tests/GH_005244_regex_escape_sequences/test.cpp
```diff
@@ -25,9 +25,7 @@ class test_regex_traits {
     using char_class_type = typename rx_traits::char_class_type;
 
     // TRANSITION, GH-995
-    using _Uelem                    = typename rx_traits::_Uelem;
-    static constexpr auto _Ch_upper = rx_traits::_Ch_upper;
```
muellerj2, line 29: The usage of `_Ch_upper` was already removed in #5403.

tests/std/tests/GH_005244_regex_escape_sequences/test.cpp
```diff
@@ -192,7 +190,11 @@ void test_gh_5244_atomescape_ecmascript() {
     g_regexTester.should_not_match("c", R"(\ca)", ECMAScript);
     g_regexTester.should_not_match("ca", R"(\ca)", ECMAScript);
     g_regexTester.should_throw(R"(\c0)", error_escape, ECMAScript);
-    g_regexTester.should_throw(R"(\c)", error_escape, ECMAScript);
```
StephanTLavavej, line 195: Why is test coverage for backslash-c at the end of the pattern being removed?

stl/inc/regex
```diff
@@ -4707,7 +4707,11 @@ bool _Parser<_FwdIt, _Elem, _RxTraits>::_CharacterEscape(bool _In_character_clas
         _Next();
     } else if (_Char == _Esc_ctrl && (_L_flags & _L_esc_ctrl)) { // handle control escape sequence
         _Next();
-        if (!_Traits.isctype(_Char, _RxTraits::_Ch_alpha)) {
+
+        using _Uelem = typename _RxTraits::_Uelem;
+        _Uelem _UCh  = static_cast<_Uelem>(_Char);
+        if ((static_cast<_Uelem>('a') > _UCh || static_cast<_Uelem>('z') < _UCh)
+            && (static_cast<_Uelem>('A') > _UCh || static_cast<_Uelem>('Z') < _UCh)) {
```
StephanTLavavej, line 4714: IMO these range tests are phrased in a really confusing manner.

tests/std/tests/GH_005244_regex_escape_sequences/test.cpp
```diff
@@ -25,9 +25,7 @@ class test_regex_traits {
     using char_class_type = typename rx_traits::char_class_type;
 
     // TRANSITION, GH-995
-    using _Uelem                    = typename rx_traits::_Uelem;
-    static constexpr auto _Ch_upper = rx_traits::_Ch_upper;
-    static constexpr auto _Ch_alpha = rx_traits::_Ch_alpha;
```
AlexGuteniev, line 30: Did this completely fix #995 ?

AlexGuteniev: Ah, there's still `_Uelem`

StephanTLavavej (approved):

tests/std/tests/GH_005244_regex_escape_sequences/test.cpp
muellerj2: The problem with `_Uelem` is that it appears in lots of different places in `<regex>`, so its removal PR would lead to non-trivial merge conflicts with parallel PRs.

The other issue is that the change is surprisingly intricate and the resulting code looks weird. Example: `if (static_cast<_Uelem>(_Ch) < 256)` will have to be replaced by something like `if (static_cast<_Elem>(static_cast<unsigned char>(_Ch)) == _Ch)`.

StephanTLavavej:

StephanTLavavej:
There is no escape for regex bugs! :joy_cat: :zany_face: :rofl:

Issue #5479, opened 2025-05-08
Use Clang's `__is_trivially_equality_comparable` intrinsic to optimize more cases in vector algorithms


AlexGuteniev:
`__is_trivially_equality_comparable`  is a trait that temms if a type is `memcmp` comarable for `==` and `!=` against itself. We can use that on Clang along with existing metaprogramming.

This will allow covering more types like `struct` containing only integers with defaulted comparison and enums without user-defined comparison. This includes user defined types, for which this cannot be achieved without "magic". Also this includes standard types, such as `chrono::duration` or `pair<int, int>` if their comparison is defaulted properly; this should be less burden that specializing traits for them all.

Some cases are already added to `GH_000431_equal_memcmp_is_safe`. Like, these ones should become `true` on Clang:
https://github.com/microsoft/STL/blob/8e9f4ee42faec6931a09bc6c1ef018cef712e52a/tests/std/tests/GH_000431_equal_memcmp_is_safe/test.compile.pass.cpp#L266-L269
For others, coverage should be added. Also, some coverage should be added to the algorithms themselves, to make sure this scenario works, in particular `_Could_compare_equal_to_value_type` is properly always `true`. I don't think separate benchmark is needed though.

A separate escape hatch could be a good idea, due to quasi-supporting old Clang and "new" Intel, which may not have `__is_trivially_equality_comparable`, or may have more bugs in it.

StephanTLavavej:
Clang supports [`__has_builtin`](https://clang.llvm.org/docs/LanguageExtensions.html#has-builtin) which is an effective way to automatically light up support. I probably should have used it instead of version checks recently.

AlexGuteniev:
> Clang supports [`__has_builtin`](https://clang.llvm.org/docs/LanguageExtensions.html#has-builtin) which is an effective way to automatically light up support.

I want to enable that only starting with Clang 19, since I fear the bugs of early implementation:
* llvm/llvm-project#63192
* llvm/llvm-project#63371
* llvm/llvm-project#63656
* llvm/llvm-project#89293
* llvm/llvm-project#107777

PR #5527, opened 2025-05-19
Vectorize equality-comparing algorithms for more types

AlexGuteniev:
Towards #5479

Use `__is_trivially_equality_comparable` starting in Clang 19, with escape hatch.
Distinguish `memcmp` and search-like vector algorithms, as the latter need 1, 2, 4, or 8 element size.

If this takes on, may improve `mismatch` to vectorize for odd element size too. 
It takes division, but it can be done in the header, where it will be optimized to big multiplication.

Also can look into `_Vector_alg_in_find_is_safe`.

---

I'm not sure though:
 * We can't have enums with this actually https://godbolt.org/z/33bsEvq1b
 * For structs need defaulted `operator==`
    - Which means `pair<int, int>` or `chrono::duration<uint64_t, std::pico>` will not be automatically engaged
 * On the other hand, here came member object pointer and member function pointer, which weren't given an invitation


StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: I fixed typos and pushed minor nitpicks.

AlexGuteniev:
I've updated the title to a more specific.
There's actually just one of "equality-comparing algorithms": `std::equal` and its `ranges` version.

AlexGuteniev:
> I've updated the title to a more specific. There's actually just one of "equality-comparing algorithms": `std::equal` and its `ranges` version.

Actually, restored previous title. `search` etc will be also optimized for the right struct size.

StephanTLavavej:

StephanTLavavej:
Thanks for supporting Clang as even more of a first-class citizen than MSVC! :joy_cat: :1st_place_medal: :tada:

PR #5532, opened 2025-05-22
Clean up `vector_algorithms.cpp` again: unsigned bit manipulation, scope guard for multiple return

AlexGuteniev:
Follow up for #5519 

1. Scope guard for multiple return. This pattern is used in other places where we have multiple return.
2. Drop `_bittestandreset`. The compiler emits `btc` for the `^= 1 << n` pattern, which has the same performance as `btr` for `_bittestandreset`.
   - For `search` could also use `_Bingo &= _Bingo - 1`, or `_blsr_u32`, but there's not enough evidence in significant enough performance advantage to use this
   - For `find_end` could try `_bzhi_u32`, but didn't even bother, here the expectation is even lower.
4. Always use unsigned 32-bit for bit masks from `movemask*`. Always 32 bit because internally there is 32-bit destination register, always unsigned because bitmask.
   - Still `int` for `cmpestr*` masks, it is compile time, and only few bits, and intrinsic takes `int`
   - In some places the cast to other bit width is needed, it is kept there

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -2884,7 +2884,7 @@ namespace {
 
                 do {
                     const __m256i _Data = _mm256_loadu_si256(static_cast<const __m256i*>(_First));
-                    int _Bingo          = _mm256_movemask_epi8(_Traits::_Cmp_avx(_Data, _Comparand));
+                    unsigned int _Bingo = _mm256_movemask_epi8(_Traits::_Cmp_avx(_Data, _Comparand));
 
                     if constexpr (_Pred == _Predicate::_Not_equal) {
                         _Bingo ^= 0xFFFF'FFFF;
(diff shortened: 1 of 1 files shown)


stl/src/vector_algorithms.cpp
```diff
@@ -4478,7 +4478,7 @@ namespace {
                         }
                     }
 
-                    if (const int _Bingo = _mm256_movemask_epi8(_Eq); _Bingo != 0) {
+                    if (const uint32_t _Bingo = _mm256_movemask_epi8(_Eq); _Bingo != 0) {
```
AlexGuteniev, line 4481: `uint32_t` instead of `unsigned int` elsewhere to match the change below

stl/src/vector_algorithms.cpp
```diff
@@ -4497,7 +4497,7 @@ namespace {
                         }
                     }
 
-                    if (const int _Bingo = _mm256_movemask_epi8(_mm256_and_si256(_Eq, _Tail_mask)); _Bingo != 0) {
+                    if (const uint32_t _Bingo = _mm256_movemask_epi8(_mm256_and_si256(_Eq, _Tail_mask)); _Bingo != 0) {
```
AlexGuteniev, line 4500: `uint32_t` instead of `unsigned int` to avoid wrapping

StephanTLavavej (approved):

StephanTLavavej:

PR #4025, opened 2023-09-13
Avoid unconditional `make_heap` for `priority_queue::push_range`

achabense:
Fixes #2826.

1. It's *dangerous* to unconditionally call `make_heap` - we will get great performance penalty if we do `push_heap` in small batches.
2. To remake the entire heap, for some types (like string), we can actually get better performance if we just do a `push_heap` loop. For those types, we can unconditionally do `push_heap` loop instead. However, for some trivial types (especially scalars) we will really get better performance if we call `make_heap`.
3. Based on 1&2, to be conservative, I'm choosing `new_size/2>old_size` as the boundary to prefer `make_heap` to `push_heap` loop. This will completely avoid the problem in 1.

About  `new_size/2>old_size`: `make_heap(size 2*o)` does `o` `_Pop_heap_hole_by_index`, and each `_Pop_heap_hole_by_index` calls a `_Push_heap_by_index`, so I think it's safe to assume doing a `o`-`push_heap` loop can be faster before/around this boundary. 

The following benchmark shows a generally better result, including those at the boundary (`Arg(vec_size / 2 + 1)`; 5001). However, I'm not sure going on in `u32/u64/float`'s `Arg(vec_size / 2 + 1)` case (in which case remaking the heap is actually faster). (This does not necessarily mean we have to set a smaller ratio boundary, as it looks size/type/sequence-sensitive. See the update part below; for more extensive benchmarks see the next comment.)

Benchmark result (`vec_size==10000`):
<details>
<summary>Previous</summary>

```
--------------------------------------------------------------------------------------
Benchmark                                            Time             CPU   Iterations
--------------------------------------------------------------------------------------
BM_push_range<uint8_t, vec_u8>/1             132314020 ns    131250000 ns            5
BM_push_range<uint8_t, vec_u8>/100             1378146 ns      1380522 ns          498
BM_push_range<uint8_t, vec_u8>/10000             23991 ns        24065 ns        29867
BM_push_range<uint8_t, vec_u8>/5001              44356 ns        43493 ns        15448

BM_push_range<uint16_t, vec_u16>/1           129688300 ns    130208333 ns            6
BM_push_range<uint16_t, vec_u16>/100           1332057 ns      1339286 ns          560
BM_push_range<uint16_t, vec_u16>/10000           21871 ns        21973 ns        32000
BM_push_range<uint16_t, vec_u16>/5001            42268 ns        42375 ns        16593

BM_push_range<uint32_t, vec_u32>/1            49569970 ns     50000000 ns           10
[...]

PR #5533, opened 2025-05-23
Make benchmarks compiling in clang-cl

AlexGuteniev:
Towards #5479

Most of the time it is warning about deprecated `benchmark::DoNotOptimize` for constants.

# 🏗️ Build steps

Build STL as usual:
```
cmake --preset x64
cmake --build --preset x64
```
Then add `-DCMAKE_CXX_COMPILER=clang-cl` to benchmark build:
```
cmake -B out\bench -S benchmarks -G Ninja -DSTL_BINARY_DIR=out\x64 -DCMAKE_CXX_COMPILER=clang-cl
cmake --build out\bench
```


benchmarks/src/move_only_function.cpp
```diff
@@ -5,6 +5,10 @@
 #include <functional>
 #include <utility>
 
+#ifdef __clang__
+#pragma clang diagnostic ignored "-Wunqualified-std-cast-call"
+#endif // defined(__clang__)
+
```
frederick-vs-ja, line 11: And here.

benchmarks/src/priority_queue_push_range.cpp
```diff
@@ -59,10 +59,8 @@ void BM_push_range(benchmark::State& state) {
 
 template <size_t L>
 void putln(const benchmark::State&) {
-    static bool b = [] {
-        puts("");
-        return true;
-    }();
+    static once_flag f;
+    call_once(f, [] { puts(""); });
```
AlexGuteniev, line 63: Unused variable. Replaced to `call_once` as it is indeed unused and confusing

benchmarks/src/search.cpp
```diff
@@ -32,7 +32,8 @@ template <size_t Size, bool Last_is_different>
 constexpr std::array fill_pattern_array = make_fill_pattern_array<Size, Last_is_different>();
 
 template <size_t Size, bool Last_is_different>
-constexpr std::string_view fill_pattern_view = fill_pattern_array<Size, Last_is_different>;
+constexpr std::string_view fill_pattern_view{
+    fill_pattern_array<Size, Last_is_different>.data(), fill_pattern_array<Size, Last_is_different>.size()};
```
AlexGuteniev, line 36: Did not want to construct `string_view` from `array` directly, this was error, not warning

benchmarks/src/locale_classic.cpp
```diff
@@ -8,7 +8,8 @@ using namespace std;
 // GH-3048 <locale>: Double-checked locking for locale::classic
 void BM_locale_classic(benchmark::State& state) {
     for (auto _ : state) {
-        benchmark::DoNotOptimize(locale::classic());
+        auto v = locale::classic();
+        benchmark::DoNotOptimize(v);
```
AlexGuteniev, line 12: temporary was **const** rvalue

AlexGuteniev (commented):
@frederick-vs-ja the rest are just const

benchmarks/src/priority_queue_push_range.cpp
frederick-vs-ja: It's desired to avoid magic statics in product code. But I wonder whether they should also be avoided in benchmarks. Side-effects-only static variable seems to be a common technique.

benchmarks/src/locale_classic.cpp
frederick-vs-ja: Attempting to add the warnings in google/benchmark#1993.

AlexGuteniev: Would be very good. It would mean that we don't need to find a way to add clang-cl compilation to CI, as MSVC would spot most of the issues fixed here.

benchmarks/src/priority_queue_push_range.cpp
StephanTLavavej: We would need to include `<mutex>` for `call_once`. But actually I have no idea what this is intended to accomplish, as no other benchmark bothers with this. I didn't question it in #4025. I'll try to remove it.

benchmarks/src/bitset_to_string.cpp
```diff
@@ -30,10 +30,11 @@ void BM_bitset_to_string(benchmark::State& state) {
     static_assert(N <= 64);
 
     for (auto _ : state) {
-        for (const auto& bits : random_bits<>) {
+        // make a copy, so that it can be potentially modified by DoNotOptimize
+        for (auto bits : random_bits<>) {
             benchmark::DoNotOptimize(bits);
             bitset<N> bs{bits};
-            benchmark::DoNotOptimize(bs.to_string<charT>());
+            benchmark::DoNotOptimize(bs.template to_string<charT>());
```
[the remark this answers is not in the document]
StephanTLavavej: You forgot to make the type dependent. This demonstrates the MSVC bug: https://godbolt.org/z/ezMcrYjcM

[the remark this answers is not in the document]
AlexGuteniev: DevCom-540547

StephanTLavavej:

StephanTLavavej:
Thanks for making the benchmark directory more portable and conformant! :heart_eyes_cat: :ship: :rocket:

Issue #619, opened 2020-03-18
Should STL specifically consider /Ob1 ?


AlexGuteniev:
This is not an issue, just a question.

Should STL be optimized for /Ob1 build?

I mean if release is built with /O2 and /Ob1, a significant part of STL is inlined, since it is inline members of template class. Still template functions defined outside classes are not marked as `inline` and thus are not inlined.

(A project may be built with /O2 + /Ob1 if it does not benefit from /Ob2, and has degraded performance comparing to /Ob1. On the other hand, it is rather unusual case)

StephanTLavavej:
The STL expects extensive inlining due to its reliance on small helper functions, and therefore I believe that /Ob1 is the wrong direction - if anything, the STL wants the new /Ob3 option (not implied by /O2, but perhaps in the future). I don’t have hard data, though.

StephanTLavavej:
@BillyONeal mentioned that if you can find performance test cases where /Ob1 is superior to /Ob2 or /Ob3, we can look into changing this option.

PR #5534, opened 2025-05-23
`<bit>`: Use popcount for `has_single_bit()` if it's always available

### LINKED ISSUE #5359: `<bit>`: Could `has_single_bit()` be faster?

pps83:
provides 1.05 to 1.1 times speed up on x64 targets.

**x64 results**
```
Benchmark                               Time 
---------------------------------------------
bm_has_single_bit_if<uint8_t>        13.4 ns 
bm_has_single_bit_if_new<uint8_t>    12.7 ns 
bm_has_single_bit_if<uint16_t>       13.9 ns 
bm_has_single_bit_if_new<uint16_t>   10.8 ns 
bm_has_single_bit_if<uint32_t>       15.9 ns 
bm_has_single_bit_if_new<uint32_t>   12.6 ns 
bm_has_single_bit_if<uint64_t>       14.3 ns 
bm_has_single_bit_if_new<uint64_t>   11.0 ns 
bm_has_single_bit<uint8_t>           6.34 ns 
bm_has_single_bit_new<uint8_t>       5.88 ns 
bm_has_single_bit<uint16_t>          6.19 ns 
bm_has_single_bit_new<uint16_t>      4.52 ns 
bm_has_single_bit<uint32_t>          5.57 ns 
bm_has_single_bit_new<uint32_t>      5.99 ns 
bm_has_single_bit<uint64_t>          6.07 ns 
bm_has_single_bit_new<uint64_t>      5.41 ns 
```

fixes #5359

### DIFF
--- a/stl/inc/bit
+++ b/stl/inc/bit
@@ -84,7 +84,14 @@ _NODISCARD constexpr int countl_zero(_Ty _Val) noexcept;
 
 _EXPORT_STD template <_Standard_unsigned_integral _Ty>
 _NODISCARD constexpr bool has_single_bit(const _Ty _Val) noexcept {
+#if defined(_POPCNT_INTRINSICS_ALWAYS_AVAILABLE) && _POPCNT_INTRINSICS_ALWAYS_AVAILABLE
+    if (_STD is_constant_evaluated()) {
+        return (_Val ^ (_Val - 1)) > _Val - 1;
+    }
+    return _Unchecked_popcount(_Val) == 1;
+#else
     return (_Val ^ (_Val - 1)) > _Val - 1;
+#endif // ! _POPCNT_INTRINSICS_ALWAYS_AVAILABLE
 }
 
 inline void _Precondition_violation_in_bit_ceil() noexcept {}
--- a/stl/inc/bit
+++ b/stl/inc/bit
@@ -85,13 +85,12 @@ _NODISCARD constexpr int countl_zero(_Ty _Val) noexcept;
 _EXPORT_STD template <_Standard_unsigned_integral _Ty>
 _NODISCARD constexpr bool has_single_bit(const _Ty _Val) noexcept {
 #if _POPCNT_INTRINSICS_ALWAYS_AVAILABLE
-    if (_STD is_constant_evaluated()) {
-        return (_Val ^ (_Val - 1)) > _Val - 1;
+    if (!_STD is_constant_evaluated()) {
+        return _Unchecked_popcount(_Val) == 1;
     }
-    return _Unchecked_popcount(_Val) == 1;
-#else // ^^^ _POPCNT_INTRINSICS_ALWAYS_AVAILABLE / !_POPCNT_INTRINSICS_ALWAYS_AVAILABLE vvv
+#endif // ^^^ _POPCNT_INTRINSICS_ALWAYS_AVAILABLE ^^^
+
     return (_Val ^ (_Val - 1)) > _Val - 1;
-#endif // ^^^ !_POPCNT_INTRINSICS_ALWAYS_AVAILABLE ^^^
 }
 
 inline void _Precondition_violation_in_bit_ceil() noexcept {}
--- a/stl/inc/bit
+++ b/stl/inc/bit
@@ -89,9 +89,9 @@ _NODISCARD constexpr bool has_single_bit(const _Ty _Val) noexcept {
         return (_Val ^ (_Val - 1)) > _Val - 1;
     }
     return _Unchecked_popcount(_Val) == 1;
-#else
+#else // ^^^ popcnt intrinsic always available / popcnt intrinsic not always available vvv
     return (_Val ^ (_Val - 1)) > _Val - 1;
-#endif // ! _POPCNT_INTRINSICS_ALWAYS_AVAILABLE
+#endif // ^^^ popcnt intrinsic not always available ^^^
 }
 
 inline void _Precondition_violation_in_bit_ceil() noexcept {}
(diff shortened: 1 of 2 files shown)


pps83:
@AlexGuteniev, do I need to update benchmarks, or simply put benchmark numbers as part of the commit message?

AlexGuteniev:
> @AlexGuteniev, do I need to update benchmarks, or simply put benchmark numbers as part of the commit message?

I think the current benchmarks are fine, just post before/after number

AlexGuteniev:
Ideally, we'd need ARM64 numbers as well. But I think we can do without them.

stl/inc/bit
```diff
@@ -84,7 +84,14 @@ _NODISCARD constexpr int countl_zero(_Ty _Val) noexcept;
 
 _EXPORT_STD template <_Standard_unsigned_integral _Ty>
 _NODISCARD constexpr bool has_single_bit(const _Ty _Val) noexcept {
+#if defined(_POPCNT_INTRINSICS_ALWAYS_AVAILABLE) && _POPCNT_INTRINSICS_ALWAYS_AVAILABLE
+    if (_STD is_constant_evaluated()) {
+        return (_Val ^ (_Val - 1)) > _Val - 1;
+    }
+    return _Unchecked_popcount(_Val) == 1;
+#else
```
AlexGuteniev, line 92: We have standard `#else` comments like 
```c++
// ^^^ defined(_POPCNT_INTRINSICS_ALWAYS_AVAILABLE) && _POPCNT_INTRINSICS_ALWAYS_AVAILABLE / !defined(_POPCNT_INTRINSICS_ALWAYS_AVAILABLE) || !_POPCNT_INTRINSICS_ALWAYS_AVAILABLE vvv
```

But that looks too verbose for this case

Maybe 

```
// ^^^ popcnt intrinsic always available / popcnt intrinsic not available or not always available vvv
```
The `#endif` comment should consistently say
```c++
// ^^^ popcnt intrinsic not available or not always available ^^^
```

AlexGuteniev:
> I think the current benchmarks are fine, just post before/after number

You also need to pass somehow `/arch:AVX` or greater to make it different, perhaps by temporarily alterint `CmakeFIle.txt` locally, or by any other working method.

pps83:
> You also need to pass somehow `/arch:AVX` or greater to make it different, perhaps by temporarily alterint `CmakeFIle.txt` locally, or by any other working method.

shouldn't unit tests automatically run with/without these switches?

AlexGuteniev:
> shouldn't unit tests automatically run with/without these switches?

Unit tests should. 
I mean benchmarks, which currently don't

pps83:
I'm getting conflicting results for the test: popcount on my pc is somehow slower by around 30%
it's a tiger-lake-u core i7-11390H, shouldn't have slow popcount :)

pps83:
> I'm getting conflicting results for the test: popcount on my pc is somehow slower by around 30% it's a tiger-lake-u core i7-11390H, shouldn't have slow popcount :)

no idea how is that it doesn't get inlined :)

image

pps83:
wow, accidentially, i had /Only __inline (/Ob1) option. Apparently, this destroys most of std code that doesn't have inlines. Imo, all these functions should be marked inline.

pps83:
popcount is on the right. So, it does improve, but will be 30% slower for projects using /Ob1

AlexGuteniev:
> Imo, all these functions should be marked inline.

No; I asked a while ago, see #619

stl/inc/bit
pps83: updated

pps83:
> So you probably mean 1.05 to 1.1 times speed up

updated. I based it on my previous commit message: https://github.com/microsoft/STL/pull/5367

(actual git history doesn't have that anyways)

AlexGuteniev:
Mixed results for `_if` version, but consistent improvement of the branchless one

i5-1235U (Alder Lake)

P Cores

Benchmark                      | Before   |  After   | Speedup
-------------------------------|----------|----------|---------
bm_has_single_bit_if<uint8_t>  | 9.78 ns  |  9.90 ns |   0.99
bm_has_single_bit_if<uint16_t> | 12.1 ns  |  10.1 ns |   1.20
bm_has_single_bit_if<uint32_t> | 11.4 ns  |  12.3 ns |   0.93
bm_has_single_bit_if<uint64_t> | 9.66 ns  |  12.7 ns |   0.76
bm_has_single_bit<uint8_t>     | 4.44 ns  |  5.56 ns |   0.80
bm_has_single_bit<uint16_t>    | 4.66 ns  |  3.59 ns |   1.30
bm_has_single_bit<uint32_t>    | 4.66 ns  |  3.59 ns |   1.30
bm_has_single_bit<uint64_t>    | 4.90 ns  |  3.47 ns |   1.41

E Cores

Benchmark                      |    Time  |    Time | Speedup
-------------------------------|----------|---------|---------
bm_has_single_bit_if<uint8_t>  | 14.7 ns  | 14.0 ns |  1.05
bm_has_single_bit_if<uint16_t> | 16.6 ns  | 14.7 ns |  1.13
bm_has_single_bit_if<uint32_t> | 14.7 ns  | 14.3 ns |  1.03
bm_has_single_bit_if<uint64_t> | 14.7 ns  | 14.1 ns |  1.04
bm_has_single_bit<uint8_t>     | 9.09 ns  | 8.84 ns |  1.03
bm_has_single_bit<uint16_t>    | 9.09 ns  | 7.24 ns |  1.26
bm_has_single_bit<uint32_t>    | 9.07 ns  | 7.25 ns |  1.25
bm_has_single_bit<uint64_t>    | 9.08 ns  | 7.27 ns |  1.25

AlexGuteniev (approved):

StephanTLavavej:

PR #5535, opened 2025-05-23
`<regex>`: Add multiline option and make non-multiline mode the default

### LINKED ISSUE #73: LWG-2503 multiline option should be added to `syntax_option_type`
LWG-2503 multiline option should be added to `syntax_option_type`

Also tracked by Microsoft-internal VSO-629739 / [AB#629739](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/629739).

muellerj2:
Resolves #73 and implements LWG-2503. Arguably also resolves [DevCom-436138](https://developercommunity.visualstudio.com/t/436138) to the degree that it is reasonable (namely, when the anchor appears in the regex before it starts branching); see the benchmark. Unblocks four libcxx tests.

This PR also aligns the multiline mode with ECMAScript's specification. The anchors now match at any of ECMAScript's line terminators: carriage returns, line feeds, line separators and paragraph separators. Before, the anchors only matched at line feeds.

The PR provides `_REGEX_MAKE_MULTILINE_MODE_DEFAULT` as an escape hatch to return to default multiline mode; if so, non-multiline mode is not available.

For POSIX grammars, the new `multiline` option has no effect. While I find this unfortunate, this behavior appears to have been specified in [\[re.synopt\]](https://eel.is/c++draft/re.synopt).

To simplify the logic in the matcher and avoid some preprocessor \#ifdefs, the matcher's internal copy of the regex syntax flags `_Sflags` is mutated before matching starts:
* The `multiline` flag is set for all grammars when the escape hatch is defined.
* The `multiline` flag is cleared for POSIX grammars when the escape hatch is not defined.

These mutations ensure that multiline mode is enabled if and only if the `multiline` flag is set in `_Sflags`.

I see a potential concern with the implementation in this PR: Even if the escape hatch is set, the matcher still changes behavior and allows anchors to match not just line feeds but all ECMAScript line terminators. It can reasonably be argued that the behavior should be completely unchanged if the escape hatch is defined. Even so, I opted to submit the implementation with ECMAScript-conforming line terminators in this PR first because this simplifies the implementation a lot.

### Benchmark

Only for pattern "^bibe" to show that this resolves [DevCom-436138](https://developercommunity.visualstudio.com/t/436138).

| Benchmark                 | Time    | CPU     | Iterations |
| ------------------------- | ------- | ------- | ---------- |
| bm_lorem_search/"^bibe"/2 | 56.4 ns | 57.8 ns |   10000000 |
| bm_lorem_search/"^bibe"/3 | 55.9 ns | 54.4 ns |   11200000 |
| bm_lorem_search/"^bibe"/4 | 55.5 ns | 56.2 ns |   10000000 |


tests/std/tests/VSO_0000000_regex_interface/test.cpp
```diff
@@ -387,9 +387,9 @@ void test_VSO_180466_regex_search_missing_Unchecked_call() {
 }
 
 void test_VSO_226914_match_prev_avail() {
-    // N.B. assumes our nonstandard multiline behavior. See also: LWG-2343, LWG-2503
+    // test assumes multiline mode
```
StephanTLavavej: `bol_anchor` is constructed with `regex_constants::multiline`, so it's not an assumption anymore.

I'm tempted to drop the comment, but changing it to say "exercises" would emphasize the option below.

stl/inc/regex
```diff
@@ -4054,30 +4071,55 @@ template <class _BidIt, class _Elem, class _RxTraits, class _It, class _Alloc>
 _BidIt _Matcher2<_BidIt, _Elem, _RxTraits, _It, _Alloc>::_Skip(_BidIt _First_arg, _BidIt _Last, _Node_base* _Node_arg) {
     // skip until possible match
     // assumes --_First_arg is valid
-    _Node_base* _Nx = _Node_arg ? _Node_arg : _Rep;
+    constexpr char _Line_terminators_char[]       = {static_cast<char>(_Meta_cr), static_cast<char>(_Meta_nl)};
+    constexpr wchar_t _Line_terminators_wchar_t[] = {static_cast<wchar_t>(_Meta_cr), static_cast<wchar_t>(_Meta_nl),
+        static_cast<wchar_t>(_Meta_ls), static_cast<wchar_t>(_Meta_ps)};
```
StephanTLavavej: These should be `static constexpr` arrays, otherwise the compiler will generate code to push each value onto the stack (so distinct arrays have distinct addresses).

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -453,48 +453,95 @@ void test_VSO_208146_regex_smoke_test_rewritten_explicit_quantifier() {
...
+            beginCd.should_search_match("cdefg", "cd");
+            beginCd.should_search_fail("cdefg", match_not_bol);
+            beginCd.should_search_match(line_terminator + "cdefg", "cd");
+            beginCd.should_search_match(line_terminator + "cdefg", "cd", match_not_bol);
+
+            beginCd.should_search_fail("ab" + line_terminator + "xcdefg");
+            beginCd.should_search_fail("ab" + line_terminator + "xcdefg", match_not_bol);
+        }
+
+        for (wstring line_terminator : {L"\u2028", L"\u2029"}) { // U+2028 LINE SEPARATOR, U+2029 PARAGRAPH SEPARATOR
```
StephanTLavavej: `_Matcher2::_Skip` is using completely separate arrays for `char` and `wchar_t`, so we should test the original recipe line terminators here, in addition to the extra crispy ones.

stl/inc/regex
```diff
@@ -1666,6 +1667,15 @@ public:
         if (_Re->_Flags & _Fl_begin_needs_d) {
             _Char_class_d = _Lookup_char_class(static_cast<_Elem>('D'));
         }
+
+// sanitize multiline mode setting
+#ifdef _REGEX_MAKE_MULTILINE_MODE_DEFAULT
```
StephanTLavavej: I'm sure there are inconsistent examples I'm forgetting, but most of our modes are "always defined to 0 or 1" (e.g. `_USE_STD_VECTOR_ALGORITHMS`). Additionally, let's phrase this to make it clear that the mode is old.

StephanTLavavej (approved):

StephanTLavavej:
Thanks for implementing this formerly-thought-to-be-impossible change! :heart_eyes_cat: I pushed fairly minor changes, including renaming/reworking the escape hatch.

StephanTLavavej:

StephanTLavavej:
:heart_eyes_cat: :green_heart: :cat:
===

PR #5540, opened 2025-05-24
`vector_algorithms.cpp`: Fix zeroupper guard for `is_sorted_until` vectrorization

AlexGuteniev:
* `is_sorted_until` is early return algorithm, so it needs scope guard
 * `minmax_element` and `minmax` are not early return algorithms, so let their guards stay as direct function call


StephanTLavavej (approved):

StephanTLavavej:
:guard: :hammer_and_wrench: :smile_cat:
===

PR #5541, opened 2025-05-24
`<algorithm>`: avoid `continue` in `ranges::includes`

AlexGuteniev:
Replace `continue` with chained `if`.

No performance differences. In fact, no codegen differences expected.


StephanTLavavej:
:chains: :arrow_right: :heavy_plus_sign:
===

PR #5551, opened 2025-05-27
Update `source_location` tests for upcoming EDG changes

StephanTLavavej:
EDG is changing their `source_location` builtins to produce output that more closely resembles C1XX's. To keep our tests passing as these changes are introduced into VS, I'm updating the tests to accept either the old output or the new output. After the changes have shipped and we can update the toolset on GitHub, I can remove the old expected outputs.

@jgoshi has validated that this passes with the upcoming changes.

### DIFF
--- a/tests/std/tests/P1208R6_source_location/test.cpp
+++ b/tests/std/tests/P1208R6_source_location/test.cpp
@@ -68,11 +68,11 @@ constexpr void local_test() {
     assert(x.column() == 37);
 #endif // ^^^ C1XX ^^^
 #if _USE_DETAILED_FUNCTION_NAME_IN_SOURCE_LOCATION
-#ifdef __EDG__
-    assert(x.function_name() == "void local_test()"sv);
-#else // ^^^ EDG / Other vvv
+#ifdef __EDG__ // TRANSITION, EDG is changing to match C1XX's output
+    assert(x.function_name() == "void __cdecl local_test(void)"sv || x.function_name() == "void local_test()"sv);
+#else // ^^^ workaround / no workaround vvv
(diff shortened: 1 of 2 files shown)


StephanTLavavej:

tests/std/tests/P1208R6_source_location/test.cpp
```diff
@@ -241,7 +244,9 @@ constexpr void lambda_test() {
 #elif defined(__clang__) // ^^^ basic / detailed Clang vvv
     assert(fun2 == "auto " THISCALL_OR_CDECL " lambda_test()::(anonymous class)::operator()(void) const"sv);
 #elif defined(__EDG__) // ^^^ detailed Clang / detailed __EDG__ vvv
-    assert(fun2 == "lambda []()->auto::operator()()->auto"sv);
+    // TRANSITION, EDG is changing to resemble C1XX's output somewhat more closely
```
davidmrdavid, line 247: nit - this "TRANSITION" comment is different from the ones above.
The variants I see are:
* EDG is changing to resemble C1XX's output somewhat more closely
* EDG is changing to almost match C1XX's output
* EDG is changing to match C1XX's output 

Were these supposed to be the same (for easy CTRL+F'ing) or is their difference meaningful?

StephanTLavavej: The difference is meaningful. For lambdas, the output differs radically, but EDG is at least adding calling conventions.

For the "almost match" ones, EDG is still missing "struct" before "std::source_location" for reasons that are unclear to me.

davidmrdavid (approved):
🚢 it

PR #5553, opened 2025-05-29
`<regex>`: Correct character translation in `icase` and `collate` mode

muellerj2:
In `icase` and `collate` mode, characters are supposed to be passed through `translate_nocase()` and `translate()` before comparing them. This PR makes sure we always apply these translations exactly once before any character or string comparison.

The test deliberately defines non-idempotent translation functions. They are very weird, but they make it easy to catch any repeated or unbalanced applications of these functions before character or string comparisons.

This PR also replaces `_Cmp_cs` by `equal_to`. While this means for now that a static call is replaced by a non-static one, there is already machinery in place to recognize when `equal_to` can be vectorized, and vectorization will help us significantly improve performance in one of the following PRs.


jovibor:
@muellerj2 
The number of [`<regex>` bug fixes](https://github.com/microsoft/STL/pulls?page=1&q=is%3Apr+label%3Aregex+author%3Amuellerj2) you've introduced for the last few months is a bit shocking.
Is `<regex>` really that buggy, how did it even functionate?🤐
Anyway, good job and thanks for making the STL better for all of us.

stl/inc/regex
```diff
@@ -4084,7 +4113,8 @@ _BidIt _Matcher2<_BidIt, _Elem, _RxTraits, _It, _Alloc>::_Skip(_BidIt _First_arg
                 _Node_str<_Elem>* _Node = static_cast<_Node_str<_Elem>*>(_Nx);
                 for (; _First_arg != _Last; ++_First_arg) { // look for starting match
                     _BidIt _Next = _First_arg;
-                    if (_Compare(_First_arg, ++_Next, _Node->_Data._Str(), _Node->_Data._Str() + 1, _Traits, _Sflags)
+                    if (_Compare_translate_left(
```
StephanTLavavej: The other calls are being `_STD` qualified.

tests/std/tests/GH_005553_regex_character_translation/test.cpp
```diff
@@ -0,0 +1,253 @@
...
+class nonidempotent_translate_regex_traits : regex_traits<charT> {
+private:
+    using rx_traits = regex_traits<charT>;
+
+public:
+    using char_type       = typename rx_traits::char_type;
+    using string_type     = typename rx_traits::string_type;
+    using locale_type     = typename rx_traits::locale_type;
+    using char_class_type = typename rx_traits::char_class_type;
+    using uchar_type      = make_unsigned_t<charT>;
```
StephanTLavavej, line 27: Should include `<type_traits>`.

tests/std/tests/GH_005553_regex_character_translation/test.cpp
```diff
@@ -0,0 +1,253 @@
...
+    check_match(subject, pattern, re, flags, false);
+}
+
+template <class Rx>
+void check_search_match(const wstring& subject, const wstring& expected, const wstring& pattern, const Rx& re,
+    match_flag_type flags = match_default) {
+    wsmatch match;
+    const bool search_result = regex_search(subject, match, re, flags);
+    if (!search_result || match[0] != expected) {
+        wprintf(LR"(Expected regex_search("%s", regex("%s", 0x%X), 0x%X) to find "%s", )", subject.c_str(),
```
StephanTLavavej, line 77: Should include `<cstdio>`.

tests/std/tests/GH_005553_regex_character_translation/test.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+#include <test_regex_support.hpp>
+
+using namespace std;
+using namespace std::regex_constants;
+
+regex_fixture g_regexTester;
+
+template <class charT>
+class nonidempotent_translate_regex_traits : regex_traits<charT> {
```
StephanTLavavej: For `class`es, we conventionally never rely on the default of `private` inheritance, and always directly say so in the code.

tests/std/tests/GH_005553_regex_character_translation/test.cpp
```diff
@@ -0,0 +1,253 @@
...
+    using rx_traits::transform;
+    using rx_traits::transform_primary;
+    using rx_traits::value;
+};
+
+template <class Rx>
+void check_match(const wstring& subject, const wstring& pattern, const Rx& re, match_flag_type flags, bool matches) {
+    if (regex_match(subject, re, flags) != matches) {
+        wprintf(LR"(Expected regex_match("%s", regex("%s", 0x%X)) to be %s.)", subject.c_str(), pattern.c_str(),
+            re.flags(), matches ? L"true" : L"false");
```
StephanTLavavej: This is directly passing `re.flags()` to `wprintf()`, missing the `static_cast<unsigned int>` that appears on the following occurrences.

tests/std/tests/GH_005553_regex_character_translation/test.cpp
```diff
@@ -0,0 +1,253 @@
...
+        g_regexTester.fail_regex();
+    }
+}
+
+template <class Rx>
+void check_match(const wstring& subject, const wstring& pattern, const Rx& re, match_flag_type flags = match_default) {
+    check_match(subject, pattern, re, flags, true);
+}
```
StephanTLavavej: This could just be a second default arg.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: I pushed cosmetic changes.

StephanTLavavej:

StephanTLavavej:
Thanks for fixing this bug, obscure though it may be! :lady_beetle: :detective: :heart_eyes_cat:

PR #5558, opened 2025-05-31
`tzdb.cpp`: Fall back to numeric offset when the current code page cannot represent the time zone short ID

### LINKED ISSUE #3158: `<chrono>`: `time_zone::get_info` fails when the current code page cannot represent the localized time zone short ID.
See:

- Issue #3097
- PR #3102
- PR #3122 

It seems likely that the result we want is for, if the abbreviation fails to convert, to instead attempt to format `UTC+{}` or `UTC-{}`.

### LINKED ISSUE #5482: `<chrono>`: `zoned_time::get_info()` and `vformat()` generate `ERROR_NO_UNICODE_TRANSLATION` exceptions when the system region is set to Slovak
See the original report DevCom-10896610. (Internal: VSO-2467874 / AB#2467874 ).

```cpp
#include <chrono>
#include <exception>
#include <format>
#include <iostream>
#include <locale>
#include <string>
using namespace std;
using namespace std::chrono;

// Set C++ Language standard to be used to ISO C++20 Standard
// Install the Slovak language first: Settings -> Time & Language -> Language -> Preferred Languages -> Add a language
// Set System Region: Settings -> Time & Language -> Region -> Region Format -> Slovak (Slovakia)

int main() {
    zoned_time<system_clock::duration> zonedTime{current_zone(), system_clock::now()};

    try { // Problem #1
        auto info = zonedTime.get_info();
        cout << "Zone abbrev: " << info.abbrev << endl;
        cout << "Exception: " << e.what() << endl;
    }

    try { // Problem #2
        string zonedTimeString = vformat(locale::classic(), "{:%Y-%m-%dT%H:%M:%S}", make_format_args(zonedTime));
        cout << "Time string: " << zonedTimeString << endl;
        cout << "Exception: " << e.what() << endl;
    }
}
```

cpplearner:
This PR adds a fallback time zone abbreviation of the form  `-05` or `+0530` when the system's code page cannot represent the localized time zone short ID produced by ICU. Previously such time zone short ID would result in `ERROR_NO_UNICODE_TRANSLATION` exceptions.

The numeric offset form (`-05` or `+0530`) is chosen because it is similar to the time zone abbreviation in [`tzdb`](https://data.iana.org/time-zones/tzdb/theory.html#abbreviations), and distinguishable from ICU's output.

The behavior depends on the current user locale and the current code page. For [ICU 77.1](https://github.com/unicode-org/icu/tree/release-77-1), the following locales have time zone short ID with non-ASCII characters, and thus might be affected by this fallback: `am`, `ar`, `as`, `bg`, `blo`, `brx`, `ckb`, `cs`, `dz`, `fa`, `ff_Adlm`, `ga`, `gd`, `he`, `hi`, `ii`, `km`, `ks`, `ks_Deva`, `ku`, `mai`, `ml`, `mni`, `nqo`, `sa`, `sat`, `sd`, `sd_Deva`, `si`, `sk`, `vi`, `xnr`.

No tests are added because the `en` locale is unaffected.

Fixes #3158 
Fixes #5482

### DIFF
--- a/stl/src/tzdb.cpp
+++ b/stl/src/tzdb.cpp
@@ -567,7 +567,7 @@ void __stdcall __std_tzdb_delete_current_zone(__std_tzdb_current_zone_info* cons
 
     _Info->_Abbrev = _Allocate_wide_to_narrow(_Abbrev.get(), _Abbrev_len, _Info->_Err);
     if (_Info->_Abbrev == nullptr) {
-        const auto _Fallback_abbrev = new (_STD nothrow) char[]{"+0000"};
+        _STD unique_ptr<char[]> _Fallback_abbrev{new (_STD nothrow) char[]{"+0000"}};
 
         if (_Fallback_abbrev == nullptr) {
             return nullptr;
@@ -590,7 +590,7 @@ void __stdcall __std_tzdb_delete_current_zone(__std_tzdb_current_zone_info* cons
         }
 
         _Info->_Err    = __std_tzdb_error::_Success;
-        _Info->_Abbrev = _Fallback_abbrev;
+        _Info->_Abbrev = _Fallback_abbrev.release();
     }
 
     return _Info.release();
--- a/stl/src/tzdb.cpp
+++ b/stl/src/tzdb.cpp
@@ -567,16 +567,16 @@ void __stdcall __std_tzdb_delete_current_zone(__std_tzdb_current_zone_info* cons
 
     _Info->_Abbrev = _Allocate_wide_to_narrow(_Abbrev.get(), _Abbrev_len, _Info->_Err);
     if (_Info->_Abbrev == nullptr) {
-        _STD unique_ptr<char[]> _Fallback_abbrev{new (_STD nothrow) char[]{"+0000"}};
-
-        if (_Fallback_abbrev == nullptr) {
-            return nullptr;
-        }
-
         const auto _Abs_offset        = _Info->_Offset < 0 ? -_Info->_Offset : _Info->_Offset;
         const auto _Offset_in_minutes = _Abs_offset / (60 * 1000);
         const auto _Hours             = _Offset_in_minutes / 60;
         const auto _Mins              = _Offset_in_minutes % 60;
+
+        _STD unique_ptr<char[]> _Fallback_abbrev{new (_STD nothrow) char[_Mins == 0 ? 4 : 6]};
+
+        if (_Fallback_abbrev == nullptr) {
+            return nullptr;
+        }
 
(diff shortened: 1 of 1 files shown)


stl/src/tzdb.cpp
```diff
@@ -567,7 +567,30 @@ void __stdcall __std_tzdb_delete_current_zone(__std_tzdb_current_zone_info* cons
 
     _Info->_Abbrev = _Allocate_wide_to_narrow(_Abbrev.get(), _Abbrev_len, _Info->_Err);
     if (_Info->_Abbrev == nullptr) {
-        return _Propagate_error(_Info);
+        const auto _Fallback_abbrev = new (_STD nothrow) char[]{"+0000"};
```
StephanTLavavej: There's no exception safety issue here, but it would still be nicer to hold this in a `unique_ptr`.

StephanTLavavej: The initializer of `"+0000"` is somewhat confusing because that's not always the form of the output. I think it would be simpler to allow the length to vary here, indicating what we'll do below.

StephanTLavavej:
Thanks! :heart_eyes_cat: I adjusted the code slightly, then manually verified that it still correctly prints both minutes and whole hours (by brain-surgery nulling out the abbrev pointer in the debugger).

StephanTLavavej:

StephanTLavavej:
Thanks for figuring out how to fix this in a reasonable way! :heart_eyes_cat: :world_map: :clock1030:

stl/src/tzdb.cpp
```diff
@@ -567,7 +567,31 @@ void __stdcall __std_tzdb_delete_current_zone(__std_tzdb_current_zone_info* cons
...
+        if (_Mins == 0) {
+            _Fallback_abbrev[3] = '\0';
+        } else {
+            _Fallback_abbrev[3] = static_cast<char>('0' + _Mins / 10);
+            _Fallback_abbrev[4] = static_cast<char>('0' + _Mins % 10);
+            _Fallback_abbrev[5] = '\0';
+        }
+
+        _Info->_Err    = __std_tzdb_error::_Success;
+        _Info->_Abbrev = _Fallback_abbrev.release();
```
TheStormN, line 594: Sorry for commenting after a merge, but does this `release()` call means we are leaking memory?

cpplearner: We aren't leaking memory. The returned `__std_tzdb_sys_info*` will be stored in a `unique_ptr<__std_tzdb_sys_info, _Tzdb_deleter<__std_tzdb_sys_info>>`, which will call `__std_tzdb_delete_sys_info` in its destructor, which will perform `delete[] _Info->_Abbrev;`

PR #5559, opened 2025-05-31
`validate.cpp`: report error lines

### LINKED ISSUE #2349: validate.cpp: display line numbers
Display violating line numbers, for some maximum line count.
Save lines numbers in arrays, print them along with an error.

AlexGuteniev:
Closes #2349


cpplearner:
Consider including the line number in the magic markup. That is, replace `linenumber=1` with the actual line number in the definition of `validation_failure`.

https://github.com/microsoft/STL/blob/2cebf24c5b16ba66590dd18776f7166419ce14ad/tools/validate/validate.cpp#L62-L63

AlexGuteniev:
> Consider including the line number in the magic markup. That is, replace `linenumber=1` with the actual line number in the definition of `validation_failure`.

Done

AlexGuteniev:
I find it questionable that the line number is `size_t` instead of `int`, but I followed the `columns` precedent.

tools/validate/validate.cpp
```diff
@@ -54,15 +55,38 @@ class BinaryFile {
...
 };
 
+struct line_and_column_type {
+    size_t line;
+    size_t column;
+};
+
+template <>
+struct std::formatter<line_and_column_type> {
+    constexpr auto parse(std::format_parse_context& ctx) {
```
StephanTLavavej: This file is `using namespace std;`.

tools/validate/validate.cpp
```diff
@@ -54,15 +55,38 @@ class BinaryFile {
...
+    template <typename FormatContext>
+    auto format(const line_and_column_type& lc, FormatContext& ctx) const {
+        return std::format_to(ctx.out(), "{}:{}", lc.line, lc.column);
+    }
+};
+
 template <class... Args>
-void validation_failure(
-    bool& any_errors, const filesystem::path& filepath, format_string<type_identity_t<Args>...> fmt, Args&&... args) {
+void validation_failure(bool& any_errors, const filesystem::path& filepath, line_and_column_type line_and_column,
```
StephanTLavavej: This is passing a modifiable value, but the formatter took `const line_and_column_type& lc`.

tools/validate/validate.cpp
```diff
@@ -54,15 +55,38 @@ class BinaryFile {
     FILE* m_file{nullptr};
 };
 
+struct line_and_column_type {
```
StephanTLavavej: Naming nitpick: `_type` is an unnecessary wart here.

tools/validate/validate.cpp
```diff
@@ -80,11 +104,18 @@ void scan_file(
...
+    array<line_and_column_type, max_error_lines_per_file> overlength_line_numbers{};
+    array<line_and_column_type, max_error_lines_per_file> tab_characters_line_numbers{};
+    array<line_and_column_type, max_error_lines_per_file> trailing_whitespace_line_numbers{};
+
     unsigned char prev      = '@';
     unsigned char previous2 = '@';
     unsigned char previous3 = '@';
 
     size_t columns = 0;
+    size_t line    = 1;
```
StephanTLavavej: It is somewhat confusing for `columns` to be 0-based, but `line` to be 1-based. I believe it would be less error-prone for both to be 0-based (and plural) here.

tools/validate/validate.cpp
```diff
@@ -80,11 +104,18 @@ void scan_file(
     size_t tab_characters            = 0;
     size_t trailing_whitespace_lines = 0;
 
+    constexpr size_t max_error_lines_per_file = 8;
+
+    array<line_and_column_type, max_error_lines_per_file> overlength_line_numbers{};
+    array<line_and_column_type, max_error_lines_per_file> tab_characters_line_numbers{};
```
StephanTLavavej: Naming nitpick: It's strange to have two plurals here. Each of these is the location of a tab character, singular.

tools/validate/validate.cpp
```diff
@@ -54,15 +55,38 @@ class BinaryFile {
...
+    size_t column;
+};
+
+template <>
+struct std::formatter<line_and_column_type> {
+    constexpr auto parse(std::format_parse_context& ctx) {
+        return ctx.begin();
+    }
+
+    template <typename FormatContext>
```
StephanTLavavej: Nitpick: This is a `class` file.

tools/validate/validate.cpp
```diff
@@ -54,15 +55,38 @@ class BinaryFile {
     FILE* m_file{nullptr};
 };
 
+struct line_and_column_type {
+    size_t line;
+    size_t column;
```
StephanTLavavej: It would be nice to have DMIs to avoid potential garbage-init (even though the code below avoids it).

tools/validate/validate.cpp
StephanTLavavej: Nitpick: Let's consistently follow the order lines, then columns - which is also how they're updated below.

tools/validate/validate.cpp
```diff
@@ -112,17 +148,23 @@ void scan_file(
                 ++disallowed_characters;
                 constexpr size_t MaxErrorsForDisallowedCharacters = 10;
                 if (disallowed_characters <= MaxErrorsForDisallowedCharacters) {
-                    validation_failure(any_errors, filepath, "file contains disallowed character 0x{:02X}.",
-                        static_cast<unsigned int>(ch));
+                    validation_failure(any_errors, filepath, {line, columns + 1},
+                        "file contains disallowed character 0x{:02X}.", static_cast<unsigned int>(ch));
                 }
```
StephanTLavavej: There was nothing magic about limiting disallowed character errors to 10. Let's use the same limit, and follow the same pattern of checking and then incrementing.

Hmm. Actually, since you now have the technology to list multiple errors on a single line, let's use it here.

tools/validate/validate.cpp
```diff
@@ -94,13 +125,18 @@ void scan_file(
                 } else {
                     has_cr = true;
                 }
+                ++line;
```
StephanTLavavej: Not that it really matters, but there's the final "file ends with CR" check too.

tools/validate/validate.cpp
```diff
@@ -80,11 +104,18 @@ void scan_file(
     size_t tab_characters            = 0;
     size_t trailing_whitespace_lines = 0;
 
+    constexpr size_t max_error_lines_per_file = 8;
+
+    array<line_and_column_type, max_error_lines_per_file> overlength_line_numbers{};
+    array<line_and_column_type, max_error_lines_per_file> tab_characters_line_numbers{};
+    array<line_and_column_type, max_error_lines_per_file> trailing_whitespace_line_numbers{};
```
StephanTLavavej: Hmm. We're recording line and column numbers, so "line numbers" is both verbose and not fully accurate. "locations" would be shorter and more accurate.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: This should make it easier than ever for contributors to find and fix their validation problems.

I pushed a bunch of nitpicks and one enhancement: using the new strategy for disallowed characters. This helps ADO present all errors on the first page, instead of hiding the rest behind more details.

tools/validate/validate.cpp
AlexGuteniev: I preferred 0 values and `{}` initialization of whole arrays to engage `memset` compiler optimization.
Despite this program is built in debug mode, I have habits.

(I briefly considered vectorization, before noticing that it runs very fast already)

tools/validate/validate.cpp
```diff
@@ -54,15 +55,57 @@ class BinaryFile {
...
+
+    template <>
+    struct formatter<character_line_column> {
+        constexpr auto parse(format_parse_context& ctx) {
+            return ctx.begin();
+        }
+
+        template <class FormatContext>
+        auto format(const character_line_column& clc, FormatContext& ctx) const {
+            return format_to(ctx.out(), "0x{:02X}@{}", static_cast<unsigned int>(clc.ch), clc.lc);
```
AlexGuteniev: `0x86@97:1` is not very readable and even confusing to some extent. What if we use `0x86 @97:1` or `0x86 @ 97:1`?

StephanTLavavej: Sure, let's add spaces on either side.

StephanTLavavej (approved):

tools/validate/validate.cpp
StephanTLavavej: I'd be fine with 0 if you feel strongly; zero is a strict subset of garbage.

AlexGuteniev: This is fine as it is.

StephanTLavavej:

StephanTLavavej:
Thanks for making it easier than ever to conform to the repo's formatting requirements! :heart_eyes_cat: :chart_with_upwards_trend: :tada:

PR #5561, opened 2025-05-31
`<filesystem>`: Fix the `path` returned when there is an error with `temp_directory_path` and simplify the validation logic

### LINKED ISSUE #5557: `<filesystem>`: `temp_directory_path(error_code&)` doesn't return an empty `path` for errors
WG21-N5008 \[fs.op.temp.dir.path\]/3:

> *Returns:* The path `p`. The signature with argument `ec` returns `path()` if an error occurs.

That doesn't appear to be what we do:

https://github.com/microsoft/STL/blob/7841cf88ff9af837fb980f2d2d4ac0f267e714c7/stl/inc/filesystem#L4042-L4056
https://github.com/microsoft/STL/blob/7841cf88ff9af837fb980f2d2d4ac0f267e714c7/stl/src/filesystem.cpp#L807-L835

YexuanXiao:
Fix #5557.

In the current implementation, `temp_directory_path` would first use `GetTempPath`/`GetTempPath2W` to obtain the temp path, then use `GetFileAttributesW` to obtain file attributes to verify if it was a directory. If the file is not a directory but is a symlink, the symlink would be opened to verify that the symlink is correct. Since `GetFileAttributesW` internally opens the file and uses `GetFileInformationByHandleEx` to obtain file attributes, the current logic contains redundancy. Instead, we should unconditionally open the path obtained by `GetTempPath`/`GetTempPath2W`, follow symlinks, and verify that it is a directory through `GetFileInformationByHandleEx`.

### DIFF
--- a/stl/inc/filesystem
+++ b/stl/inc/filesystem
@@ -618,6 +618,7 @@ namespace filesystem {
         friend inline path absolute(const path& _Input, error_code& _Ec);
         friend inline __std_win_error _Canonical(path& _Result, const wstring& _Text);
         friend inline path temp_directory_path(error_code& _Ec);
+        friend inline path temp_directory_path();
         friend inline path current_path(error_code& _Ec);
         friend inline void current_path(const path& _To);
         friend inline void current_path(const path& _To, error_code& _Ec) noexcept;
@@ -3945,7 +3946,7 @@ namespace filesystem {
         if (_Error == __std_win_error::_Success) {
             _Result = file_time_type{file_time_type::duration{_Stats._Last_write_time}};
         } else {
-            _Result = (file_time_type::min)();
+            _Result = (file_time_type::min) ();
         }
 
         return _Error;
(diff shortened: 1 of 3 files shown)


YexuanXiao:
It looks like two tests failed, and I'll debug them later.

YexuanXiao:
https://github.com/microsoft/STL/blob/7841cf88ff9af837fb980f2d2d4ac0f267e714c7/stl/inc/filesystem#L4049-L4052

@StephanTLavavej What's the significance of using different categories here? The standard doesn't seem to require it.

StephanTLavavej:
What Windows error code would we synthesize to indicate "not a directory"?

YexuanXiao:
> What Windows error code would we synthesize to indicate "not a directory"?

If the path points to a symlink but the target does not exist then `CreateFile` will report `_File_not_found`; or the target's class does not match the symlink, then `CreateFile` will report `_Directory_name_is_invalid`. For consistency, [L830](https://github.com/microsoft/STL/pull/5561/files#diff-27a6050161dbeaacef8db1e1a236a18b9762936515f4af930283d8ddb16666baR830) might need to be the same.

The File APIs does not seem to strictly distinguish between files and directories in its error codes, and the error codes appear to be unstable. Unless necessary, identifying or conversion should be avoided.

StephanTLavavej:
There are a lot of logic changes here, much more than are needed to simply clear the path on error. Can you please update your PR description to explain how the code is changing? This is both for my benefit while reviewing, and for future maintainers/contributors to understand what changed.

Edit: Indeed, it looks like these are unrelated behavioral changes. We try to avoid mixing together bugfixes with other behavioral changes (or even cleanups) unless they are inextricably linked.

YexuanXiao:
> There are a lot of logic changes here, much more than are needed to simply clear the path on error. Can you please update your PR description to explain how the code is changing? This is both for my benefit while reviewing, and for future maintainers/contributors to understand what changed.
> 
> Edit: Indeed, it looks like these are unrelated behavioral changes. We try to avoid mixing together bugfixes with other behavioral changes (or even cleanups) unless they are inextricably linked.

Should I split it into two PRs?

StephanTLavavej:
> Should I split it into two PRs?

Yes, please.

PR #5563, opened 2025-06-01
`<xstring>`: Suppress code analysis warning C6510 for `basic_string`

muellerj2:
When instantiating `basic_string` with character types of class type, code analysis warning C6510 is produced. See here: https://gcc.godbolt.org/z/8dhaG8bd4

I noticed this while working on #995, when I tried to instantiate `basic_regex` (and hence `basic_string`) with a character type of class type to extend test coverage.

#3032 already resolved this issue for `basic_string_view`. This PR applies the same fix (warning suppression) to `basic_string` and extends the existing test coverage to `basic_string`. (Strictly speaking, the extended test isn't run in C++14 mode, but I think this is still good enough to validate a warning suppression applied unconditionally. And there will soon be positive indirect coverage in C++14 mode via the PRs for #995.)


tests/std/tests/P0220R1_string_view/test.cpp
```diff
@@ -1247,18 +1253,76 @@ struct std::char_traits<char_wrapper> {
         return strlen(reinterpret_cast<const char*>(a));
     }
 
+    static char_wrapper* copy(char_wrapper* const first1, const char_wrapper* const first2, const size_t count) {
+        copy_n(first2, count, first1);
```
StephanTLavavej, line 1258: Should include `<algorithm>`.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for extending this suppression! :shushing_face: :mute: :zipper_mouth_face:

PR #5565, opened 2025-06-01
`filesystem.cpp`: Avoid loading `GetTempPath2W` for OneCore

StephanTLavavej:
[`GetModuleHandleW`](https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulehandlew) is a "desktop apps only" API. We avoid calling it for OneCore:

https://github.com/microsoft/STL/blob/7841cf88ff9af837fb980f2d2d4ac0f267e714c7/stl/src/winapisupp.cpp#L179-L194

and so we can't load [`GetTempPath2W`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppath2w) for OneCore and must always use the [`GetTempPathW`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppathw) fallback instead:

https://github.com/microsoft/STL/blob/7841cf88ff9af837fb980f2d2d4ac0f267e714c7/stl/src/winapisupp.cpp#L164-L175

`winapisupp.cpp` is built into the STL's DLL, and the non-dllexported `__crtGetTempPath2W` is used by `filesys.cpp` in the DLL, implementing the non-Standard/highly-deprecated `<experimental/filesystem>`.

For Standard `<filesystem>`, we inject `filesystem.cpp` into the STL's import LIB, but we weren't similarly guarding `GetModuleHandleW`. (This is the only other call in the STL.) Somehow, nobody has hissed about `<filesystem>` being incompatible with UWP apps, but I suspect maybe nobody happened to be calling `temp_directory_path()` in their UWP apps, and so the linker discarded this unused machinery from the import lib. In any event, we should fix the code to be consistent and avoid potential problems for UWP.

### DIFF
--- a/stl/src/filesystem.cpp
+++ b/stl/src/filesystem.cpp
@@ -789,18 +789,21 @@ struct __std_fs_file_id { // typedef struct _FILE_ID_INFO {
 namespace {
     _Success_(return > 0 && return < nBufferLength) DWORD WINAPI _Stl_GetTempPath2W(
         _In_ DWORD nBufferLength, _Out_writes_to_opt_(nBufferLength, return +1) LPWSTR lpBuffer) noexcept {
+#if !defined(_ONECORE)
         // See GH-3011: This is intentionally not attempting to cache the function pointer.
         // TRANSITION, ABI: This should use __crtGetTempPath2W after this code is moved into the STL's DLL.
-        using _Fun_ptr = decltype(&::GetTempPath2W);
 
+        // use GetTempPath2W if it is available (only on Windows 11+)...
         const auto _Kernel32 = ::GetModuleHandleW(L"kernel32.dll");
         _Analysis_assume_(_Kernel32);
-        _Fun_ptr _PfGetTempPath2W = reinterpret_cast<_Fun_ptr>(::GetProcAddress(_Kernel32, "GetTempPath2W"));
-        if (!_PfGetTempPath2W) {
-            _PfGetTempPath2W = &::GetTempPathW;
+        const auto _Pf = reinterpret_cast<decltype(&::GetTempPath2W)>(::GetProcAddress(_Kernel32, "GetTempPath2W"));
+        if (_Pf) {
+            return _Pf(nBufferLength, lpBuffer);
         }
+#endif // ^^^ !defined(_ONECORE) ^^^
 
-        return _PfGetTempPath2W(nBufferLength, lpBuffer);
+        // ...otherwise use GetTempPathW.
+        return GetTempPathW(nBufferLength, lpBuffer);
     }
 } // unnamed namespace
 


YexuanXiao:
I suspect UWP apps prefer using WinRT over STL to obtain this information because the STL lacks support for access permissions and lacks APIs related to application data folders.

stl/src/filesystem.cpp
```diff
@@ -789,18 +789,21 @@ struct __std_fs_file_id { // typedef struct _FILE_ID_INFO {
 namespace {
     _Success_(return > 0 && return < nBufferLength) DWORD WINAPI _Stl_GetTempPath2W(
         _In_ DWORD nBufferLength, _Out_writes_to_opt_(nBufferLength, return +1) LPWSTR lpBuffer) noexcept {
+#if !defined(_ONECORE)
```
DHowett, line 792: FWIW, `onecore` and `store` are two different targets that may be worth considering distinctly here.

I observe that my machine has _four_ copies of each arch STL object library: `[nothing]`, `onecore`, `uwp` and `store`.

Would this change--which is specific to `uwp` and `store`--break `onecore`?

StephanTLavavej: I don't pretend to understand the distinction; all I know is that our usage is inconsistent, and the rest of the codebase certainly believes that checking for OneCore is how we remain UWP-friendly.

DHowett: Works for me! 🙃

stl/src/filesystem.cpp
```diff
@@ -789,18 +789,21 @@ struct __std_fs_file_id { // typedef struct _FILE_ID_INFO {
...
-        if (!_PfGetTempPath2W) {
-            _PfGetTempPath2W = &::GetTempPathW;
+        const auto _Pf = reinterpret_cast<decltype(&::GetTempPath2W)>(::GetProcAddress(_Kernel32, "GetTempPath2W"));
+        if (_Pf) {
+            return _Pf(nBufferLength, lpBuffer);
         }
+#endif // ^^^ !defined(_ONECORE) ^^^
 
-        return _PfGetTempPath2W(nBufferLength, lpBuffer);
+        // ...otherwise use GetTempPathW.
```
davidmrdavid, line 805: non-blocking nit: I'm not sure if this is a pre-existing convention in the codebase, but I'm not the biggest fan of "split comments" separated by "...". As the code changes, it can get lost that these were part of the same sentence.

In any case, this is a subjective stylistic nit. The change seems correct, thank you!

stl/src/filesystem.cpp
```diff
@@ -789,18 +789,21 @@ struct __std_fs_file_id { // typedef struct _FILE_ID_INFO {
...
         // TRANSITION, ABI: This should use __crtGetTempPath2W after this code is moved into the STL's DLL.
-        using _Fun_ptr = decltype(&::GetTempPath2W);
 
+        // use GetTempPath2W if it is available (only on Windows 11+)...
         const auto _Kernel32 = ::GetModuleHandleW(L"kernel32.dll");
         _Analysis_assume_(_Kernel32);
-        _Fun_ptr _PfGetTempPath2W = reinterpret_cast<_Fun_ptr>(::GetProcAddress(_Kernel32, "GetTempPath2W"));
-        if (!_PfGetTempPath2W) {
-            _PfGetTempPath2W = &::GetTempPathW;
+        const auto _Pf = reinterpret_cast<decltype(&::GetTempPath2W)>(::GetProcAddress(_Kernel32, "GetTempPath2W"));
```
davidmrdavid, line 799: for my own learning: is `_Pf` convention for "pointer function" or "process function" or something like that?

davidmrdavid (approved):

stl/src/filesystem.cpp
StephanTLavavej: "pointer to function", yes.

stl/src/filesystem.cpp
StephanTLavavej: Yeah, they're pretty rare, we probably ought to make them rarer. At least the ellipses make it somewhat clear that they're partial (I absolutely agree that they would be abominations without ellipses).

StephanTLavavej:

StephanTLavavej:
I'm going to recharacterize this as a consistency enhancement instead of a bugfix. @amyw-msft informed me that the banned API restrictions were lifted back in 2020, and while certain APIs still can't be physically called (e.g. user32 UI), there are no deployment time blocks for common Win32 APIs. In the future I'll try to eradicate all of the OneCore checks, but for now I'll merge this.

PR #5570, opened 2025-06-03
`<filesystem>`: Fix the `path` returned when there is an error with `temp_directory_path` 

### LINKED ISSUE #5557: `<filesystem>`: `temp_directory_path(error_code&)` doesn't return an empty `path` for errors

YexuanXiao:
Fix #5557.

### DIFF
--- a/stl/inc/filesystem
+++ b/stl/inc/filesystem
@@ -617,7 +617,7 @@ namespace filesystem {
         friend class _Path_iterator;
         friend inline path absolute(const path& _Input, error_code& _Ec);
         friend inline __std_win_error _Canonical(path& _Result, const wstring& _Text);
-        friend inline path temp_directory_path(error_code& _Ec);
+        friend inline path _Temp_directory_path_impl(error_code& _Ec);
         friend inline path current_path(error_code& _Ec);
         friend inline void current_path(const path& _To);
         friend inline void current_path(const path& _To, error_code& _Ec) noexcept;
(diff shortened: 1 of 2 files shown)


stl/inc/filesystem
```diff
@@ -4039,9 +4039,8 @@ namespace filesystem {
         _STD filesystem::permissions(_Target, _Perms, perm_options::replace, _Ec);
     }
 
-    _EXPORT_STD _NODISCARD inline path temp_directory_path(error_code& _Ec) {
+    _NODISCARD inline path _Temp_directory_path_impl(error_code& _Ec) {
```
AlexGuteniev, line 4042: What are advantages of wrapping the existent function instead of clearing that string on error right here?

YexuanXiao: When an error occurs, `_impl` does not clear the path, but the `error_code` version does. Therefore, the exception version needs to use `_impl` to reuse code instead of the `error_code` version in order to report the path.

AlexGuteniev: Oh, I see the path is the part of the exception. Then it makes sense!

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for fixing this long-standing bug! :floppy_disk: :hammer_and_wrench: :smile_cat:

PR #5575, opened 2025-06-07
`<semaphore>`: use `steady_clock`

### LINKED ISSUE #5574: `<semaphore>`: `counting_semaphore::try_acquire_for()` does not block long enough for shorter durations
Originally reported as DevCom-10913616 / VSO-2495823.

My cleaned-up repro:

```
C:\Temp>type meow.cpp
```
```cpp
#include <chrono>
#include <print>
#include <semaphore>
#include <stop_token>
#include <thread>
using namespace std;
using namespace std::chrono;

void f(stop_token stop) {
    binary_semaphore new_timer{0};
    for (int i = 0; i < 50 && !stop.stop_requested(); ++i) {
        const auto s = steady_clock::now();
        (void) new_timer.try_acquire_for(800us);
        const auto e = steady_clock::now();
        println("{}", duration_cast<nanoseconds>(e - s));
    }
}
int main() {
    jthread t(f);
    this_thread::sleep_for(1s);
}
```
```
C:\Temp>cl /EHsc /nologo /W4 /std:c++latest /MT /O2 meow.cpp
meow.cpp

C:\Temp>meow
200ns
100ns
0ns
0ns
0ns
0ns
100ns
0ns
[...]
```

I haven't fully analyzed this, but I observe that we still have some calls to `GetTickCount64()`, which IIRC uses the system clock instead of a steady clock:

https://github.com/microsoft/STL/blob/7841cf88ff9af837fb980f2d2d4ac0f267e714c7/stl/src/atomic_wait.cpp#L180-L207

# :hammer_and_wrench: Previous fixes for this category of bugs
The other uses of `GetTickCount64()` in sharedmutex.cpp appear to be properly "checked" in the headers against `steady_clock` or the user's clock. This is after the series of fixes:

* VS 2022 17.9: #3914
  + Fixed `this_thread::sleep_until()` to consistently use the given `time_point`'s clock type, instead of also depending on the system clock. This also fixed `this_thread::sleep_for()` to consistently use `steady_clock`.
* VS 2022 17.11: #4457
  + Fixed `condition_variable::wait_until()` and `condition_variable_any::wait_until()` to consistently use the given `time_point`'s clock type, instead of also depending on the system clock. This also fixed `condition_variable::wait_for()` to consistently use `steady_clock`.
* VS 2022 17.12: #4755`
  + Fixed `condition_variable_any::wait_for()` to consistently use `steady_clock`.

AlexGuteniev:
Fixes #5574 

Abandon `__std_atomic_wait_get_deadline` and `__std_atomic_wait_get_remaining_timeout`. Originally they were used as fast and convenient steady clock, that is just as precise as the wait itself. But we have to use the `steady_clock`, not _a steady clock_, and be as precise as it is.

No need in `_Clamped_rel_time_ms_count`. The existing `_Semaphore_remaining_timeout` is good enough, although it diverges in that it clamps to 10 days, not to 1 day. Importantly. both use `chrono::ceil` to get larger rather than smaller timeout when rounding.

### DIFF
--- a/stl/inc/semaphore
+++ b/stl/inc/semaphore
@@ -29,12 +29,6 @@ _STL_DISABLE_CLANG_WARNINGS
 
 _STD_BEGIN
 
-template <class _Rep, class _Period>
-_NODISCARD unsigned long long _Semaphore_deadline(const chrono::duration<_Rep, _Period>& _Rel_time) {
-    return __std_atomic_wait_get_deadline(
-        chrono::duration_cast<chrono::duration<unsigned long long, milli>>(_Rel_time).count());
-}
-
 template <class _Clock, class _Duration>
 _NODISCARD unsigned long _Semaphore_remaining_timeout(const chrono::time_point<_Clock, _Duration>& _Abs_time) {
     const auto _Now = _Clock::now();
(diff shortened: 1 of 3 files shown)


StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:vertical_traffic_light: :stopwatch: :balance_scale:
===

Issue #5393, opened 2025-04-06
`<regex>`: What names can and should `regex_traits::lookup_collatename()` recognize?


muellerj2:
The title states this as one question, but it's really two.

First is an actual choice: Should the [portable character names in the POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap06.html#tag_06_01) be recognized and translated by`lookup_collatename()`?

Cons:
* These aren't actually names for locale-specific collating elements (which are defined in [Chapter 7](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap07.html)); they are symbolic names of characters for internal use in the POSIX standard.
* TR1 was deliberately changed to not mandate that these are recognized by `lookup_coolatename()` (see the end of Section 2 of [N1623](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2004/n1623.html)).

Pro:
* Boost.Regex and the other standard libraries recognize these names as an extension.
* After #5392, these names are the final puzzle piece to make a few more libcxx tests pass.

The second question is a technical issue that I don't have an answer to yet: Can we actually access the set of locale-specific collating elements or recognize them in a reasonable way using some Windows API? Or is there some other reasonable approach to recognize locale-specific collating elements like "ch" in Czech or "dzs" in Hungarian?

muellerj2:
I have been looking a bit into the recognition of locale-specific collating elements:

* I failed to find a simple Windows API to query collating elements of a locale.
* In ICU, there is the concept of [*contractions*](https://unicode-org.github.io/icu/userguide/collation/concepts.html#contractions), which is similar to POSIX' collating elements. ICU supports querying the list of contractions for a specific locale, so maybe we could solve this by querying ICU. But do we want to make regex dependent on ICU?
[...]

PR #5587, opened 2025-06-14
Categorize remaining regex-related libcxx test failures

muellerj2:
There are two remaining reasons for failing libcxx tests related to `regex`:

* libc++'s implementation of `match_prev_avail` is broken, see llvm/llvm-project#74838.
* the remaining failing tests are related to the recognition of collating elements, so are covered by issue #5393. (If preferred, we could make an additional distinction: the "locale" test failures are about the "can" in the issue question, the failures of the other grammar-specific tests are about the "should", and the failure of the `lookup_collatename()` test is about both.)


StephanTLavavej:

StephanTLavavej:
Thanks for investigating what remaining work is needed! :detective: :scientist: :bulb:

PR #5591, opened 2025-06-15
Vectorize `mismatch` for clang-cl for odd element sizes

AlexGuteniev:
Towards #5479

Unlike other comparing algorithms, `mismatch` vectorization can be expanded towards arbitrary element size easily, without even adding separately-compiled code. This is currently possible for clang-cl only, and it requires defaulted `operator==`, and trivially-comparable types, and unique object representation.

The mismatch result can be divided by element size in the header, thus truncating any offset into the middle, giving the element index. The division is likely to be optimized to large multiplication by the compiler.

Rather than using just `__std_mismatch_1`, attempt to use as large element mismatch as possible, for the following reasons:
 * Primarily, optimal tail processing. The reason to have `_2` / `_4` / `_8` versions in the first place.
 * Also potentially smaller divisor is more likely to be optimized to the multiplication by the compiler. Though I haven't checked whether the compiler can get rid of power-of-two factor on its own; probably it can.

`lexicographical_compare` vectorization cannot be expanded this way, as it requires a different flavor of trivial comparability.

## Benchmark results

Benchmark                                    | Before   |  After    | Speedup
---------------------------------------------|----------|-----------|----------
`bm<color, op::mismatch, c1, c2>/8/3`        |  1.97 ns |  1.92 ns  |   1.03
`bm<color, op::mismatch, c1, c2>/24/22`      |  11.4 ns |  4.22 ns  |   2.70
`bm<color, op::mismatch, c1, c2>/105/-1`     |  52.6 ns |  11.4 ns  |   4.61
`bm<color, op::mismatch, c1, c2>/4021/3056`  |  1456 ns |   294 ns  |   4.95

### DIFF
--- a/stl/inc/xutility
+++ b/stl/inc/xutility
@@ -454,7 +454,15 @@ size_t _Mismatch_vectorized(const void* const _First1, const void* const _First2
     } else if constexpr (_Element_size == 8) {
         return __std_mismatch_8(_First1, _First2, _Count);
     } else {
-        _STL_INTERNAL_STATIC_ASSERT(false); // unexpected size
+        if constexpr ((_Element_size % 8) == 0) {
+            return __std_mismatch_8(_First1, _First2, _Count * (_Element_size / 8)) / (_Element_size / 8);
+        } else if constexpr ((_Element_size % 4) == 0) {
+            return __std_mismatch_4(_First1, _First2, _Count * (_Element_size / 4)) / (_Element_size / 4);
+        } else if constexpr ((_Element_size % 2) == 0) {
+            return __std_mismatch_2(_First1, _First2, _Count * (_Element_size / 2)) / (_Element_size / 2);
+        } else {
+            return __std_mismatch_1(_First1, _First2, _Count * _Element_size) / _Element_size;
+        }
     }
 }
 _STD_END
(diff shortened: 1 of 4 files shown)


stl/inc/xutility
```diff
@@ -454,7 +454,15 @@ size_t _Mismatch_vectorized(const void* const _First1, const void* const _First2
     } else if constexpr (_Element_size == 8) {
         return __std_mismatch_8(_First1, _First2, _Count);
     } else {
-        _STL_INTERNAL_STATIC_ASSERT(false); // unexpected size
+        if constexpr ((_Element_size % 8) == 0) {
```
StephanTLavavej: As a multiplicative operator, `%` has far higher precedence than `==` and doesn't need parentheses here.

StephanTLavavej: We can actually use the mod cases to handle the equal cases. Because `_Element_size` is a template parameter, the compiler will easily handle multiplying and dividing by `1` (and I am not concerned about the trivial impact on debug codegen).

tests/std/tests/VSO_0000000_vector_algorithms_mismatch_and_lex_compare/test.cpp
```diff
@@ -130,6 +138,65 @@ void test_mismatch_and_lex_compare_family(mt19937_64& gen) {
...
 }
 
+#if _HAS_CXX20
+template <class T>
+struct triplet {
+    T x;
+    T y;
+    T z;
+
+    bool operator==(const triplet<T>&) const = default;
```
StephanTLavavej: We conventionally take advantage of the injected-class-name.

StephanTLavavej (approved):

stl/inc/xutility
AlexGuteniev: Great idea!

StephanTLavavej:

StephanTLavavej:
:3rd_place_medal: :five: :seven:
===

PR #5592, opened 2025-06-15
`<regex>`: Remove usage of non-standard `_Uelem` from parser

muellerj2:
Towards #995. A second PR in the future will remove `_Uelem` from the matcher.

## Requirements on the character type `_Elem` in the standard

Effectively, the standard currently spells out the following guarantees for the character type:
* The regex traits class can handle the character type for all the operations in [\[re.traits\]](https://eel.is/c++draft/re.traits).
* Indirectly, through the reference to `string_type` as a `basic_string` in the regex traits class requirements, the `traits_type` of the `string_type` (which should be `char_traits<_Elem>`) must support all operations in [\[char.traits.require\]](https://eel.is/c++draft/char.traits.require).
* According to [\[strings.general\]/1](https://eel.is/c++draft/strings.general#1) referenced by [\[re.general\]/2](https://eel.is/c++draft/re.general#2), `_Elem` must be a non-array trivially copyable standard-layout and trivially default-constructible type.

Alas, this is woefully underspecified because `regex` has to convert and compare code points (integers) to characters (see also LWG-3835). There is a de facto requirement that `regex` must be able to convert and compare integers to characters, otherwise regexes couldn't be parsed or line endings couldn't be matched. There is also a de facto requirement that `regex` must be able to convert characters to integers again to implement [\[re.grammar\]/12](https://eel.is/c++draft/re.grammar#12). But how these conversions and comparisons actually work is not specified at all.

One idea out might be to rely on the existing `int_type` in the character traits type, but the standard (a) does not actually specify any property for this type other than that it can somehow represent all characters + `eof()` (see [\[char.traits.typedefs\]/1](https://eel.is/c++draft/char.traits.typedefs#1)) and (b) immediately goes on to violate this only requirement in the specializations for Unicode character types (see LWG-2959). Moreover, there doesn't appear to be an actual requirement that `int_type` is an integer type -- there is only a guarantee that it can be used through the API of the character traits class. So `int_type` does not seem helpful, rather, relying on it would just open its own can of worms.

## Requirements on `_Elem` in the `regex` implementations before and after this PR

The following fundamental requirements on `_Elem` remain unchanged by this PR:

* All integral conversions must be consistent with some implicit one-to-one-mapping between unsigned integers (code points) and characters (up to lossiness and sign extension), and the natural ordering of these unsigned integers must be consistent with the `lt()` and `eq()` functions in the character traits class.
* Lossy casts to and from integers are allowed and behave as one would expect for casts between integers.
* There must not be any gaps between code points, i.e., if a character has code point `c`, then `0, .., c-1` must be valid code points of characters as well (in the sense that we can obtain a different character object for each of them through casting).
* The code points for the basic character set (to the degree they have special meaning in regexes) plus the paragraph and line separator must agree with the corresponding Unicode code points. For the paragraph and line separator, this must only be the case if the `_Elem` type is large enough to represent these code points.
* `_Elem{}` must represent NUL (with code point 0).

## Requirements on `_Elem` in the `regex` implementation before this PR
The current implementation makes at least the following additional assumptions on convertibility and comparability:

* `_Elem` must be equality-comparable to itself, `char`, `int` and the internal enum type `_Meta_type` (maybe including some implicit conversion).
* `_Elem` type must be implicitly convertible to `int` and `_Meta_type` must be implicitly convertible to `_Elem`.
* `char`, `int` and `unsigned int` must be explicitly convertible to `_Elem`.
* `_Elem` must be explicitly convertible to `_Meta_type`.
* The code point for any character can fit into an `int`, i.e., conversion `_Elem` -> `int` -> `_Elem` must yield the original character again.
* The character traits must provide an unsigned integral type `_Uelem` such that explicit conversion to this type yields the code point for any character, i.e., conversion `_Elem` -> `_Uelem` -> `_Elem` produces the original character again *and* the natural ordering of the `_Uelem` values after conversion must be consistent with the `lt()` and `eq()` functions in the character traits class. (Note that conversion to an arbitrary but big unsigned integral type does not achieve this if the character type behaves like a signed integral type because the converted value will be sign-extended.)
* `make_unsigned<_Elem>` must be well-defined. (This mostly defeats the purpose of `_Uelem`, because specializing `make_unsigned<_Elem>` for user-defined types is forbidden.)

This list might be non-exhaustive.

## Requirements on `_Elem` in the `regex` implementation after this PR

This PR imposes the following requirements on comparisons and conversions:
* `_Elem` must be equality-comparable to itself.
* `_Elem` must be explicitly convertible to `unsigned char` and `unsigned int`.
  * If `_Elem` is an integral or enum type, it must also be explicitly convertible to and from `make_unsigned<_Elem>`. Explicit conversion to this character type thus yields the code points of the characters.
  * Otherwise, conversion to `unsigned int` must yield the unsigned code point for a character, if the code point can fit into an `unsigned int`.
    * This means that `_Elem` must behave like an unsigned integer type when converted to `unsigned int`.
* `char`, `unsigned char` and `unsigned int` must be explicitly convertible to `_Elem`.

This list should be exhaustive; the new test checks that we don't do any conversions not listed above.

We cannot just drop most of the special logic for signed integral and enum types because we must support `char`.

But if desired, we could drop explicit convertibility of `_Elem` from and to `char` and `unsigned char` in favor of `unsigned int` (and `make_unsigned<_Elem>` for integral or enum `_Elem`) only. But this will mean we will have to add even more casts in `<regex>`.

## Differences in requirements

Essentially, we have the following new requirements:
* `_Elem` must be explicitly convertible to `unsigned int`.
* `_Elem` must support explicit conversion to and from `unsigned char`.
* `_Elem` must convert like an unsigned integer type when explicitly converted to `unsigned int`, if `_Elem` is not an integral or enum type. (This requirement is only kind of new: Before this PR, `regex` didn't even compile for such types, except when entering UB territory by specializing `make_unsigned`.)

In exchange, the following requirements will be dropped when the changes are completed for the matcher as well:
* No more equality-comparability between `_Elem` and other integral types.
* `_Elem` does not have to be convertible to and from `int` and `_Meta_type`.
* No more implicit conversions from and to `_Elem`.
* The code points for `_Elem` no longer have to fit into an `int`.
* No more `_Uelem` in the regex traits class.
* No more reliance on `make_unsigned<_Elem>` for types that are not integral or enum.

## Changes
* Add new `_Unescaped_char` member to `_Parser2` to represent the character represented by some kind of escape sequence.
  * This allows us to drop the "must fit into an `int`" requirement, because we no longer have to represent such characters in the `_Val` member.
* Lots of `static_cast`'s added to avoid implicit conversions.
  * For many `_Meta_type` values, we first have to cast them to `char` before casting them to `_Elem`.
  * In two cases, I avoided casts from `_Meta_type` values by temporarily saving the character from the input character sequence.
* To decide whether a character can be stored in the bitmap, the test checks now whether a character code point fits into an `unsigned char`.
  * Such tests are done by roundtrip casting: `_Elem` -> `unsigned char` -> `_Elem` yields the original character again if and only if the code point is less than 256.
* Only call `strchr()` in `_Parser2::_Trans()` when the character code point fits into an `unsigned char`.
  * It's actually surprising that this logic worked for `wregex` (probably by accident). It uses the fact that casting to `_Meta_type` for non-ASCII code points produces meta values the parser doesn't know about, and the parser treats unknown meta values as non-special. In any case, this change is also necessary to drop the "must fit into an `int`" requirement.
* Replace some ordering checks by calls to the `lt()` function of the character traits class.
  *  This lets us avoid some casting and case distinction between integral/enum types and other types. We can also avoid the addition of a requirement that the character code points must fit into an `unsigned int` or similar.
* Rework the logic for the bitmap optimization and small character range optimization for the new requirements.
  * The code for the small character range optimization is no longer generated if `sizeof(_Elem) = 1U`: The bitmap optimization already handles all characters with code point < 256, so this is just dead code.
  * Checks whether character code points are too large (to store the character in the bitmap or the code point in `unsigned int`) are done using roundtrip casting.
  * We have to differentiate between integral/enum types and other user-defined types for the small character range optimization, as we have to cast the characters to different unsigned integer types.
    * When `_Elem` is not an integral or enum type, the small character range optimization is not performed for code points unrepresentable by `unsigned int`.
* Store the named character classes and equivalence classes if `sizeof(_Elem) > 1U`.
  * This is sufficient (due to the requirement that code points must have no gaps), but not necessary: We could have a character type with `sizeof(_Elem) > 1U` with maximal code point < 256. A more accurate check appears much more ugly to me, though, for little gain.
* Revise logic to cast integer value (from escape sequence) to an `_Elem` in `_CharacterEscape()`:
  * Again, we have to differentiate between integral/enum types and other user-defined types to perform the correct casts.
  * We check whether the integer value corresponds to a code point by performing roundtrip and testing whether it yields the same integer value again.
*  No more `switch` on `_Elem`. All switches are now performed on `unsigned char` (after checking that the code point is less than 256 while handling the case for code points >= 256 separately) or `int` (when it is on `_Mchar`). In the one switch on `_Mchar`, the value is cast to `int` first to prevent a warning that there aren't case labels for some of the enum identifiers.

## Test

For now, the test only checks that the parser compiles and doesn't crash on two user-defines character types. We can only check semantic correctness after adjusting the matcher similarly.

The test currently suppresses warning C6510 because #5563 hasn't been merged yet.

### DIFF
--- a/tests/std/tests/GH_000995_regex_custom_char_types/env.lst
+++ b/tests/std/tests/GH_000995_regex_custom_char_types/env.lst
@@ -1,4 +1,4 @@
 # Copyright (c) Microsoft Corporation.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-RUNALL_INCLUDE ..\usual_matrix.lst
+RUNALL_INCLUDE ..\impure_matrix.lst


tests/std/tests/GH_000995_regex_custom_char_types/test.cpp
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    locale_type getloc() const {
+        return inner.getloc();
+    }
+
+    regex_traits<UnderlyingChar> inner;
+};
+
+enum class signed_wchar_enum : signed short {
```
StephanTLavavej: Style nitpick: We conventionally say `short`, never `signed short`.

tests/std/tests/GH_000995_regex_custom_char_types/test.cpp
```diff
@@ -0,0 +1,262 @@
...
+    return !(lhs == rhs);
+}
+
+template <class Elem>
+struct custom_char_traits {
+
+    using char_type  = Elem;
+    using int_type   = int;
+    using pos_type   = streampos;
+    using off_type   = streamoff;
```
StephanTLavavej, line 148: Should include `<ios>`.

tests/std/tests/GH_000995_regex_custom_char_types/test.cpp
```diff
@@ -0,0 +1,262 @@
...
+    using state_type = char_traits<wchar_t>::state_type;
+
+    static Elem* copy(Elem* const first1, const Elem* const first2, const size_t count) noexcept {
+        std::copy_n(first2, count, first1);
```
StephanTLavavej: Should include `<algorithm>`, and this test is already `using namespace std;` (and there is no `copy_n()` member function here).

tests/std/tests/GH_000995_regex_custom_char_types/test.cpp
```diff
@@ -0,0 +1,262 @@
...
+        while (*first != Elem{}) {
+            ++count;
+            ++first;
+        }
+
+        return count;
+    }
+
+    static const Elem* find(const Elem* first, size_t count, const Elem& ch) noexcept /* strengthened */ {
+        // look for _Ch in [_First, _First + _Count)
```
StephanTLavavej: Copy-pasted ugly comment.

tests/std/tests/GH_000995_regex_custom_char_types/test.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+    locale_type imbue(locale_type l) {
+        return inner.imbue(l);
+    }
+
+    locale_type getloc() const {
+        return inner.getloc();
+    }
+
+    regex_traits<UnderlyingChar> inner;
```
StephanTLavavej: Suggestion: There appears to be no reason for `inner` to be public, and we could use `rx_traits` here.

tests/std/tests/GH_000995_regex_custom_char_types/test.cpp
```diff
@@ -0,0 +1,262 @@
...
+    for (; first != last; ++first) {
+        str.push_back(static_cast<UnderlyingChar>(*first));
+    }
+    return str;
+}
+
+template <class StringType, class UnderlyingChar>
+StringType convert_from_underlying_string(const basic_string<UnderlyingChar>& str) {
+    StringType result;
+    for (auto ch : str) {
```
StephanTLavavej: Style: I always recommend iterating in-place when possible.

tests/std/tests/GH_000995_regex_custom_char_types/test.cpp
```diff
@@ -0,0 +1,262 @@
...
+    wchar_t character;
+};
+
+bool operator!=(const wrapped_wchar& lhs, const wrapped_wchar& rhs) {
+    return !(lhs == rhs);
+}
+
+template <class Elem>
+struct custom_char_traits {
+
```
StephanTLavavej: Nitpick: Unnecessary newline.

tests/std/tests/GH_000995_regex_custom_char_types/test.cpp
```diff
@@ -0,0 +1,262 @@
...
+    using off_type   = streamoff;
+    using state_type = char_traits<wchar_t>::state_type;
+
+    static Elem* copy(Elem* const first1, const Elem* const first2, const size_t count) noexcept {
+        std::copy_n(first2, count, first1);
+        return first1;
+    }
+
+    static Elem* move(Elem* const first1, const Elem* const first2, const size_t count) noexcept /* strengthened */ {
+        std::copy_n(first2, count, first1);
```
StephanTLavavej: WG21-N5008 \[tab:char.traits.req\] says "`X::move(s,p,n)` Copies correctly even where the ranges `[p,p+n)` and `[s,s+n)` overlap.", but LWG-3089 "`copy_n` should require non-overlapping ranges".

In practice, the code gets away with this because `copy_n` activates its `_Copy_memmove_n` optimization, but we can see the problem if we use a highly nontrivial element type like `string` (obviously not a valid regex element type). I manually tested a fix:

```
C:\Temp>type meow.cpp
```
```cpp
#include <algorithm>
#include <cstddef>
#include <print>
#include <string>
#include <vector>
using namespace std;

using Elem = string;

Elem* MyMoveAlgorithm(Elem* const result, const Elem* const first2, const size_t count) noexcept /* strengthened */ {
#ifdef FIX
    if (result == first2) {
        // nothing to do
    } else if (first2 <= result && result < first2 + count) {
        copy_backward(first2, first2 + count, result + count);
    } else {
#endif
        copy_n(first2, count, result);
#ifdef FIX
    }
#endif

    return result;
}

int main() {
    const vector<string> orig{"A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R"};

    vector<string> v;

    println("non-overlapping, data copied to right:");
    v = orig;
    println("{}", v);
    MyMoveAlgorithm(v.data() + 10, v.data() + 3, 5);
    println("{}", v);

    println("non-overlapping, data copied to left:");
    v = orig;
    println("{}", v);
    MyMoveAlgorithm(v.data() + 3, v.data() + 10, 5);
    println("{}", v);

    println("overlapping, data copied to right:");
    v = orig;
    println("{}", v);
    MyMoveAlgorithm(v.data() + 5, v.data() + 3, 5);
    println("{}", v);

    println("overlapping, data copied to left:");
    v = orig;
    println("{}", v);
    MyMoveAlgorithm(v.data() + 3, v.data() + 5, 5);
    println("{}", v);

    println("same:");
    v = orig;
    println("{}", v);
    MyMoveAlgorithm(v.data() + 3, v.data() + 3, 5);
    println("{}", v);
}
```
```
meow.cpp
non-overlapping, data copied to right:
["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R"]
["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "D", "E", "F", "G", "H", "P", "Q", "R"]
non-overlapping, data copied to left:
["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R"]
["A", "B", "C", "K", "L", "M", "N", "O", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R"]
overlapping, data copied to right:
["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R"]
["A", "B", "C", "D", "E", "D", "E", "D", "E", "D", "K", "L", "M", "N", "O", "P", "Q", "R"]
overlapping, data copied to left:
["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R"]
["A", "B", "C", "F", "G", "H", "I", "J", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R"]
same:
["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R"]
["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R"]

C:\Temp>cl /EHsc /nologo /W4 /std:c++latest /MTd /Od /DFIX meow.cpp && meow
meow.cpp
non-overlapping, data copied to right:
["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R"]
["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "D", "E", "F", "G", "H", "P", "Q", "R"]
non-overlapping, data copied to left:
["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R"]
["A", "B", "C", "K", "L", "M", "N", "O", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R"]
overlapping, data copied to right:
["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R"]
["A", "B", "C", "D", "E", "D", "E", "F", "G", "H", "K", "L", "M", "N", "O", "P", "Q", "R"]
overlapping, data copied to left:
["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R"]
["A", "B", "C", "F", "G", "H", "I", "J", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R"]
same:
["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R"]
["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R"]
```

tests/std/tests/GH_000995_regex_custom_char_types/test.cpp
```diff
@@ -0,0 +1,262 @@
...
+    template <class FwdIt>
+    string_type transform(FwdIt first, FwdIt last) const {
+        auto str = convert_to_underlying_string<UnderlyingChar>(first, last);
+        return convert_from_underlying_string<string_type>(inner.transform(str.begin(), str.end()));
+    }
+
+    template <class FwdIt>
+    string_type transform_primary(FwdIt first, FwdIt last) const {
+        auto str = convert_to_underlying_string<UnderlyingChar>(first, last);
+        return convert_from_underlying_string<string_type>(inner.transform(str.begin(), str.end()));
```
StephanTLavavej: It looks like a copy-paste mistake that `transform_primary` calls `inner.transform`. (If it was intended, it should be commented.)

StephanTLavavej:
Thanks as always for the extremely detailed writeup and careful changes! :heart_eyes_cat: I pushed a conflict-free merge with `main`, followed by minor nitpicks, a theoretical correctness fix for the test's custom char traits `move`, and a fix for what I believe was a copy-paste typoed function call - please double-check.

StephanTLavavej:

StephanTLavavej:
I had to push an additional commit to disable the test for `/clr:pure`. It's throwing an exception and I don't know or care why.

```
Unhandled Exception: System.Runtime.InteropServices.SEHException: External component has thrown an exception.
   at _CxxThrowException(Void* , _s__ThrowInfo* )
   at std._Xregex_error(error_type _Code)
   at std._Builder2<enum signed_wchar_enum const *,enum signed_wchar_enum,test_regex_traits<enum signed_wchar_enum,wchar_t> >._Add_equiv(_Builder2<enum signed_wchar_enum const \*\,enum signed_wchar_enum\,test_regex_traits<enum signed_wchar_enum\,wchar_t> >* , signed_wchar_enum* A_0, signed_wchar_enum* A_1)
   at std._Parser2<enum signed_wchar_enum const *,enum signed_wchar_enum,test_regex_traits<enum signed_wchar_enum,wchar_t> >._Do_ex_class(_Parser2<enum signed_wchar_enum const \*\,enum signed_wchar_enum\,test_regex_traits<enum signed_wchar_enum\,wchar_t> >* , _Meta_type A_0)
   at std._Parser2<enum signed_wchar_enum const *,enum signed_wchar_enum,test_regex_traits<enum signed_wchar_enum,wchar_t> >._ClassAtom(_Parser2<enum signed_wchar_enum const \*\,enum signed_wchar_enum\,test_regex_traits<enum signed_wchar_enum\,wchar_t> >* , Boolean A_0)
   at std._Parser2<enum signed_wchar_enum const *,enum signed_wchar_enum,test_regex_traits<enum signed_wchar_enum,wchar_t> >._ClassRanges(_Parser2<enum signed_wchar_enum const \*\,enum signed_wchar_enum\,test_regex_traits<enum signed_wchar_enum\,wchar_t> >* )
   at std._Parser2<enum signed_wchar_enum const *,enum signed_wchar_enum,test_regex_traits<enum signed_wchar_enum,wchar_t> >._CharacterClass(_Parser2<enum signed_wchar_enum const \*\,enum signed_wchar_enum\,test_regex_traits<enum signed_wchar_enum\,wchar_t> >* )
   at std._Parser2<enum signed_wchar_enum const *,enum signed_wchar_enum,test_regex_traits<enum signed_wchar_enum,wchar_t> >._Alternative(_Parser2<enum signed_wchar_enum const \*\,enum signed_wchar_enum\,test_regex_traits<enum signed_wchar_enum\,wchar_t> >* )
   at std._Parser2<enum signed_wchar_enum const *,enum signed_wchar_enum,test_regex_traits<enum signed_wchar_enum,wchar_t> >._Disjunction(_Parser2<enum signed_wchar_enum const \*\,enum signed_wchar_enum\,test_regex_traits<enum signed_wchar_enum\,wchar_t> >* )
   at std._Parser2<enum signed_wchar_enum const *,enum signed_wchar_enum,test_regex_traits<enum signed_wchar_enum,wchar_t> >._Compile(_Parser2<enum signed_wchar_enum const \*\,enum signed_wchar_enum\,test_regex_traits<enum signed_wchar_enum\,wchar_t> >* )
   at std.basic_regex<enum signed_wchar_enum,test_regex_traits<enum signed_wchar_enum,wchar_t> >._Reset<enum signed_wchar_enum const *>(basic_regex<enum signed_wchar_enum\,test_regex_traits<enum signed_wchar_enum\,wchar_t> >* , signed_wchar_enum* _First, signed_wchar_enum* _Last, syntax_option_type _Flags)
   at std.basic_regex<enum signed_wchar_enum,test_regex_traits<enum signed_wchar_enum,wchar_t> >.{ctor}<struct std::char_traits<enum signed_wchar_enum>,class std::allocator<enum signed_wchar_enum> >(basic_regex<enum signed_wchar_enum\,test_regex_traits<enum signed_wchar_enum\,wchar_t> >* , basic_string<enum signed_wchar_enum\,std::char_traits<enum signed_wchar_enum>\,std::allocator<enum signed_wchar_enum> >* _Str, syntax_option_type _Flags)
   at test_gh_5592()
   at mainCRTStartup(String[] arguments)
```

muellerj2:
Ah, the test is trying to exert lots of different paths in the parser with custom types to make sure that the parser doesn't unexpectedly crash. The regex in the test includes an equivalence, but I remember now that we had to disable `regex_traits::transform_primary` for /clr:pure because of its reliance on RTTI, so parsing the regex fails for /clr:pure.

If you think it's worth it, I will create a follow-up PR that replaces `transform_primary` in the custom traits class by some dummy implementation that doesn't rely on `regex_traits::transform_primary`.

StephanTLavavej:
Thanks for the explanation. I don't think it's worth the effort - if the impact is limited to the test alone, I don't care about `/clr:pure` coverage at all. (We're keeping that mode in life support for one legacy project.)

StephanTLavavej:
Thanks again for working towards the elimination of this old non-Standard requirement! :heart_eyes_cat: :cat: :cat2:

PR #5625, opened 2025-06-24
`<xloctime>`: Silence CodeQL warning SM03231 about leap year arithmetic

StephanTLavavej:
This is cpp/microsoft/public/leap-year/unchecked-after-arithmetic-year-modification, "Year field changed using an arithmetic operation without checking for leap year".

Only the line with the conditional operator appeared to be emitting this warning, but to be safe I'm restructuring the code to use if-statements consistently (the comments make it clearer anyways), and to suppress the CodeQL warning whenever we're performing arithmetic.

Should prevent a blizzard of internal bugs like OS-57986329 from being created in the future.

### DIFF
--- a/stl/inc/xloctime
+++ b/stl/inc/xloctime
@@ -418,11 +418,13 @@ protected:
         if (!(_Res & ios_base::failbit)) {
             if (_Digits_read <= 2) {
                 if (_Ans < 69) {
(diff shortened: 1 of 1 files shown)


davidmrdavid (approved):

StephanTLavavej:

Issue #5596, opened 2025-06-16
`wmemchr` optimization and its throughput impact


AlexGuteniev:
This include:

https://github.com/microsoft/STL/blob/313964b78a8fd5a52e7965e13781f735bcce13c5/stl/inc/xutility#L15

Is used for `wmemchr` optimization, it was introcuded in 2024.

The `<cwchar>` brings in `<intrin.h>`, which we have always been avoided for throughput reasons. Probably it started bringing it in specifically for `wmemchr` optimization.

What should we do now? I see three opposite paths forward:
 * Drop `#include <cwchar>` and current `wmemchr` optimizations. Almost no impact as `vector_algorithms.cpp` is preferred most of the time anyway.
 * Get SDK fixed to use `<intrin0.h>`.
 * Do nothing, accept throughput impact.

StephanTLavavej:
See #3631:

https://github.com/microsoft/STL/blob/313964b78a8fd5a52e7965e13781f735bcce13c5/stl/inc/limits#L12

I am not super inclined to do anything here. `intrin0.h` was added for the STL, and I am not sure if the UCRT should be using it, but `<intrin.h>` is really quite big. I wonder what intrinsics `<wchar.h>` actually needs.

AlexGuteniev:
> I am not super inclined to do anything here

I don't agree. I think, if we are not willing to remove `wmemchr` optimization to get the throughput back, then we should at least take the full advantage of it, and use it in all paths.

> // TRANSITION, see GH-3631: Lots of user code assumes that <limits> drags in <cwchar>

Still it is `<limits>`, which is heavier and less frequent than `<xutility>`, and it is only been there for about a year in `<xutility>`

> I wonder what intrinsics `<wchar.h>` actually needs

I suspect it needs intrinsics specifically to vectorize `wmemchr` for x86, x64, AR6M64, ARM64EC and nothing much else. 
For x86/x64 it is the spread, load, compare, and move mask for AVX2 and SSE2, SSE2 path additionally has 8-bytes load. It doesn't do AVX2 mask stuff on its AVX2 path.

AlexGuteniev:
I've edited the issue description to match the current situation.

PR #5628, opened 2025-06-27
Restore `wmemchr` optimization in `ranges::find`

AlexGuteniev:
`ranges::find` is not using `wmemchr` after #5284. Yet, `std::find` uses it.

For `wmemchr` we drag in `<cwchar>` which is heavy, specifically due to `wmemchr` optimization. So it seems silly not to take the full advantage of `wmemchr`. The PR addresses that by making `ranges::find` also using `wmemchr`.

This affects ARM64 (non-EC), and builds with vector algorithms disabled.

See also #5596. Not sure if this PR closes that issue, because merging the PR does not preclude doing also Windows SDK throughput optimization or otherwise addressing the throughput in the future.

### DIFF
--- a/stl/inc/xutility
+++ b/stl/inc/xutility
@@ -6266,12 +6266,13 @@ namespace ranges {
     template <input_iterator _It, sentinel_for<_It> _Se, class _Ty, class _Pj = identity>
         requires indirect_binary_predicate<ranges::equal_to, projected<_It, _Pj>, const _Ty*>
     _NODISCARD constexpr _It _Find_unchecked(_It _First, const _Se _Last, const _Ty& _Val, _Pj _Proj = {}) {
-        constexpr bool _Elements_are_1_byte = sizeof(_Iter_value_t<_It>) == 1;
-        constexpr bool _Is_sized            = sized_sentinel_for<_Se, _It>;
+        constexpr bool _Elements_are_1_byte  = sizeof(_Iter_value_t<_It>) == 1;
+        constexpr bool _Elements_are_2_bytes = sizeof(_Iter_value_t<_It>) == 2;
+        constexpr bool _Is_sized             = sized_sentinel_for<_Se, _It>;
 
         if constexpr (_Vector_alg_in_find_is_safe<_It, _Ty>
                       && (_Elements_are_1_byte ? _Is_sized || same_as<_Se, unreachable_sentinel_t>
-                                               : _Is_sized && _USE_STD_VECTOR_ALGORITHMS)
+                                               : _Is_sized && (_Elements_are_2_bytes || _USE_STD_VECTOR_ALGORITHMS))
                       && same_as<_Pj, identity>) {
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
This is restoring it for sized calls only, which seems safe. The problem was with unsized.

stl/inc/xutility
```diff
@@ -6266,12 +6266,13 @@ namespace ranges {
...
-        constexpr bool _Elements_are_1_byte = sizeof(_Iter_value_t<_It>) == 1;
-        constexpr bool _Is_sized            = sized_sentinel_for<_Se, _It>;
+        constexpr bool _Elements_are_1_byte  = sizeof(_Iter_value_t<_It>) == 1;
+        constexpr bool _Elements_are_2_bytes = sizeof(_Iter_value_t<_It>) == 2;
+        constexpr bool _Is_sized             = sized_sentinel_for<_Se, _It>;
 
         if constexpr (_Vector_alg_in_find_is_safe<_It, _Ty>
                       && (_Elements_are_1_byte ? _Is_sized || same_as<_Se, unreachable_sentinel_t>
-                                               : _Is_sized && _USE_STD_VECTOR_ALGORITHMS)
+                                               : _Is_sized && (_Elements_are_2_bytes || _USE_STD_VECTOR_ALGORITHMS))
```
StephanTLavavej: This would be clearer if refactored into "1 byte, 2 byte, more bytes".

stl/inc/xutility
```diff
@@ -6295,24 +6296,36 @@ namespace ranges {
                 } else
```
StephanTLavavej, line 6296: The comment above, `// When _Is_sized && _Elements_are_1_byte, prefer this over memchr() for performance`, needs to be updated.

stl/inc/xutility
```diff
@@ -6295,24 +6296,36 @@ namespace ranges {
...
-                    // C23 7.27.5.2 "The memchr generic function"/2 says "The implementation shall behave as if
-                    // it reads the characters sequentially and stops as soon as a matching character is found."
-                    // C23 7.32.4.6.9 "The wmemchr generic function"/2 lacks such wording,
-                    // so we don't use wmemchr(), avoiding issues with unreachable_sentinel_t.
-                    _Result = static_cast<_Ptr_t>(_CSTD memchr(_First_ptr, static_cast<unsigned char>(_Val), _Count));
-
-                    if constexpr (_Is_sized) {
                         if (_Result == nullptr) {
                             return _RANGES next(_STD move(_First), _Last);
                         }
```
StephanTLavavej: This could be extracted, avoiding the duplication of handling null results from `memchr()/wmemchr()`.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for getting us to the right place here! :dart: :rocket: :smile_cat:

PR #5634, opened 2025-07-01
Update `source_location` tests for upcoming EDG changes, one more time

StephanTLavavej:
Followup to #5551 and #5580.

It turns out that a couple more changes are necessary to accept EDG's new behavior on x86. With these changes, Elizabeth Morrow has been able to actually update our internal copy of EDG (in MSVC-PR-645029), so this will be the final change.

This PR is a reverse mirror of the MSVC-PR, followed by a clang-format whitespace change that I'll need to forward mirror.


StephanTLavavej:

PR #5636, opened 2025-07-02
Cleanup: Mostly remove `_VCRT_ALLOW_INTERNALS`

StephanTLavavej:
Defining `_VCRT_ALLOW_INTERNALS` allows `<vcruntime_internal.h>` to be dragged in:

```cpp
#if !defined _VCRT_BUILD && !defined _VCRT_ALLOW_INTERNALS
    // This is an internal C Runtime header file.  It is used when building the
    // C Runtime only.  It is not to be used as a public header file.
    #error ERROR: Use of C Runtime library internal header file.
#endif
```

The entire STL build (in CMake and MSBuild) defines this macro, but only two translation units need it (for EH machinery). To prevent unintentionally using VCRuntime internals in the future, we should limit the use of the macro to where it's needed.

Ordinarily, I would hiss at defining macros in source files, as opposed to in the build system. However, because this merely guards an `#error` (instead of affecting something ODR-relevant), it's very safe to define this way. Defining per-file macros in the build system(s) is not worth the trouble here.

I've validated this with an internal build.

### DIFF
--- a/stl/CMakeLists.txt
+++ b/stl/CMakeLists.txt
@@ -419,7 +419,7 @@ set(CMAKE_SHARED_LINKER_FLAGS_DEBUG "")
 set(CMAKE_SHARED_LINKER_FLAGS_RELEASE "")
 
 # Toolset options must appear before add_library() for the options to take effect.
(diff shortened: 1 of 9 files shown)


JamesMcNellis (approved):
LGTM 😄

StephanTLavavej:

PR #5637, opened 2025-07-02
Optimize `_Div_ceil<_Signed128>`

### LINKED ISSUE #2948: `<ranges>`: `_Div_ceil` can be implemented more efficiently for `_Signed128` and `_Unsigned128`
The current implementation:

https://github.com/microsoft/STL/blob/main/stl/inc/ranges#L5238-L5245

performs two divisions, when all we need to do is to simply divide and add 1 to the quotient if the remainder is non-zero.

JMazurkiewicz:
* Rename `_Size` template parameter to `_Int`.
* Add extra internal assertions - this function is always called with non-negative values of signed *integer-like* types.
* Add `_Signed128::_Div_ceil` function, which is simply a combination of [`_Base128::_Divide`](https://github.com/microsoft/STL/blob/313964b78a8fd5a52e7965e13781f735bcce13c5/stl/inc/__msvc_int128.hpp#L428-L552) and [`_Base128::_Modulo`](https://github.com/microsoft/STL/blob/313964b78a8fd5a52e7965e13781f735bcce13c5/stl/inc/__msvc_int128.hpp#L554-L683). For better performance, this function does not accept negative arguments.
* Add tests.
* Close #2948.

| benchmark | before | after | speedup |
|--------|--------|--------|--------|
| bm/div_ceil_int128_uint32 | 106,54 | 40,43 | 2,64 |
| bm/div_ceil_int128_uint64 | 103,46 | 40,75 | 2,54 |
| bm/div_ceil_int128_int128 | 109,22 | 43,96 | 2,48 |

### DIFF
--- a/stl/inc/__msvc_int128.hpp
+++ b/stl/inc/__msvc_int128.hpp
@@ -715,7 +715,7 @@ struct alignas(16) _Base128 {
             }
 
             if (_Num._Word[1] == 0) {
-                _Base128 _Result = _Num._Word[0] / _Den._Word[0];
+                uint64_t _Result = _Num._Word[0] / _Den._Word[0]; // with 64-bit inputs, the ceiling is also 64-bit
                 if (_Num._Word[0] % _Den._Word[0] != 0) {
                     ++_Result;
                 }


stl/inc/__msvc_int128.hpp
```diff
@@ -682,6 +682,168 @@ struct
...
+            } else {
+                _Rem = _Num;
+            }
+
+            _Base128 _Result = _Num._Word[1] == 0 ? _Num._Word[0] / _Den._Word[0] : _Num._Word[0] >= _Den._Word[0];
+            if (_Rem != 0) {
+                ++_Result;
+            }
+
+            return _Result;
```
statementreply: I think the code would become cleaner when `_Num._Word[1] == 0` (do 64-bit division) and `_Num._Word[1] != 0` (return 1 or 2) cases are separated.

stl/inc/__msvc_int128.hpp
```diff
@@ -682,6 +682,168 @@ struct
...
+            // 4-digit by 4-digit base 2^32 division
+            _Knuth_4_3_1_D(__u, 5, __v, 4, __q);
+        }
+
+        _Result       = (static_cast<uint64_t>(__q[1]) << 32) | __q[0];
+        _Num._Word[0] = (static_cast<uint64_t>(__u[1]) << 32) | __u[0];
+        _Num._Word[1] = (static_cast<uint64_t>(__u[3]) << 32) | __u[2];
+#endif // _STL_128_INTRINSICS
+
+        _Num >>= __d;
```
statementreply: This right shift is unnecessary. The lower `__d` bits of `_Num` here, which is the remainder of `(uint192)_Num << __d` divided by `_Den << __d` (the original `_Num` and `_Den`), is already zero.

stl/inc/ranges
```diff
@@ -5470,13 +5470,21 @@ namespace ranges {
...
-    template <class _Size>
-    _NODISCARD constexpr _Size _Div_ceil(const _Size _Num, const _Size _Denom) noexcept {
-        _Size _Result = _Num / _Denom;
-        if (_Num % _Denom != 0) {
-            ++_Result;
+    template <class _Int>
+    _NODISCARD constexpr _Int _Div_ceil(const _Int _Num, const _Int _Denom) noexcept {
+        _STL_INTERNAL_STATIC_ASSERT(_Signed_integer_like<_Int>);
+        _STL_INTERNAL_CHECK(_Num >= 0);
+        _STL_INTERNAL_CHECK(_Denom >= 0);
```
statementreply: `_Denom` shouldn't ever be `0`.

stl/inc/__msvc_int128.hpp
```diff
@@ -678,6 +678,169 @@ struct alignas(16) _Base128 {
...
+
+        return _Result;
+    }
+#endif // !_STL_128_DIV_INTRINSICS
+
+    _NODISCARD static constexpr _Base128 _Div_ceil(const _Base128& _Num, const uint64_t _Den) noexcept {
+        _Base128 _Result;
+        _Result._Word[1] = _Num._Word[1] / _Den;
+        uint64_t _Rem;
+        _Result._Word[0] = _UDiv128(_Num._Word[1] % _Den, _Num._Word[0], _Den, _Rem);
```
StephanTLavavej, line 702: Comparing `_Divide` to `_Div_ceil`, this is purely using `_Rem` as an out-param, which I think is clearer, so we should go back and apply this to `_Divide` to reduce source divergence. (This also makes it follow `_Modulo` more closely.)

stl/inc/__msvc_int128.hpp
```diff
@@ -678,6 +678,169 @@ struct alignas(16) _Base128 {
...
+
+                return _Result;
+            }
+
+            _Base128 _Result = _Num._Word[0] >= _Den._Word[0];
+            if (_Num._Word[0] - _Den._Word[0] != 0) {
+                ++_Result;
+            }
+
+            return _Result;
```
StephanTLavavej: This is correct but phrased confusingly and inefficiently, possibly because it started with `_Divide`'s `_Num._Word[0] >= _Den._Word[0]` return value.

Here we have the high words being equal and nonzero. There are 3 cases for comparing the lower words: if the numerator's lower word is smaller, then 0.x gets ceilinged up to 1. If the numerator's lower word is equal, then the result is exactly 1. And if the numerator's lower word is larger, than 1.x gets ceilinged up to 2.

So there are actually only two possible return values (1 versus 2), but the current logic performs two tests for what initially looks like four cases, and a general 128-bit increment.

We should replace this with a single test and a return of 1 versus 2.

stl/inc/__msvc_int128.hpp
```diff
@@ -678,6 +678,169 @@ struct alignas(16) _Base128 {
...
+                return static_cast<_Base128>(_Num != 0);
+            }
+
+            if (_Num._Word[1] == 0) {
+                _Base128 _Result = _Num._Word[0] / _Den._Word[0];
+                if (_Num._Word[0] % _Den._Word[0] != 0) {
+                    ++_Result;
+                }
+
+                return _Result;
```
StephanTLavavej: This is slightly inefficient. Here, we have the high words being equal and zero, so we can just look at the low words.

We're performing a general 128-bit increment for the ceiling, but that's not necessary. The largest possible result of the 64-bit division is 2^64 - 1, when the numerator is maximum and the denominator is 1. But in that case, maximum modulo 1 is 0, so we don't need to increment.

Thus we should do this work in 64-bit, and only at the end widen to 128-bit.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: I pushed a few simplifications/optimizations, please double-check.

StephanTLavavej:

StephanTLavavej:
Thanks for mathing faster! :abacus: :rocket: :heart_eyes_cat:

PR #5640, opened 2025-07-09
Optimize core loop of `_Count_vbool`

localspook:
To handle the difference between counting ones and counting zeros, `_Count_vbool`'s core loop conditionally complements every block it processes: that's O(n) work. We can do better: we can unconditionally count ones, then, after the loop, if we actually wanted zeros, subtract the number of ones from the number of bits we processed: O(1) work.

### DIFF
--- a/stl/inc/vector
+++ b/stl/inc/vector
@@ -3790,7 +3790,7 @@ _NODISCARD _CONSTEXPR20 _Iter_diff_t<_VbIt> _Count_vbool(_VbIt _First, const _Vb
             _Iter_diff_t<_VbIt> _Count = _Popcount_impl(_FirstVal);
             ++_VbFirst;
 
-            const _Iter_diff_t<_VbIt> _Bits = (_VbLast - _VbFirst) * _VBITS;
+            const auto _Bits                = static_cast<_Iter_diff_t<_VbIt>>(_VbLast - _VbFirst) * _VBITS;
             _Iter_diff_t<_VbIt> _Count_ones = 0;
 
             for (; _VbFirst != _VbLast; ++_VbFirst) {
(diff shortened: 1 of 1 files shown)


localspook:
All CI jobs are failing with "The agent did not connect within the alloted time of 45 minute(s)." 🤔

StephanTLavavej:
If that happens, ping me on `#stalled-checks` on the [STL Discord](https://discord.gg/XWanNww) and I can rerun the checks. Unfortunately, while our CI is highly reliable when it runs, we're at the mercy of Azure deciding to give us any VMs, and it's been rough lately.

AlexGuteniev:
Can you post benchmark results before / after?
(I guess there is such benchmark already)

AlexGuteniev:
I mean intuitively it looks like an improvement, but without benchmarking I wouldn't be sure that the compiler doesn't do something clever on its own. 

This PR optimization is not what compiler is likely to do, but it might be able to achieve the same by duplication the loop for `*_VbFirst` and `~*_VbFirst` and making the condition out of loop.

I don't think it is very likely to happen, and that's why we implemented the `_Select_popcount_impl` thing, but would be good to prove with measurement.

StephanTLavavej:
I took a look at the code and it looked good to me, but I would also like to see benchmark results before slightly complicating the code in this way.

StephanTLavavej:
@localspook Do you have time to add a benchmark here?

Alternatively I would accept an independent PR from another contributor adding a benchmark.

Apologies for how long this has taken, I have been severely overloaded, otherwise I would have added a benchmark myself.

StephanTLavavej:

StephanTLavavej:
Thanks for noticing and improving this performance! :rocket: :smile_cat: :tada:

Issue #5651, opened 2025-07-23
cmake build: missing /CETCOMPAT linker flag for x86/x64


matbech:
Compared to the msvcp140*.dll files which come with Visual Studio, the cmake build is missing the /CETCOMPAT linker flag for the x64 and x86 builds.

StephanTLavavej:
Yep. Strongly related to #5646.

I think we're missing a significant amount of stuff here. Internal `src/vctools/VCTools.DevDiv.Cpp.props` has, among other things:

```xml

  <!-- Compile all the libraries with /guard:ehcont option for amd64 and arm64 architecture. VSO #1073832 and #1513316 -->
  <PropertyGroup Condition="'$(BuildArchitecture)' == 'amd64' or '$(BuildArchitecture)' == 'arm64' or '$(BuildArchitecture)' == 'arm64ec'">
    <ClAdditionalOptions>$(ClAdditionalOptions) /d2guardehcont</ClAdditionalOptions>
    <AssembleAmd64Flags>$(AssembleAmd64Flags) /guard:ehcont</AssembleAmd64Flags>
    <AssembleArm64Flags>$(AssembleArm64Flags) /guard:ehcont</AssembleArm64Flags>
  </PropertyGroup>

  <!-- Link binaries with /guard:ehcont option for amd64 and arm64 architecture only if requested. -->
  <PropertyGroup Condition="'$(GuardEHContMetadata)' == 'true' and ('$(BuildArchitecture)' == 'amd64' or '$(BuildArchitecture)' == 'arm64' or '$(BuildArchitecture)' == 'arm64ec')">
    <LinkAdditionalOptions>$(LinkAdditionalOptions) /guard:ehcont /guard:cf</LinkAdditionalOptions>
  </PropertyGroup>

  <!-- Link binaries for CET if ehcont is enabled. -->
  <PropertyGroup Condition="'$(GuardEHContMetadata)' == 'true' and '$(BuildArchitecture)' == 'amd64'">
    <LinkAdditionalOptions>$(LinkAdditionalOptions) /CETCOMPAT</LinkAdditionalOptions>
  </PropertyGroup>
```

I think we need a detailed audit to figure out what missing options we've accumulated.

Issue #5646, opened 2025-07-15
cmake build: missing /guard:ehcont compiler flag


matbech:
Compared to the internal stl build (msbuild), the cmake build is missing the following compiler option for 64-bit builds: /guard:ehcont

To reproduce, use cmake to build the x64 target and then link to the .lib (shared dll). This results in the following warning:
`msvcprt.lib(vector_algorithms.cpp.obj) : warning LNK4291: module may contain '__except' (Structured Exception Handling) but was not compiled with /guard:ehcont; generating conservative metadata`

When linking to the msvcprt.lib library provided by Visual Studio, there is no warning.

StephanTLavavej:
Yep, great catch. This is controlled by internal `src/vctools/crt/crt_build.settings.targets` (outside of the STL's sources which are mirrored in `src/vctools/crt/github`):

```xml
<!-- Compile all the libraries with /guard:ehcont option for amd64 architecture. VSO #1073832 -->
<GuardEhContMetadata>true</GuardEhContMetadata>
```

As the comment indicates, this was VSO-1073832 resolved by MSVC-PR-241996 merged on 2020-04-29, after we created the CMake build system for the STL in 2019, which is why we missed this.

PR #5652, opened 2025-07-23
cmake build: Adds /CETCOMPAT for x86/x64

matbech:
Fixes https://github.com/microsoft/STL/issues/5651


stl/CMakeLists.txt
```diff
@@ -415,6 +415,9 @@ set(CMAKE_STATIC_LINKER_FLAGS "/WX")
 set(CMAKE_STATIC_LINKER_FLAGS_DEBUG "")
 set(CMAKE_STATIC_LINKER_FLAGS_RELEASE "")
 set(CMAKE_SHARED_LINKER_FLAGS "/DEBUG:FULL /WX /RELEASE /SUBSYSTEM:Console /NODEFAULTLIB /INCREMENTAL:NO /MANIFEST:NO /DLL /profile /guard:cf /DEBUGTYPE:cv,fixup /LARGEADDRESSAWARE")
+if(VCLIBS_TARGET_ARCHITECTURE MATCHES "^(x86|x64)$")
```
StephanTLavavej: Thanks! However, I am not sure that this is correct (I see x86 mentioned here, but not in the internal MSBuild machinery) or sufficient (this is apparently related to `/guard:ehcont`, see #5646 and https://github.com/microsoft/STL/issues/5651#issuecomment-3109475789).

I think we need to compare the internal MSBuild machinery to the options that the GitHub CMake build is emitting, to figure out what has been added (by the compiler backend team) to the central MSVC options, that wasn't picked up during our original conversion of MSBuild to CMake. Unfortunately this central MSVC stuff is in internal files that you can't see (the `stl/msbuild` subdirectory here is the other half of that internal machinery, but it's STL-specific; there is more machinery that applies to the entire MSVC toolset including the compiler and other libraries).

matbech: You are correct. I have inspected the x86 msvcp140*.dll files and they do not have the /CETCOMPAT linker flag set. I have pushed the change to match the internal msbuild.

StephanTLavavej: Thanks! I think I'm going to need to do that audit of the internal MSBuild machinery to figure out what other options we're missing, so I'm moving this to Work In Progress and I'll try to get it to later. Sorry about that - I always prefer to accept simple PRs that are strict improvements, but in this case I think we need to fix the divergence systematically.

stl/CMakeLists.txt
```diff
@@ -415,6 +415,9 @@ set(CMAKE_STATIC_LINKER_FLAGS "/WX")
 set(CMAKE_STATIC_LINKER_FLAGS_DEBUG "")
 set(CMAKE_STATIC_LINKER_FLAGS_RELEASE "")
 set(CMAKE_SHARED_LINKER_FLAGS "/DEBUG:FULL /WX /RELEASE /SUBSYSTEM:Console /NODEFAULTLIB /INCREMENTAL:NO /MANIFEST:NO /DLL /profile /guard:cf /DEBUGTYPE:cv,fixup /LARGEADDRESSAWARE")
+if(VCLIBS_TARGET_ARCHITECTURE MATCHES "x64")
```
StephanTLavavej: `MATCHES` is a regex match, but we should use the more specific `STREQUAL` here.

StephanTLavavej (approved):

StephanTLavavej:
My apologies for how long it took to get around to this. I audited the MSBuild machinery and extracted what I believe are the most important missing options. I verified both the GitHub and MSVC-internal builds, checking that the expected options are present or not for all architectures.

There are probably a few more missing flags but I believe this is a strict improvement.

StephanTLavavej:

StephanTLavavej:
Thanks for noticing and fixing this very subtle build divergence! :eagle: :eye: :heart_eyes_cat:

PR #5656, opened 2025-07-26
`VSO_0000000_instantiate_iterators_misc`: Suppress deprecated volatile operation warnings

YexuanXiao:
At line 1596 of this file, a `volatile tuple<int, int, int>` is used, but `utility` L217 marks it as deprecated, causing the test to fail. This PR uses `#define _SILENCE_CXX20_VOLATILE_DEPRECATION_WARNING` to suppress the warning and allow the test to pass.

https://github.com/microsoft/STL/blob/f015db5833e8daf3a5534c4ce2b6c9acc9689b0c/tests/std/tests/VSO_0000000_instantiate_iterators_misc/test.compile.pass.cpp#L1596

https://github.com/microsoft/STL/blob/f015db5833e8daf3a5534c4ce2b6c9acc9689b0c/stl/inc/utility#L616-L617


StephanTLavavej (approved):

StephanTLavavej:
Interesting - why is this passing with our current compilers? Are newer versions of Clang noticing this?

Regardless, change looks like a slam dunk. :bouncing_ball_person:

YexuanXiao:
> Interesting - why is this passing with our current compilers? Are newer versions of Clang noticing this?
> 
> Regardless, change looks like a slam dunk. ⛹️

I believe there is a bug in MSVC where using a static member of a deprecated struct does not generate a warning, see [Godbolt](https://godbolt.org/z/1s1srEsq9).

YexuanXiao:
I reported this issue on [DevCom-10946291](https://developercommunity.visualstudio.com/t/The-deprecated-attribute-of-a-class/10946291). Before MSVC fixes it, the `_CXX20_DEPRECATE_VOLATILE` macro had no effect.

YexuanXiao:
Bad news: MSVC can generate warnings, but enabling /w14996 or /W3 manually is required. By default, MSVC does not respect `[[deprecated]]`.

StephanTLavavej:
Oh, I expect /W3 as an absolute minimum of effort from users (I also believe it's the IDE default). /W1 and /W2 are jokes.

YexuanXiao:
I found that even with /W3, MSVC doesn't respect the `[[deprecated]]` attribute of **template specializations**.

```cpp
template <typename T>
struct foo_trait
{
    static inline constexpr bool value = false;
};

template <typename T>
struct [[deprecated]] foo_trait<const T>
{
    static inline constexpr bool value = true;
};

int main()
{
    static_assert(foo_trait<const int>::value);
}

```

I re-reported it as [DevCom-10948908](https://developercommunity.visualstudio.com/t/10948908).

StephanTLavavej:

StephanTLavavej:
:warning: :cat2: :warning:
===

PR #5653, opened 2025-07-26
`<xlocale>`: Use `static_cast` to eliminate implicit conversions between different `charN_t` types

YexuanXiao:
The merged PR [#138708](https://github.com/llvm/llvm-project/pull/138708) in Clang will generate warnings for implicit conversions between different `charN_t` types. This PR eliminates the implicit conversions that would produce warnings in `<xlocale>`. Some tests will fail due to this reason, such as GH_002769_handle_deque_block_pointers (though not all).

PR #5657, opened 2025-07-26
`VSO_0180466_algorithm_overhauls`: Suppress `-Wcharacter-conversion` for Clang 21

YexuanXiao:
Lines 35-84 of this test involve comparisons between different `charN_t` types, so similar to #5653, Clang will also generate warnings for this.

https://github.com/microsoft/STL/blob/f015db5833e8daf3a5534c4ce2b6c9acc9689b0c/tests/std/tests/VSO_0180466_algorithm_overhauls/test.cpp#L35-L84

### DIFF
--- a/tests/std/tests/VSO_0180466_algorithm_overhauls/test.cpp
+++ b/tests/std/tests/VSO_0180466_algorithm_overhauls/test.cpp
@@ -6,8 +6,10 @@
 #pragma warning(disable : 4389) // signed/unsigned mismatch in arithmetic
 
 #ifdef __clang__
-#pragma clang diagnostic ignored "-Wsign-compare"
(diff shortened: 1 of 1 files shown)


tests/std/tests/VSO_0180466_algorithm_overhauls/test.cpp
```diff
@@ -7,6 +7,7 @@
 
 #ifdef __clang__
 #pragma clang diagnostic ignored "-Wsign-compare"
+#pragma clang diagnostic ignored "-Wcharacter-conversion"
```
StephanTLavavej, line 10: We need to guard this for Clang 21. Sorted order would also be nice.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:warning: :zipper_mouth_face: :warning:
===

Issue #2794, opened 2022-06-16
LLVM lit sharding is non-deterministic


StephanTLavavej:
Repros with `main` as of ea32e86deed6e8a8cd6116dc275e358a901d2b50, where our submodule is https://github.com/llvm/llvm-project/commit/b8d38e8b4fcab071c5c4cb698e154023d06de69e.

We use LLVM lit's "sharding" feature to split our large test suite across 8 VMs per architecture:

https://github.com/microsoft/STL/blob/ea32e86deed6e8a8cd6116dc275e358a901d2b50/azure-devops/native-build-test.yml#L21

We expected that for a given set of tests, this would exactly partition them across the VMs, with no duplicates and no missed tests. That is, we expected the sharding algorithm to be *deterministic*, even across different machines (because these VMs run independently), as long as the set of tests is the same on each machine (which it is, because they've checked out the same commit). There should be no sensitivity to filesystem enumeration order, time of day, or anything else. (However, it's okay if adding/removing a single test radically changes the subset shards, and it's okay for each subset shard to be run in a totally randomized/shuffled order.)

We're observing non-deterministic behavior from lit. This was originally observed in #2793, whose initial version mistakenly had 2 tests that were XPASSing - but the XPASSes showed up only for x86, when nothing was architecture-sensitive. Looking at x64, we didn't see the affected tests running in *any* of the 8 shards, even though other tests from that subdirectory ran.

Eventually, we found that this repros locally! No machine variation is needed - simply two consecutive runs on the same machine, at the same commit.

<details>
<summary>Click to expand example:</summary>

```
D:\GitHub\STL\out\build\x64>python tests\utils\stl-lit\stl-lit.py ..\..\..\llvm-project\libcxx\test\std\language.support\support.limits\support.limits.general --num-shards=8 --run-shard=1 -o testing_x64.log
stl-lit.py: D:\GitHub\STL\llvm-project\llvm\utils\lit\lit\main.py:193: note: Selecting shard 1/8 = size 15/114 = tests #(8*k)+1 = [1, 9, 17, ...]
-- Testing: 15 of 114 tests, 15 workers --
XFAIL: libc++ :: std/language.support/support.limits/support.limits.general/version.version.pass.cpp:0 (1 of 15)
XFAIL: libc++ :: std/language.support/support.limits/support.limits.general/tuple.version.pass.cpp:0 (2 of 15)
[...]

PR #1414, opened 2020-11-02
Accelerate tests by dividing them into shards

cbezault:

PR #5670, opened 2025-08-08
Tests for P1831 and workaround for DevCom-10948908

YexuanXiao:
Even with /W3, MSVC doesn't respect the `[[deprecated]]` attribute of template specializations ([DevCom-10948908](https://developercommunity.visualstudio.com/t/10948908)). 
Therefore, the macro `_CXX20_DEPRECATE_VOLATILE` has no effect with `tuple_size`, `tuple_elements`, `variant_size` and `variant_alternatives`.

Due to another bug of MSVC, applying the `[[deprecated]]` attribute directly to `type` still has no effect, so it can only be applied to `_Mybase`. Since my attempt to reproduce it failed, I cannot report it as another bug. When DevCom-10948908 is fixed, this bug may still persist, but all two workaround can be removed.

### DIFF
--- a/stl/inc/utility
+++ b/stl/inc/utility
@@ -616,7 +616,7 @@ struct tuple_size<const _Tuple> : _Tuple_size_sfinae<_Tuple> {}; // ignore cv
 // TRANSITION, DevCom-10948908, Clang fixed in 21: the [[deprecated]] attribute of a class template specialization
 // has no effect
 // See the specialization of tuple_size for volatile _Tuple
-#if 1 // vvv workaround vvv
+#if _HAS_CXX20 // vvv workaround vvv
 template <class _Tuple>
 struct _Tuple_size_sfinae<volatile _Tuple, void_t<decltype(tuple_size<_Tuple>::value)>>
     : integral_constant<size_t, tuple_size<_Tuple>::value> {
@@ -655,7 +655,7 @@ struct _CXX20_DEPRECATE_VOLATILE _MSVC_KNOWN_SEMANTICS tuple_element<_Index, vol
     : tuple_element<_Index, _Tuple> {
     // TRANSITION, DevCom-10948908, Clang fixed in 21: the [[deprecated]] attribute of a class template specialization
     // has no effect
-#if 1
+#if _HAS_CXX20 // vvv workaround vvv
     using _Mybase _CXX20_DEPRECATE_VOLATILE = tuple_element<_Index, _Tuple>;
 #else // ^^^ workaround / no workaround vvv
     using _Mybase = tuple_element<_Index, _Tuple>;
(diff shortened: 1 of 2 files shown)


YexuanXiao:
Apart from VSO_0000000_instantiate_iterators_misc (as expected), a few tr1 tests failed. I'll fix them later.

YexuanXiao:
This issue also exists in Clang19 and Clang20, but it will be fixed in Clang21. See [commit](https://github.com/llvm/llvm-project/commit/abd10578653a4f16c81c6a164f6367af64f21194).

YexuanXiao:
@StephanTLavavej What is the difference between checks numbered 3, 4, and others in Build and Test x64 x64? They unexpectedly passed some new tests.

MattStephanson:
>What is the difference between checks numbered 3, 4, and others in Build and Test x64 x64? They unexpectedly passed some new tests.

#1414 originally introduced this feature, but #2794 concisely explains it:
>We use LLVM lit's "sharding" feature to split our large test suite across 8 VMs per architecture

If you go to the pipeline run webpage and click on the [results tab](https://dev.azure.com/vclibs/STL/_build/results?buildId=19135&view=ms.vss-test-web.build-test-results-tab), you'll see a list of the failed tests.

YexuanXiao:
> > What is the difference between checks numbered 3, 4, and others in Build and Test x64 x64? They unexpectedly passed some new tests.
> 
> #1414 originally introduced this feature, but #2794 concisely explains it:
> 
> > We use LLVM lit's "sharding" feature to split our large test suite across 8 VMs per architecture
> 
> If you go to the pipeline run webpage and click on the [results tab](https://dev.azure.com/vclibs/STL/_build/results?buildId=19135&view=ms.vss-test-web.build-test-results-tab), you'll see a list of the failed tests.

Based on your explanation, I believe I have found a clue to reproduce the issue. Thanks!

YexuanXiao:
The workaround for `tuple_element` and `variant_alternative` does not work for Clang 19, even when applying the `[[deprecated]]` attribute directly to `type` instead of `_Mybase`. It's as tricky to simplify to reproduce as the MSVC bug. I can't figure out how to implement a workaround for it.

YexuanXiao:
The MSVC team believes [a CWG issue is needed](https://developercommunity.visualstudio.com/t/MSVC-does-not-respect-the-deprecated/10948908#T-ND10951088), so it may not be fixed in the short term. How should STL adapt to this situation? @StephanTLavavej

StephanTLavavej:
@YexuanXiao Thanks for looking into this. Because MSVC's behavior is unlikely to change in the moderate future and the impact of missing deprecation warnings for volatile machinery is highly limited, I believe I would prefer to abandon this attempt. I appreciate the effort you've put into it, but at this point I don't think making changes here is worth the complexity, risk, and time.

PR #5672, opened 2025-08-10
`<regex>`: Speed up skip optimization for default `regex_traits` in `collate` mode

muellerj2:
In the default `std::regex_traits<char>` and `std::regex_traits<wchar_t>` implementation, the `translate()` function just returns its argument unchanged ([see \[re.traits\]/4](https://eel.is/c++draft/re.traits#4)). This means that the call to this function can be skipped for these traits classes.

This PR implements this change, and as a consequence, the skip optimization now engages the vectorized `std::search` implementation in `collate` mode as well.

The tests added by #5553 already check correct handling of `translate()` in the skip optimization for user-provided regex traits, so this PR just adds a short variant of this test for default `std::regex` and `std::wregex`.

Benchmark (only relevant lines):

name | before | after | speedup
-- | -- | -- | --
bm_lorem_search/"bibe"/2 | 2825 ns | 2915 ns | 0.97
bm_lorem_search/"bibe"/3 | 5625 ns | 5580 ns | 1.01
bm_lorem_search/"bibe"/4 | 10986 ns | 11160 ns | 0.98
bm_lorem_search/"bibe".collate/2 | 15346 ns | 2915 ns | 5.26
bm_lorem_search/"bibe".collate/3 | 30482 ns| 5625 ns | 5.42
bm_lorem_search/"bibe".collate/4 | 61384 ns | 11718 ns | 5.24

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -3693,6 +3693,8 @@ _BidIt1 _Compare_translate_both(_BidIt1 _Begin1, _BidIt1 _End1, _BidIt2 _Begin2,
     // compare character ranges, translating characters in both ranges according to syntax options
     if (_Sflags & regex_constants::icase) {
         return _STD _Cmp_chrange(_Begin1, _End1, _Begin2, _End2, _Cmp_icase<_RxTraits>{_Traits});
+    } else if constexpr (is_same_v<_RxTraits, regex_traits<char>> || is_same_v<_RxTraits, regex_traits<wchar_t>>) {
+        return _STD _Cmp_chrange(_Begin1, _End1, _Begin2, _End2, equal_to<typename _RxTraits::char_type>{});
     } else if (_Sflags & regex_constants::collate) {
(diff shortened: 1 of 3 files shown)


stl/inc/regex
```diff
@@ -3693,6 +3693,8 @@ _BidIt1 _Compare_translate_both(_BidIt1 _Begin1, _BidIt1 _End1, _BidIt2 _Begin2,
     // compare character ranges, translating characters in both ranges according to syntax options
     if (_Sflags & regex_constants::icase) {
         return _STD _Cmp_chrange(_Begin1, _End1, _Begin2, _End2, _Cmp_icase<_RxTraits>{_Traits});
+    } else if constexpr (is_same_v<_RxTraits, regex_traits<char>> || is_same_v<_RxTraits, regex_traits<wchar_t>>) {
```
StephanTLavavej: Suggestion: `_Is_any_of_v` would make this slightly more readable.

StephanTLavavej:

StephanTLavavej:
:rocket: :rocket: :rocket:
===

PR #5675, opened 2025-08-13
`<regex>`: Compare against `char` directly

### LINKED ISSUE #5673: `<regex>`: After `_Uelem` changes, Build2 asserts in the Real World Code test suite
After merging #5592, our test team reported VSO-2548012 "\[RWC\]\[prod/fe\]\[Regression\] Build2 test failed with Assertion failed: `type () == line_type::special`, file C:\gitP\Build2\build2-toolchain-0.17.0\build2\libbuild2/script/regex.hxx, line 208". The affected code is:

https://github.com/build2/build2/blob/05d68f47aa044da9161d9bfabdd153602169d902/libbuild2/script/regex.hxx#L200-L210
```cpp
        // Provide basic_regex (such as from msvcrt) with the ability to
        // explicitly cast line_chars to implementation-specific numeric
        // types (enums, msvcrt's _Uelem, etc).
        //
        template <typename T>
        explicit
        operator T () const
        {
          assert (type () == line_type::special);
          return static_cast<T> (special ());
        }
```

I haven't yet attempted to repro this myself. I don't know if it's a bug in Build2 (i.e. a non-Standard assumption that we've broken), or a bug in #5592. I also don't know if this is further affected by the currently open #5671.

FYI @muellerj2.

StephanTLavavej:
Fixes #5673 and fixes VSO-2548012 in conjunction with small Build2 changes.

Build2 is doing something extremely complicated and fragile with user-defined character types and `regex`. When we construct user-defined `_Elem` from ordinary `char`s, that triggers assertions in their machinery, even though the resulting homogeneous comparisons between `_Elem` and `_Elem` are arguably more cromulent.

This PR slightly backs off from the purity of such comparisons, resulting in code that's somewhat closer to our old way of doing things, but still avoiding the internal `_Uelem` machinery. Now we directly compare `_Elem` to `char`. Build2 has heterogeneous comparison operators that handle this.

It will still be necessary to update Build2 to round-trip their user-defined character type instead of asserting, but that's a less invasive change: https://github.com/build2/build2/pull/478

This makes no difference to the vast, vast majority of users who simply use `char` or `wchar_t`.

### DIFF
--- a/tests/std/tests/GH_000995_regex_custom_char_types/test.cpp
+++ b/tests/std/tests/GH_000995_regex_custom_char_types/test.cpp
@@ -104,21 +104,21 @@ private:
 namespace signed_wchar_ns {
     enum class signed_wchar_enum : short {};
 
-    bool operator==(const signed_wchar_enum left, const char right) {
-        return static_cast<short>(left) == right;
+    bool operator==(const signed_wchar_enum swe, const char ch) {
+        return static_cast<unsigned short>(swe) == static_cast<unsigned char>(ch);
     }
 
 #if !_HAS_CXX20
-    bool operator!=(const signed_wchar_enum left, const char right) {
-        return static_cast<short>(left) != right;
+    bool operator!=(const signed_wchar_enum swe, const char ch) {
+        return !(swe == ch);
     }
 
-    bool operator==(const char left, const signed_wchar_enum right) {
-        return left == static_cast<short>(right);
+    bool operator==(const char ch, const signed_wchar_enum swe) {
+        return swe == ch;
     }
 
-    bool operator!=(const char left, const signed_wchar_enum right) {
-        return left != static_cast<short>(right);
+    bool operator!=(const char ch, const signed_wchar_enum swe) {
+        return !(swe == ch);
     }
 #endif // !_HAS_CXX20
 } // namespace signed_wchar_ns
--- a/tests/std/tests/GH_000995_regex_custom_char_types/test.cpp
+++ b/tests/std/tests/GH_000995_regex_custom_char_types/test.cpp
@@ -101,7 +101,29 @@ private:
     rx_traits inner;
 };
 
-enum class signed_wchar_enum : short {};
+namespace signed_wchar_ns {
+    enum class signed_wchar_enum : short {};
+
+    bool operator==(const signed_wchar_enum left, const char right) {
+        return static_cast<short>(left) == right;
+    }
+
+#if !_HAS_CXX20
+    bool operator!=(const signed_wchar_enum left, const char right) {
+        return static_cast<short>(left) != right;
+    }
+
+    bool operator==(const char left, const signed_wchar_enum right) {
+        return left == static_cast<short>(right);
+    }
+
+    bool operator!=(const char left, const signed_wchar_enum right) {
+        return left != static_cast<short>(right);
+    }
+#endif // !_HAS_CXX20
+} // namespace signed_wchar_ns
+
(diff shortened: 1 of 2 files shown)


tests/std/tests/GH_000995_regex_custom_char_types/test.cpp
```diff
@@ -101,7 +101,29 @@ class test_regex_traits {
...
+    bool operator!=(const signed_wchar_enum left, const char right) {
+        return static_cast<short>(left) != right;
+    }
+
+    bool operator==(const char left, const signed_wchar_enum right) {
+        return left == static_cast<short>(right);
+    }
+
+    bool operator!=(const char left, const signed_wchar_enum right) {
+        return left != static_cast<short>(right);
```
davidmrdavid: just curious - any reason not to define this based on `operator==`. As in

```C++
    return !(left == right);
```

tests/std/tests/GH_000995_regex_custom_char_types/test.cpp
```diff
@@ -101,7 +101,29 @@ class test_regex_traits {
     rx_traits inner;
 };
 
-enum class signed_wchar_enum : short {};
+namespace signed_wchar_ns {
```
davidmrdavid: for my own learning - `_ns` here stands for namespace, right?

StephanTLavavej: Yes.

tests/std/tests/GH_000995_regex_custom_char_types/test.cpp
StephanTLavavej: More or less stylistic. The direct definition is better in terms of debug codegen, not that it matters in a test. For anything complicated (e.g. an `operator<`), defining one in terms of another is definitely more robust.

tests/std/tests/GH_000995_regex_custom_char_types/test.cpp
```diff
@@ -101,7 +101,29 @@ class test_regex_traits {
     rx_traits inner;
 };
 
-enum class signed_wchar_enum : short {};
+namespace signed_wchar_ns {
+    enum class signed_wchar_enum : short {};
+
+    bool operator==(const signed_wchar_enum left, const char right) {
+        return static_cast<short>(left) == right;
```
muellerj2: It would actually be more correct to do the following casts in this operator:

```
return static_cast<unsigned short>(left) == static_cast<unsigned char>(right);
```

But it doesn't really matter much for this test, since the parser only compares positive `char`s.

StephanTLavavej: I can do that.

StephanTLavavej:
@davidmrdavid @muellerj2 I've pushed changes to the test, improving the casts and centralizing the operators. Thanks for the reviews! :heart_eyes_cat:

muellerj2 (approved):

StephanTLavavej:

StephanTLavavej:
I've pushed a non-trivial merge with `main`, reverse-mirroring what I did in MSVC-PR-662263.

This fixes the conflict in `GH_000995_regex_custom_char_types` with #5671 (which had moved and altered code in the test). Now, `signed_wchar_enum` and `ullong_enum` are gaining comparison operators through the same technique, `wrapped_character`'s operators have been adjusted now that it's templated on `Elem`, and they're all implemented in terms of each other. (I suppose I was ultimately persuaded that it's the simpler thing to do, thanks for the review comment.)

PR #5676, opened 2025-08-13
`P0811R3_midpoint_lerp`: Suppress warning C4756

StephanTLavavej:
This mirrors the STL part of the compiler backend MSVC-PR-661055 "Inlining Related Fixes" by @manbearian:

> Disable warning about overflow in constant math that shows up if the inliner does more aggressive inlining.

Although this MSVC-PR is targeted at internal `prod/be`, it's safe to mirror this suppression to GitHub and our usual branch `prod/fe`, where they'll merge harmlessly in MSVC `main`.

### DIFF
--- a/tests/std/tests/P0811R3_midpoint_lerp/test.cpp
+++ b/tests/std/tests/P0811R3_midpoint_lerp/test.cpp
@@ -1031,6 +1031,8 @@ bool test_lerp() {
     return true;
 }
 
+#pragma warning(push)
+#pragma warning(disable : 4756) // ignore constant arithmetic overflow warning
 void test_gh_1917() {
     // GH-1917 <cmath>: lerp(1e+308, 5e+307, 4.0) spuriously overflows
     using bit_type       = unsigned long long;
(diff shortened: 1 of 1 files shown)


tests/std/tests/P0811R3_midpoint_lerp/test.cpp
```diff
@@ -1031,6 +1031,8 @@ bool test_lerp() {
     return true;
 }
 
+#pragma warning(push)
+#pragma warning(disable : 4756) // ignore constant arithmetic overflow warning
```
davidmrdavid, line 1035: if the optimizer is triggering this warning, doesn't that mean this is a faulty warning, or a faulty optimizer?

davidmrdavid (commented):
Just a question

tests/std/tests/P0811R3_midpoint_lerp/test.cpp
manbearian: Neither. it means the code is doing dumb stuff and the optimizer is smart enough to discover this.

You can see below that this test is specifically testing a case that has overflow in it.
The compiler detects this only if it does enough inlining to find the overflow.

davidmrdavid: Ah, the age old trick of looking further down the diff. Thanks, this checks out.

StephanTLavavej:

PR #5590, opened 2025-06-15
(Semi-)vectorize `includes`

AlexGuteniev:
# ❓ Draft PR
Not sure if this is good, because it optimizes best cases, whereas pessimizing some worse cases.
Unlike `std::search` vectorization:
 * The set of worse cases seems larger.
 * For `std::search` there's intuitive understanding of an "usual" case: it is "usual" language text search. The "usual" case for `includes` is not known.

I'm not implementing the following for a draft PR, would like the approach to be approved or rejected first:
 * ➕ unsigned types (matter of few adjustments, although this means that unsigned will be less efficient)
 * ✔️ test coverage (the benchmark gives some minimal coverage by checking the expected result for every case)

# ½ Half-vectorization approach

 * Iterate haystack by vector-sized units
 * Skip the whole vector, if all elements are less than the current needle element
 * Skip the vector part is some elements are less than the current needle element
 * Proceed the rest of the current vector in a scalar way

# 😾 Which full vectorization approaches considered and failed

Generally, any of these aproaches result in major slowdown for the following case:
 * Hay: `0,0,1,1,2,2,...999,999`
 * Needle: `0,1,2,...999`

The approaches considered:
* Compare the needle for equality using vector comparison, skip all equal elements. This ends up being close to scalar but using inefficient instructions, when the needle is large and spread nearly equally
* Use `*cmpestr*` in any-to-any match mode. Could possibly have worked if there were no duplicates to consider
* Use `remove`/`unique` approach in a loop. When the needle is large and spread nearly equally, it takes up to half vector size iterations, which is very slow

# ⏱️ Benchmark results

Benchmark                                              |    Before    |   After    | Speedup
-------------------------------------------------------|--------------|------------|----------
bm_includes<int8_t, alg_type::std_fn>/3000/3/0/1       |    370 ns    |   36.8 ns  |   10.05
bm_includes<int8_t, alg_type::std_fn>/3000/22/0/1      |    586 ns    |   60.0 ns  |   9.77
bm_includes<int8_t, alg_type::std_fn>/3000/105/0/1     |    558 ns    |   86.4 ns  |   6.46
bm_includes<int8_t, alg_type::std_fn>/3000/1504/0/1    |    954 ns    |   1026 ns  |   0.93
bm_includes<int8_t, alg_type::std_fn>/3000/2750/0/1    |   1372 ns    |   1488 ns  |   0.92
[...]

PR #5683, opened 2025-08-17
Implement P3349R1

### LINKED ISSUE #5295: P3349R1 Converting Contiguous Iterators To Pointers
WG21-P3349R1 Converting Contiguous Iterators To Pointers

We can accept PRs to implement and test this change unconditionally. This mostly just limits what users can rely on, but there's an exception as the paper notes:

> The wording in this R1 revision requires the library to perform `i+n` on the original iterator (not only on a pointer `to_address(i)`), which gives the iterator a chance to check for a bounds overrun.

I believe we're good about this bounds check in debug mode, but I suspect that we don't always conform to this new requirement for implementers in release mode.

NylteJ:
This PR audits and updates all calls to `to_address` (or `_To_address`) to ensure that both `to_address(i)` and `to_address(i + (s - i))` (which should be equivalent to the standard-mandated `to_address(i + ranges::distance(i, s))`) are evaluated before using the pointer converted from the iterator. Fixes #5295.

Additional Notes:

1. For algorithms accepting an iterator pair `[i, s)`, this change evaluates only `to_address(s)` instead of `to_address(i + (s - i))`.   
While the proposal specifies converting `[i, s)` to `[to_address(i), to_address(i + (s - i)))` (implying at least one `operator+` invocation before conversion), iterator requirements guarantee that `i + (s - i)` must be equal to `s`, and functions like `ranges::advance` also do not actually "advance" the iterator in this case.    
Based on the proposal’s intent, this should not affect iterator validation: if `i + (s - i)` is invalid, then `s` is already invalid and should have been checked before the library call.   
This could be revised later if necessary.

2. The constructors of `string_view` and `span`, as well as the implementation of `views::counted`, are intentionally unchanged, as the standard explicitly specifies their behavior ([[string.view.cons]/9, 13](https://eel.is/c++draft/string.view.cons), [[span.cons]/6, 11, 19](https://eel.is/c++draft/span.cons), [[range.counted]/2.1](https://eel.is/c++draft/range.counted#2.1)). This may be worth submitting as a potential defect.

3. Added some previously missing `static_cast`s to pass the tests.


NylteJ:

stl/inc/yvals_core.h
```diff
@@ -83,6 +83,7 @@
 // P3223R2 Making istream::ignore() Less Surprising
 // P3323R1 Forbid atomic<cv T>, Specify atomic_ref<cv T>
 //     (for atomic<cv T>)
+// P3349R1 Converting Contiguous Iterators To Pointers
```
frederick-vs-ja: Maybe just a question: Is there, or should there be, any chance that the changes affect pre-C++20 modes? The paper only mentions C++20 `contiguous_iterator`, so perhaps the should be listed in the "`_HAS_CXX20` directly controls:" section.

Before C++20, we only consider some non-pointer iterators provided by the implementation to be contiguous (e.g. iterators of `vector`, `string`, and `string_view`). I wonder whether we've already done implementable checks for these implementation-specific iterators.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/env.lst
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+RUNALL_INCLUDE ..\usual_latest_matrix.lst
```
frederick-vs-ja: The matrix should be either `usual_20_matrix.lst` or `usual_matrix.lst`. (I guess the former is more appropriate.)

stl/inc/yvals_core.h
NylteJ: Indeed, they don’t do anything before C++20. Updated.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/env.lst
NylteJ: Done. I also updated the test code to match this change.

AlexGuteniev:
If both this and #5590 are accepted, there's stealth conflict: `includes` dispatch also uses `_First1_ptr + static_cast<ptrdiff_t>(_Last1 - _First1)` pattern.

StephanTLavavej:
I am reviewing this now (and will resolve the `<algorithm>` merge conflict), sorry for the delay!

stl/inc/xutility
```diff
@@ -4703,6 +4703,35 @@ _NODISCARD constexpr auto _To_address(const move_iterator<_Iter>& _Val) noexcept
...
+_NODISCARD constexpr _Iter_diff_t<_Iter> _Contiguous_iter_distance(const _Iter& _First, const _Sent& _Last)
+    noexcept(is_same_v<_Iter, _Sent> || noexcept(_First + (_Last - _First))) {
+    _STL_INTERNAL_STATIC_ASSERT(_Iterator_is_contiguous<_Iter>);
+    if constexpr (is_same_v<_Iter, _Sent>) {
+        return static_cast<_Iter_diff_t<_Iter>>(_STD _To_address(_Last) - _STD _To_address(_First));
+    } else {
+        return static_cast<_Iter_diff_t<_Iter>>(_STD _To_address(_First + (_Last - _First)) - _STD _To_address(_First));
+    }
+}
+template <class _Iter, class _Sent>
```
StephanTLavavej, line 4815: Style: We conventionally place a newline between function definitions unless they are extremely short and repetitive, which isn't really the case here.

stl/inc/xutility
```diff
@@ -6336,21 +6383,21 @@ namespace ranges {
...
 #endif // ^^^ _USE_STD_VECTOR_ALGORITHMS ^^^
                 {
                     if constexpr (_Elements_are_2_bytes) {
                         _STL_INTERNAL_STATIC_ASSERT(_Is_sized);
-                        _Result = reinterpret_cast<_Ptr_t>(
-                            const_cast<wchar_t*>(_CSTD wmemchr(reinterpret_cast<const wchar_t*>(_First_ptr),
-                                _STD _Find_arg_cast<wchar_t>(_Val), static_cast<size_t>(_Last - _First))));
+                        _Result = reinterpret_cast<_Ptr_t>(const_cast<wchar_t*>(_CSTD wmemchr(
+                            reinterpret_cast<const wchar_t*>(_First_ptr), _STD _Find_arg_cast<wchar_t>(_Val),
+                            static_cast<size_t>(_STD _Contiguous_iter_distance(_First, _Last)))));
```
StephanTLavavej: Style: Mostly pre-existing, but this is getting verbose enough that we should extract the arguments for clarity. I'm willing to pay more lines to make the steps easier to follow.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,463 @@
...
+
+struct safe_iter_out_of_bounds_err {};
+
+enum range_type { range_large = 3, range_medium = 4, range_small = 5 };
+
+bool safe_iter_nothrow_OOB_sign = false;
+
+// When `nothrow` is true, disables bounds checking during iterator increment/decrement (to verify `to_address` calls),
+// logging out-of-bounds accesses instead of throwing. When false, throws immediately on bounds violation.
+template <class T, bool nothrow = false>
```
StephanTLavavej: Style: We're somewhat less strict about naming conventions in tests (although other conventions, like "brace all control flow", remain in full force), but we still generally capitalize template parameters. Here I would expect `Nothrow`. I find this makes them easier to recognize as compile-time constants.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,463 @@
...
+#include <numeric>
+#include <ranges>
+#include <string>
+#include <vector>
+
+using namespace std;
+
+struct safe_iter_out_of_bounds_err {};
+
+enum range_type { range_large = 3, range_medium = 4, range_small = 5 };
```
StephanTLavavej: I recommend upgrading this to an `enum class`, which is more type-safe. (The less we have to worry about weird stuff happening in test code, the more time we have to worry about the weird stuff in product code.)

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,463 @@
...
+
+// When `nothrow` is true, disables bounds checking during iterator increment/decrement (to verify `to_address` calls),
+// logging out-of-bounds accesses instead of throwing. When false, throws immediately on bounds violation.
+template <class T, bool nothrow = false>
+class safe_iter {
+public:
+    using iterator_concept  = contiguous_iterator_tag;
+    using iterator_category = random_access_iterator_tag;
+    using value_type        = T;
+    using difference_type   = ptrdiff_t;
```
StephanTLavavej, line 39: Technically need `<cstddef>` for `ptrdiff_t`.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,463 @@
...
+    }
+
+    static constexpr safe_iter begin(ranges::contiguous_range auto&& rng) noexcept {
+        return {to_address(ranges::begin(rng)), to_address(ranges::begin(rng)), to_address(ranges::end(rng))};
+    }
+    static constexpr safe_iter end(ranges::contiguous_range auto&& rng) noexcept {
+        return {to_address(ranges::end(rng)), to_address(ranges::begin(rng)), to_address(ranges::end(rng))};
+    }
+
+    static constexpr pair<safe_iter, safe_iter> get_iters(ranges::contiguous_range auto&& rng, range_type type) {
```
StephanTLavavej, line 143: Need `<utility>` for `pair`.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,463 @@
...
+        const auto n = static_cast<underlying_type_t<range_type>>(type);
```
StephanTLavavej, line 144: Need `<type_traits>` for `underlying_type_t`.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,463 @@
...
+
+#undef TESTW
+#undef TESTR
+    }
+}
+
+template <bool nothrow>
+void test_matrix() {
+    // Ensures vectorization (and iterator-to-pointer conversion)
+    test<int8_t, 63, nothrow>();
```
StephanTLavavej: Need `<cstdint>` for `int8_t`.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,463 @@
...
+    using iterator_concept  = contiguous_iterator_tag;
+    using iterator_category = random_access_iterator_tag;
+    using value_type        = T;
+    using difference_type   = ptrdiff_t;
+    using pointer           = T*;
+    using reference         = T&;
+
+    constexpr safe_iter(pointer ptr, pointer range_first, pointer range_last) noexcept
+        : current_ptr{ptr}, range_first_ptr{range_first}, range_last_ptr{range_last} {
+        assert(current_ptr >= range_first_ptr && current_ptr <= range_last_ptr);
```
StephanTLavavej: It's better to assert one condition at a time, so that if the assertion fails, we'll see in the test log what the problem was.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,463 @@
...
+    GEN_RANGE_ITERS(r2, container_read_unique, const_iter)
+    GEN_RANGE_ITERS(w, container_write, iter)
+    GEN_RANGE_ITERS(w2, container_write_2, iter)
+
+    assert(r_first_s < r_first_m && r_first_m < r_first_l && r_last_s < r_last_m && r_last_m < r_last_l);
+
+#undef GEN_RANGE_ITERS
+#undef GEN_RANGE_ITER
+
+    static constexpr auto tests_base = [](auto&& func, auto&& first, auto&& last_or_dist) {
```
StephanTLavavej: Non-arrays can be plain `constexpr`, as the `static` doesn't make a difference. (For arrays it makes a considerable difference.)

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,463 @@
...
+        FAIL(ranges::remove_copy(src_first, src_last_ptr, dst_first, 1));
+
+        PASS(unique_copy(src_first, src_last, dst_first));
+        FAIL(unique_copy(src_first, src_last, dst_first + 1));
+        PASS(ranges::unique_copy(src_first, src_last_ptr, dst_first));
+        FAIL(ranges::unique_copy(src_first, src_last_ptr, dst_first + 1));
+    }
+
+#if _HAS_CXX23
+    // format {:s} & {:?s}
```
StephanTLavavej: Style: Because `&` has special meaning in C++, we generally write out "and" in comments to avoid confusion.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,463 @@
...
+    {
+        const vector chars = ranges::to<vector<char>>("\"Hello, world!\"");
```
StephanTLavavej: Style: It's not a huge deal here, but I generally recommend raw string literals whenever possible to avoid having to think about backslash-escaping.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
StephanTLavavej: The default `false` never appears to be used; it should be omitted to avoid confusion.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,463 @@
...
+        return *(*this + n);
+    }
+
+    constexpr safe_iter& operator+=(difference_type offset) noexcept(nothrow) {
+        current_ptr += offset;
+        if constexpr (!nothrow) {
+            if (current_ptr < range_first_ptr || current_ptr > range_last_ptr) {
+                throw safe_iter_out_of_bounds_err{};
+            }
+        }
```
StephanTLavavej: I expect this to log if it can't throw.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,463 @@
...
+    }
+    friend constexpr difference_type operator-(pointer left, const safe_iter& right) noexcept {
+        return left - right.current_ptr;
+    }
+    constexpr bool operator==(pointer right) const noexcept {
+        return current_ptr == right;
+    }
+    friend constexpr bool operator==(pointer left, const safe_iter& right) noexcept {
+        return left == right.current_ptr;
+    }
```
StephanTLavavej: `operator==(pointer left, const safe_iter& right)` shouldn't be necessary, as C++20 equality rewriting will call `operator==(pointer right)`.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,463 @@
...
+    {
+        const vector chars = ranges::to<vector<char>>("\"Hello, world!\"");
+
+        const auto [chars_first, chars_last] = safe_iter<const char, nothrow>::get_iters(chars, range_medium);
+
+        const ranges::subrange good_range{chars_first, to_address(chars_last)};
+        const ranges::subrange bad_range{chars_first, to_address(chars_last) + 1};
+        const ranges::subrange bad_range_2 = [&] {
+            if constexpr (nothrow) {
+                return ranges::subrange{chars_first - 1, to_address(chars_last)};
```
StephanTLavavej: This is problematic for consistently setting the out-of-band `bool`, because it performs a bad operation while constructing the bad subrange.

I looked into fixing this (with an unchecked ctor) and unifying the throwing and non-throwing cases, but fundamentally I don't think it's providing useful coverage. The pattern of `[to_address(i), to_address(i + ranges::distance(i, s)))` mandated by the paper *doesn't* catch this, because `i + ranges::distance(i, s)` is within bounds (and `operator+` doesn't assert that the *input* is valid). So the "nothrow" case was detecting badness purely because `chars_first - 1` here was checked, and the "throw" case was detecting badness because it was just duplicating `bad_range`.

I believe we should remove this.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,463 @@
...
+        if (current_ptr < range_first_ptr || current_ptr >= range_last_ptr) {
+            if constexpr (nothrow) {
+                safe_iter_nothrow_OOB_sign = true;
+            } else {
+                throw safe_iter_out_of_bounds_err{};
+            }
+        }
+        return *current_ptr;
+    }
+    constexpr pointer operator->() const noexcept {
```
StephanTLavavej: I spent a lot of time figuring out why `operator->` can't mirror `operator*`, I believe this should be commented.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,463 @@
...
+
+#define PASS(expr) single_test_pass<nothrow>([&]() { (void) (expr); })
+#define FAIL(expr) single_test_fail<nothrow>([&]() { (void) (expr); })
+
+#define EXPR_WRAP_IS(expr)       ([&](auto&& i, auto&& s) { (void) (expr); })
+#define EXPR_WRAP_IN(expr)       ([&](auto&& i, auto&& n) { (void) (expr); })
+#define EXPR_WRAP_ISD(expr)      ([&](auto&& i, auto&& s, auto&& dst) { (void) (expr); })
+#define EXPR_WRAP_IS_DI_DS(expr) ([&](auto&& i, auto&& s, auto&& dst_i, auto&& dst_s) { (void) (expr); })
+
+template <class T, size_t range_size, bool nothrow>
```
StephanTLavavej: Similarly `range_size` should be `RangeSize`.

StephanTLavavej (approved):

StephanTLavavej:
Thanks for the major work here! :heart_eyes_cat: I pushed moderate changes, please meow if I messed anything up.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
NylteJ: I don't think we can do that; see：
https://github.com/NylteJ/STL/blob/d11fa885caeca00cf3fffd77e9d2d903a0c613c7/tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp#L28-L29

The standard requires us to call `to_address` on the incremented iterator. Logging here would make that call unobservable.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
NylteJ: Because of the reasons mentioned above, `chars_first - 1` doesn’t trigger logging in the nothrow case, so I think it can be used for detecting the `to_address(i)` part (although it’s almost certain to be called).

NylteJ:
Also, it seems to have some issues on ARM64. I'm trying to investigate...

stl/inc/algorithm
```diff
@@ -465,7 +465,8 @@ constexpr bool _Vector_alg_in_unique_is_safe = _Vector_alg_in_search_is_safe<_It
 // Can we use this output iterator for remove_copy and unique_copy?
 template <class _Out, class _In>
 constexpr bool _Output_iterator_for_vector_alg_is_safe() {
-    if constexpr (_Iterator_is_contiguous<_Out>) {
+    if constexpr (is_pointer_v<_Out>) { // P3349R1: Non-pointer contiguous iterators prevent vectorization (unknown
```
AlexGuteniev: I don't understand this part.. But also we don't usually cite papers, except in yvals_core.h

NylteJ: `remove_copy` and `unique_copy` only copy a subset of the input range to the output range, and there’s no way to know the size of that subset in advance. Therefore, we can’t replace a non‑pointer `i + [0, n)` with `[to_address(i), to_address(i + n))` (since `n` is unknown), which prevents vectorization. See also: https://github.com/NylteJ/STL/blob/d11fa885caeca00cf3fffd77e9d2d903a0c613c7/tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp#L407-L431.  

However, if we perform an additional pass to determine the size of the output range, we could re‑enable vectorization here. I’m just not sure whether that would actually yield better performance…

As for the citation, I’ll remove it.

AlexGuteniev: Thanks for the explanation! 

I'm not sure if the citation should be removed, I'd like to ask @StephanTLavavej that.

frederick-vs-ja: I think the citation should be removed. P3349R1 should fall into the same category as that for LWG issues, and when an LWG issue is accepted into the standard, we generally remove, but not add, the citation to it.

AlexGuteniev: Regarding extra pass. 

Currently `remove_copy` and `unique_copy` use internal temporary buffer to allow writing more than the detonation size: the vectorization algorithm wants to write by up to 8 elements, even if there's only one actually. With extra pass and knowing size, we could have avoided temporary buffer most of the time.

Still I think that we should avoid extra pass due to both additional complexity, and extra memory access that might degrade the performance for very large inputs.

AlexGuteniev: This also has some relation to LWG-4262. There are comment of that one being NAD, 

If this is resolved as NAD, then we should probably consider whether we would add the source range size, or keep vetoing that part.

NylteJ: It looks like this part has changed in the latest working draft:

[[alg.remove]/9, 11](https://eel.is/c++draft/alg.remove#9):
> 9. Let:
>    - M be the number of iterators i in [first, last) for which E(i) is false;
>    - result_last be result + M for the overloads with no parameter result_last or result_r;
>    - N be min(M,  result_last - result).

> 11. Preconditions: The ranges [first, last) and [result, result + N) do not overlap.

[[alg.unique]/7, 9](https://eel.is/c++draft/alg.unique#7):
> 7. Let:
>    - M be the number of iterators i in the range [first + 1, last) for which E(i) is false;
>    - result_last be result + M + 1 for the overloads with no parameter result_last or result_r;
>    - N be min(M + 1, result_last - result).

> 9. Preconditions:
>    - The ranges [first, last) and [result, result + N) do not overlap.

If I understand correctly, this means that in any case we can only add the final output range size.

AlexGuteniev:
> Also, it seems to have some issues on ARM64. I'm trying to investigate...

You probably need to cross-test with non-vectorization, like here:
https://github.com/microsoft/STL/blob/1118f375859015034d221c6fdba73a9605c4a086/tests/std/tests/VSO_0000000_vector_algorithms/env.lst#L4-L7

stl/inc/algorithm
AlexGuteniev: > It looks like this part has changed in the latest working draft:

Interesting! I want to know which paper(s) cause this change. LWG-4262 appears to be no longer needed.

AlexGuteniev: It was WG21-P3179 that changed this part.

NylteJ:
@StephanTLavavej I’ve pushed some test fixes and additions, and reverted the two changes mentioned above. Please review again to make sure I didn’t mess anything up.

StephanTLavavej:
Thank you, I'll take a look! :heart_eyes_cat:

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
StephanTLavavej: Oh, I see. I didn't understand the comment due to its positioning at the top of the class instead of above `operator+=`.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
StephanTLavavej: Thanks, I think this makes sense now, my attempted change to add logging to `operator+=` is what led me down this path.

stl/inc/algorithm
StephanTLavavej: Thanks, I'll adjust the comment; I agree that we generally don't cite papers (which becomes increasingly confusing as the Standard is edited over the years). I think that the actual impact of limiting vectorization to pointers after unwrapping, instead of arbitrary contiguous iterators that aren't aware of unwrapping, is limited - 99% of use cases will be with pointers, arrays, and STL iterators that participate in unwrapping (`array`, `string`, `string_view`, `vector` iterators).

StephanTLavavej: Going with:
```
    // Before running remove_copy and unique_copy, the output size is unknown. This limits vectorization
    // to pointers (after unwrapping). N5014 [iterator.requirements.general]/13 allows us to convert
    // a range of contiguous iterators to pointers if and only if we know the size of the range ahead of time.
```

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
StephanTLavavej: I'll move the comment since this confused my cat-sized brain.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,506 @@
...
+        const const_iter rng_last{valid_last_ptr, valid_first_ptr, valid_last_ptr};
+        const auto rng_last_ptr = to_address(rng_last);
+
+        PASS(search_n(rng_first, rng_last, 2, 3));
+        FAIL(search_n(rng_first, rng_last + 1, 3, 3));
+        FAIL(search_n(rng_first - 1, rng_last, 2, 1));
+
+        PASS(ranges::search_n(rng_first, rng_last_ptr, 2, 3));
+        FAIL(ranges::search_n(rng_first, rng_last_ptr + 1, 3, 3));
+        FAIL(ranges::search_n(rng_first - 1, rng_last_ptr, 2, 1));
```
StephanTLavavej: In test code, I recommend "unique numbers for unique purposes". When all of your example values are small integers, it's hard to tell whether `2` is a value in the range, or a count of values. Also, `1` appears as a value in the range, and an amount to adjust iterators by.

For this reason, we commonly use values like `10, 20, 30` or `11, 22, 33` which are distinctive without being overly verbose.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,506 @@
...
+        FAIL(unique_copy(src_first, src_last, dst_first + 1));
+        PASS(ranges::unique_copy(src_first, src_last_ptr, dst_first));
+        FAIL(ranges::unique_copy(src_first, src_last_ptr, dst_first + 1));
+    }
+
+    // search_n
+    {
+        using const_iter = safe_iter<const int, Nothrow>;
+
+        const auto rng             = {1, 1, 2, 3, 3, 3};
```
StephanTLavavej: This is creating an `initializer_list` which is a quasi-container. The code isn't wrong, but we usually use real containers for test data. I think this would be clearer if replaced with an `array`.

tests/std/tests/P3349R1_contiguous_iterators_to_pointers/test.cpp
```diff
@@ -0,0 +1,506 @@
...
+        const auto valid_first_ptr = rng.begin() + 1;
+        const auto valid_last_ptr  = rng.end() - 1;
+
+        const const_iter rng_first{valid_first_ptr, valid_first_ptr, valid_last_ptr};
+        const const_iter rng_last{valid_last_ptr, valid_first_ptr, valid_last_ptr};
+        const auto rng_last_ptr = to_address(rng_last);
```
StephanTLavavej: I don't understand the purpose of `rng_last_ptr` - it's just `valid_last_ptr` and this isn't exercising any product code.

StephanTLavavej:
Thanks again, pushed another round of changes. Hopefully I didn't mess anything up this time! :joy_cat:

StephanTLavavej:

StephanTLavavej:
Thanks for implementing this C++26 feature, and congratulations on your first microsoft/STL commit! :heart_eyes_cat: :tada: :birthday:

This feature will ship in the MSVC Build Tools 14.51 in a future VS 2026 update (I don't know exactly which one yet).

PR #5684, opened 2025-08-19
Benchmark `count` for `vector<bool>`

AlexGuteniev:
There is `count` for `vector<bool>` optimization that uses `popcnt` on the integer elements of the `vector<bool>` internal representation, originally added in #1131. There's open PR #5640 to enhance that optimization further.

This PR adds benchmark to measure the results.

I've started off copying `vector_bool_copy.cpp` to mimic the existing style, then left only the `_algined` case, since unalignment doesn't make significant impact (unlike copying), still left the same name (as just `count` matches the STL algorithm name), and added `DoNotOptimize` where necessary. The value to count is alternating to explore both branches without adding extra benchmarks.

----

The results for  #5640 are mixed for me. 

On P cores of i5-1235U I see no improvement:

Benchmark              |   Before   |   After   | Speedup
-----------------------|------------|-----------|--------
count_aligned/64       |  17.0 ns   |  17.0 ns  |  1.00
count_aligned/4096     |  61.5 ns   |  59.6 ns  |  1.03
count_aligned/65536    |   718 ns   |   747 ns  |  0.96

On E cores I see some improvement, which is not too little for such a small change:

Benchmark              |  Before    |  After    | Speedup
-----------------------|------------|-----------|--------
count_aligned/64       |  21.3 ns   |  21.6 ns  |  0.99
count_aligned/4096     |   114 ns   |  90.6 ns  |  1.26
count_aligned/65536    |  1505 ns   |  1092 ns  |  1.38


StephanTLavavej:
5950X results:

Benchmark           |  Before |   After | Speedup
--------------------|---------|---------|--------
count_aligned/64    | 18.8 ns | 18.8 ns | 1.00
count_aligned/4096  | 77.7 ns | 62.8 ns | 1.24
count_aligned/65536 |  903 ns |  786 ns | 1.15

benchmarks/CMakeLists.txt
```diff
@@ -134,5 +134,6 @@ add_benchmark(sv_equal src/sv_equal.cpp)
 add_benchmark(swap_ranges src/swap_ranges.cpp)
 add_benchmark(unique src/unique.cpp)
 add_benchmark(vector_bool_copy src/vector_bool_copy.cpp)
+add_benchmark(vector_bool_count src/vector_bool_count.cpp)
 add_benchmark(vector_bool_copy_n src/vector_bool_copy_n.cpp)
```
StephanTLavavej: Should be sorted.

benchmarks/src/vector_bool_count.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include <cstddef>
+#include <random>
+#include <vector>
+
+using namespace std;
+
+vector<bool> createRandomVector(const size_t size) {
+    static mt19937 gen;
+    vector<bool> result(size);
+    generate_n(result.begin(), size, [] { return bernoulli_distribution{0.5}(gen); });
```
StephanTLavavej: It would be slightly more elegant to call `generate()` instead of repeating the size.

Also, constructing a distribution for every element is unnecessary, even if this distribution happens to be cheap. Storing it alongside the engine would be simpler.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: I pushed minor stylistic changes.

benchmarks/src/vector_bool_count.cpp
AlexGuteniev: The reason it was written this way is that I wanted to avoid diverging from this for now:
https://github.com/microsoft/STL/blob/07045f469c862fdb442873c3f7c66cb35a3486d9/benchmarks/src/vector_bool_copy.cpp#L13-L18

I also disliked this distribution construction. But I think it should capture by reference instead. If that wasn't cheap, it would have been better to capture by reference than use magic static (which would have been engaged then).

`static mt19937 gen` should also be captured by reference. Avoiding `static` in addition to small perf improvement would also make the benchmark cases not dependent on previously executed ones, so the pattern would be repeated in case of filtering some cases via `--benchmark_filter`.

I think all those need a follow up, so that existing benchmarks are fixed along with the existing ones. We can make it right in this PR, if you'd like.

StephanTLavavej: This is small enough that a followup commit in this PR would be fine, before I prepare a mirror.

StephanTLavavej:

StephanTLavavej:
Thanks for measuring this performance! :stopwatch: :rocket: :heart_eyes_cat:

PR #5690, opened 2025-08-23
Fix consteval case for _STL_REPORT_ERROR

enginelesscc:
Fixes repro case: https://godbolt.org/z/YPex3q69z
Short description: `_STL_REPORT_ERROR` did not consider constant evaluation and emitted code that would still cause an expected build failure, but hid the intended error message
The change now explicitly uses a `static_assert` for this case.

### DIFF
--- a/stl/inc/yvals.h
+++ b/stl/inc/yvals.h
@@ -259,9 +259,13 @@ extern "C" __declspec(noreturn) void __fastfail(unsigned int); // declared by <i
 #endif // choose "doom function"
 #endif // ^^^ !defined(_MSVC_STL_DOOM_FUNCTION) ^^^
 
-#define _STL_REPORT_ERROR(mesg) \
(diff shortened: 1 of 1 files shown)


enginelesscc:
Okay this is more difficult than expected.
`if (_Is_constant_evaluated())` wont hide the static assert, causing false positives in runtime-code.
Limiting to `#if _HAS_CXX23` + using `if consteval` does not work either.

cpplearner:
Unfortunately `static_assert` is the wrong tool for this problem. `static_assert` reports the error at instantiation time, even if the enclosing branch is not evaluated.

I think we'll have to wait for @brevzin's [P2758](https://wg21.link/P2758) "Emitting messages at compile time".

StephanTLavavej:
@cpplearner Yep, that's what I expected.

@enginelesscc Thanks for looking into improving our diagnostics. I'm going to close this PR as there's no feasible way to achieve this right now.

Issue #1024, opened 2020-07-09
<charconv>: instead of memcpy'ing the result you can calculate necessary buffer size ahead of time for integers


toughengineer:
For integers instead of filling the buffer from right to left and `memcpy`'ing the result back so the most significant digit (that is written last) ends up in the leftmost position, you can calculate the necessary buffer size ahead of time.

See example in libc++:
https://github.com/llvm/llvm-project/blob/master/libcxx/include/charconv#L166
straight from the Bit Twiddling Hacks:
https://graphics.stanford.edu/~seander/bithacks.html#IntegerLog10

`memcpy` usage here:
https://github.com/microsoft/STL/blob/550713eba27c5803addb6f615d3526400cb2df37/stl/inc/charconv#L137-L146

miscco:
As far as I know libc++ charconv implementation is derived from the STLs one. 

The reason that code is not using the builtins is most likely that their support in MSVC was only merged recently (See #795)

StephanTLavavej:
Thanks for the performance idea! MSVC's and libc++'s integer `<charconv>` implementations were independently written, which is why we're lacking this technique.

toughengineer:
I'm glad to help.

After I submitted this issue I realized that writing to a buffer on the stack and then `memcpy`'ing to destination buffer may still be faster than calculating the needed length in advance.

The calculation uses log2 as counting bits instruction (which is several clocks on modern CPUs) and some multiplication+shifting+table lookup and fixup.

I tried to ask people who compared different implementations (e.g. `to_chars`/`to_string` and that of fmtlib, maybe @vitaut can comment on this) and got no definitive answer. So, I guess, one should investigate if it's really beneficial to spend time calculating the size in advance, and possibly take other considerations into account.

vitaut:
I ran some benchmarks in http://www.zverovich.net/2020/06/13/fast-int-to-string-revisited.html. Size computation (https://github.com/fmtlib/fmt/blob/60c43e870387affb07de32ed4fc2794ac7a9f60d/include/fmt/format.h#L787) is a relatively insignificant factor compared to other optimizations.

StephanTLavavej:
Also, I have considered having different codepaths for "definitely enough space for the worst case" and "maybe not enough space, need to do precise bounds checking".

AlexGuteniev:
I've attempted the following function to determine length:

<details>
<summary>_UIntegral_digits function</summary>

```c++
// Copyright (c) Microsoft Corporation.
[...]

PR #5691, opened 2025-08-23
Use division by 100 in `to_string` for integers

### LINKED ISSUE #3857: `<string>`: `to_string()` for integers could be faster
Related to #1024 and #3805 but not a duplicate. 

In DevCom-10402949, a user has suggested using the divide-by-100 approach (already used by Ryu/`<charconv>`) to print digits faster. (Note: Due to licensing concerns, please **DO NOT** refer to libstdc++'s implementation linked by the DevCom suggestion.) They have also suggested calling `memcpy()` for 2 bytes, although I find this surprising and would want to confirm via benchmarking.

AlexGuteniev:
# ⚙️ Optimization

Resolves #3857. Divides by 100 instead of by 10, as proposed.

There's similar place in `to_chars`, skipped for now.

# 🏁 Benchmark

Large and small numbers., like numbers naturally seen when counting things.
Generated via log-normal distribution, as @statementreply suggested.
Picked some arbitrary parameters, to approximately fit in the integer ranges.

Benchmarked also `std::_UIntegral_to_buff` separetely as well to see how much the optimization helps on its own, avoiding #1024 limitation.

# ⏱️ Benchmark results

i5-1235U  P cores:

Benchmark                                      |  Before  |  After   | Speedup
-----------------------------------------------|----------|----------|-------
internal_integer_to_buff<uint8_t, 2.5, 1.5>    | 2.30 ns  | 3.42 ns  | 0.67
internal_integer_to_buff<uint16_t, 5.0, 3.0>   | 3.70 ns  | 2.64 ns  | 1.40
internal_integer_to_buff<uint32_t, 10.0, 6.0>  | 4.69 ns  | 2.86 ns  | 1.64
internal_integer_to_buff<uint64_t, 20.0, 12.0> | 10.5 ns  | 5.29 ns  | 1.98
integer_to_string<uint8_t, 2.5, 1.5>           | 5.87 ns  | 5.44 ns  | 1.08
integer_to_string<uint16_t, 5.0, 3.0>          | 6.79 ns  | 6.32 ns  | 1.07
integer_to_string<uint32_t, 10.0, 6.0>         | 8.11 ns  | 7.28 ns  | 1.11
integer_to_string<uint64_t, 20.0, 12.0>        | 14.5 ns  | 14.2 ns  | 1.02
integer_to_string<int8_t, 2.5, 1.5>            | 6.64 ns  | 5.96 ns  | 1.11
integer_to_string<int16_t, 5.0, 3.0>           | 6.23 ns  | 5.88 ns  | 1.06
integer_to_string<int32_t, 10.0, 6.0>          | 7.58 ns  | 6.33 ns  | 1.20
integer_to_string<int64_t, 20.0, 12.0>         | 17.8 ns  | 18.8 ns  | 0.95

i5-1235U  E cores:

Benchmark                                      |  Before  | After    | Speedup
-----------------------------------------------|----------|----------|-------
internal_integer_to_buff<uint8_t, 2.5, 1.5>    | 4.14 ns  | 4.79 ns  | 0.86
internal_integer_to_buff<uint16_t, 5.0, 3.0>   | 8.08 ns  | 4.76 ns  | 1.70
internal_integer_to_buff<uint32_t, 10.0, 6.0>  | 11.4 ns  | 5.41 ns  | 2.11
internal_integer_to_buff<uint64_t, 20.0, 12.0> | 23.8 ns  | 13.9 ns  | 1.71
integer_to_string<uint8_t, 2.5, 1.5>           | 17.2 ns  | 12.7 ns  | 1.35
integer_to_string<uint16_t, 5.0, 3.0>          | 17.1 ns  | 13.6 ns  | 1.26
integer_to_string<uint32_t, 10.0, 6.0>         | 18.3 ns  | 14.0 ns  | 1.31
integer_to_string<uint64_t, 20.0, 12.0>        | 36.6 ns  | 29.4 ns  | 1.24
integer_to_string<int8_t, 2.5, 1.5>            | 17.8 ns  | 12.0 ns  | 1.48
integer_to_string<int16_t, 5.0, 3.0>           | 20.0 ns  | 13.4 ns  | 1.49
integer_to_string<int32_t, 10.0, 6.0>          | 21.5 ns  | 15.1 ns  | 1.42
integer_to_string<int64_t, 20.0, 12.0>         | 39.7 ns  | 35.0 ns  | 1.13

# 🥉 Results interpretation

I'm not even sure if this is worth doing.

Allocating the string and copying the result there takes roughly half of the time, so the effect of micro-optimization in digits generation is small.

However, the internal function seem to show improvement. This looks like an indication that #1024 improvement would help here. It could be that the performance is limited due to failed store-to-load forwarding, as individual character stores are followed by bulk memcpy; in this case, the improvement may be somewhat negated by a longer stall.

### DIFF
--- a/benchmarks/src/integer_to_string.cpp
+++ b/benchmarks/src/integer_to_string.cpp
@@ -4,6 +4,7 @@
 #include <algorithm>
 #include <array>
 #include <benchmark/benchmark.h>
+#include <cmath>
 #include <cstdint>
 #include <limits>
 #include <random>
@@ -18,8 +19,16 @@ auto generate_array() {
 
     mt19937_64 gen;
     lognormal_distribution<double> dis(M, S);
-    constexpr auto max_val = static_cast<double>(numeric_limits<T>::max());
-    ranges::generate(a, [&] { return static_cast<T>(clamp(dis(gen), 0.0, max_val)); });
+    auto get_clamped_value = [&] {
+        for (;;) {
+            const double dbl       = floor(dis(gen));
+            constexpr auto max_val = static_cast<double>(numeric_limits<T>::max());
+            if (dbl <= max_val) {
+                return static_cast<T>(dbl);
+            }
+        }
+    };
+    ranges::generate(a, get_clamped_value);
 
     if constexpr (is_signed_v<T>) {
         bernoulli_distribution b(0.5);


stl/inc/xmemory
```diff
@@ -2786,6 +2787,15 @@ _NODISCARD _Elem* _UIntegral_to_buff(_Elem* _RNext, _UTy _UVal) { // used by bot
     auto _UVal_trunc = static_cast<unsigned long>(_UVal);
 #endif // ^^^ !defined(_WIN64) ^^^
 
+    if constexpr (_Is_any_of_v<_Elem, char, wchar_t>) {
+        while (_UVal_trunc >= 100) {
+            const unsigned long _UVal_trunc_part = _UVal_trunc % 100;
+            _UVal_trunc /= 100;
+            *--_RNext = static_cast<_Elem>(__DIGIT_TABLE<_Elem>[_UVal_trunc_part * 2 + 1]);
+            *--_RNext = static_cast<_Elem>(__DIGIT_TABLE<_Elem>[_UVal_trunc_part * 2]);
```
AlexGuteniev: This can be further optimized to
```c++
            _RNext -= 2;
            _CSTD memcpy(_RNext, __DIGIT_TABLE<_Elem> + _UVal_trunc_part * 2, 2 * sizeof(_Elem))
```
still without much better results in the benchmark.

AlexGuteniev:
It might be that doing #1024 equivalent for `to_sting()` would give better improvement, and after that, this one would be noticeable.

Whereas normally `memcpy` that follows buffer forming is fast, the `memcpy` followed be fast one-by-one elements placement is slow due to failed store-to-load forwarding stall, and making the initial buffer fill faster will make that stall longer.

Edit: separate benchmark of internal `std::_UIntegral_to_buff` confirms that avoiding `memcpy`, while possibly helpful on its own, wouldn't make this optimization more helpful.

StephanTLavavej:
I think I would be more comfortable with this if we didn't have to drag in the Ryu tables into `<xmemory>`.

It would be fairly simple to completely reimplement the digit tables with a constexpr helper to emit the digits.

AlexGuteniev:
> It would be fairly simple to completely reimplement the digit tables with a constexpr helper to emit the digits.

Do we want that? I've inserted predefined array for now. which should be better for throughput.

AlexGuteniev:
> I think I would be more comfortable with this if we didn't have to drag in the Ryu tables into `<xmemory>`.

After doing this, and eliminating `if constexpr`, I was able to do some code rearrangement, the results are now somewhat better, updated the tables in description.

StephanTLavavej:
My concern was around keeping Ryu-derived code in separate files. If we generate the tables with a freshly written constexpr function, then there's no question that they aren't a derived work. That said, the digit tables are very obviously data/facts with no creativity.

AlexGuteniev:
Ok, generated the table.

stl/inc/xmemory
```diff
@@ -2762,10 +2762,25 @@ namespace ranges {
...
+    }
+
+    return _Result;
+}
+
 template <class _Elem, class _UTy>
 _NODISCARD _Elem* _UIntegral_to_buff(_Elem* _RNext, _UTy _UVal) { // used by both to_string and thread::id output
     // format _UVal into buffer *ending at* _RNext
     static_assert(is_unsigned_v<_UTy>, "_UTy must be unsigned");
+    static constexpr auto _Table = _STD _UIntegral_to_buff_digits<_Elem>();
```
StephanTLavavej: This emits a separate table per `_UTy`, and I'm not confident that the linker will always fold it. Better to extract.

stl/inc/xmemory
```diff
@@ -2762,10 +2762,25 @@ namespace ranges {
 } // namespace ranges
 #endif // _HAS_CXX23
 
+template <class _Elem>
+constexpr auto _UIntegral_to_buff_digits() {
+    struct {
+        _Elem _Data[200];
```
StephanTLavavej: Since we control the type, we could use a multi-dimensional array to avoid having to do manual arithmetic for the indices.

stl/inc/xmemory
```diff
@@ -2776,20 +2791,34 @@ _NODISCARD _Elem* _UIntegral_to_buff(_Elem* _RNext, _UTy _UVal) { // used by bot
...
+            _RNext -= 2;
+            _CSTD memcpy(_RNext, _Table._Data + _UVal_trunc_part * 2, 2 * sizeof(_Elem));
+        } while (_UVal_trunc >= 10);
+
+        if (_UVal_trunc == 0) {
+            return _RNext;
+        }
+    }
+
+    *--_RNext = static_cast<_Elem>('0' + _UVal_trunc);
```
StephanTLavavej: This control flow is undesirably complicated (note the dual tests of `_UVal_trunc >= 10`) and I think we can improve it.

stl/inc/xmemory
```diff
@@ -2762,10 +2762,25 @@ namespace ranges {
...
 #endif // _HAS_CXX23
 
+template <class _Elem>
+constexpr auto _UIntegral_to_buff_digits() {
+    struct {
+        _Elem _Data[200];
+    } _Result = {};
+
+    for (int _Idx = 0; _Idx != 100; ++_Idx) {
+        _Result._Data[_Idx * 2 + 0] = static_cast<_Elem>(_Idx / 10 + '0');
```
StephanTLavavej: Consistency: Below, we write `'0' + stuff`.

benchmarks/src/integer_to_string.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include <random>
+#include <type_traits>
+
+using namespace std;
+
+template <class T, double M, double S>
+auto generate_array() {
+    array<T, 2000> a;
+
+    mt19937 gen;
```
StephanTLavavej: We should prefer to use `mt19937_64`.

benchmarks/src/integer_to_string.cpp
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    return a;
+}
+
+template <class T, double M, double S>
+void internal_integer_to_buff(benchmark::State& state) {
+    auto a = generate_array<T, M, S>();
+
+    char buff[24];
```
StephanTLavavej: This is unnecessarily large. `18446744073709551615` and `-9223372036854775807` are both 20 chars.

benchmarks/src/integer_to_string.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+    auto it = a.begin();
+    for (auto _ : state) {
+        auto i = *it;
+        benchmark::DoNotOptimize(i);
+        auto s = std::_UIntegral_to_buff(buff, i);
```
StephanTLavavej: This is writing out of bounds: https://github.com/microsoft/STL/blob/cbe2ee99caf122555a305d18f69e57c75b3fe5ec/stl/inc/xmemory#L2765-L2767

benchmarks/src/integer_to_string.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+template <class T, double M, double S>
+void integer_to_string(benchmark::State& state) {
+    auto a = generate_array<T, M, S>();
+
+    auto it = a.begin();
+    for (auto _ : state) {
+        auto i = *it;
+        benchmark::DoNotOptimize(i);
+        auto s = to_string(i);
```
StephanTLavavej, line 70: Need `<string>`.

benchmarks/src/integer_to_string.cpp
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <algorithm>
+#include <array>
+#include <benchmark/benchmark.h>
+#include <cstddef>
+#include <cstdint>
+#include <cstdlib>
```
StephanTLavavej: Unless I'm missing something, `<cstddef>` and `<cstdlib>` are unused.

benchmarks/src/integer_to_string.cpp
StephanTLavavej: We should exercise `wchar_t` at least for the internal helper. (The `to_wstring` layer is what's relevant to users, I suppose, but will have different allocation characteristics.)

stl/inc/xmemory
```diff
@@ -2776,20 +2791,34 @@ _NODISCARD _Elem* _UIntegral_to_buff(_Elem* _RNext, _UTy _UVal) { // used by bot
...
-                *--_RNext = static_cast<_Elem>('0' + _UVal_chunk % 10);
-                _UVal_chunk /= 10;
+            for (int _Idx = 0; _Idx != 4; ++_Idx) {
+                const unsigned long _UVal_chunk_part = _UVal_chunk % 100;
+                _UVal_chunk /= 100;
+                _RNext -= 2;
+                _CSTD memcpy(_RNext, _Table._Data + _UVal_chunk_part * 2, 2 * sizeof(_Elem));
             }
+
+            *--_RNext = static_cast<_Elem>('0' + _UVal_chunk);
```
StephanTLavavej: Benchmarking indicates that this 32-bit chunk optimization remains extremely valuable, and that printing 8 digits at a time instead of 9 is a minor win (in addition to simplifying the code).

StephanTLavavej:
Thanks! I pushed major changes, please double-check.

AlexGuteniev:
Looks good. 3708bbff4699431aca5f20cf1df2750cae2dadbb might change codegen, but I assume it is fine, as you have benchmarked (at least for `wchar_t` and on x86)

StephanTLavavej:
Yeah, I checked and the differences between your codegen and mine appeared to be lost in the noise.

StephanTLavavej:

StephanTLavavej:
I've pushed commits to fix a bug found by internal AI code review. The `lognormal_distribution` was generating out-of-range values up to 4% of the time, triggering UB when we `static_cast` from `double` to integer. I've replaced this with a retry loop, so that we preserve the distribution's behavior, without generating the maximum value unusually often (which would happen if we just used `std::clamp()`). This carefully uses `floor()` to actually generate the maximum value some of the time. Needless to say, my code was written without any AI assistance.

benchmarks/src/integer_to_string.cpp
```diff
@@ -0,0 +1,100 @@
...
+auto generate_array() {
+    array<T, 2000> a;
+
+    mt19937_64 gen;
+    lognormal_distribution<double> dis(M, S);
+    auto get_clamped_value = [&] {
+        for (;;) {
+            const double dbl       = floor(dis(gen));
+            constexpr auto max_val = static_cast<double>(numeric_limits<T>::max());
+            if (dbl <= max_val) {
```
MattStephanson, line 26: As pointed out by @statementreply on Discord, for `uint_64` this rounds to $2^{64}$, so to be rigorous you would need to either special-case that or make the test `dbl < max_val`.

StephanTLavavej: Thanks, I can fix this in a followup.

StephanTLavavej:
:100: :100: :100:
===

Issue #1243, opened 2020-08-28
<charconv>: Report compiler optimization issues


StephanTLavavej:
`<charconv>` is very fast, but it could be faster. One issue is that MSVC's optimizer often generates slower code for `<charconv>` than Clang/LLVM. We should create reduced test cases and file bugs on [Developer Community](https://developercommunity.visualstudio.com/spaces/62/index.html) for the optimizer team to investigate and fix.

At the bottom of this issue, I've prepared a self-contained test case using `<charconv>`. Here's a table of my timing results (Intel Core i7-8700, VS 2019 16.8 Preview 2, Clang/LLVM 10.0.0). The measured times are average nanoseconds per floating-point value.

| Scenario | MSVC x86 | LLVM x86 | LLVM x86 Speedup Ratio | MSVC x64 | LLVM x64 | LLVM x64 Speedup Ratio |
|----------|-------:|---------:|-----------------------:|-------:|---------:|-----------------------:|
| `float` plain shortest | 48.3 | 45.5 | 1.06 | 41.9 | 35.7 | :warning: 1.17 |
| `double` plain shortest | 92.5 | 71.8 | :x: 1.29 | 49.1 | 39.4 | :x: 1.25 |
| `float` scientific shortest | 47.2 | 41.4 | :warning: 1.14 | 38.3 | 31.6 | :x: 1.21 |
| `double` scientific shortest | 92.6 | 72.5 | :x: 1.28 | 48.0 | 39.6 | :x: 1.21 |
| `float` fixed shortest | 58.9 | 51.1 | :warning: 1.15 | 47.6 | 38.8 | :x: 1.23 |
| `double` fixed shortest | 338.5 | 298.4 | :warning: 1.13 | 145.9 | 111.6 | :x: 1.31 |
| `float` general shortest | 48.8 | 42.5 | :warning: 1.15 | 39.3 | 32.5 | :x: 1.21 |
| `double` general shortest | 92.9 | 72.4 | :x: 1.28 | 48.6 | 38.5 | :x: 1.26 |
| `float` hex shortest | 25.8 | 26.4 | :heavy_check_mark: 0.98 | 21.0 | 23.8 | :heavy_check_mark: 0.88 |
| `double` hex shortest | 109.3 | 53.8 | :boom: **2.03** | 27.7 | 26.6 | 1.04 |
| `float` scientific 8 | 116.5 | 102.3 | :warning: 1.14 | 63.8 | 58.0 | :warning: 1.10 |
[...]

PR #5700, opened 2025-08-28
`<charconv>`: floating benchmark

AlexGuteniev:
Port benchmark from https://github.com/microsoft/STL/issues/1243#issue-687888565 comment.

Changes:
 * Don't hard-code the number of repetitions, this is now determined by the framework
 * Use `DoNotOptimize` instead of a dummy variable
 * Measure only one conversion rather than whole vector. Primarily because Google Benchmark still shows these bigger times in ns, so they are inconveniently big when measuring them for the whole vector
 * Underscore in benchmark names, as they are now function names
 * `N` -> `n`, `RT` -> `Rt`, avoiding macro-like names
 * Randomly generated vectors do not persist between tests. I find this easier to maintain this way, despite some runtime cost, feel free to disagree
 * Avoided `.size()` in a loop. Perf habit: calling `vector<T>::size()` in a loop may inhibit certain optimizations

Example output:
```
2025-08-28T09:43:39+03:00
Running out\bench\benchmark-charconv_floats.exe
Run on (12 X 2496 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 1280 KiB (x6)
  L3 Unified 12288 KiB (x1)
-------------------------------------------------------------------------
Benchmark                               Time             CPU   Iterations
-------------------------------------------------------------------------
STL_float_plain_shortest             36.2 ns         36.1 ns     19478261
STL_double_plain_shortest            43.8 ns         44.5 ns     17920000
STL_float_scientific_shortest        34.8 ns         34.9 ns     22400000
STL_double_scientific_shortest       43.0 ns         39.1 ns     16000000
STL_float_fixed_shortest             43.0 ns         41.4 ns     16592593
STL_double_fixed_shortest             110 ns          110 ns      6400000
STL_float_general_shortest           35.4 ns         33.5 ns     22400000
STL_double_general_shortest          43.3 ns         43.3 ns     16592593
STL_float_hex_shortest               18.6 ns         18.4 ns     37333333
STL_double_hex_shortest              22.4 ns         22.0 ns     32000000
STL_float_scientific_8               59.7 ns         55.8 ns     11200000
STL_double_scientific_16             68.4 ns         68.4 ns     11200000
STL_float_fixed_6_lossy              40.6 ns         40.1 ns     17920000
STL_double_fixed_6_lossy             92.8 ns         94.2 ns      8960000
STL_float_general_9                  71.6 ns         69.8 ns      8960000
STL_double_general_17                81.5 ns         77.4 ns      7466667
STL_float_hex_6                      18.2 ns         18.0 ns     37333333
STL_double_hex_13                    21.2 ns         20.4 ns     34461538
```


benchmarks/src/charconv_floats.cpp
```diff
@@ -0,0 +1,194 @@
...
+    constexpr size_t n = 2'000'000; // how many floating-point values to test
+
+    constexpr size_t BufSize = 2'000; // more than enough
+
+    mt19937_64 mt64;
+
+    vector<Floating> vec;
+
+    vec.reserve(n);
+    for (size_t i = 0; i != n; ++i) {
```
AlexGuteniev: Need `while (vec.size() < n)` loop here

benchmarks/src/charconv_floats.cpp
```diff
@@ -0,0 +1,192 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <benchmark/benchmark.h>
+#include <charconv>
+#include <random>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
```
StephanTLavavej: We conventionally include `<cmeow>` everywhere, so this should be adjusted.

benchmarks/src/charconv_floats.cpp
```diff
@@ -0,0 +1,192 @@
...
+    case RoundTrip::Lossy:
+    default:
+        puts("FAIL");
+        exit(EXIT_FAILURE);
+    }
+}
+
+template <RoundTrip Rt, typename Floating, typename... Args>
+void test_to_chars(benchmark::State& state, const Args&... args) {
+    constexpr size_t n = 2'000'000; // how many floating-point values to test
```
StephanTLavavej, line 45: Technically `<cstddef>` for `size_t`.

benchmarks/src/charconv_floats.cpp
```diff
@@ -0,0 +1,192 @@
...
+    while (vec.size() < n) {
+        using Integral             = conditional_t<sizeof(Floating) == 4, uint32_t, uint64_t>;
+        const Integral val         = static_cast<Integral>(mt64());
+        constexpr Integral inf_nan = sizeof(Floating) == 4 ? 0x7F800000U : 0x7FF0000000000000ULL;
+        if ((val & inf_nan) == inf_nan) {
+            continue; // skip INF/NAN
+        }
+        Floating flt;
+        static_assert(sizeof(flt) == sizeof(val));
+        memcpy(&flt, &val, sizeof(flt));
```
StephanTLavavej: We can use `bit_cast` now, which I didn't have access to when I wrote that benchmark.

benchmarks/src/charconv_floats.cpp
```diff
@@ -0,0 +1,192 @@
...
+        }
+    }
+}
+
+constexpr auto STL_float_plain_shortest  = test_to_chars<RoundTrip::Gen, float>;
+constexpr auto STL_double_plain_shortest = test_to_chars<RoundTrip::Gen, double>;
+
+void STL_float_scientific_shortest(benchmark::State& state) {
+    test_to_chars<RoundTrip::Sci, float>(state, chars_format::scientific);
+}
```
StephanTLavavej: We can use Google Benchmark machinery to avoid having to define and mention these wrappers separately.

Because the optional arguments are compile-time constants, we can actually pass them as non-type template parameters.

benchmarks/src/charconv_floats.cpp
```diff
@@ -0,0 +1,192 @@
...
+void verify(const bool b) {
+    if (!b) {
+        puts("FAIL");
+        exit(EXIT_FAILURE);
+    }
+}
+
+enum class RoundTrip { Sci, Fix, Gen, Hex, Lossy };
+
+constexpr chars_format chars_format_from_RoundTrip(const RoundTrip rt) {
```
StephanTLavavej: This could be upgraded to `consteval` which would detect misuse at compile-time.

benchmarks/src/charconv_floats.cpp
```diff
@@ -0,0 +1,192 @@
...
+        puts("FAIL");
+        exit(EXIT_FAILURE);
+    }
+}
+
+template <RoundTrip Rt, typename Floating, typename... Args>
+void test_to_chars(benchmark::State& state, const Args&... args) {
+    constexpr size_t n = 2'000'000; // how many floating-point values to test
+
+    constexpr size_t BufSize = 2'000; // more than enough
```
StephanTLavavej: Because this is optimized, we can use `end(buf)` instead of having to extract `BufSize` and perform pointer arithmetic.

benchmarks/src/charconv_floats.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+    mt19937_64 mt64;
```
StephanTLavavej: It's not terribly important, but `mt64` isn't needed after preparing the `vector`, so we can reduce its scope.

benchmarks/src/charconv_floats.cpp
```diff
@@ -0,0 +1,192 @@
...
+        Floating flt;
+        static_assert(sizeof(flt) == sizeof(val));
+        memcpy(&flt, &val, sizeof(flt));
+        vec.push_back(flt);
+    }
+
+    char buf[BufSize];
+
+    auto it = vec.begin();
+    for (auto _ : state) {
```
StephanTLavavej: `it` isn't needed for the verification loop, so reducing its scope would be nice.

StephanTLavavej (approved):

StephanTLavavej:
Thanks, this is awesome! :heart_eyes_cat: I pushed some simplifications, and verified that the benchmark results weren't degraded.

StephanTLavavej:

benchmarks/src/charconv_floats.cpp
AlexGuteniev: Whoa, didn't notice that there was such possibility.

You can now restore spaces instead of underscores, if you liked them more.

StephanTLavavej:
Thanks for modernizing this benchmark and letting me relive the glory days of `<charconv>`! :heart_eyes_cat: :stopwatch: :rocket:

PR #5704, opened 2025-08-29
Build the benchmark with clang-cl only on x64

AlexGuteniev:
There's infrequent but not very rara access violation when linking with the benchmark using clang-cl on x86.

Apparently x64 works always.

Let's keep only x64 built on CI. It might worth giving another try when clang-cl is updated to the next major.

Note that this change does not prevent building x86 benchmark with clang-cl locally, only makes it possible for this configuration to bit rot.

### DIFF
--- a/azure-devops/build-benchmarks.yml
+++ b/azure-devops/build-benchmarks.yml
@@ -29,10 +29,10 @@ steps:
   displayName: 'Configure the benchmarks for ${{ parameters.compiler }}'
   timeoutInMinutes: 2
   env: { TMP: $(tmpDir), TEMP: $(tmpDir) }
-  # TRANSITION, we currently don't build the benchmarks with Clang for ARM64 or ARM64EC
+  # TRANSITION, we currently only build the benchmarks with Clang for x64
   condition: >
     and(succeeded(), ${{ parameters.buildBenchmarks }},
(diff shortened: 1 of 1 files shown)


azure-devops/build-benchmarks.yml
```diff
@@ -29,18 +29,20 @@ steps:
   displayName: 'Configure the benchmarks for ${{ parameters.compiler }}'
   timeoutInMinutes: 2
   env: { TMP: $(tmpDir), TEMP: $(tmpDir) }
-  # TRANSITION, we currently don't build the benchmarks with Clang for ARM64 or ARM64EC
+  # TRANSITION, we currently only build the benchmarks with Clang for x64
   condition: >
     and(succeeded(), ${{ parameters.buildBenchmarks }},
-      not(and(eq('${{ parameters.compiler }}', 'clang-cl'), startsWith('${{ parameters.targetPlatform }}', 'arm64'))))
+      not(and(eq('${{ parameters.compiler }}', 'clang-cl'),
+        not(startsWith('${{ parameters.targetPlatform }}', 'x64')))))
```
StephanTLavavej: `startsWith` was being used to handle both `'arm64'` and `'arm64ec'`. We should use `eq` for `'x64'`.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for improving CI reliability! :shield: :mechanical_arm: :100:

Issue #5698, opened 2025-08-27
Questions about the precondition of std::any's assignment operators


achabense:
In https://github.com/microsoft/STL/pull/3965, I tried to fix `std::any`'s self-assignment. The obvious way is to compare the address first, but I chose a different design (by always making a copy first) to support the `if(1)` case below.

The implementation cannot be as efficient as address comparison. **And worse**, I find that `any::swap` is implemented based on `std::exchange`, which calls `operator=(any&&)` inside. As a result, that PR made `any::swap(any&)` a lot less efficient.

```cpp
// What's the precondition of any-assignment and swap?
void scenarios() {
    std::any a = std::make_any<std::any>(42);
    std::any& a_o = *std::any_cast<std::any>(&a);
    if (1) { // Should this be supported?
        a = std::move(a_o);
        assert(std::any_cast<int>(a) == 42);
    } else { // And what about this?
        a_o = std::move(a);
    }
    // And these?
    // a.swap(a_o);
    // a_o.swap(a);
}
```

**So the question is, what's the precondition for `std::any`'s swap and assignment operators? Should we really support those use cases, i,e. can we just compare the address directly?**

Even if we cannot compare the address, I think we should still try to improve `any::swap`. By inlining the code, it turns out at least one copy is unnecessary.

Related:

The test case added in the original PR, and another PR trying to fix self-assignment:
https://github.com/microsoft/STL/pull/5413
https://github.com/microsoft/STL/pull/3965/files#diff-a0f3b054031c8ff90d8a2ad3fc1903f67bb1477d1a1fbf4217c02b55d2301dbe

The current implementation of `operator=` and `swap`:
https://github.com/microsoft/STL/blob/37d575ede5ade50ad95b857f22ed7f1be4b1f2df/stl/inc/any#L183-L186
https://github.com/microsoft/STL/blob/37d575ede5ade50ad95b857f22ed7f1be4b1f2df/stl/inc/any#L232-L234

https://eel.is/c++draft/any.assign#4.sentence-1
```
any& operator=(any&& rhs) noexcept;
Effects: As if by any(std​::​move(rhs)).swap(*this)[.](https://eel.is/c++draft/any.assign#4.sentence-1)
```

https://eel.is/c++draft/any.modifiers#19.sentence-1
```
void swap(any& rhs) noexcept;
Effects: Exchanges the states of *this and rhs[.](https://eel.is/c++draft/any.modifiers#19.sentence-1)
```

[...]

PR #5710, opened 2025-09-03
Optimize `any::swap`

achabense:
This pr removes an unnecessary copy in `any::swap`.

Hopefully fixes https://github.com/microsoft/STL/issues/5698.

Benchmark result:
```
------------------------------------------------------
Benchmark            Time             CPU   Iterations
------------------------------------------------------
bm<trivial>       14.5 ns         14.8 ns     49777778
bm<small>         21.1 ns         21.0 ns     32000000
bm<large>         15.2 ns         15.0 ns     44800000

->
------------------------------------------------------
Benchmark            Time             CPU   Iterations
------------------------------------------------------
bm<trivial>       10.1 ns         9.52 ns     64000000
bm<small>         15.9 ns         15.3 ns     40727273
bm<large>         10.6 ns         10.7 ns     64000000

```


stl/inc/any
```diff
@@ -230,7 +230,10 @@ public:
     }
 
     void swap(any& _That) noexcept {
-        _That = _STD exchange(*this, _STD move(_That));
+        any _Old = _STD move(*this);
+        *this    = _STD move(_That);
```
achabense, line 234: We are still paying an additional copy for some insane self-swaps here...

```cpp
std::any a = std::make_any<std::any>(42);
std::any& a_o = *std::any_cast<std::any>(&a);
a.swap(a_o);
a_o.swap(a);
```

benchmarks/src/any_swap.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include <benchmark/benchmark.h>
+#include <type_traits>
+
+using trivial = std::array<int, 2>;
+static_assert(std::is_trivially_copyable_v<trivial>);
+
+struct small {
+    std::array<int, 4> c{};
+    small()                                  = default;
+    small(const small& other)                = default;
```
achabense: Should remove the unused names (`other`).

benchmarks/src/any_swap.cpp
```diff
@@ -0,0 +1,50 @@
...
+BENCHMARK(bm<trivial, trivial>);
+BENCHMARK(bm<trivial, small>);
+BENCHMARK(bm<trivial, large>);
+
+BENCHMARK(bm<small, trivial>);
+BENCHMARK(bm<small, small>);
+BENCHMARK(bm<small, large>);
+
+BENCHMARK(bm<large, trivial>);
+BENCHMARK(bm<large, small>);
```
achabense: There is no significant difference between `bm<large, small>` and `bm<small, large>` etc.

stl/inc/any
```diff
@@ -230,7 +230,10 @@ public:
     }
 
     void swap(any& _That) noexcept {
-        _That = _STD exchange(*this, _STD move(_That));
+        any _Old = _STD move(*this);
+        _Assign(_STD move(_That));
```
achabense: We are still paying an additional copy for some insane self-swaps here...

```cpp
std::any a = std::make_any<std::any>(42);
std::any& a_o = *std::any_cast<std::any>(&a);
a.swap(a_o);
a_o.swap(a);
```

(Also, I think `_Assign(any-prvalue)` is somewhat misleading (for example https://github.com/microsoft/STL/pull/5413)), and it's better to create the temporal `any` explicitly in each function. For example, this would be v.)

```cpp
any _Old_this = _STD move(*this);
{
    any _Old_that = _STD move(_That);
    reset();
    _Move_from(_Old_that);
}
_That.reset();
_That._Move_from(_Old_this);

```

benchmarks/src/any_swap.cpp
```diff
@@ -0,0 +1,42 @@
...
+    for (auto _ : state) {
+        a.swap(b);
+        benchmark::DoNotOptimize(a);
+        benchmark::DoNotOptimize(b);
+    }
+}
+
+BENCHMARK(bm<trivial>);
+BENCHMARK(bm<small>);
+BENCHMARK(bm<large>);
```
achabense, line 40: The benchmark is very simple, but I think it's enough to show the compiler cannot optimize away the copy.

stl/inc/any
StephanTLavavej: I'm not concerned about the performance of pathological self-swaps, only their correctness.

StephanTLavavej: This move-constructs `_Assign`'s value parameter (as an aside, I also don't like that pattern), which seems to be unnecessary. Self-swap correctness should already be ensured by capturing `any _Old` above.

StephanTLavavej:
Thanks! :heart_eyes_cat: I pushed a change to further simplify and optimize `any::swap()`.

Here are benchmarks comparing `main`, your change (v2), and my change (v3). The speedups I measured on my 5950X for v2 were similar to what you measured (1.44, 1.33, 1.43).

Benchmark     |  `main` |      v2 |      v3 | v2 Speedup | v3 Speedup
--------------|---------|---------|---------|------------|-----------
`bm<trivial>` | 13.6 ns | 8.74 ns | 6.19 ns |       1.56 | 2.20
`bm<small>`   | 18.6 ns | 13.5 ns | 9.41 ns |       1.38 | 1.98
`bm<large>`   | 9.65 ns | 7.89 ns | 4.27 ns |       1.22 | 2.26

I believe that v3 passes our test coverage and handles self-swap scenarios equally well.

StephanTLavavej:

StephanTLavavej:
Thanks for optimizing and benchmarking this! :stopwatch: :racing_car: :tada:

achabense:
Initially, I didn't dare to remove that additional copy in `_Assign` in v2, as I thought that might change the behavior for some (cursed) self-swaps (swapping with internal `any`; see `a.swap(b)` and `b.swap(a)` below). Since the copy is removed finally (v3), this becomes more disturbing so I rechecked what really happens in v2, v3 and the original version (before this pr).

Summary:
- No version (even the original version) can actually deal with those swaps. (And the net effect of v2, v3 and the original version are the same.)
- It's because for such cases, the "correct behavior" doesn't exist at all, i.e. these swaps cannot be well defined.
- So v3 is correct and optimal for the current implementation.

(I'm only talking about cursed swaps here. All these versions are supposed to handle normal swaps correctly (including normal self-swaps).)

---

Suppose `any a` contains `any b`, which contains `auto c`.
- `a` is "large", as it contains another `any`. (large ~ contains only an owning pointer, and when moved-from the pointer gets exchanged.)
- `b` may contain whatever possible.
- (If `a` contains `b` via `unique_ptr<any>` instead, it's "small" but also exchanged when moved-from, so the following analysis applies too.)

v2 impl (equivalent version):
```cpp
any _Old_this = _STD move(*this);
{
    any _Old_that = _STD move(_That);
    reset();
    _Move_from(_Old_that);
}
_That.reset();
_That._Move_from(_Old_this);
```
Effects:
```cpp
// If `a.swap(b)` happens first:
any old_a = move(a);     // old_a contains (owns) &b, a is moved-from (exchanged / emptied).
{
    any old_b = move(b); // contains move(c), b is moved-from.
    a.reset();           // noop.
    a._Move_from(old_b); // a contains move(move(c))
}
b.reset();               // b is emptied.
b._Move_from(old_a);     // b owns &b. (!!)

// If `b.swap(a)` happens first:
any old_b = move(b);     // contains move(c), b is moved-from.
{
    any old_a = move(a); // owns &b, a is emptied.
    b.reset();           // b is emptied.
    b._Move_from(old_a); // b owns &b. (!!)
}
a.reset();               // noop.
a._Move_from(old_b);     // a contains move(move(c)).
```

v3 impl:
```cpp
any _Old = _STD move(*this);
reset();
_Move_from(_That);
_That.reset();
_That._Move_from(_Old);
```
Effects:
```cpp
// If `a.swap(b)` happens first:
any old_a = move(a); // old_a owns &b, a is emptied.
a.reset();           // noop.
a._Move_from(b);     // a contains move(c), b is moved-from.
b.reset();           // b is emptied.
b._Move_from(old_a); // b owns &b. (!!)

// If `b.swap(a)` happens first:
any old_b = move(b); // old_b contains move(c), b is moved-from.
b.reset();           // b is emptied.
b._Move_from(a);     // b owns &b, a is emptied. (!!)
a.reset();           // noop.
a._Move_from(old_b); // b contains move(c).
```

In all these cases (either v2 or v3, either `a.swap(b)` or `b.swap(a)`), `a` contains `move(c)` (or `move(move(c))`), but `b` "owns" itself finally (lines containing `(!!)` comment). The swap will not crash immediately as after the swap, no entity is responsible for destroying `b`, so its dtor is never called.

This seems a serious problem, but it turns out that, at least this pr doesn't make anything worse - the original `exchange`-based implementation already have the same problem.

Take the following examples for example. On my visual studio (using the original `exchange`-based version), the first example compiles to a program that quickly reaches several GBs. (Which suggests the same leak pattern.) The second program tries to destroy `b` and crash due to stack-overflow. (Which suggests the same self-owning pattern.)

```cpp
// 1
#include<any>
int main()
{
    for (;;) {
        std::any a = std::make_any<std::any>(42);
        std::any& b = *std::any_cast<std::any>(&a);
        a.swap(b);
    }
}

// 2
#include<any>
int main()
{
    std::any a = std::make_any<std::any>(42);
    std::any& b = *std::any_cast<std::any>(&a);
    a.swap(b);
    b.~any();
}
```

Initially, I thought this may be due to the implementation chooses to hold data in-place for trivial/small types. However, it turns out even if `std::any` always hold data via a pointer `p`, the same problem happens. For pointer-only implementation, the only reasonable way to implement `any::swap` is to swap their pointer:

```cpp
std::swap(a.p, b.p);
// (and swap type info)
```

Note that after this, the same issue happens - now `b` "owns" itself.

---

I believe these cursed swaps cannot be well-defined.

 Conceptually, `a` must be implicitly owned by something else, either on stack, allocated or inside some other classes, so the actual case is:

`a's unknown owner (can be stack)` contains `any a`, which contains `any b`, which contains `auto c`.

If the swap must be defined for `a` and `b`, the only meaningful effect (I can imagine) is to transfer `b`'s ownership to `a`'s initial owner, and let `b` contain `a` instead (and `a` contain `c`). But these are impossible to implement within `any::swap`. (Not generally implementable in C++.)

As to the current C++ standard, it only says the effect of `any::swap` being "exchanging the state". ([Ref.](https://eel.is/c++draft/any.modifiers#19.sentence-1)). (Which doesn't look quite helpful; I'm not familiar with the standard, so maybe some rules listed elsewhere already prohibit these swaps?)
```
void swap(any& rhs) noexcept;
Effects: Exchanges the states of *this and rhs.
```
(Copy and move assignments are not cursed likewise as they are explicitly defined in terms of creating a temporary then swap.)

StephanTLavavej:
Thanks for the careful analysis!

PR #5716, opened 2025-09-12
`<regex>`: Handle overflow and allocation failure when parsing repetitions

muellerj2:
This fixes two obscure bugs when parsing repetitions:

* The `_Loops` counter can potentially overflow, which will result in a crash during matching. The fix is to throw a `regex_error` (of suitable type `error_space`) when overflow is about to occur during parsing.
* The `_Node_end_rep` object can potentially be leaked  if the subsequent `new _Node_rep(...)` throws an exception. The simplest fix is to link the `_Node_end_rep` object into the NFA before allocating and constructing the `_Node_rep` object, so the `_Node_end_rep` object will be cleaned up together with the rest of the NFA by the `_Tidy_guard` in `_Parser2::_Compile` if an exception occurs.

No test coverage added because testing these bugs is impractical. In particular, the first bug occurs only after more than 500 GB have been allocated to represent the NFA.


StephanTLavavej:
Thanks for noticing these subtle problems and clearly explaining them! :heart_eyes_cat:

StephanTLavavej:

StephanTLavavej:
Thanks for noticing and fixing these corner cases! :eagle: :eye: :hammer_and_wrench:

PR #5718, opened 2025-09-14
<array>: Remove Unused Boolean

Xphalnos:
This variable was added in **https://github.com/microsoft/STL/pull/735** and seems to be useless.

PS: This is my first contribution here so please excuse me if I make any mistakes ;)

### DIFF
--- a/stl/inc/array
+++ b/stl/inc/array
@@ -115,8 +115,6 @@ public:
         return _Ptr;
     }
 
-    static constexpr bool _Unwrap_when_unverified = true;
-
     constexpr void _Seek_to(pointer _It) noexcept {
         _Ptr = _It;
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
Thanks for looking into this. It isn't dead, as indicated by the failing tests. It's used here:

https://github.com/microsoft/STL/blob/0cb8132f7eb6e26c3a79899bd0b61953bfbfc3d0/stl/inc/xutility#L1459-L1468

In general, I recommend using global search (Ctrl+Shift+F in VSCode) when trying to understand if something is used and, if so, where. Not all usage is local.

Closing this PR without merging, as it would break the repo, but we welcome future contributions! :cat:

PR #5721, opened 2025-09-15
`<complex>`: use hardware FMA on Clang too

AlexGuteniev:
Yes it does not compile without `[[gnu::target("fma")]]` magic spell.

### DIFF
--- a/stl/inc/complex
+++ b/stl/inc/complex
@@ -21,8 +21,6 @@
 // https://learn.microsoft.com/en-us/cpp/build/arm64-windows-abi-conventions?view=msvc-170#base-requirements
 // Both floating-point and NEON support are presumed to be present in hardware.
 #define _FMP_USING_STD_FMA
-#elif defined(__clang__) // ^^^ defined(_M_ARM64) || defined(_M_ARM64EC) ^^^
-// TRANSITION, not using x86/x64 FMA intrinsics for Clang yet
 #elif defined(_M_IX86) || defined(_M_X64)
 #define _FMP_USING_X86_X64_INTRINSICS
 #include <__msvc_bit_utils.hpp>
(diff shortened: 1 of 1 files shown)


stl/inc/complex
```diff
@@ -167,7 +165,11 @@ namespace _Float_multi_prec {
     }
 
 #ifdef _FMP_USING_X86_X64_INTRINSICS
-    _NODISCARD inline double _Sqr_error_x86_x64_fma(const double _Xval, const double _Prod0) noexcept {
+#ifdef __clang__
+    [[gnu::target("fma")]]
```
philnik777:
```suggestion
    [[__gnu__::__target__("fma")]]

```

StephanTLavavej:
Thanks for figuring out the magic incantation here! :heart_eyes_cat:

I observe that this is inconsistent with the spelling of https://github.com/microsoft/STL/blob/43e96b29a971a7a26aceaa539d22cbedfcf13687/stl/inc/atomic#L33 but I don't care enough to iterate.

AlexGuteniev:
> I observe that this is inconsistent with the spelling of

If you want consistency, I'd change `<atomic>` actually.

The spell in `<atomic>` allows cx16 insn in any function, and this theoretically permits generating one out of thin air. In contrast, the approach in this PR allows fma only in specific function, for which we have runtime check.

`<atomic>` does not centralize cx16, and there would be few functions, so it takes more changes, but still may be worth doing.

StephanTLavavej:
Oh, all I meant was the spelling of `__gnu__::__target__`.

AlexGuteniev:
Aha! It was suggested by @philnik777. Makes sense as a defense from user's macro, `#define gnu 1` is allowed. And in pragma there's no preprocessor, since pragma itself is considered preprocessor.

StephanTLavavej:

StephanTLavavej:
Thanks for multiplying and adding to the compilers where we use these intrinsics! :joy_cat: :heavy_multiplication_x: :heavy_plus_sign:

PR #5725, opened 2025-09-16
`<vector>`: Fix UB in `vector<bool>` by rotating instead of shifting

### LINKED ISSUE #5720: `<vector>`: The `vector<bool>` optimization for `copy()` performs a forbidden negative shift
Found by Clang 20 UBSan in `GH_000625_vector_bool_optimization`. Reduced:

```
```
```cpp

#include <algorithm>
#include <cassert>
#include <cstddef>
#include <vector>
using namespace std;

const vector<bool> source = { //
    true, false, true, false, true, true, true, false, //
    true, false, true, false, true, true, true, false, //
    true, false, true, false, true, true, true, false, //
    true, false, true, false, true, true, true, false, //
    true};

void test_copy_offset_source(const size_t length) {
    const vector<bool> result = {//
        /***/ false, true, false, true, true, true, false, //
        true, false, true, false, true, true, true, false, //
        true, false, true, false, true, true, true, false, //
        true, false, true, false, true, true, true, false, //
        true};

    vector<bool> dest(length, false);
    const auto first    = source.begin() + 1;
    const auto last     = first + length;
    const auto output   = dest.begin();
    const auto res_copy = copy(first, last, output);
    assert(dest == result);
    assert(res_copy == dest.end());
}

int main() {
    test_copy_offset_source(32);
}
```
```
D:\GitHub\STL\out\x64>clang-cl /EHsc /nologo /W4 /std:c++latest -fsanitize=undefined /Zi /Fdmeow.pdb meow.cpp /link /ignore:4217 && meow
   Creating library meow.lib and object meow.exp
D:\GitHub\STL\out\x64\out\inc\vector:3863:70: runtime error: shift exponent 18446744073709551615 is too large for 32-bit type '_Vbase' (aka 'unsigned int')
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior D:\GitHub\STL\out\x64\out\inc\vector:3863:70
```

This line is (please ignore the line number difference due to clang-formatting):

https://github.com/microsoft/STL/blob/0cb8132f7eb6e26c3a79899bd0b61953bfbfc3d0/stl/inc/vector#L3861-L3862

Here, `_LastShift` is `-1`, which UBSan has correctly detected.

In many other places we have logic to detect whether we need to shift left or right. Instead of spot-patching this location, we should think deeply about whether any other places could be affected, or if larger changes are necessary. We should also think about whether we could get different test inputs to actually fail (because right now, the test happens to pass without UBSan).

The one bit of good news is that this appears to be the only occurrence of UB found in this optimization.

AlexGuteniev:
Fixes #5720.
The UB manifested as correctly working machine code.
The fix imitates the UB manifestation without running into the UB,

The negative big shift happened to turn into positive large shift.
The rotate is performance equivalent<sup>1</sup>, that can accept negative amount.

Beware, intrinsic! Seems to work even in `/clr:pure` though, as I don't see any attempts to escape it in `<bit>`.

___
<sup>1</sup> On x64. Apparently, on ARM64, `rotr` and swapping the subtraction would be performance equivalent, and `rotl` is slightly slower. Still keeping `rotl` to minimize the amount of changes.

---

This seems to open possibility of considering rotates elsewhere, collapsing conditionals and branches. May look into this later.

### DIFF
--- a/stl/inc/vector
+++ b/stl/inc/vector
@@ -3859,8 +3859,8 @@ _CONSTEXPR20 _OutIt _Copy_vbool(_VbIt _First, _VbIt _Last, _OutIt _Dest) {
 
         const auto _DestMask = _FirstDestMask | (_DestEnd._Myoff == 0 ? 0 : _LastDestMask);
         if (_Last._Myoff != 0) {
-            const auto _LastShift     = _DestEnd._Myoff - _Last._Myoff;
-            const auto _LastSourceVal = (*_VbLast & _LastSourceMask) << _LastShift;
+            const int _LastRotate     = static_cast<int>(_DestEnd._Myoff) - static_cast<int>(_Last._Myoff);
+            const auto _LastSourceVal = _rotl(*_VbLast & _LastSourceMask, _LastRotate);
             *_VbDest                  = (*_VbDest & _DestMask) | _SourceVal | _LastSourceVal;
         } else {
             *_VbDest = (*_VbDest & _DestMask) | _SourceVal;


AlexGuteniev:
Closed as introducing intrinsic usage is not the best way to fix this bug.

PR #4709, opened 2024-06-03
Block building STL in wrong command prompt

fsb4000:
Fixes #4639 

I didn't test ARM command prompts.

изображение
изображение
изображение
изображение

PR #5731, opened 2025-09-19
Block building for the wrong target arch in Developer Command Prompt

cpplearner:
#4709 and #4717 make it impossible to build in the wrong command prompt (e.g. if someone builds the STL in an "x64 Native Tools Command Prompt" and runs the tests in an "x86 Native Tools Command Prompt"). But they are ineffective in a "Developer Command Prompt" because they examine the `Platform` env var, which is not defined there.

This PR examines `VSCMD_ARG_TGT_ARCH` instead, which is defined in both the Developer Command Prompt and the Native Tools Command Prompt.

It's unclear to me whether it is safe to use such an undocumented env var. But anyway, if someone encounters an unexpected error, they can just manually unset `VSCMD_ARG_TGT_ARCH` when building.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -22,7 +22,7 @@
       },
       "condition": {
         "type": "inList",
-        "string": "$env{Platform}",
+        "string": "$env{VSCMD_ARG_TGT_ARCH}",
         "list": [
           "x64",
           ""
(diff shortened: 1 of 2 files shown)


StephanTLavavej:

StephanTLavavej:
Thanks for noticing that Diablo Canyon 1 wasn't like Diablo Canyon 2! :radioactive: :eagle: :eye:

PR #5732, opened 2025-09-19
Improve existing `vector<bool>` benchmarks

AlexGuteniev:
The following minor changes:
 - Extract `createRandomVector`, use the version that doesn't use `static`, and rename it to `random_bool_vector` for consitency.
 - Use optimization barriers, and avoid `const` on `source` to make them effective.
 - Remove one stray newline.


StephanTLavavej:

StephanTLavavej:
:broom: :stopwatch: :smile_cat:
===

PR #5734, opened 2025-09-22
`<regex>`: Process negative lookahead assertions non-recursively

muellerj2:
Towards #997 and #1528. Negative lookahead assertions can be handled very similarly to positive ones, but two things have to be kept in mind after the assertion is processed: (a) Failure must be inverted and (b) one must reset not just the position (as is sufficient for positive lookahead assertions), but also the capture groups.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -1674,7 +1674,10 @@ public:
     }
 };
 
-enum class _Rx_unwind_ops { _After_assert = _N_end + 1 };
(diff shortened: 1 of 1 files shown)


StephanTLavavej:

StephanTLavavej:
Thanks for looking ahead to a future of negative stack space consumption! :joy_cat: :tada: :rofl:

PR #5742, opened 2025-09-26
Fix non-inline `random_bool_vector` in benchmarks

AlexGuteniev:
Fuse `random_bool_vector` into  `random_vector`

### DIFF
--- a/benchmarks/inc/utility.hpp
+++ b/benchmarks/inc/utility.hpp
@@ -13,12 +13,11 @@
 template <class Contained, template <class> class Alloc = std::allocator>
 std::vector<Contained, Alloc<Contained>> random_vector(size_t n) {
     std::vector<Contained, Alloc<Contained>> res(n);
+    std::mt19937_64 prng;
 
     if constexpr (std::is_same_v<Contained, bool>) {
-        std::mt19937 prng;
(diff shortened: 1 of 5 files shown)


benchmarks/inc/utility.hpp
```diff
@@ -7,30 +7,27 @@
...
 #include <vector>
 
 template <class Contained, template <class> class Alloc = std::allocator>
 std::vector<Contained, Alloc<Contained>> random_vector(size_t n) {
-    std::mt19937_64 prng;
-
     std::vector<Contained, Alloc<Contained>> res(n);
 
+    if constexpr (std::is_same_v<Contained, bool>) {
+        std::mt19937 prng;
```
StephanTLavavej: I'd actually like us to just always use `mt19937_64`; IIRC it consumes more stack space but it's not slower on x64 and can actually be faster (e.g. when generating 64-bit indices, although that's obviously not the case here). That would simplify this function further by extracting it.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for cleaning up and unifying this machinery! :broom: :heart_eyes_cat: :rocket:

PR #5743, opened 2025-09-26
`vector_algorithms.cpp`: Improve conformance by adding missing `template` and `typename` for dependent things

AlexGuteniev:
Found by compiling with `clang-cl stl\src\vector_algorithms.cpp -std:c++latest -march=x86-64-v3`.

Whereas we don't build STL with that, there are still reason why compiling this particular file with `clang-cl` may be useful:
 * Catch instructions not available on a particular arch. If `-march=x86-64-v3` is dropped, there would be a lot of target errors, that are solved by placing `gnu::target` attributes. When they are placed expectedly, unexpected instruction on an unverified properly code path will trigger an error.
 * Run with `-fsanitize=undefined` to avoid any UB.
 * Conformance by itself is good.

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -4398,21 +4398,21 @@ namespace {
                 if constexpr (_Needle_length_el_magnitude >= 1) {
                     _Eq = _Traits::_Cmp_avx(_Data1, _Data2s0);
                     if constexpr (_Needle_length_el_magnitude >= 2) {
-                        const __m256i _Data2s1 = _Traits::_Shuffle_avx<1>(_Data2s0);
+                        const __m256i _Data2s1 = _Traits::template _Shuffle_avx<1>(_Data2s0);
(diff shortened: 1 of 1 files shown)


AlexGuteniev:
Thanks WG21-P0634R3 for making this only mildly annoying!

StephanTLavavej (approved):

StephanTLavavej:
I'm surprised that MSVC doesn't enforce these in strict mode - I think it'd be worth reporting a DevCom bug.

AlexGuteniev:
> I think it'd be worth reporting a DevCom bug.

DevCom-540547

<details>
<summary>Reported in 2019</summary>

a7brk3
</details>

StephanTLavavej:

StephanTLavavej:
Thanks for cleaning up all the broken glass from where MSVC let us be bad kitties, jumping up on the table where we're not allowed and batting objects off! :joy_cat: :cat2: :boom:

PR #5744, opened 2025-09-26
`vector_algorithms.cpp`: remove unused functions

AlexGuteniev:
Discovered while experimenting with `clang-cl stl\src\vector_algorithms.cpp -std:c++latest` command.

A couple of functions are non-unique, and should be removed.

Two other ones are also exiting this file, as scope guard is used instead.


StephanTLavavej:

StephanTLavavej:
:low_battery: :wastebasket: :magic_wand:
===

PR #5745, opened 2025-09-26
`<regex>`: Process disjunctions non-recursively

muellerj2:
Towards #997 and #1528.

We have to handle disjunctions (represented by `_N_if` nodes) differently depending on longest mode: When `_Longest` is true, we have to evaluate all alternatives during unwinding no matter if they matched successfully or not. Otherwise, we always may (and usually must) stop evaluating alternatives after one succeeded. The former is implemented by the case `_Disjunction_eval_alt_always`, the latter by `_Disjunction_eval_alt_on_failure`. We can implement the latter case by falling through to the former.

During unwinding, `_Frame` already holds the correct values to evaluate further alternatives in the disjunctions except for the NFA node it points to. So we can just update the node pointer and "push" `_Frame` on the stack again by incrementing `_Frames_count`.

In longest mode, `_Match_pat()` now returns false when matching the last evaluated alternative failed, even if some alternative evaluated earlier matched successfully. `_Matcher3::_Match()` can deal with this behavior change by checking the member `_Matched` as well, which is set to true in longest mode when the first successful match was found. (Note that in longest mode, the return value of `_Match_pat()` mostly doesn't matter when possible trajectories are evaluated by the matcher. The only exceptions are the `_Match_pat(_Node->_Next)` calls found in `_Do_rep0()`, which handles simple loops. But the return values of these calls are unaffected by this change in the evaluation of `_N_if` nodes because simple loops are guaranteed to be branchless.)

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -1677,6 +1677,8 @@ public:
 enum class _Rx_unwind_ops {
     _After_assert = _N_end + 1,
     _After_neg_assert,
+    _Disjunction_eval_alt_on_failure,
+    _Disjunction_eval_alt_always,
 };
 
 template <class _BidIt>
@@ -1753,7 +1755,7 @@ public:
 
         _Matched = false;
 
-        bool _Succeeded = _Match_pat(_Rep);
+        bool _Succeeded = _Match_pat(_Rep) || _Matched;
         _STL_INTERNAL_CHECK(_Frames_count == 0);
 
         if (!_Succeeded) {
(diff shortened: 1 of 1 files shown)


AlexGuteniev:
What is the performance impact?

muellerj2:
There has been no noticeable performance impact due to changes since #5714, as far as I can tell: The `regex_search` benchmarks and test runs seem to take basically the same amount of time.

But there will be a major change soon: Eliminating `_Do_rep0()` should generally improve performance for greedy simple loops (think `a*`), in some cases drastically.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:heart_eyes_cat: :green_heart: :100:
===

PR #5740, opened 2025-09-25
`<random>`: Prevent full-width shifts in `independent_bits_engine`

MattStephanson:
> _It's perfectly stable, so long as tha don't shift it about too much...._
>
> &mdash;Thomas Pynchon, _Mason & Dixon_

Fixes #5719.

The second loop in `operator()` is fine because it's only executed if multiple random values are needed (modulo the rejection sampling), i.e., `_Nx > 1`. In such cases, `_Wx0 <= _Wx / _Nx <= sizeof(_Eres)*8 / 2`, so `_Wx0+1 < sizeof(_Eres)*8` as long as `_Eres` is more than two bits.

Issue #5719, opened 2025-09-14
`<random>`: `independent_bits_engine` performs forbidden full shifts


StephanTLavavej:
Found by Clang 20 UBSan in `P0952R2_new_generate_canonical`'s usage of `independent_bits_engine`. Reduced:

```
D:\GitHub\STL\out\x64>type meow.cpp
```
```cpp
#include <cstddef>
#include <cstdint>
#include <print>
#include <random>
#include <string_view>
using namespace std;

template <class Engine, size_t W>
void test(const string_view name) {
    println("test<{}, {}>():", name, W);
    independent_bits_engine<Engine, W, uint64_t> ibe;
    const auto val = ibe();
    println("val: {}", val);
    println();
}

int main() {
    test<mt19937_64, 64>("mt19937_64");
    test<mt19937, 64>("mt19937");
    test<mt19937, 32>("mt19937");
}
```
```
D:\GitHub\STL\out\x64>clang-cl /EHsc /nologo /W4 /std:c++latest -fsanitize=undefined meow.cpp /Zi /Fdmeow.pdb /link /ignore:4217 && meow
   Creating library meow.lib and object meow.exp
test<mt19937_64, 64>():
D:\GitHub\STL\out\x64\out\inc\random:1977:25: runtime error: shift exponent 64 is too large for 64-bit type '_Eres' (aka 'unsigned long long')
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior D:\GitHub\STL\out\x64\out\inc\random:1977:25
D:\GitHub\STL\out\x64\out\inc\random:1977:34: runtime error: shift exponent 64 is too large for 64-bit type '_Eres' (aka 'unsigned long long')
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior D:\GitHub\STL\out\x64\out\inc\random:1977:34
D:\GitHub\STL\out\x64\out\inc\random:1978:27: runtime error: shift exponent 64 is too large for 64-bit type '_Eres' (aka 'unsigned long long')
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior D:\GitHub\STL\out\x64\out\inc\random:1978:27
D:\GitHub\STL\out\x64\out\inc\random:1978:42: runtime error: shift exponent 64 is too large for 64-bit type '_Eres' (aka 'unsigned long long')
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior D:\GitHub\STL\out\x64\out\inc\random:1978:42
[...]

PR #5717, opened 2025-09-13
Toolset update: MSVC Compiler 19.50 Preview 1, Clang 20

StephanTLavavej:
# :scroll: Changelog
- Improved test coverage:
  * Enabled ARM64EC test coverage for Clang.
- Code cleanups:
  * Removed compiler bug workarounds.
- Infrastructure improvements:
  * Updated dependencies.
    + Updated MSVC Compiler to 19.50 Preview 1 (now required).
    + Updated Clang to 20.1.8 (now required).
    + Updated Python to 3.13.7.

# :world_map: Overview
This updates our toolset to the MSVC Compiler 19.50 Preview 1 and Clang 20 in VS 2026 Insiders. As a new major version, we have a greater-than-usual number of things that need to be updated. I'm overhauling the README, simplifying the installer instructions. I've centralized the path to VsDevCmd.bat in our Azure Pipelines machinery, making it easier to update in the future. Both MSVC and Clang accumulated a lot of bugfixes, so I can remove a lot of workarounds, but in a few cases I had to report new bugs and add workarounds. Enough constexpr misbehavior was fixed that the remaining list wasn't as daunting, so I went through our libcxx skips and reported all of the remaining MSVC-specific bugs we're working around. Clang 20 allows us to enable ARM64EC which went surprisingly well (only coroutines were blocked). EDG shipped behavior changes to `<source_location>`, allowing us to remove workarounds that were dealing with the difference between the internal and public versions. CUDA rejects the 19.5x compiler version, so once again we need to activate their escape hatch. (We've informed NVIDIA and whenever CUDA begins allowing MSVC again, we'll update our minimum required version in the STL.)

As usual with a new major version of Clang, I'm updating our `.clang-format` baseline and regenerating the formatting. (This time I didn't see any new options that were worth configuring, although I experimented with parenthesis removal, which identified a few places to clean up.) I also took this opportunity to review all of our clang-format suppressions, removing them when possible. All of the remaining suppressions are now marked with a reason, using their supported syntax.

# :gear: Commits
* Python 3.13.7.
* Update VS URL from 17 Preview Enterprise to 18 Insiders Community.
* New pool.
* Generalize to "Windows 11 SDK (10.0.26100) or later".
  + VS is currently shipping 10.0.26100.4654, but the patch versions aren't significant to us.
* Require MSVC Compiler 19.50 Preview 1.
[...]

PR #5746, opened 2025-09-29
Reactivate Clang UBSan test coverage

### LINKED ISSUE #3568: Investigate why Clang/LLVM UBSan doesn't link internally

StephanTLavavej:
Fixes #3568.

This was briefly enabled for Clang 15 by #3452 on 2023-03-13, then disabled for Clang 16 by #3711 on 2023-05-18. After updating to Clang 20 by #5717 on 2025-09-15, we can re-enable this test coverage.

Note that `tests/utils/stl/test/features.py` disabled the configurations. We accumulated a couple of changes to `tests/std/tests/VSO_0000000_vector_algorithms_floats/env.lst` and `tests/std/tests/usual_matrix.lst` commenting out configurations, but I believe those changes were unnecessary. This PR updates `features.py` to enable the feature, uncomments the commented-out configurations, and doesn't change the large number of already-uncommented configurations across other files.

I audited our `env.lst` and related files, and I believe our UBSan coverage is reasonably comprehensive now. For the tests it's missing from, they either have good reasons (e.g. compile-only, `/clr` specific, incompatible with Clang for other reasons, etc.) or already have a blizzard of configurations where UBSan wouldn't be worth the extra cost.

Notably, UBSan has never been enabled for libcxx, and I am not attempting to do so here. libcxx configurations are extremely expensive (that's why we have only three). It may be reasonable to explore this in the future, but I don't want to spend developer and machine time on that now.

I'll leave Clang UBSan disabled for the MSVC-internal test harness (this was `src/qa/VC/shared/testenv/bin/run.pl` in MSVC-PR-457627, for future reference). I don't know if it works now, but it would provide negative value by burning a lot of machine time when GitHub is providing sufficient coverage of the STL's product and test code.

We've already merged a couple of fixes for bugs in `vector<bool>` and `independent_bits_engine` found by this impending PR:

* #5720
  + Fixed by @AlexGuteniev with #5726
* #5719
  + Fixed by @MattStephanson with #5740

# :gear: Commits
* Activate UBSan.
* Avoid `duration_cast()` UB, tracked by LWG-3921.
  + `D:\GitHub\STL\out\x64\out\inc\__msvc_chrono.hpp:451:52: runtime error: signed integer overflow: 9223372036854775806 * 1000000 cannot be represented in type '_CR' (aka 'long long')`
* Avoid UB caused by deliberate misalignment with `#pragma pack`.
  + `D:\GitHub\STL\tests\std\tests\VSO_0000000_vector_algorithms_mismatch_and_lex_compare\test.cpp:177:9: runtime error: reference binding to misaligned address 0x002ec5cfe1bb for type 'short[1]', which requires 2 byte alignment`
* Permanently avoid `wmemcmp()` UB, reported as VSO-2583476 "`wmemcmp()` triggers UB with a misaligned load".
  + `C:\Program Files (x86)\Windows Kits\10\include\10.0.26100.0\ucrt\wchar.h:458:37: runtime error: load of misaligned address 0x7ff647c3b7bc for type 'unsigned long long', which requires 8 byte alignment`
  + The offending line is: `unsigned __int64 __v1 = *(unsigned __int64*)__s1;`
* Properly detect UBSan in test code.
  + This is necessary because MSVC doesn't understand `__has_feature`. (I initially missed this because I was testing Clang only.)

### DIFF
--- a/tests/std/tests/VSO_0000000_vector_algorithms_mismatch_and_lex_compare/test.cpp
+++ b/tests/std/tests/VSO_0000000_vector_algorithms_mismatch_and_lex_compare/test.cpp
@@ -172,9 +172,13 @@ namespace test_mismatch_sizes_and_alignments {
 
     template <class T, size_t Size, size_t PadSize>
     char stack_array_various_alignments_impl() {
+#if defined(__clang__) && __has_feature(undefined_behavior_sanitizer)
+        // Avoid testing misaligned inputs, which UBSan would reject.
+#else // ^^^ UBSan / no UBSan vvv
         with_pad<T, Size + 1, PadSize + 1> a = {};
         with_pad<T, Size + 1, PadSize + 1> b = {};
         assert(mismatch(begin(a.v), end(a.v), begin(b.v), end(b.v)) == make_pair(end(a.v), end(b.v)));
+#endif // ^^^ no UBSan ^^^
         return 0;
     }
 
--- a/tests/std/tests/VSO_0000000_vector_algorithms_mismatch_and_lex_compare/test.cpp
+++ b/tests/std/tests/VSO_0000000_vector_algorithms_mismatch_and_lex_compare/test.cpp
@@ -172,7 +172,13 @@ namespace test_mismatch_sizes_and_alignments {
 
     template <class T, size_t Size, size_t PadSize>
     char stack_array_various_alignments_impl() {
-#if defined(__clang__) && __has_feature(undefined_behavior_sanitizer)
+#ifdef __clang__
+#if __has_feature(undefined_behavior_sanitizer)
+#define TESTING_UBSAN
+#endif
+#endif
+
+#ifdef TESTING_UBSAN
         // Avoid testing misaligned inputs, which UBSan would reject.
 #else // ^^^ UBSan / no UBSan vvv
         with_pad<T, Size + 1, PadSize + 1> a = {};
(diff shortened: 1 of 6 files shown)


tests/std/tests/VSO_0000000_vector_algorithms_mismatch_and_lex_compare/test.cpp
```diff
@@ -172,9 +172,19 @@ namespace test_mismatch_sizes_and_alignments {
 
     template <class T, size_t Size, size_t PadSize>
     char stack_array_various_alignments_impl() {
+#ifdef __clang__
+#if __has_feature(undefined_behavior_sanitizer)
```
AlexGuteniev, line 176: Do you prefer nested preprocessor instead of preprocessor logical and here?

cpplearner: MSVC treats `defined(__clang__) && __has_feature(undefined_behavior_sanitizer)` as a syntax error, so it's necessary to use nested preprocessor here.

StephanTLavavej: Yep. Unknown object-like macros expand to `0` (with an off-by-default warning) but unknown function-like macros are a syntax error, as I discovered.

davidmrdavid (approved):
LGTM, ty

StephanTLavavej:

PR #5753, opened 2025-10-03
Final cleanups for `<experimental/filesystem>`

StephanTLavavej:
Before I place `<experimental/filesystem>` into cryostasis, there are a couple of things that I want to clean up, in case we ever need to look at it again while trying to understand legacy code:

* `_FS_DLL` => `extern "C" _CRTIMP2_PURE`
  + This macro obscured how we're dllexporting the separately compiled machinery. Nothing else in the STL used this pattern. It was also never overridden, so the `#ifndef _FS_DLL` was pointless.
* `_Pchar` => `wchar_t`
  + This alias obscured the `path`'s character type, and wasn't consistently used.


StephanTLavavej:

MahmoudGSaleh (approved):
LGTM

PR #5759, opened 2025-10-04
`vector_algorithms.cpp`: avoid cast to integral in `__std_bitset_from_string_N`

AlexGuteniev:
This avoids potential misaligned access, which can only happen if the user has managed to misalign the bitset itself...

Meh, but it is still cleaner and consistent with using `memcpy` elsewhere.

The benchmark results are the same.


StephanTLavavej:

StephanTLavavej:
Thanks for cleaning up this unnecessarily scary cast! :ghost: :jack_o_lantern: :vampire:

PR #5761, opened 2025-10-04
`vector_algorithms.cpp`: remove incorrect `__assume`

AlexGuteniev:
The idea behind that `__assume` was to tell the compiler that `memcpy` is small. 

Unfortunately, the compilers are still only able to optimize `memcpy` only for exactly known size, not "small" size or other size characteristics.

The `__assume` was also incorrect. as the comparison was bits vs bytes.


StephanTLavavej:

StephanTLavavej:
Goodbye, single line of code! :tada: :cat: :tada:

PR #5765, opened 2025-10-04
Remove non-Standard `<experimental/filesystem>`

StephanTLavavej:
# Overview
This permanently removes our non-Standard `<experimental/filesystem>` header, which was superseded by C++17 `<filesystem>`.

For 6 years ([VS 2019 16.3](https://devblogs.microsoft.com/cppblog/c20-features-and-fixes-in-vs-2019-16-1-through-16-6/) in September 2019), `<experimental/filesystem>` has been "hard deprecated" with an impossible-to-ignore compiler error telling users that "The `<experimental/filesystem>` header providing `std::experimental::filesystem` is deprecated by Microsoft and will be REMOVED." It's finally time to deliver on that promise. (Also, the name inherently indicated that the technology was tentative and not eternal.)

Eliminating this code removes unnecessary complexity from this fiendishly complex library. For users, it removes ways to write non-portable code. For library implementers, it slightly reduces our maintenance burden, making it easier for us to reason about the remaining Standard code.

The interface of `<experimental/filesystem>` is fairly close to that of Standard `<filesystem>`, making migration easy in most cases. However, the implementation of `<experimental/filesystem>` was riddled with deficiencies and limitations (including `MAX_PATH` limitations). This is a major reason to push users to migrate to Standard `<filesystem>`, whose implementation was written from scratch to be conforming. Unfortunately, we can't *completely* remove the `<experimental/filesystem>` implementation, as its separately compiled component must be retained for binary compatibility. But we can still remove the header itself, and some support logic in other headers.

Because of the retained-for-bincompat caveat, I am preserving test coverage for `<experimental/filesystem>`. Most of our retained-for-bincompat functions have no test coverage, and we simply rely on not changing them, but the implementation of `<experimental/filesystem>` was large and complicated enough that I think preserving test coverage is worth the effort.

# Commits
* Move: `tests/std/include/experimental_filesystem.hpp`
* Add a cryostasis banner.
* Remove `<experimental/filesystem>` from product code.
  + If someone's looking at Standard `filesystem.cpp`, we don't need to keep a comment pointing out Experimental `filesys.cpp`.
* Remove the feature-test macro `__cpp_lib_experimental_filesystem`.
  + I don't think it's necessary to transfer this to the cryopreserved `experimental_filesystem.hpp`.
* Remove `_FSTREAM_SUPPORTS_EXPERIMENTAL_FILESYSTEM`, simplify Standard `filesystem::path` detection.
  + After removing support for `experimental::filesystem::path`, we can simplify away `_Is_any_path`. Now, those constructors can be guarded for `_HAS_CXX17`, and they can use `is_same_v` exactly as depicted by WG21-N5014 \[ifstream.cons\]/4, \[ofstream.cons\]/4, \[fstream.cons\]/4. They remain templates, so there's no ABI impact to their definitions completely vanishing in C++14 mode.
* Update `src/filesys.cpp` to compile standalone.
  + Add a TRANSITION, ABI comment.
  + We don't need to define the silencing macro, because we can't include `<experimental/filesystem>` anymore.
  + We use `wstring`, so we need to include `<string>`. (This was previously dragged in, at least partially.)
  + We use `uint64_t` and `uintmax_t`, so we should include `<cstdint>` to be good kitties.
  + We need the `_MAX_FILESYS_NAME`, `_FS_BEGIN`, and `_FS_END` macros.
  + Within the namespace, we need `file_type`, `perms` and its `_BITMASK_OPS`, and `space_info`.
* Drop coverage in `include_each_header_alone_matrix.lst`.
* Update test coverage to include `experimental_filesystem.hpp`.
  + `tests/tr1/tests/filesystem1/test.cpp` is still allowed to say that it's testing `<experimental/filesystem>`;
rephrasing this wasn't worth the effort.
* Update `Dev09_172666_tr1_tuple_odr` to include `experimental_filesystem.hpp` and dual-mode test Standard filesystem.
  + It was previously dragging in `<experimental/filesystem>` via `<__msvc_all_public_headers.hpp>`, which is why I missed it in #5749.


StephanTLavavej:

StephanTLavavej:
I've pushed a merge with `main` to resolve trivial adjacent-delete conflicts with #5764 in `stl/inc/__msvc_all_public_headers.hpp` and `tests/std/tests/include_each_header_alone_matrix.lst`.

Issue #1754, opened 2021-03-17
<algorithm>: investigate sharing code between std and ranges algorithms


CaseyCarter:
There's an enormous amount of overlap in implementations of the "same" algorithm in namespaces `std` and `std::ranges`. For example, `std::copy`:

https://github.com/microsoft/STL/blob/2d890fdadde44bdb8eced76f29803ce2ecbe0378/stl/inc/xutility#L4119-L4146

vs. `ranges::copy`:

https://github.com/microsoft/STL/blob/2d890fdadde44bdb8eced76f29803ce2ecbe0378/stl/inc/algorithm#L1507-L1547

There are substantial differences in the interface, but eventually they boil down to identical or nearly identical loops. It would be nice if we could reduce the total amount of code we must maintain by using a common back-end for such algorithms.

Note that in doing so we absolutely must not regress performance or codesize, and should not regress compile throughput or increase the complexity of the overall code.

PR #5769, opened 2025-10-07
Optimize `std::transform` for `vector<bool>`

AlexGuteniev:
Towards #625, specifically https://github.com/microsoft/STL/issues/625#issuecomment-3307182440 items 1 and 2.

# 🦖 Optimization

When a standard functor, either transparent or integer-specialized, passed to `transfrom`, along with all `vector<bool>` iterators, map that functor and operator on the underlying type.

Only do this for zero-offset, supporting all possible offset combination is much complexity for a little gain.

Extract pointers from iterators to help the compiler auto-vectorize.

Don't call `transform` again, to avoid unnecessary recursion, the operation is simple.

Don't process tails explicitly, yield to the existing loop for now.

Don't do ranges yet. Other `vector<bool>` optimization don't do them either. It is getting complicated, so instead of doing ranges separately, need to look into #1754 at last.

# 🏁 Benchmark

Feed the randomizer with some seed to make sure the inputs are different.

Since (auto-)vectorization is engaged, use alignment controlling allocator.

# ⏱️ Benchmark results

Benchmark                                           |   Before  |  After  | Speedup
----------------------------------------------------|-----------|---------|----------
`transform_two_inputs_aligned<logical_and<>>/64`    |    108 ns | 2.55 ns |    42.35
`transform_two_inputs_aligned<logical_and<>>/4096`  |  13869 ns | 9.44 ns | 1,469.17
`transform_two_inputs_aligned<logical_and<>>/65536` | 416424 ns |  115 ns | 3,621.08
`transform_two_inputs_aligned<logical_or<>>/64`     |    123 ns | 2.59 ns |    47.49
`transform_two_inputs_aligned<logical_or<>>/4096`   |  14377 ns | 9.07 ns | 1,585.12
`transform_two_inputs_aligned<logical_or<>>/65536`  | 409012 ns |  112 ns | 3,651.89
`transform_one_input_aligned<logical_not<>>/64`     |   83.7 ns | 2.14 ns |    39.11
`transform_one_input_aligned<logical_not<>>/4096`   |   6891 ns | 7.28 ns |   946.57
`transform_one_input_aligned<logical_not<>>/65536`  | 264957 ns | 82.7 ns | 3,203.83


stl/inc/functional
```diff
@@ -258,6 +258,56 @@ struct bit_not<void> {
...
+};
+
+template <class _Ty>
+struct _Map_vb_functor<bit_xor<_Ty>> {
+    using _Type = conditional_t<_Is_vbool_functor_arg<_Ty>, bit_xor<>, void>;
+};
+
+template <class _Ty>
+struct _Map_vb_functor<equal_to<_Ty>> {
+    using _Type = conditional_t<_Is_vbool_functor_arg<_Ty>, _Bit_xnor, void>;
```
AlexGuteniev: Alternatively, we can map to `_Map_vb_functor` itself and have `operator()` right here to save one `struct`.

stl/inc/vector
```diff
@@ -4004,6 +4004,48 @@ _CONSTEXPR20 _OutIt _Copy_vbool(_VbIt _First, _VbIt _Last, _OutIt _Dest) {
     return _DestEnd;
 }
 
+template <class _VbIt, class _OutIt, class _Mapped_fn>
+_CONSTEXPR20 _OutIt _Transform_vbool_aligned(
```
AlexGuteniev, line 4008: I moved this out to `<vector>` from `<algorithms>` because other algorithms are moved out.
However, I'm not sure if it is useful. 

For accessing `vector<bool>` representation it is not strictly necessary. Most of things are template-dependent member functions and datas. The only exception is `_Vbase`, which can be still deduced from iterators.

For throughput it does not look useful either. `<vector>` is more frequent than `<algorithm>` so it appears more useful to off-load `<vector>` instead.

For reference, 0f24d45309f2282fe296a95ee843bb084b6130d6 is the commit where this movement was made.

benchmarks/src/vector_bool_transform.cpp
```diff
@@ -0,0 +1,52 @@
...
+        transform(source1.begin(), source1.end(), source2.begin(), dest.begin(), Pred{});
+        benchmark::DoNotOptimize(dest);
+    }
+}
+
+void common_args(auto bm) {
+    bm->RangeMultiplier(64)->Range(64, 64 << 10);
+}
+
+BENCHMARK(transform_two_inputs_aligned<logical_and<>>)->Apply(common_args);
```
StephanTLavavej, line 49: Need `<functional>`.

stl/inc/xutility
```diff
@@ -4889,6 +4889,14 @@ _OutCtgIt _Copy_memmove_n(_CtgIt _First, const size_t _Object_count, _OutCtgIt _
     return _STD _Copy_memmove_tail(_First_ch, _STD move(_Dest), _Byte_count, _Object_count);
 }
 
+template <class _Fn>
+struct _Map_vb_functor {
+    using _Type = void;
```
StephanTLavavej: One of the rare exceptions to using `_Ugly` names is that we use plain `type` and `value`. They definitely can't be macroized, and we got burned by `_Type`/`type` confusion in the past.

stl/inc/functional
```diff
@@ -258,6 +258,56 @@ struct bit_not<void> {
     using is_transparent = int;
 };
 
+struct _Bit_xnor {
+    template <class _Ty1, class _Ty2>
+    _NODISCARD constexpr auto operator()(_Ty1&& _Left, _Ty2&& _Right) const -> decltype(~(_Left ^ _Right)) {
+        return ~(_Left ^ _Right);
+    }
+};
```
StephanTLavavej, line 269: This is similar, but not identical, to a transparent functor. Although it's only used internally, I think we should exactly imitate the Standard types.

stl/inc/functional
```diff
@@ -258,6 +258,56 @@ struct bit_not<void> {
...
+
+template <class _Ty>
+struct _Map_vb_functor<equal_to<_Ty>> {
+    using _Type = conditional_t<_Is_vbool_functor_arg<_Ty>, _Bit_xnor, void>;
+};
+
+template <class _Ty>
+struct _Map_vb_functor<logical_not<_Ty>> {
+    using _Type = conditional_t<_Is_vbool_functor_arg<_Ty>, bit_not<>, void>;
+};
```
StephanTLavavej, line 312: We should comment why `bit_not` isn't mapped to itself.

stl/inc/algorithm
```diff
@@ -4063,13 +4063,30 @@ _CONSTEXPR20 void iter_swap(_FwdIt1 _Left, _FwdIt2 _Right) { // swap *_Left and
     swap(*_Left, *_Right); // intentional ADL
 }
 
+template <class _VbIt, class _OutIt, class _Mapped_fn>
+_CONSTEXPR20 _OutIt _Transform_vbool_aligned(
+    const _VbIt _First, const _VbIt _Last, _OutIt _Dest, const _Mapped_fn _Mapped_func);
+
+template <class _VbIt, class _OutIt, class _Mapped_fn>
+_CONSTEXPR20 _OutIt _Transform_vbool_aligned(
+    const _VbIt _First1, const _VbIt _Last1, const _VbIt _First2, _OutIt _Dest, const _Mapped_fn _Mapped_func);
```
StephanTLavavej: We conventionally omit `const` on value parameters in declarations.

StephanTLavavej: Bug: This signature doesn't handle varying `vector<bool>::iterator` versus `vector<bool>::const_iterator` on the inputs to binary `transform()`. Our test coverage (including libcxx) missed this compile-time regression.

stl/inc/vector
```diff
@@ -4004,6 +4004,48 @@ _CONSTEXPR20 _OutIt _Copy_vbool(_VbIt _First, _VbIt _Last, _OutIt _Dest) {
...
+        *_Dest_ptr = _Mapped_func(*_First_ptr);
+    }
+
+    if (_Last._Myoff != 0) {
+        const _Vbase _Mask = (_Vbase{1} << _Last._Myoff) - 1;
+        *_Dest_ptr         = (*_Dest_ptr & ~_Mask) | (_Mapped_func(*_First_ptr) & _Mask);
+        _Dest._Myoff       = _Last._Myoff;
+    }
+
+    return _Dest;
```
StephanTLavavej, line 4025: Bug: This doesn't update `_Dest._Myptr`, indicating a lack of test coverage.

stl/inc/functional
```diff
@@ -258,6 +258,56 @@ struct bit_not<void> {
...
+template <class _Ty>
+constexpr bool _Is_vbool_functor_arg = is_same_v<_Ty, void> || is_integral_v<_Ty>;
+
+template <class _Ty>
+struct _Map_vb_functor<logical_and<_Ty>> {
+    using _Type = conditional_t<_Is_vbool_functor_arg<_Ty>, bit_and<>, void>;
+};
+
+template <class _Ty>
+struct _Map_vb_functor<bit_and<_Ty>> {
```
StephanTLavavej, line 280: We should test `bit_meow`.

stl/inc/functional
StephanTLavavej: That would be weird, I'm happy with the current scheme.

StephanTLavavej:
Thanks! I pushed a conflict-free merge with `main` and fixes for two regressions, plus test coverage to catch them.

StephanTLavavej:

StephanTLavavej:
:t-rex: :sauropod: :chicken:
===

PR #5774, opened 2025-10-10
`<regex>`: Process non-greedy and longest-mode simple loops non-recursively

muellerj2:
Towards #997 and #1528. This PR implements non-recursive non-greedy matching for simple loops. Since non-greedy matching minimizes the amount of storage we will need in `_Frames` (at least for now) and leftmost-longest matching basically means exhaustively trying all possible trajectories, we also do non-greedy matching of simple loops in leftmost-longest mode. This also keeps memory usage before and after this PR roughly the same in leftmost-longest mode.

Non-greedy matching means that after each repetition at or beyond the minimum number of reps (or before the first repetition if the minimum number of reps is zero), we first try to match the remainder of the pattern and if it doesn't match (or matching is in leftmost-longest mode), we try another rep if permitted.

This means: After reaching the minimum number of repetitions, we first set up tail-matching (by assigning the `_Next` member of the `_N_end_rep` node to the `_Next` variable) after each repetition or when encountering the `_N_rep` node and push a frame to potentially match one more repetition if the maximum number of repetitions hasn't been reached yet.

Because of the properties of simple loops (see #5762), we don't have to worry about resetting `_Sav._Loop_index` during stack unwinding, but have to increment it. (We also avoid potential UB due to signed integer flow if `_REGEX_MAX_COMPLEXITY_COUNT` is set to `0`.)
During stack unwinding, we also have to reset the next position to match in the input and the set of valid capture groups (but not the limits of capture groups because the loop is not contained in any loops that are repeated more than once).

Again, I do not try to imitate `_Do_rep0()`'s increment of the stack usage count by one while evaluating the simple loop because it adds a lot of complexity for basically no gain.

I added a few more simple tests because there was basically no test coverage for non-greedy matching of simple loops that have an upper bound on the number of repetitions.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -1586,7 +1586,7 @@ public:
           _Loop_number(_Number) {}
 
     const int _Min;
-    const int _Max;
+    const int _Max; // non-negative if bounded, -1 if unbounded
     _Node_end_rep* _End_rep;
     unsigned int _Loop_number;
     int _Simple_loop = -1; // -1 undetermined, 0 contains if/do, 1 simple
@@ -4137,7 +4137,7 @@ bool _Matcher3<_BidIt, _Elem, _RxTraits, _It, _Alloc>::_Match_pat(_Node_base* _N
                             _Next          = _Node->_End_rep->_Next;
 
                             // set up stack unwinding for non-greedy matching if at least one rep is allowed
-                            if (_Node->_Max == -1 || 0 < _Node->_Max) {
+                            if (_Node->_Max != 0) {
                                 _Push_frame(_Rx_unwind_ops::_Loop_simple_nongreedy, _Node);
                             }
                         } else {
@@ -4179,7 +4179,7 @@ bool _Matcher3<_BidIt, _Elem, _RxTraits, _It, _Alloc>::_Match_pat(_Node_base* _N
                         } else if (_Longest || !_Greedy) {
                             _Increase_complexity_count();
                             // set up stack unwinding for non-greedy matching if one more rep is allowed
-                            if (_Nr->_Max == -1 || _Sav._Loop_idx < _Nr->_Max) {
+                            if (_Sav._Loop_idx != _Nr->_Max) {
                                 _Push_frame(_Rx_unwind_ops::_Loop_simple_nongreedy, _Nr);
                             }
                             // _Next is already assigned correctly for matching tail


stl/inc/regex
```diff
@@ -3413,22 +3414,18 @@ void _Matcher3<_BidIt, _Elem, _RxTraits, _It, _Alloc>::_Increase_complexity_coun
 }
 
 template <class _BidIt, class _Elem, class _RxTraits, class _It, class _Alloc>
-bool _Matcher3<_BidIt, _Elem, _RxTraits, _It, _Alloc>::_Do_rep0(_Node_rep* _Node, bool _Greedy) {
+bool _Matcher3<_BidIt, _Elem, _RxTraits, _It, _Alloc>::_Do_rep0(_Node_rep* _Node) {
     // apply repetition to loop with no nested if/do
     int _Ix                                   = _Node->_Min;
     const size_t _Frame_idx                   = _Loop_vals[_Node->_Loop_number]._Loop_frame_idx;
-    _Loop_vals[_Node->_Loop_number]._Loop_idx = _Ix + 1;
+    _Loop_vals[_Node->_Loop_number]._Loop_idx = _Ix + 2;
```
muellerj2, line 3421: The only point of this change is to make `_Sav._Loop_idx == 1` false when processing the `_N_end_rep` node, so that recursive `_Match_pat(_Node->_Next)` calls in `_Do_rep0()` are still handled correctly. (The next PR will remove `_Do_rep0()` including this line.)

StephanTLavavej: Thanks! BTW, I am trying to prioritize your PRs to get as many in as possible, but if the one-merge-batch-per-week that I usually end up doing is too slow for you, feel free to have multiple fine-grained commits in a PR; this adds no additional difficulty for my review process (I just load the PR up in VSCode and look at each commit in sequence). In any event I appreciate these step-by-step transformations instead of a monolithic mega-commit! :heart_eyes_cat:

muellerj2: To be honest, I haven't prepared any further code changes yet, so there was nothing more to include in this PR.

The next PR will be a bit different because it will actually not just refactor existing logic and split it into the two big switch statements. It will replace the currently implemented "greedy matching simulated by exhaustive non-greedy matching" by actual greedy matching when greedy matching is requested for a simple loop.

I might do the whole `_Do_rep()` replacement in one PR, as the new code inside the large switch statements should be quite similar to (a) the code for handling simple loops (except for some additional state restoration during unwinding and the handling of empty matches in later repetitions) and (b) the old code in `_Do_rep()`. But that will depend on how straightforward the changes are and how far I get in preparing them by the time the previous PR has been merged.

In any case, I expect that there will be at least two and at most four more PRs until #997 and #1528 are fixed. But this will be followed by a clean-up phase and I'm a bit doubtful that we will complete it before the MSVC Build Tools 15.41 branch off. But I hope that there will be enough time to remove the internal vectors from `_Rx_state_frame` to reduce the number of allocations for regexes with capture groups and to make `_Rx_state_frame` trivial iff the (unwrapped) iterator type is trivial.

stl/inc/regex
```diff
@@ -4155,29 +4156,41 @@ bool _Matcher3<_BidIt, _Elem, _RxTraits, _It, _Alloc>::_Match_pat(_Node_base* _N
...
+                            _Increase_complexity_count();
+
+                            _Next = _Nr->_Next;
+                            // GH-5365: We have to reset the capture groups from the second iteration on.
+                            _Tgt_state._Grp_valid = _Frames[_Sav._Loop_frame_idx]._Match_state._Grp_valid;
+                            ++_Sav._Loop_idx;
+                        } else if (_Longest || !_Greedy) {
+                            _Increase_complexity_count();
+                            // set up stack unwinding for non-greedy matching if one more rep is allowed
+                            if (_Nr->_Max == -1 || _Sav._Loop_idx < _Nr->_Max) {
```
muellerj2: I only realized this now (and the matcher implementation hasn't taken advantage of this observation anywhere else yet), but I think this condition can actually be simplified to `_Sav._Loop_idx !=_Nr->_Max`: 
* `_Sav._Loop_idx` is non-negative, so if  `_Nr->_Max` is `-1`, `_Sav._Loop_idx != _Nr->_Max` is true as well.
* If `_Nr->_Max` is non-negative, `_Sav._Loop_idx` can't exceed `_Nr->_Max`, else we would have stopped repeating the looped pattern, so `_Sav._Loop_idx < _Nr->_Max` compares the same as `_Sav._Loop_idx != _Nr->_Max`.

muellerj2: Change applied. I also added a comment that explicitly documents the value range of `_Max` because future versions of the matcher might interpret any negative value to mean "unbounded", not just `-1`. However, previous versions of the matcher will only work correctly for `-1` and not any other negative value.

stl/inc/regex
StephanTLavavej: Sounds good, thanks!

StephanTLavavej:

StephanTLavavej:
Thanks again for the ongoing overhaul and clear PR descriptions! :heart_eyes_cat: :green_heart: :100:

PR #5775, opened 2025-10-10
`<xstring>`: Add more casts to `basic_string` for ASan

StephanTLavavej:
Followup to #5562.

I forgot that ASan activates conditionally compiled code, which also needs casts to avoid compiler warnings.

To simplify things, I'm also consistently extracting `_New_size` so we don't need to repeat the cast expressions. I verified that this isn't introducing any shadowing.

Thanks to @davidmrdavid and @zacklj89 for bringing this to my attention.


StephanTLavavej:
ASan passed: https://dev.azure.com/vclibs/STL/_build/results?buildId=19418&view=results

PR #5781, opened 2025-10-14
`_Locinfo`: Use `_wsetlocale` to query and restore locales

### LINKED ISSUE #5780: `<iostream>`: Printing with std::cout changes the global locale permanently
# Describe the bug

Using the streams provided by `<iostream>` can change the locale permanently

# Command-line test case
```
D:\>type test-locale-stl.cpp
#include <iostream>
#include <clocale>

char *
query_locale_name()
{
  return std::setlocale(LC_ALL, nullptr);
}

int main()
{
  std::setlocale(LC_ALL, "norwegian-bokmal.UTF-8");

  std::cout << "locale before: " << query_locale_name() << std::endl;
  std::cout << 0.1 << std::endl;
  std::cout << "locale after: " << query_locale_name() << std::endl;
}

D:\>cl /EHsc /W4 /WX /std:c++latest .\test-locale-stl.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.44.35217 for x64

test-locale-stl.cpp
Microsoft (R) Incremental Linker Version 14.44.35217.0

/out:test-locale-stl.exe
test-locale-stl.obj

D:\>test-locale-stl
locale before: Norwegian Bokmål_Norway.utf8
0.1
locale after: C
```

# Expected behavior

Locale after should be `Norwegian Bokmål_Norway.utf8`

# STL version

`Microsoft (R) C/C++ Optimizing Compiler Version 19.44.35217 for x64`

# Additional context

lb90:
`_Locinfo` changes the locale temporarily and then reverts to the previous locale on destruction. The sequence of `setlocale` calls are as follows:

1. `oldlocname = setlocale(LC_ALL, nullptr)` to query the locale string
2. `setlocale(LC_ALL, newlocname)` to set the temporary locale
3. `setlocale(LC_ALL, oldlocname)` to restore the previous locale

However there's a catch here: the fully-qualified locale names returned by setlocale are not always ASCII strings (more on that below). This creates challenges because the oldlocname is encoded depending on the "outer" locale, while the `setlocale` call at point 3) expects an encoding which follows the "inner" locale, and the two may not match.

To solve this issue, use the wide variant of setlocale: _wsetlocale. This way all strings are UTF-16 and there's no issue with varying narrow string encodings.

Addendum:

Actually, the C RunTime library does its best to use ASCII strings! It queries the english name of the locale using `GetLocaleInfoEx`. MSDN says that the returned string is always ASCII [1], but that's not always the case [2].

Fixes #5780

References:
 1. https://learn.microsoft.com/en-us/windows/win32/intl/locale-senglish-constants
 2. https://developercommunity.visualstudio.com/t/GetLocaleInfoEx-w-LOCALE_SENGLISHLANGUA/10981789

I made a similar change in `libc++`: https://github.com/llvm/llvm-project/pull/160479

### DIFF
--- a/tests/std/tests/GH_005780_non_ascii_locales/test.cpp
+++ b/tests/std/tests/GH_005780_non_ascii_locales/test.cpp
@@ -11,13 +11,13 @@
 using namespace std;
 
 string set_locale(const string& locale_name) {
-    const char* ret = setlocale(LC_ALL, locale_name.c_str());
+    const char* const ret = setlocale(LC_ALL, locale_name.c_str());
     assert(ret != nullptr);
     return ret;
 }
 
 string query_locale() {
-    const char* ret = setlocale(LC_ALL, nullptr);
+    const char* const ret = setlocale(LC_ALL, nullptr);
     assert(ret != nullptr);
     return ret;
 }
@@ -30,7 +30,7 @@ void assert_string_non_ascii(const string& str) {
 // GH-5780 <iostream>: Printing with std::cout changes the global locale permanently
 void test_gh_5780() {
     // https://learn.microsoft.com/en-us/cpp/c-runtime-library/language-strings#supported-language-strings
-    string locale_name = set_locale("norwegian-bokmal.437");
+    const string locale_name = set_locale("norwegian-bokmal.437");
     assert_string_non_ascii(locale_name);
 
     cerr.imbue(locale::classic());


stl/inc/xlocinfo
```diff
@@ -382,7 +382,7 @@ private:
     _Yarn<char> _Months; // month names
     _Yarn<wchar_t> _W_Days; // wide weekday names
     _Yarn<wchar_t> _W_Months; // wide month names
-    _Yarn<char> _Oldlocname; // old locale name to revert to on destruction
+    _Yarn<wchar_t> _Oldlocname; // old locale name to revert to on destruction
```
muellerj2: I think this breaks ABI: If the linker ends up choosing old `_Locinfo::_Locinfo_ctor` and new `_Locinfo::_Locinfo_dtor` or vice versa, one piece of code assumes that `_Oldlocname` holds a `char*` and the other assumes it holds a `wchar_t*`.

The changed class must receive a new name.

But we can't just outright remove the old class and change function signatures to use the new name, because it seems the old name appears in some exported symbols.

StephanTLavavej: Agreed, this is an ABI break. The `_CRTIMP2_PURE_IMPORT` in `extern "C++" class _CRTIMP2_PURE_IMPORT _Locinfo` indicates that the entire class is separately compiled, and its member functions live in `locale.cpp` and `locale0.cpp` which are being modified here. (This is an old, bad practice that we've strongly moved away from. When separately compiled code is necessary, we prefer to use "flat C exports" that accept plain C arguments. This gives us the most flexibility in the headers.)

This case is very interesting because aside from the definition in `<xlocinfo>`, `_Oldlocname` is never mentioned by the headers. It's only manipulated by `locale.cpp` and `locale0.cpp`. There's an important fact: There Is Only One STL (dll or static lib), and we never have to worry about mismatch between the STL's separately compiled components, only between outdated user code and newer STL dll/libs.

So while we can't mess with `_Locinfo`'s name (at most, we could add a new name, but then we'd need to maintain a separate implementation, and I'm not sure if it could be header-only, or if it could be injected into the import lib without causing headaches especially for `/clr` and `/clr:pure`), and we can't change `_Oldlocname`'s type without also changing `_Locinfo`'s name, I think we *could* change `_Oldlocname`'s content. Specifically, if `_Locinfo::_Locinfo_ctor()` stored UTF-8, then `_Locinfo::_Locinfo_dtor()` could decode and use it. Old user binaries running against the new STL DLL (or old user objects/libs linked against the new STL static LIB) would get matching new `_Locinfo::_Locinfo_ctor()` / `_Locinfo::_Locinfo_dtor()` talking UTF-8 to each other, but at no point would the user code actually care about the content of `_Oldlocname`, only its type of `_Yarn<char>`.

StephanTLavavej:
Also, what about `_Yarn<char> _Newlocname`? It appears to get its content from `setlocale`. That one is trickier because it is exposed to the headers via `_Getname()`.

muellerj2:
> Also, what about _Yarn<char> _Newlocname? It appears to get its content from setlocale. That one is trickier because it is exposed to the headers via _Getname().

Its value is effectively returned by `locale:name()`, so there might be some users out there relying on its encoding.

stl/inc/xlocinfo
frederick-vs-ja: @lb90 I opened lb90#1 for this.

frederick-vs-ja:
Is it possible to add test cases?

stl/inc/xlocinfo
lb90: Fantastic, thanks a lot!

lb90:
> Is it possible to add test cases?

Sure :)

P.S: will get back to this within 1 or 2 days...

tests/std/tests/GH_005780_non_ascii_locales/test.cpp
```diff
@@ -0,0 +1,39 @@
...
+    const char *ret = std::setlocale(LC_ALL, nullptr);
+    assert(ret != nullptr);
+    return ret;
+}
+
+void test_gh_5780() {
+    // https://learn.microsoft.com/en-us/cpp/c-runtime-library/language-strings#supported-language-strings
+    std::string locale_name = set_locale("norwegian-bokmal.437");
+
+    const auto& char_not_ascii = [](char c){ return (c & 0x80) != 0; };
```
AlexGuteniev: There's no point to use lifetime-prolonging reference-binding-to-temporary here.

There's almost never need to use this obscure feature.

Just `const auto char_not_ascii` would be better.

AlexGuteniev: `const char c`

tests/std/tests/GH_005780_non_ascii_locales/test.cpp
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <algorithm>
+#include <string>
+#include <clocale>
+#include <cassert>
+
+#include <xlocinfo>
```
AlexGuteniev: We should avoid internal headers, and include corresponding standard headers.

tests/std/tests/GH_005780_non_ascii_locales/env.lst
AlexGuteniev, line 1: Need to update `/tests/std/test.lst` to list the new test (in sorted order) to make sure this test is picked by internal (non-Github) harness.

lb90: Ah, indeed! Fixed

StephanTLavavej:
By the way, after code review has started, we recommend avoiding force-pushes, as that makes it harder to see what's changed. Not a big deal, just wanted to mention it for the future.

lb90:
Hi! Should I rebase to solve the conflict?

AlexGuteniev:
> Hi! Should I rebase to solve the conflict?

You should merge instead of rebasing.
This is not mandatory; maintainers would be able to do this for you too (be sure to allow maintainers making changes).

StephanTLavavej:
If it's just a `test.lst` conflict I agree that there's no need to push a merge and rerun the checks, I can do that when I review the PR again. (I will try to get to this soon since I've been working through the PR backlog and this is now the second-oldest.) When there are more significant conflicts, I do appreciate when contributors resolve the conflicts, locally rerun affected tests, and push normally.

StephanTLavavej:
My bosses want me to focus on `<flat_map>` and `<flat_set>` so my review may be delayed.

lb90:
No problem at all! 😊

mikekaganski:
Does this fix https://developercommunity.visualstudio.com/t/setlocale-may-crash-CRT-in-some-cases/10603395 and https://github.com/Exiv2/exiv2/issues/3225 ?

lb90:
Hello @mikekaganski! https://github.com/Exiv2/exiv2/issues/3225 is fixed by the LLVM/LibC++ PR linked above. On the other hand, issue https://developercommunity.visualstudio.com/t/setlocale-may-crash-CRT-in-some-cases/10603395 must be fixed in the UCRT. The UCRT is source-available but not open source, so can only be fixed by MS developers :)

stl/inc/xlocinfo
```diff
@@ -134,6 +134,28 @@ private:
...
 
+template <class _Elem>
+_Elem* _Ntcts_dup_dbg(const _Elem* const _Ptr) noexcept {
+    _STL_INTERNAL_STATIC_ASSERT(_Is_any_of_v<_Elem, char, wchar_t, unsigned short>);
+
+    auto _Iter = _Ptr;
+    while (*_Iter != _Elem{}) {
+        ++_Iter;
+    }
+    const size_t _Count = static_cast<size_t>(_Iter - _Ptr + 1) * sizeof(_Elem);
```
cpplearner: Can we use `char_traits::length` here?

```suggestion
    const size_t _Count = (char_traits<_Elem>::length(_Ptr) + 1) * sizeof(_Elem);
```

frederick-vs-ja: I guess the original author wanted to avoid using non-Standard `char_traits<unsigned short>` even if it could just work.

stl/inc/xlocinfo
```diff
@@ -382,6 +400,7 @@ private:
     _Yarn<char> _Months; // month names
     _Yarn<wchar_t> _W_Days; // wide weekday names
     _Yarn<wchar_t> _W_Months; // wide month names
+    // TRANSITION, ABI, `_Oldlocname._Myptr` is reinterpreted as `wchar_t*`. `wchar` should be wrapped instead.
```
StephanTLavavej: Nitpick: `wchar` is not a thing. Let's explicitly say `_Yarn<wchar_t>`.

stl/inc/xlocinfo
StephanTLavavej: This should be fine, I can make that change.

StephanTLavavej: This is pre-existing but `_Count` is a bad name for something that's a number of bytes instead of a number of elements.

stl/inc/xlocinfo
```diff
@@ -134,6 +134,28 @@ private:
     void* _Timeptr; // pointer to time information
 };
 
+template <class _Elem>
+_Elem* _Ntcts_dup_dbg(const _Elem* const _Ptr) noexcept {
```
StephanTLavavej: Let's add `_NODISCARD` to be good kitties.

tests/std/tests/GH_005780_non_ascii_locales/test.cpp
```diff
@@ -0,0 +1,41 @@
...
+    return ret;
+}
+
+std::string query_locale() {
+    const char* ret = std::setlocale(LC_ALL, nullptr);
+    assert(ret != nullptr);
+    return ret;
+}
+
+void assert_string_non_ascii(const std::string& string) {
```
StephanTLavavej: It's technically not shadowing due to the lack of a `using`-directive, but naming the variable `string` is visually confusing. It would be better to `using namespace std;` and rename this to `str`.

tests/std/tests/GH_005780_non_ascii_locales/test.cpp
```diff
@@ -0,0 +1,41 @@
...
+    const auto char_not_ascii = [](const char c) { return (c & 0x80) != 0; };
+    assert(std::any_of(string.begin(), string.end(), char_not_ascii));
+}
+
+void test_gh_5780() {
+    // https://learn.microsoft.com/en-us/cpp/c-runtime-library/language-strings#supported-language-strings
+    std::string locale_name = set_locale("norwegian-bokmal.437");
+    assert_string_non_ascii(locale_name);
+
+    std::cerr.imbue(std::locale::classic());
```
StephanTLavavej: Should include `<locale>`.

tests/std/tests/GH_005780_non_ascii_locales/test.cpp
```diff
@@ -0,0 +1,41 @@
...
+    assert(ret != nullptr);
+    return ret;
+}
+
+void assert_string_non_ascii(const std::string& string) {
+    const auto char_not_ascii = [](const char c) { return (c & 0x80) != 0; };
+    assert(std::any_of(string.begin(), string.end(), char_not_ascii));
+}
+
+void test_gh_5780() {
```
StephanTLavavej, line 31: Let's record the bug title here.

tests/std/tests/GH_005780_non_ascii_locales/test.cpp
```diff
@@ -0,0 +1,41 @@
...
+    // https://learn.microsoft.com/en-us/cpp/c-runtime-library/language-strings#supported-language-strings
+    std::string locale_name = set_locale("norwegian-bokmal.437");
```
StephanTLavavej: Could add more const.

StephanTLavavej:
Thank you! :heart_eyes_cat: (And apologies for taking so long to review this, `<flat_meow>` took a while.) I am cautious because locales/iostreams burned me repeatedly at the beginning of my career, but this looks like a solid fix and if it goes wrong it'll be in a surprising way which is all I can ask for. :joy_cat:

StephanTLavavej:

StephanTLavavej:
:earth_africa: :tada: :heart_eyes_cat:
===

PR #5783, opened 2025-10-15
Toolset update: MSVC Compiler 19.50.35702, test `<filesystem>` long path support

### LINKED ISSUE #1923: `<filesystem>`: Consider setting up the 1ES Hosted Pool with long path support for test coverage
We don't have automated test coverage of `<filesystem>`'s long path support - if I recall correctly, this is because our internal test machines aren't under the STL's control, so we can't assume anything about their OS support. Now that we're on GitHub, I believe we could add long path support to our 1ES Hosted Pool scripts, and then conditionally add test coverage.

Prompted by #1921.

### LINKED ISSUE #5733: libcxx: Flaky timing assumption in `std/thread/thread.semaphore/timed.pass.cpp`
The [CI run](https://dev.azure.com/vclibs/STL/_build/results?buildId=19313&view=results) for #5714 encountered a spurious failure in libcxx's [`std/thread/thread.semaphore/timed.pass.cpp`](https://github.com/llvm/llvm-project/blob/c075feef3715f26faee010c65110a0acd45c0e0a/libcxx/test/std/thread/thread.semaphore/timed.pass.cpp#L40):

```
Assertion failed: s.try_acquire_until(start + std::chrono::seconds(2)), file C:\a\_work\1\s\llvm-project\libcxx\test\std\thread\thread.semaphore\timed.pass.cpp, line 40
```

This looks like a flaky timing assumption.

StephanTLavavej:
# Changelog
- Improved test coverage:
  * Added test coverage for `<filesystem>`'s long path support.
- Code cleanups:
  * Updated dependencies.
    + Updated MSVC Compiler to 19.50.35702 (now required).
    + Updated CMake to 4.1.1 (now required).
    + Updated Python to 3.14.0 (now required).
    + Updated Boost.Math to 1.89.0.

# Overview
In this toolset update, I attempted to add an ARM64 pool. While I encountered roadblocks and abandoned the attempt for now,
I was able to extend provision-image.ps1 to automatically detect when it's being run in an ARM64 VM, and switch to ARM64 installers. I also have small restructurings to create-1es-hosted-pool.ps1 that will make it easier to add ARM64 as an option in the future.

I was also able to enable long paths in the VMs, and write a test to exercise one of `<filesystem>`'s functions. It's written
to detect the absence of long path support and do nothing, so it won't break local test runs.

# Commits
* Azure extended the deadline for default outbound access.
* Extract `$PoolSize`.
  + Because ARM64 will use a smaller pool.
* Centralize StlBuild, use double quotes instead of plus concat.
  + Making it easier for everything to mention the arch.
* Improve log name, put arch in resource group name.
  + This is a significant improvement for the log files (previously, we'd concat to an existing log file). I'm shrinking "StlBuild" to just "Stl" since "Build" adds nothing, and I'm adding the arch which makes things longer.
* PowerShell 7.5.3.
* Python 3.14.0.
  + I now require the specific patch version we've upgraded to (or later). No reason to let people get away with outdated dependencies, even slightly.
* Extend provision-image.ps1 to detect ARM64.
  + Manually verified. This was actually quite easy; we don't need to pass parameters, since the script can just look at the machine architecture. ARM64 builds are available for all of our dependencies except CUDA, which we simply skip.
  + The one headache was that I encountered a failure to immediately delete the installer, presumably because the ARM64 VM was somewhat slower (I only had access to v5 for manual testing, not v6). This is potentially an issue for x64 too, we just haven't encountered yet. I felt that a 5-second sleep was the simplest way to avoid this.
* MSVC Compiler 19.50.35702.
  + I am no longer attempting to name Insiders releases of the compiler in the error message.
* CMake 4.1.1.
  + Again, requiring the specific patch version (or newer).
* Set `CMAKE_MSVC_RUNTIME_CHECKS` to the empty string.
  + This was added in CMake 4.0: https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_CHECKS.html
  + We clear out all of CMake's default options before setting our own.
  + Fixes: cl : Command line error D8016 : '/RTC1' and '/O2' command-line options are incompatible
  + This matches internal src/vctools/crt/crt_build.settings.targets, which sets `<NoRtc>true</NoRtc>`.
* CMakeLists.txt: Update comment to no longer mention "MSVC version 19.32.31328".
* Boost.Math 1.89.0.
  + I will replace this with an MSVC-internal commit, already tested.
* special_math.cpp: Silence warning C4714: function marked as `__forceinline` not inlined.
* VSO_0744055_atomic_load_8_bytes_readonly: Update comment to no longer mention "MSVC 19.24".
* Overhaul VSO_0744055_atomic_load_8_bytes_readonly.
  + Don't `printf()`, instead `assert()` to check correctness.
  + Cite documentation explaining why the bytes are 0.
  + Use `std::uint64_t` instead of `unsigned long long` for clarity.
  + There's no need to `static_cast<unsigned char*>` the result of `MapViewOfFileEx`.
  + Use `std::align()` as I don't see any documented alignment guarantees.
* Skip flaky libcxx test to resolve #5733.
* Remove workarounds for VSO-1008447, VSO-2066340.
* Enable long paths to resolve #1923.
* Extract `test_temp_directory` with minor changes:
  + Make `directoryPath` a `const` data member. (It probably should be private with an accessor, but that required too many changes for my level of laziness.)
  + Add qualification.
  + Use `std::endl` as I am slightly paranoid about flushing output for logs.
  + If we can't create the temp directory, `assert` instead of simply warning. This will block further testing.
* Add GH_001923_filesystem_long_path_support.
  + We need `custom_format.py` to link against `Advapi32.lib`, but especially to pass `/MANIFESTINPUT:{manifestFile}`. The Python script can get the path to `long_path_aware.manifest` which lives next to `test.cpp`. Attempting to do this through `PM_LINK` in `env.lst` won't work, because the compile-and-link step is run in a subdirectory (and I felt that hardcoding `..` was gross).
* Extract `$Provisioning_x64`, for clarity.
* Use `Get-AzContext` to avoid annoying warnings.
  + `Set-AzContext` was emitting an annoying warning about an apparently inaccessible unrelated "tenant". After logging in, the subscription appears to stay fixed to `CPP_STL_GitHub`, so we just need a check that we're using the right subscription. I now also print a clear message if the check fails. (`Connect-AzAccount` pops up a UI dialog, so I don't want to automatically run it.)
* New pool.

STL-ASan-CI [passed](https://dev.azure.com/vclibs/STL/_build/results?buildId=19428&view=results).

### DIFF
--- a/tests/std/tests/GH_001923_filesystem_long_path_support/custombuild.pl
+++ b/tests/std/tests/GH_001923_filesystem_long_path_support/custombuild.pl
@@ -1,0 +1,15 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+use strict;
+use warnings;
+
+use Run;
+
+sub CustomBuildHook()
+{
+    my $cwd = Run::GetCWDName();
+
+    Run::ExecuteCL(join(" ", "test.cpp", "/Fe$cwd.exe", "/link", "/MANIFESTINPUT:long_path_aware.manifest"));
+}
+1


tests/std/tests/GH_001923_filesystem_long_path_support/custom_format.py
```diff
@@ -0,0 +1,23 @@
...
+
+from stl.test.format import STLTestFormat, TestStep
+from stl.test.tests import TestType
+
+class CustomTestFormat(STLTestFormat):
+    def getBuildSteps(self, test, litConfig, shared):
+        exeSource = test.getSourcePath()
+        manifestFile = os.path.join(os.path.dirname(exeSource), 'long_path_aware.manifest')
+
+        _, outputBase = test.getTempPaths()
```
davidmrdavid, line 14: ultra nit / not worth resetting, but it seems this tis only used in the `elif` path

StephanTLavavej: Yeah, I copy-pasted this structure from another test with `custom_format.py`, hence the unused `outputBase` on one codepath. Good catch!

StephanTLavavej:

StephanTLavavej:
I pushed two additional commits because my technique to skip for MSVC-internal wasn't working. Now we `#pragma comment` the lib, which works on both sides, and I added a Perl script to embed the manifest. Some MSVC-internal test machines have the regkey, so this allows the test to work (or skip) as needed.

Issue #1902, opened 2021-05-04
`<chrono>`: Handle overflow cases in `[time.format]`


mnatsuhara:
Observed by @statementreply.

There are two specific cases of overflow that we are not currently handling correctly: 
```
> format("{:%Q%q}", nanoseconds::min())
< --9223372036854775808ns
> format("{:%T}", duration<float, ratio<1, 10'000'000>>{179999981568.0f})
< 05:00:00.-016384
```

We should also audit for other cases of overflow.  

Discovered while implementing #12.

PR #5786, opened 2025-10-16
Add evil tests for negative duration formatting

achabense:
This pr adds some evil tests for negative duration formatting. The background is https://github.com/microsoft/STL/issues/1902. The first case is not very hard to fix in itself, but I noticed something very tricky about it. Initially I thought it can be fixed by removing both signedness checks...

https://github.com/microsoft/STL/blob/85e4b574ca1f127993538673d3560dd4fbb4e27d/stl/inc/chrono#L5469-L5470
https://github.com/microsoft/STL/blob/85e4b574ca1f127993538673d3560dd4fbb4e27d/stl/inc/chrono#L5612

But but after looking into the standard, I found myself wrong. The standard says (https://eel.is/c++draft/time.format#4):
```
The result of formatting a std​::​chrono​::​duration instance holding a negative value, or an hh_mm_ss object h
for which h.is_negative() is true, is equivalent to the output of the corresponding positive value, with a
STATICALLY-WIDEN<charT>("-") character sequence placed before the replacement of the initial conversion specifier.
```

Which means the '-' is added unconditionally, regardless of conversion specifiers. Thus this pr. I call these tests "evil" as as a user I won't expect the formatting to work this way (for example, when it comes to `"%%%Q"`, I would expect `"%-1"` instead of `-%1`.)

Neither gcc nor clang can pass this test. https://godbolt.org/z/aqYYWhv75

---

(Additionally, the second case in the issue is not directly due to formatting impl, but the converted `hh_mm_ss` having negative `subseconds()`.)
```cpp
#include <chrono>
#include <cassert>

int main() {
    const auto dur = std::chrono::duration<float, std::ratio<1, 10'000'000>>{ 179999981568.0f };
    // format("{:%T}", dur); // This calls `_Write_seconds(... , dur)`, which converts dur to hh_mm_ss...
    const auto dur_hh_mm_ss = std::chrono::hh_mm_ss{ dur };
    assert(dur_hh_mm_ss.subseconds().count() < 0);
}
```

### DIFF
--- a/tests/std/tests/P0355R7_calendars_and_time_zones_formatting/test.cpp
+++ b/tests/std/tests/P0355R7_calendars_and_time_zones_formatting/test.cpp
@@ -287,6 +287,20 @@ void test_duration_formatter() {
     // GH-4247: <chrono>: format() should accept %X and %EX for duration and hh_mm_ss
     assert(format(STR("{:%X}"), 9h + 7min + 5s) == STR("09:07:05"));
     assert(format(STR("{:%EX}"), 9h + 7min + 5s) == STR("09:07:05"));
+
+    // N4950 [time.format]/4
+    // "The result of formatting a std::chrono::duration instance holding a negative value, or an hh_mm_ss object h for
+    // which h.is_negative() is true, is equivalent to the output of the corresponding positive value, with a
+    // STATICALLY-WIDEN <charT>("-") character sequence placed before the replacement of the initial conversion
+    // specifier."
+    const auto minus_1s = -1s;
+    assert(format(STR("{:%Q}"), minus_1s) == STR("-1"));
+    assert(format(STR("{:%q}"), minus_1s) == STR("-s"));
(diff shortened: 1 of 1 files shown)


cpplearner:
This is LWG-3844.

achabense:
Oops... then are these tests meaningful?

StephanTLavavej:
Thanks @cpplearner!

@achabense I think your proposed tests are correctly verifying what the current Standard mandates, but as you noted, the Standardese is wacky. However LWG-3844 ends up being resolved, it should fix this specification.

Since the examples are pathological, I think we shouldn't add test coverage here. I'm going to close this PR without merging, but thanks for looking into this!

As an aside, I am somewhat concerned that this LWG issue hasn't made apparent progress since Feb 2023 despite having what appears to be a reasonable proposed resolution. What needs to happen here for the Standard to actually get fixed soon?

PR #5790, opened 2025-10-18
`<regex>`: Process greedy simple loops non-recursively

muellerj2:
Towards #997 and #1528.

This implements processing of greedy simple loops in a non-recursive way. After this PR, we actually do greedy matching when greedy quantifiers are used. But in exchange, we have to store more frames in the `_Frames` vector. But I think we should now take the chance to actually process the regex as the drafter of the regex requested. (Before this PR, "more memory" would have meant "more stack", so the correct processing order would have resulted in many more stack overflows.)

The greedy processing is implemented as follows:
* In `_N_rep`, we first try to match the repeated pattern, if the maximum number of reps is greater than zero. We also push a frame with opcode `_Loop_simple_greedy` in this case if the minimum number of reps is zero. If the maximum number of reps is zero, we instead try to match the remainder of the regex.
* In `_N_end_rep`, we also first try to match the repeated pattern, if the maximum number of reps hasn't been reached yet, and push a frame with opcode `_Loop_simple_greedy` if the minimum number of reps has been reached. If the maximum number of reps has been reached, we instead try to match the remainder of the regex. We also avoid UB now by checking for potential overflow of `_Loop_idx`.
* During backtracking, we try to match the tail when processing `_Loop_simple_greedy` if matching has failed, resetting the position in the input and the state of capturing groups appropriately. (We do not have to handle `_Longest` here because we always perform non-greedy matching when `_Longest` is true.) As already argued in #5762, we don't have to restore loop state here because simple loops are branchless and non-reentrant.

The increase in size of the `_Frames` vector is no longer accurately reflected by the preexisting stack usage counter. But this is deliberate choice to preserve backwards compatibility: If the stack usage count were increased as well, we might throw a `regex_error(error_stack)` on inputs that were previously accepted.

It is possible to avoid this increase in size of the `_Frames` vector. But this would majorly complicate this PR and a few more changes I consider more relevant, so I would like to defer this to a later PR.

I felt again that the existing test coverage was a bit insufficient, so I added a few more tests. This includes tests that greedy quantifiers and non-greedy quantifiers (and leftmost-longest mode) can yield matches of different length when searching.

### DIFF
--- a/tests/std/tests/VSO_0000000_regex_use/test.cpp
+++ b/tests/std/tests/VSO_0000000_regex_use/test.cpp
@@ -2133,6 +2133,7 @@ void test_gh_5774() {
     // GH-5774: Process non-greedy and longest-mode simple loops non-recursively.
     // This extends our test coverage on non-greedy simple loops with bounded number of repetitions.
     g_regexTester.should_not_match("", "a+?");
+    g_regexTester.should_match("b", "a{0}?b");
     g_regexTester.should_not_match("ab", "a{0}?b");
     g_regexTester.should_match("ab", "a{0,1}?b");
     g_regexTester.should_not_match("aab", "a{0,1}?b");
@@ -2147,6 +2148,7 @@ void test_gh_5790() {
     // GH-5790: Process greedy simple loops non-recursively.
     // This extends our test coverage on (mainly greedy) simple loops.
     g_regexTester.should_not_match("", "a+");
+    g_regexTester.should_match("b", "a{0}b");
     g_regexTester.should_not_match("ab", "a{0}b");
     g_regexTester.should_match("ab", "a{0,1}b");
     g_regexTester.should_not_match("aab", "a{0,1}b");
@@ -2208,6 +2210,11 @@ void test_gh_5790() {
     {
         test_regex lower_bounded_nongreedy_a_rep(&g_regexTester, "a{4,1000}?");
         lower_bounded_nongreedy_a_rep.should_search_match("aaaaaaaaaa", "aaaa");
+    }
+
+    {
+        test_regex lower_and_upper_bounded_nongreedy_a_rep(&g_regexTester, "a{2,5}?");
+        lower_and_upper_bounded_nongreedy_a_rep.should_search_match("aaaaaaaaaa", "aa");
     }
 
     {


stl/inc/regex
```diff
@@ -4163,31 +4096,36 @@ bool _Matcher3<_BidIt, _Elem, _RxTraits, _It, _Alloc>::_Match_pat(_Node_base* _N
...
-                        } else if (_Longest || !_Greedy) {
-                            _Increase_complexity_count();
+                        } else if (_Greedy && !_Longest && _Sav._Loop_idx != _Nr->_Max) { // one more rep to try next
+                            // set up stack unwinding for greedy matching
+                            _Push_frame(_Rx_unwind_ops::_Loop_simple_greedy, _Nr);
+
+                            _Next = _Nr->_Next;
+                            // GH-5365: We have to reset the capture groups from the second iteration on.
+                            _Tgt_state._Grp_valid = _Frames[_Sav._Loop_frame_idx]._Match_state._Grp_valid;
+                            if (_Sav._Loop_idx < INT_MAX) { // avoid overfloading _Loop_idx
```
StephanTLavavej: Typo.

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -2143,6 +2143,84 @@ void test_gh_5774() {
...
+    g_regexTester.should_match("ab", "a{0,1}b");
+    g_regexTester.should_not_match("aab", "a{0,1}b");
+    g_regexTester.should_match("aab", "a{0,2}b");
+    g_regexTester.should_match("aab", "a{1,2}b");
+    g_regexTester.should_not_match("aab", "a{1}b");
+    g_regexTester.should_not_match("aaab", "a{1,2}b");
+    g_regexTester.should_match("aaab", "a{1,3}b");
+
+    // Check that greedy and non-greedy search find the appropriate match.
+    // for the following regexes, greedy and leftmost-longest search yield the same matches.
```
StephanTLavavej: Nitpick: This is a full sentence so it should be capitalized.

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -2143,6 +2143,84 @@ void test_gh_5774() {
     g_regexTester.should_match("aaab", "a{1,3}?b");
 }
 
+void test_gh_5790() {
+    // GH-5790: Process greedy simple loops non-recursively.
+    // This extends our test coverage on (mainly greedy) simple loops.
+    g_regexTester.should_not_match("", "a+");
+    g_regexTester.should_not_match("ab", "a{0}b");
```
StephanTLavavej, line 2152: We can also positively test that `"b"` should match `"a{0}b"`.

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -2143,6 +2143,84 @@ void test_gh_5774() {
...
+            lower_and_upper_bounded_greedy_a_rep.should_search_match("aaaaaaaaaa", "aaaaa");
+        }
+
+        {
+            test_regex lower_and_upper_bounded_greedy_a_rep(&g_regexTester, "a{2,5}", options);
+            lower_and_upper_bounded_greedy_a_rep.should_search_match("aaaaaaaaaa", "aaaaa");
+        }
+
+        {
+            test_regex too_large_min_greedy_a_rep(&g_regexTester, "a{11,1000}");
```
StephanTLavavej: This one is missing `options`.

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -2143,6 +2143,84 @@ void test_gh_5774() {
...
+
+        {
+            test_regex lower_and_upper_bounded_greedy_a_rep(&g_regexTester, "a{2,5}", options);
+            lower_and_upper_bounded_greedy_a_rep.should_search_match("aaaaaaaaaa", "aaaaa");
+        }
+
+        {
+            test_regex lower_and_upper_bounded_greedy_a_rep(&g_regexTester, "a{2,5}", options);
+            lower_and_upper_bounded_greedy_a_rep.should_search_match("aaaaaaaaaa", "aaaaa");
+        }
```
StephanTLavavej: This is exactly repeated.

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -2143,6 +2143,84 @@ void test_gh_5774() {
...
+
+    {
+        test_regex lower_bounded_nongreedy_a_rep(&g_regexTester, "a{4,1000}?");
+        lower_bounded_nongreedy_a_rep.should_search_match("aaaaaaaaaa", "aaaa");
+    }
+
+    {
+        test_regex too_large_min_nongreedy_a_rep(&g_regexTester, "a{11,1000}?");
+        too_large_min_nongreedy_a_rep.should_search_fail("aaaaaaaaaa");
+    }
```
StephanTLavavej, line 2223: It's not super important, but it would be reasonable to test non-greedy `"a{2,5}?"` here, for symmetry with the greedy cases above.

StephanTLavavej:
Thanks as always for the exceptionally clear writeup, careful changes, and added test coverage! :heart_eyes_cat:

I really appreciate the attention to avoiding disruption here:

> The increase in size of the `_Frames` vector is no longer accurately reflected by the preexisting stack usage counter. But this is deliberate choice to preserve backwards compatibility: If the stack usage count were increased as well, we might throw a `regex_error(error_stack)` on inputs that were previously accepted.

I pushed a comment typo fix and minor test fixes. Also I noticed how you predict what the PR number is going to be, so your test mentions it correctly on the first try :joy_cat:

StephanTLavavej:

StephanTLavavej:
Thanks - I'm greedy for your amazing regex PRs! :money_mouth_face: :moneybag: :money_with_wings:

PR #5793, opened 2025-10-20
`<regex>`: Fix `regex_error` when matching a repeated pattern containing a lookahead assertion

### LINKED ISSUE #5792: `<regex>`: `regex_match()` throws `regex_error(error_stack)` when a repeated pattern contains a lookahead assertion
# Describe the bug
Since MSVC Build Tools 14.50, a lookahead assertion inside a repeated pattern results in a `regex_error(error_stack)`.

# Test case
```
#include <iostream>
#include <regex>
#include <string>

using namespace std;

int main() {
    try {
        regex re("(?:(?=ab))*");
        (void) regex_match("xyz", re);
        cout << "no exception\n";
    } catch (const regex_error& ex) {
        cout << "regex_error with code " << ex.code() << "\n";
    }

    return 0;
}
```
This prints "regex_error with code 12".

# Expected behavior
Should print "no exception".

# STL version
MSVC Build Tools 18.50 up to current head.

# Additional context
I introduced this bug in #5456. But I expect that its practical impact is very limited: It is very difficult to reason about a repeated pattern that contains a lookahead assertion.

muellerj2:
Fixes #5792.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -1681,6 +1681,7 @@ enum class _Rx_unwind_ops {
     _Disjunction_eval_alt_always,
     _Do_nothing,
     _Loop_simple_nongreedy,
+    _Loop_simple_greedy,
 };
 
 template <class _BidIt>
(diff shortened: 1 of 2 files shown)


StephanTLavavej (approved):

StephanTLavavej:
Thanks for fixing this regression and adding test coverage! :cat:

StephanTLavavej:

StephanTLavavej:
Resolved trivial adjacent-add conflicts with #5790 in `tests/std/tests/VSO_0000000_regex_use/test.cpp`.

StephanTLavavej:
:derelict_house: :hammer_and_wrench: :house_with_garden:
===

PR #5796, opened 2025-10-22
Optimize `transform` of `vector<bool>` for more predicates

AlexGuteniev:
Towards #625, specifically https://github.com/microsoft/STL/issues/625#issuecomment-3307182440 item 3

Follow up to #5769

# ➡️ Optimization

@rodiers asked for set subtraction. 
> 3. A substract like method
>     /// Subtract the sets of true flags in both CFlags.
>     /// S1 = {i where *this[i] == true}, S2 = {i where ac_Other[i] == true}, Result[i] == true <=> i element of S with S=S1\S2

Out of the standard predicates `std::greater` does the trick: it is only true if the element is in first set and not in second set. Or `std::less` if converse. For completeness, let's add all four remaining comparisons.

# ❓ What it even is 

Apparently, there's no universally accepted name for the corresponding integer operation, unlike `xnor` for `equal_to`.

In the benchmark, we can keep comparison names, like `less_equal`. But we also need to name functors. Here are some ideas:
 * In Wikipedia, these operations defined as implication and nonimplication, and the gates are `imply` and `nimply`. The other two, when the other input is negated, are converse implication and converse nonimplication,
 * In our favorites ISA there are instructions to do one of these operation: `andn` (scalar, BMI1), `pandn` (MMX, SSE2), and more flavors of `pandn` for bigger vectors and element masks, all accessible as intrinsics, and generated automatically by the compiler. We can call these ops `andn` and `orn` therefore. With slight confusion with `nand` / `nor` that we don't use anyway, and with no obvious way how to mark one vs the other inputs as negated. 
 * A historian in a Stack Overflow comment mentions such names as "selective clear" and "selective set". Still without obvious way to define which arg is inverted.

In the PR I went with the Wikipedia naming, but I'm open to any other option

# ⏱️ Benchmark results

Benchmark                                    |    Before   |   After   | Speedup
---------------------------------------------|-------------|-----------|----------
`transform_two_inputs_aligned<less<>>/64`      |     130 ns  |  3.03 ns  | 42.9
`transform_two_inputs_aligned<less<>>/4096`    |   11462 ns  |  10.8 ns  | 1060
`transform_two_inputs_aligned<less<>>/65536`   |  264332 ns  |   141 ns  | 1870

# 🚗 Drive-by

There's a concise way to test for `void`, done that.


StephanTLavavej:
The 1870x speedup exceeds all but the literally exponential one :joy_cat: :heart_eyes_cat:

Thanks for the thorough test coverage and I agree with the names!

StephanTLavavej:

StephanTLavavej:
:ghost: :vampire: :bat:
===

PR #5798, opened 2025-10-22
`<regex>`: Process generic loops non-recursively

### LINKED ISSUE #997: <regex>: Grouping within repetition causes regex stack error

### LINKED ISSUE #1528: <regex>: regex_match gets caught in recursive loop until stack overflow occurs

muellerj2:
Fixes #997. Fixes #1528.

This is the final change to make the matcher non-recursive, but it will be followed by a few more PRs to clean up the code and simplify the data stored in `_Rx_state_frame`s.

The new code in `_Match_pat()` is closer to the structure of the original code than for simple loops, but also quite similar to the non-recursive code for simple loops, so I chose not to split it into several PRs.

Differences compared to simple loops:
* Since non-simple loops are potentially re-entrant and repeated patterns can branch, the loop variables have to be reset whenever we backtrack. This means we have to add more members to `_Rx_state_frame` to store the original values of these loop variables and have to add an unwinding opcode just for resetting these loop variables. Moreover, we have to completely restore the match state whenever we backtrack after failure (ECMAScript) or while matching a POSIX regex.
  * In the case of greedy matching, we can restore the loop variables in the same opcode that also sets up matching of the tail when matching another repetition failed. This has the slight advantage that we need to maintain fewer frames for backtracking while matching the tail.
* I kept the structure of conditions mostly the same between simple and generic loops, except that I exchanged the order of conditions for "empty match" and "minimum number of repetitions not reached yet". This order aligns more closely with `_Do_rep()` and is a bit simpler for generic loops, I think, because the logic for handling the minimum number of repetitions isn't affected much by whether the last encountered match is empty or not (except for the optimization of increasing `_Loop_idx` by more than one when a match is empty). This is different from simple loops, where the easy case was an empty match: Because simple loops are branchless, one empty match means all are empty, so the minimum number of repetitions essentially becomes meaningless when an empty match appears.

Most of the test coverage is derived from the new coverage for simple loops, with a minor change to make the loops non-simple by making them branch. I also added another test to make sure we correctly handle an empty match that only appears in the final repetition to reach the minimum number of repetitions.

As for the two tests to verify that stack overflows no longer occur (copied from the original issues), I deliberately set one of them up to fail when matching no longer results in an exception, because the code should be changed to use the regex tester when an exception is no longer thrown. (I can't easily do this for the other test case because it currently succeeds for x86 but fails for x64. It seems out of proportion to add preprocessor logic just to make the test fail on 64-bit systems when it succeeds unexpectedly.)

### DIFF
--- a/tests/std/tests/VSO_0000000_regex_use/test.cpp
+++ b/tests/std/tests/VSO_0000000_regex_use/test.cpp
@@ -904,7 +904,7 @@ void test_gh_997() {
     // GH-1528: <regex>: regex_match gets caught in recursive loop until stack overflow occurs
 
     try {
-        (void) regex_match(std::string(1025, 'a'), regex("(?:a)+"));
+        (void) regex_match(string(1025, 'a'), regex("(?:a)+"));
         assert(false); // adjust test when matching succeeds
     } catch (const regex_error& ex) {
         assert(ex.code() == error_stack);


stl/inc/regex
```diff
@@ -4078,8 +4014,33 @@ bool _Matcher3<_BidIt, _Elem, _RxTraits, _It, _Alloc>::_Match_pat(_Node_base* _N
                             }
                         }
                     } else {
-                        _Failed = !_Do_rep(_Node, _Greedy, 0);
-                        _Next   = nullptr;
+                        if (_Node->_Min > 0 || (_Greedy && !_Longest && _Node->_Max != 0)) { // try a rep first
+                            // set up stack unwinding for greedy matching or loop val restoration
+                            auto _Frame_idx = _Push_frame(
+                                _Node->_Min == 0 ? _Rx_unwind_ops::_Loop_greedy : _Rx_unwind_ops::_Loop_restore_vals,
```
StephanTLavavej: Suggestion: Extracting the code would make this easier to read (even though it wouldn't save a line).

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -899,6 +899,36 @@ void test_gh_993() {
...
+        assert(false); // adjust test when matching succeeds
+    } catch (const regex_error& ex) {
+        assert(ex.code() == error_stack);
+    }
+
+    try {
+        wregex rgx(L"^http[s]?\\:\\/\\/([^.]+\\.)*github\\.com\\/.*$", icase);
+
+        assert(!regex_match(
+            L"https://www.google.com/"
```
StephanTLavavej: Although this was taken from the original bug, we should use example domains instead of one that we own and one that we don't. I also don't know what's encoded in the remainder of the URL and it doesn't appear to be important to the scenario.

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -899,6 +899,36 @@ void test_gh_993() {
...
+
+    try {
+        (void) regex_match(std::string(1025, 'a'), regex("(?:a)+"));
+        assert(false); // adjust test when matching succeeds
+    } catch (const regex_error& ex) {
+        assert(ex.code() == error_stack);
+    }
+
+    try {
+        wregex rgx(L"^http[s]?\\:\\/\\/([^.]+\\.)*github\\.com\\/.*$", icase);
```
StephanTLavavej: I recommend using raw string literals whenever possible, to avoid having to mentally reverse C++ string literal escapes.

StephanTLavavej: Also, there's unnecessary escaping here (colons and forward slashes don't need to be escaped). Again I know this came from the user repro, and usually messing with a repro is unwise, but the unnecessary escaping obscures what the regex is doing and isn't important for the repro.

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -899,6 +899,36 @@ void test_gh_993() {
...
+
+        assert(!regex_match(
```
StephanTLavavej, line 918: Let's also add a positive test to make sure the regex can match something.

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -2232,6 +2262,89 @@ void test_gh_5792() {
     g_regexTester.should_match("bc", "(?:(?!ab))+bc");
 }
 
+void test_gh_5798() {
+    // GH-5792: <regex>: Process generic loops non-recursively.
```
StephanTLavavej: Copy-pasted citation of GH-5792.

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -899,6 +899,36 @@ void test_gh_993() {
     }
 }
 
+void test_gh_997() {
+    // GH-997: <regex>: Grouping within repetition causes regex stack error
+    // GH-1528: <regex>: regex_match gets caught in recursive loop until stack overflow occurs
+
+    try {
+        (void) regex_match(std::string(1025, 'a'), regex("(?:a)+"));
```
StephanTLavavej: This test is `using namespace std;`.

StephanTLavavej (approved):

StephanTLavavej:
Thanks, this is incredible! :heart_eyes_cat: I pushed some changes to the test, please double-check that I didn't mess anything up.

If the CLA Bot is snoozy I may need to close and reopen this PR.

StephanTLavavej:

StephanTLavavej:
An incredible achievement, thank you! :heart_eyes_cat: :tada: :smile_cat:

PR #5801, opened 2025-10-24
Add algorithm-specific detection macros for vectorization

StephanTLavavej:
No functional change for users; this keeps `_USE_STD_VECTOR_ALGORITHMS` as our [documented control macro](https://learn.microsoft.com/en-us/cpp/standard-library/vectorized-stl-algorithms?view=msvc-170).

To prepare for adding ARM64 and eventually ARM64EC vectorized implementations, this begins by adding a layer of architecture-specific detection macros. (This layer isn't strictly necessary, but it makes life easier because it (1) allows the control macro to shut everything down, (2) deals with how ARM64EC defines both `_M_ARM64EC` and `_M_X64`, and (3) centralizes how x64/x86 vectorization is always paired.)

Then this introduces algorithm-specific detection macros. This makes connections throughout the codebase easier to understand, because a given vectorized algorithm needs declarations of separately compiled functions (`__std_meow_4`), a wrapper template (`_Meow_vectorized`), callsites, and occasionally helper type traits (`_Vector_alg_in_meow_is_safe`). This will also make it *significantly* easier to implement algorithms one at a time for ARM64, since this PR is decomposing what was previously a monolithic mode.

The `_VECTORIZED_MEOW` macros are named to avoid confusion with any control macros (we usually use the `_HAS_BLAH` or `_USE_BLAH` patterns for control). They're deliberately the mirror image of the `_Meow_vectorized` wrappers, to further avoid confusion.

In simple cases, the macros are 1:1 with the wrapper templates. However, there are several "algorithm families" implemented in `vector_algorithms.cpp`, like the `minmax` and `minmax_element` families and the various `find` variants. I tried to introduce macros that corresponded to these algorithm families, since that's how they'll be implemented (and there would be no benefit to introducing an even finer-grained system).

There are a couple of places where I had to slightly adjust `if constexpr` logic to handle the `minmax` and `minmax_element` families potentially being variously active.

Finally, while removing the old monolithic macro, I allowed the code to expand to empty `extern "C" { ... }` blocks when no vectorization is active, because this is harmless and avoids cluttering up the code with even more verbosity.


tests/std/tests/GH_000431_equal_memcmp_is_safe/test.compile.pass.cpp
```diff
@@ -529,10 +529,10 @@ STATIC_ASSERT(test_equal_memcmp_is_safe_for_types<magic, DefaultComparisonOddSiz
 // The only difference between _Equal_memcmp_is_safe and _Vector_alg_in_search_is_safe is how the magic works
 STATIC_ASSERT(_Equal_memcmp_is_safe<DefaultComparison*, DefaultComparison*, equal_to<>> == magic);
 STATIC_ASSERT(_Equal_memcmp_is_safe<DefaultComparisonOddSize*, DefaultComparisonOddSize*, equal_to<>> == magic);
-#if _USE_STD_VECTOR_ALGORITHMS
+#if _VECTORIZED_FOR_X64_X86
```
AlexGuteniev: I think this deserves a comment like "should match `_Vector_alg_in_search_is_safe` visibility"

tests/std/tests/GH_000431_lex_compare_memcmp_classify/test.compile.pass.cpp
```diff
@@ -178,7 +178,7 @@ void test_lex_compare_memcmp_classify_for_types() {
 #endif // _HAS_CXX20
 }
 
-constexpr bool vec_alg = _USE_STD_VECTOR_ALGORITHMS;
+constexpr bool vec_alg = _VECTORIZED_FOR_X64_X86;
```
AlexGuteniev: I think this deserves a comment like "should match `_Mismatch_vectorized` availability"

stl/inc/xutility
```diff
@@ -41,17 +41,47 @@ _STL_DISABLE_CLANG_WARNINGS
...
 #endif // ^^^ _USE_STD_VECTOR_ALGORITHMS != 0 ^^^
-#endif // ^^^ no support for vector algorithms ^^^
+#else // ^^^ vector algorithms unsupported for /clr:pure or CHPE / vector algorithms supported by default vvv
+#ifndef _USE_STD_VECTOR_ALGORITHMS
+#define _USE_STD_VECTOR_ALGORITHMS 1
+#endif // !defined(_USE_STD_VECTOR_ALGORITHMS)
+#endif // ^^^ vector algorithms supported by default ^^^
+
+// _USE_STD_VECTOR_ALGORITHMS is our documented control macro:
+// https://learn.microsoft.com/en-us/cpp/standard-library/vectorized-stl-algorithms?view=msvc-170
```
AlexGuteniev, line 57: Maybe drop `?view=msvc-170` and never question whether this needs fixup for a newer toolset version

StephanTLavavej: Sure. I'll clean up other URLs in another PR.

AlexGuteniev: I'm actually hesitating now. Maybe older VS should keep trying to like older docs, and that's why the version needs to be saved in the URL?

StephanTLavavej: I'll revert since I don't care very much.

tests/std/tests/GH_000431_lex_compare_memcmp_classify/test.compile.pass.cpp
StephanTLavavej: Thanks. This is now `_VECTORIZED_MISMATCH` so no comment is necessary.

tests/std/tests/GH_000431_equal_memcmp_is_safe/test.compile.pass.cpp
StephanTLavavej: Thanks. This is now guarded by `_VECTORIZED_SEARCH` etc. with a preprocessor comment "`_Vector_alg_in_search_is_safe` is available" so the link should be clearer now.

stl/inc/xutility
```diff
@@ -41,17 +41,69 @@ _STL_DISABLE_CLANG_WARNINGS
...
+#error Unknown architecture.
+#endif // ^^^ unknown architecture ^^^
+
+#define _VECTORIZED_ADJACENT_FIND      _VECTORIZED_FOR_X64_X86
+#define _VECTORIZED_BITSET_FROM_STRING _VECTORIZED_FOR_X64_X86
+#define _VECTORIZED_BITSET_TO_STRING   _VECTORIZED_FOR_X64_X86
+#define _VECTORIZED_COUNT              _VECTORIZED_FOR_X64_X86
+#define _VECTORIZED_FIND               _VECTORIZED_FOR_X64_X86
+#define _VECTORIZED_FIND_END           _VECTORIZED_FOR_X64_X86
+#define _VECTORIZED_FIND_FIRST_OF      _VECTORIZED_FOR_X64_X86
```
AlexGuteniev, line 87: Have you intentionally combined all three flavors of this (free algorithm, string member, string member with `not`), but for _last of_ kept separate two?

stl/inc/__msvc_string_view.hpp
```diff
@@ -37,17 +37,23 @@ __declspec(noalias) size_t __stdcall __std_find_first_of_trivial_pos_4(
...
 __declspec(noalias) size_t __stdcall __std_find_last_of_trivial_pos_2(
     const void* _Haystack, size_t _Haystack_length, const void* _Needle, size_t _Needle_length) noexcept;
+#endif // ^^^ _VECTORIZED_FIND_LAST_OF ^^^
 
+#if _VECTORIZED_FIND
 const void* __stdcall __std_find_not_ch_1(const void* _First, const void* _Last, uint8_t _Val) noexcept;
 const void* __stdcall __std_find_not_ch_2(const void* _First, const void* _Last, uint16_t _Val) noexcept;
 const void* __stdcall __std_find_not_ch_4(const void* _First, const void* _Last, uint32_t _Val) noexcept;
 const void* __stdcall __std_find_not_ch_8(const void* _First, const void* _Last, uint64_t _Val) noexcept;
+#endif // ^^^ _VECTORIZED_FIND ^^^
```
AlexGuteniev, line 54: I see inconsistency here. These go to just find, but `__std_find_last_not_ch_pos_N` go to `_VECTORIZED_FIND_LAST_OF` rather than `_VECTORIZED_FIND_LAST`

stl/inc/xutility
StephanTLavavej: `_VECTORIZED_FIND_FIRST_OF` guards 
`__std_find_first_of_trivial_pos_[1248]`, `__std_find_first_not_of_trivial_pos_[12]`, `__std_find_first_of_trivial_[1248]`.

Function | Calls
---|---
`__std_find_first_of_trivial_pos_1` | `_Find_meow_of::_First_of::_Dispatch_pos<uint8_t, _Find_meow_of::_Predicate::_Any_of>`
`__std_find_first_not_of_trivial_pos_1` | `_Find_meow_of::_First_of::_Dispatch_pos<uint8_t, _Find_meow_of::_Predicate::_None_of>`
`__std_find_first_of_trivial_1` | `_Find_meow_of::_First_of::_Dispatch_ptr<uint8_t>`

As far as I could tell, `_Dispatch_ptr` and `_Dispatch_pos` share the same implementation although they differ in details.

---

`_VECTORIZED_FIND_LAST_OF` guards `__std_find_last_of_trivial_pos_[12]` and `__std_find_last_not_of_trivial_pos_[12]`.

Function | Calls
---|---
`__std_find_last_of_trivial_pos_1` | `_Find_meow_of::_Last_of::_Dispatch_pos<uint8_t, _Find_meow_of::_Predicate::_Any_of>`
`__std_find_last_not_of_trivial_pos_1` | `_Find_meow_of::_Last_of::_Dispatch_pos<uint8_t, _Find_meow_of::_Predicate::_None_of>`

---

`_VECTORIZED_FIND_LAST_NOT_OF` guards `__std_find_last_not_ch_pos_[1248]`.

Function | Calls
---|---
`__std_find_last_not_ch_pos_1`  | `_Finding::_Find_last_pos_impl<_Finding::_Find_traits_1, _Finding::_Predicate::_Not_equal>`

---

It appeared to me that `_Find_meow_of::_Last_of::_Dispatch_pos` didn't share commonality with `_Finding::_Find_last_pos_impl`. But you're right that the asymmetry is confusing, especially since `_VECTORIZED_FIND_LAST_OF` does control a "not" function, `__std_find_last_not_of_trivial_pos_[12]`.

I should probably unify them.

stl/inc/__msvc_string_view.hpp
StephanTLavavej: Guard | Function | Calls
---|---|---
`_VECTORIZED_FIND` | `__std_find_trivial_1` | `_Finding::_Find_impl<_Finding::_Find_traits_1, _Finding::_Predicate::_Equal>`
`_VECTORIZED_FIND` | `__std_find_not_ch_1` | `_Finding::_Find_impl<_Finding::_Find_traits_1, _Finding::_Predicate::_Not_equal>`

It is intentional that these two are unified, because they share exactly the same implementation modulo the predicate.

---

Guard | Function | Calls
---|---|---
`_VECTORIZED_FIND_LAST` | `__std_find_last_trivial_1` | `_Finding::_Find_last_impl<_Finding::_Find_traits_1, _Finding::_Predicate::_Equal>`
`_VECTORIZED_FIND_LAST_OF` | `__std_find_last_of_trivial_pos_1` | `_Find_meow_of::_Last_of::_Dispatch_pos<uint8_t, _Find_meow_of::_Predicate::_Any_of>`
`_VECTORIZED_FIND_LAST_OF` | `__std_find_last_not_of_trivial_pos_1` | `_Find_meow_of::_Last_of::_Dispatch_pos<uint8_t, _Find_meow_of::_Predicate::_None_of>`
`_VECTORIZED_FIND_LAST_NOT_OF` | `__std_find_last_not_ch_pos_1`  | `_Finding::_Find_last_pos_impl<_Finding::_Find_traits_1, _Finding::_Predicate::_Not_equal>`

There appeared to be 3 groups here, hence the three guards. (But confusing that the "find last not of" guard doesn't cover one of the "find last not of" functions.)

StephanTLavavej: Oh, I see, `_Find_last_pos_impl` just calls `_Find_last_impl` and translates the pointer result to a position, so they are the same.

AlexGuteniev: I see.

T mee it seems that `_VECTORIZED_FIND_LAST_NOT_OF` should go to `_VECTORIZED_FIND_LAST` the same way as `_Find_not_ch_vectorized` is guarded by just `_VECTORIZED_FIND`. 

The `__std_find_last_not_ch_pos_1` / `__std_find_last_trivial_1` pair and `__std_find_trivial_1` / `__std_find_not_ch_1` pair have equal amount of commonality.

StephanTLavavej: Yes, that's what I'm about to push after locally validating.

StephanTLavavej:

stl/inc/xutility
```diff
@@ -41,17 +41,47 @@ _STL_DISABLE_CLANG_WARNINGS
...
+#elif defined(_M_X64) || defined(_M_IX86)
+#define _VECTORIZED_FOR_X64_X86               1
+#define _VECTORIZED_FOR_X64_X86_ARM64         1
+#define _VECTORIZED_FOR_X64_X86_ARM64_ARM64EC 1
+#elif defined(_M_ARM64)
+#define _VECTORIZED_FOR_X64_X86               0
+#define _VECTORIZED_FOR_X64_X86_ARM64         1
+#define _VECTORIZED_FOR_X64_X86_ARM64_ARM64EC 1
+#else // ^^^ known architecture / unknown architecture vvv
+#error Unknown architecture.
```
davidmrdavid, line 78: non-blocking nit: 'Unknown architecture' is not a super actionable error message. It would be great if there was more we could say, such as 'supported architectures can be found here: X' or 'try Y`, etc.

But I realize that's going above and beyond, and that this error is super rare. So we can dismiss. Just wanted to call this out.

stl/inc/xutility
```diff
@@ -5905,13 +5910,13 @@ namespace ranges {
...
-#if _VECTORIZED_FOR_X64_X86
+#if _VECTORIZED_MISMATCH
     is_integral_v<_Elem1> && is_integral_v<_Elem2> && sizeof(_Elem1) == sizeof(_Elem2)
     && is_unsigned_v<_Elem1> == is_unsigned_v<_Elem2>;
-#else // ^^^ _VECTORIZED_FOR_X64_X86 / !_VECTORIZED_FOR_X64_X86 vvv
+#else // ^^^ _VECTORIZED_MISMATCH / !_VECTORIZED_MISMATCH vvv
     conjunction_v<_Is_character_or_bool<_Elem1>, _Is_character_or_bool<_Elem2>, is_unsigned<_Elem1>,
         is_unsigned<_Elem2>>;
-#endif // ^^^ !_VECTORIZED_FOR_X64_X86 ^^^
+#endif // ^^^ !_VECTORIZED_MISMATCH ^^^
```
davidmrdavid: for my own learning: - the [commit](https://github.com/microsoft/STL/pull/5801/commits/11a385eb9a71d731978f3fa80f46052e375b21cf#diff-980098f27df5ad089b2a93a8489574cfcab50f4b9d9c55524b0a57444a4c664eR5913-R5919) here asks me to note _something_ about `_Lex_compare_memcmp_classify_elements`, but it's unclear to me what you're trying to highlight. Can you call that out?

stl/inc/algorithm
```diff
@@ -11315,19 +11324,25 @@ namespace ranges {
...
+#endif // ^^^ _VECTORIZED_MINMAX ^^^
+                {
+                    if constexpr (_Is_min_max_optimization_safe<decltype(_UFirst), _Pr>) {
+                        if (!_STD is_constant_evaluated()) {
+                            const auto _First_ptr = _STD to_address(_UFirst);
+                            const auto _Last_ptr  = _First_ptr + static_cast<ptrdiff_t>(_ULast - _UFirst);
+                            const auto _Result    = _STD _Minmax_element_vectorized(_First_ptr, _Last_ptr);
+                            return {
+                                *static_cast<const _Vty*>(_Result.first), *static_cast<const _Vty*>(_Result.second)};
+                        }
```
davidmrdavid: I don't have enough context to understand why previously it was 'ok' for us to have this if-else chain without any macros, and now we need to 'erase' the body of the one of cases depending on `_VECTORIZED_MINMAX`. Can you please elaborate on what's changed here?

davidmrdavid (commented):
Almost ready to approve, just want to understand a few things.

stl/inc/xutility
AlexGuteniev: Note that it is also changed from `_VECTORIZED_FOR_X64_X86` to `_VECTORIZED_MISMATCH`. 
Specifically, that with fine-grained vectorization macros, this trait goes to `_VECTORIZED_MISMATCH` bucket and not elsewhere, and not separately.

stl/inc/xutility
StephanTLavavej: Thanks. This is a pre-existing pattern:

https://github.com/microsoft/STL/blob/1717ac0bde5412309975b1782efe2a73364cfba5/stl/inc/__msvc_sanitizer_annotate_container.hpp#L165-L167
https://github.com/microsoft/STL/blob/1717ac0bde5412309975b1782efe2a73364cfba5/stl/src/xonce2.cpp#L37-L39

I'm not concerned about the actionability of the error message - it is easy to look up the line number and figure out what happened. This should also be a "can't happen" error - we know all of the architectures we target, and unless we got confused (or we introduce a new one in the future), we will have exhaustively handled them.

stl/inc/xutility
StephanTLavavej: Yes, it's just a note to myself that this type trait, which on the surface isn't talking about `mismatch`, is actually related to `mismatch` vectorization.

(This was unusual enough to record in my commit messages; most of the other type traits / helpers / etc. are obviously linked to each macro being introduced.)

stl/inc/algorithm
StephanTLavavej: Previously, vectorization was all-or-nothing. Either we enabled STL vectorization and had both cases (minmax value optimization, minmax pointer optimization), or we disabled STL vectorization and had neither case.

Now I am preparing the codebase to have individual algorithm families independently vectorized, because I don't know in which order ARM64 vectorization will be implemented. So I had to adjust the control flow here to handle zero, one or the other, or both of these vectorizations being enabled.

davidmrdavid: Thanks. I think part of what made this hard to read for me was the weird/clever poisition of the macros. I expected them to neatly fall inside either the `if`- or the `else`- case, but the first macro (`_VECTORIZED_MINMAX`) ends slightly after the `else` keyword is declared, effectively removing it,  leaving an extra `{ }` scope around the cases. 

I see the intent now, I just needed to work out the cases. Approving shortly, thanks!

davidmrdavid (approved):

PR #5804, opened 2025-10-24
Deprecate experimental coroutine headers

StephanTLavavej:
In the MSVC Build Tools 14.50 shipping in VS 2026 18.0, the [`/await` compiler option](https://learn.microsoft.com/en-us/cpp/build/reference/await-enable-coroutine-support?view=msvc-170) was officially deprecated and emits a driver warning. However, driver warnings don't terminate builds, even with `/WX`:

```
C:\Temp>type meow.cpp
```
```cpp
#include <cstdio>
int main() {
    std::puts("Hello, deprecated world!");
}
```
```
C:\Temp>cl /EHsc /nologo /W4 /WX /await meow.cpp && meow
cl : Command line warning D9047 : option 'await' has been deprecated and will be removed in a future release.
    Standard C++ coroutines are available by default in C++20 or later, or use '/await:strict' in earlier language modes.
meow.cpp
Hello, deprecated world!
```

To turn up the heat, I want to "hard deprecate" the experimental coroutine headers, matching what we did with `<experimental/filesystem>`, `<hash_map>`, and `<hash_set>`. This means adding an impossible-to-miss compiler **error**, with an escape hatch. This will ship in MSVC Build Tools 14.51, making MSVC Build Tools 14.52 the earliest we could completely remove the headers (it remains to be seen whether we can/will be that aggressive).

I am also improving our various warning and error messages to guide users towards the modern Standard ways.

### DIFF
--- a/stl/inc/experimental/generator
+++ b/stl/inc/experimental/generator
@@ -17,8 +17,9 @@
 _EMIT_STL_ERROR(STL1009, "The <experimental/coroutine>, <experimental/generator>, and <experimental/resumable> "
                          "headers do not support Clang, but the C++20 <coroutine> and C++23 <generator> headers do.");
 #elif !defined(_SILENCE_EXPERIMENTAL_COROUTINE_DEPRECATION_WARNINGS)
-_EMIT_STL_ERROR(STL1011, "<experimental/generator> is deprecated by Microsoft and will be REMOVED IN THE NEAR FUTURE. "
-                         "It is superseded by the C++23 <generator> header. You can define "
+_EMIT_STL_ERROR(STL1011, "The /await compiler option, <experimental/coroutine>, <experimental/generator>, and "
+                         "<experimental/resumable> are deprecated by Microsoft and will be REMOVED SOON. "
+                         "They are superseded by the C++20 <coroutine> and C++23 <generator> headers. You can define "
                          "_SILENCE_EXPERIMENTAL_COROUTINE_DEPRECATION_WARNINGS to suppress this error for now.");
 #elif defined(__cpp_impl_coroutine)
 #include <coroutine>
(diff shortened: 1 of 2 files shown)


stl/inc/experimental/generator
```diff
@@ -13,12 +13,22 @@
...
 #include <memory>
 
-#if defined(__cpp_impl_coroutine)
+#ifdef __clang__
+_EMIT_STL_ERROR(STL1009, "The <experimental/coroutine>, <experimental/generator>, and <experimental/resumable> "
+                         "headers do not support Clang, but the C++20 <coroutine> and C++23 <generator> headers do.");
+#elif !defined(_SILENCE_EXPERIMENTAL_COROUTINE_DEPRECATION_WARNINGS)
+_EMIT_STL_ERROR(STL1011, "<experimental/generator> is deprecated by Microsoft and will be REMOVED IN THE NEAR FUTURE. "
+                         "It is superseded by the C++23 <generator> header. You can define "
+                         "_SILENCE_EXPERIMENTAL_COROUTINE_DEPRECATION_WARNINGS to suppress this error for now.");
```
AlexGuteniev, line 23: The same macro is used to suppress two different messages. They are different on which header is deprecated.

StephanTLavavej: I felt it was simpler to have a single silencing macro, although I suppose I could unify the messages.

AlexGuteniev: Even if you want them different, I think both messages should mention all headers, if they are suppressed by the same macro. You can probably start each of the messages with headers they are triggered by though.

StephanTLavavej: Unified, thanks.

davidmrdavid (approved):
I'll miss the "YOU HAVE BEEN WARNED" message, but agreed it's probably best without it.
"[...] FOR NOW" 😈 

LGTM.

StephanTLavavej:

kobykahane:
One reason we're still on the experimental generator header is DevCom-10920618 (apparently the MSVC equivalent of https://github.com/llvm/llvm-project/issues/56507) - using the new implementation is blocked on 32-bit x86 MSVC with /guard:cf. It'd be appreciated if its final removal from STL would be gated on this compiler bug being fixed.

StephanTLavavej:
Thanks, I've marked that bug as "STL-requested" and commented internally. A couple of back-end devs were analyzing it earlier this month, but I haven't seen a PR yet.

PR #5808, opened 2025-10-25
Unwrap call of `function` passed to `move_only_function`

AlexGuteniev:
Towards #5504.

One of the parts of the issue that can be addressed without having `copyable_function`.

# ✅ Coverage

Added just copy counter to see that the optimization works. It is copied zero times due to some core language optimization, but without the optimization it is copied once in the cases where it was missing.

Also added allocation counters to see that the other part of the optimization works.

Exercise different wrapping and also null functions.

# ⚙️ Optimization

Extract the inner `_Func_impl` / `_Func_impl_no_alloc` and use directly it without the containing `function`. This both gets call unwrapping and allocation avoidance.

# ⚠️ ABI and alignment issues

This change is only complicated in the current ABI. If in a future ABI it is possible to implement `function` the same way that `move_only_function` is implemented, things would be way easier.

For 32-bit we can't easily avoid allocation of small function. The buffer of `function` is aligned to `max_align_t` and the buffer inside `move_only_function` is aligned only as `void*` if it is next to vTable. `move_only_function` can contain objects of `max_align_t` alignment, but in that case the buffer size would be smaller. As type erasure already happened by the time we have `function`, we can't query actual size or alignment.

We probably could change `move_only_function` to make is emulated vtable the last pointer instead of the first, as we're not ABI-bound for `/std:c++latest` yet. But it doesn't look like a good idea, considering other, more common `move_only_function` usage.

# ❄️ Special cases

We throw `bad_function_call`, and don't make `move_only_function` containing null `function` null itself. Because the Standard is spelled out this way currently. 

If `function` is constructed inside `move_only_function` using `in_place_type`, this optimization is not engaged.

Also see #5806

### DIFF
--- a/tests/std/tests/GH_005504_avoid_function_call_wrapping/test.cpp
+++ b/tests/std/tests/GH_005504_avoid_function_call_wrapping/test.cpp
@@ -125,7 +125,7 @@ void test_plain_null(const bool throws) {
 }
 
 template <class OuterWrapper, class InnerWrapper>
-void test_wrapped_call(const bool outer_is_null, const bool outer_throws) {
+void test_wrapped_null(const bool outer_is_null, const bool outer_throws) {
     InnerWrapper inner{};
     OuterWrapper outer{move(inner)};
     assert(!inner);
@@ -160,7 +160,7 @@ int main() {
     alloc_checker{0}, test_plain_null<move_only_function<fn_type>>(false);
 
     // wrapped nulls
-    alloc_checker{0}, test_wrapped_call<function<fn_type>, function<fn_type>>(true, true);
-    alloc_checker{0}, test_wrapped_call<move_only_function<fn_type>, move_only_function<fn_type>>(true, false);
-    alloc_checker{0}, test_wrapped_call<move_only_function<fn_type>, function<fn_type>>(false, true);
+    alloc_checker{0}, test_wrapped_null<function<fn_type>, function<fn_type>>(true, true);
+    alloc_checker{0}, test_wrapped_null<move_only_function<fn_type>, move_only_function<fn_type>>(true, false);
+    alloc_checker{0}, test_wrapped_null<move_only_function<fn_type>, function<fn_type>>(false, true);
 }


tests/std/test.lst
```diff
@@ -270,6 +270,7 @@ tests\GH_005402_string_with_volatile_range
 tests\GH_005421_vector_algorithms_integer_class_type_iterator
 tests\GH_005472_do_not_overlap
 tests\GH_005546_containers_size_type_cast
+tests\GH_005504_avoid_function_call_wrapping
```
StephanTLavavej: Should be sorted.

stl/inc/functional
```diff
@@ -1467,6 +1494,22 @@ inline void __stdcall _Function_deallocate_large_default_aligned(_Move_only_func
     ::operator delete(_Self._Large_fn_ptr<void>());
 }
 
+template <class _Fn>
+inline void __stdcall _Function_destroy_old_large(_Move_only_function_data& _Self) noexcept {
```
StephanTLavavej: After `template` we don't say `inline` anymore.

stl/inc/functional
```diff
@@ -1441,6 +1448,18 @@ _NODISCARD _Rx __stdcall _Function_inv_large(const _Move_only_function_data& _Se
     }
 }
 
+template <class _Fn, class _Rx, bool _Noex, class... _Types>
+_NODISCARD _Rx __stdcall _Function_inv_old_large(const _Move_only_function_data& _Self, _Types&&... _Args)
+    noexcept(_Noex) {
+    return _Self._Large_fn_ptr<_Fn>()->_Do_call(_STD forward<_Types>(_Args)...);
```
AlexGuteniev: Taking advantage of void return here and just below, instead of handling `void` separately.
This is not inmost invoke, these signatures always match exactly.

stl/inc/functional
```diff
@@ -1965,16 +2064,20 @@ public:
         using _Vt = decay_t<_Fn>;
         static_assert(is_constructible_v<_Vt, _Fn>, "_Vt should be constructible from _Fn. "
                                                     "(N4950 [func.wrap.move.ctor]/6)");
+        using _VtInvQuals = _Call::template _VtInvQuals<_Vt>;
 
-        if constexpr (is_member_pointer_v<_Vt> || is_pointer_v<_Vt> || _Is_specialization_v<_Vt, move_only_function>) {
-            if (_Callable == nullptr) {
-                this->_Reset_to_null();
-                return;
+        if constexpr (is_same_v<_Vt, function<_Signature...>>) {
```
AlexGuteniev, line 2042: This optimization requires exactly the same signature and it avoids an invoke call (with potential objects copying) and often an allocation.

There's very similar but different optimization possible that would accept different but compatible signature, and will avoid an allocation, but not a call.

One thing at a time.

AlexGuteniev: My detailed analysis is in https://github.com/microsoft/STL/issues/5504#issuecomment-2893240109

stl/inc/functional
```diff
@@ -1577,9 +1628,38 @@ public:
...
+            if constexpr (alignof(max_align_t) == alignof(void*)) {
+                // 64-bit target, can put small function into small move_only_function directly
+                _Data._Impl = _Create_impl_ptr<_Impl_kind::_Old_fn_small, _Vt, _VtInvQuals>();
+                _Old_fn_impl->_Move(_Data._Buf_ptr<void*>());
+                _Func._Tidy();
+            } else {
+                // 32-bit target, cannot small function into small move_only_function directly
+                // due to potentially not enough alignment. Allocate large function
+                void* _Where = ::operator new((_Small_object_num_ptrs - 1) * sizeof(void*));
+                _Old_fn_impl->_Move(_Where);
```
AlexGuteniev: `_Move` does not throw for `function`. If the underlying callable has throwing move, it is not used, and the large function is engaged.

`_Tidy()` just below may call destructor of the underlying callable, if it is not trivial, but that destructor may not throw.

So the allocated memory doesn't need guard. 

(`move_only_function` shares this approach: its `_Move` does not throw either, because large function is always engaged for objects with throwing move.

Apparently there's no way to support small function optimization for callables with throwing move.)

stl/inc/functional
```diff
@@ -1467,6 +1494,23 @@ inline void __stdcall _Function_deallocate_large_default_aligned(_Move_only_func
...
+template <class _Fn>
+void __stdcall _Function_destroy_old_large(_Move_only_function_data& _Self) noexcept {
+    _Self._Large_fn_ptr<_Fn>()->_Delete_this(true);
+}
+
+template <class _Fn>
+void __stdcall _Function_destroy_old_small_as_large(_Move_only_function_data& _Self) noexcept {
+    _Fn* const _Old_fn_impl = _Self._Large_fn_ptr<_Fn>();
+    _Old_fn_impl->_Delete_this(false);
+    ::operator delete(static_cast<void*>(_Old_fn_impl));
```
AlexGuteniev, line 1506: The reason why "_small_as_large" exists as a separate mode rather than being just "_large" is that we avoid inner `_Deallocate` of `_Delete_this` by passing `false` there, and free the memory with direct `::operator delete` paired with direct `::operator new`.

We could call `_Allocate` to match `_Deallocate`, but we can't call it with exactly matching size and alignment, because we don't have that information, it was type-erased.

stl/inc/functional
```diff
@@ -1423,6 +1424,12 @@ template <class _Rx, class... _Types>
     _STL_UNREACHABLE; // no return value available for "continue on error"
 }
 
+template <class _Rx, bool _Noex, class... _Types>
+[[noreturn]] _Rx __stdcall _Function_old_not_callable(const _Move_only_function_data&, _Types&&...)
+    noexcept(_Noex) /* terminates if _Noex */ {
+    _Xbad_function_call();
```
AlexGuteniev, line 1438: In case of `_Noex` let the exception be thrown and let it trigger the `terminate` call, instead of calling `terminate` ourselves. This may aid debugging.

tests/std/tests/GH_005504_avoid_function_call_wrapping/test.cpp
```diff
@@ -0,0 +1,84 @@
...
+    }
+    {
+        move_only_function<function_type> fn{move_only_function<function_type>{large_callable{}}};
+        assert(fn(copy_counter{}) == 0);
+    }
+
+    // Moves from function to move_only_function
+    {
+        move_only_function<function_type> fn{function<function_type>{small_callable{}}};
+        assert(fn(copy_counter{}) == 0);
```
AlexGuteniev: When this optimization is not implemented, this and the next one returns 1.

stl/inc/functional
AlexGuteniev: It may also feel like we defeated the purpose of inner allocator of `function`, and also likely have to allocate more memory than we need.

But remember that before this optimization we did exactly the same: `move_only_function` does not have inner allocators, and would have allocated the whole `sizeof(std::function)`.

So that's the scenario when this optimization is not useful to avoid allocation, but it doesn't make things any worse either, and the extra invoke is still avoided.

AlexGuteniev: Another view on this situation is that old `std::function` implementation should not have taken that `bool` in `_Delete_this`. It could know if it is small or large at compile time. The deleting path for small functions is essentially dead code. 

(`move_only_function` is superior here: it uses different implementations of deleters, including no deleter at all for trivially-destructible small functions and generic deallocator for trivially-destructible and not overaligned large functions (shared for all `move_only_function`s and all callables). It both improves runtime performance and reduces code bloat)

stl/inc/functional
AlexGuteniev: Actually, this will not happen with the current change.
The current optimization is for exactly matching signatures, and `std::function` template can't take `noexcept`.

AlexGuteniev: Removed `noexcept` from old invokes for now.

stl/inc/functional
```diff
@@ -1097,6 +1102,10 @@ protected:
 #endif // _HAS_STATIC_RTTI
 
 private:
+#if _HAS_CXX23
+    friend class _Move_only_function_base<_Ret, false, _Types...>;
```
StephanTLavavej: We can use extended `friend` here (i.e. omit `class` and avoid the *elaborated-type-specifier*) because the friendship itself is not templated.

stl/inc/functional
```diff
@@ -1568,9 +1625,38 @@ public:
...
+        if (_Old_fn_impl == nullptr) {
+            _Data._Impl = _Create_impl_ptr<_Impl_kind::_Old_fn_null, _Vt, _VtInvQuals>();
+        } else if (_Func._Local()) {
+            if constexpr (alignof(max_align_t) == alignof(void*)) {
+                // 64-bit target, can put small function into small move_only_function directly
+                _Data._Impl = _Create_impl_ptr<_Impl_kind::_Old_fn_small, _Vt, _VtInvQuals>();
+                _Old_fn_impl->_Move(_Data._Buf_ptr<void*>());
+                _Func._Tidy();
+            } else {
+                // 32-bit target, cannot small function into small move_only_function directly
```
StephanTLavavej: Missing "put".

stl/inc/functional
```diff
@@ -1661,10 +1747,30 @@ public:
...
+                _Impl._Destroy = _Function_destroy_old_large<_Fn>;
+            } else if constexpr (_Kind == _Impl_kind::_Old_fn_small_as_large) {
+                _Impl._Invoke  = _Function_inv_old_large<_Fn, _Rx, _Types...>;
+                _Impl._Move    = nullptr;
+                _Impl._Destroy = _Function_destroy_old_small_as_large<_Fn>;
+            } else if constexpr (_Kind == _Impl_kind::_Old_fn_small) {
+                _Impl._Invoke  = _Function_inv_old_small<_Fn, _Rx, _Types...>;
+                _Impl._Move    = _Function_move_old_small<_Fn>;
+                _Impl._Destroy = _Function_destroy_old_small<_Fn>;
+            }
```
StephanTLavavej: Let's `static_assert` that the cases are exhaustive.

stl/inc/functional
StephanTLavavej: You've gone further than I would have - I might not have bothered to optimize 32-bit at all. Thanks for the explanation.

stl/inc/functional
StephanTLavavej: Agreed. Yeah, we have to reject throwing moves, because we need to be nothrow swappable. I remember implementing that years ago.

stl/inc/functional
StephanTLavavej: Starting with matching signatures is definitely good. I'm not sure if handling varying signatures is worth additional implementation complexity, but I could be persuaded.

stl/inc/functional
```diff
@@ -1568,9 +1625,38 @@ public:
...
     }
 
+    template <class _Vt, class _VtInvQuals, class _Fn>
+    void _Construct_with_old_fn(_Fn&& _Func) {
+        const auto _Old_fn_impl = _Func._Getimpl();
+        if (_Old_fn_impl == nullptr) {
+            _Data._Impl = _Create_impl_ptr<_Impl_kind::_Old_fn_null, _Vt, _VtInvQuals>();
+        } else if (_Func._Local()) {
+            if constexpr (alignof(max_align_t) == alignof(void*)) {
+                // 64-bit target, can put small function into small move_only_function directly
```
StephanTLavavej: This is weird - we almost never handle 32-bit vs. 64-bit with `if constexpr`. I would probably be willing to accept this unusual pattern, except for how it leaves significant parts of code unused for each architecture.

I think we should be more conventional and have `_WIN64` codepaths that clearly mark who uses what.

StephanTLavavej:
Thanks! :heart_eyes_cat: I've fully reviewed the product code, pushed a `_WIN64` reorganization, and it looks good now. Back to you to enhance the test code until you're happy with it, then send it back to me.

stl/inc/functional
AlexGuteniev: I don't know how much cases `move_only_function<int(int) noexcept>(function<int(int)>(...))` or `move_only_function<void(int)>(function<int(int)>(...))` are expected to be common.

But I expect that it will not add much complexity on top of this one.

The main part is adjusting `_Function_inv_old_large` and `_Function_inv_old_small` to do `invoke` with `void` special case instead of direct propagation.. And propagating the info whether we going exact signature path or inexact signature path.

I expect for the rest of stuff to stay the same.

I admit the "small as large" case will be completely unuseful for inexact signatures, as neither an allocation, nor a call is avoided, but still not making special exclusion seems cleaner.

AlexGuteniev:
> Back to you to enhance the test code until you're happy with it, then send it back to me.

I've pushed some simplifications to product code, and also expanded test coverage to actually test null functions and count allocations.

tests/std/tests/GH_005504_avoid_function_call_wrapping/test.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+using namespace std;
+
+#pragma warning(disable : 4324) // 'large_callable': structure was padded due to alignment specifier
+#pragma warning(disable : 28251) // Inconsistent annotation for 'new': this instance has no annotations.
+
+int alloc_count   = 0;
+int dealloc_count = 0;
+
+size_t adjust_alloc_size(const size_t size) {
```
StephanTLavavej, line 20: Technically `<cstddef>`.

tests/std/tests/GH_005504_avoid_function_call_wrapping/test.cpp
```diff
@@ -0,0 +1,162 @@
...
+void* check_alloc(void* const result) {
+    if (!result) {
+        throw bad_alloc{};
+    }
+    return result;
+}
+
+void* operator new(const size_t size) {
+    ++alloc_count;
+    return check_alloc(malloc(adjust_alloc_size(size)));
```
StephanTLavavej, line 33: `<cstdlib>`.

tests/std/tests/GH_005504_avoid_function_call_wrapping/test.cpp
```diff
@@ -0,0 +1,162 @@
...
+        assert(context == 42);
+        return counter.count;
+    }
+};
+
+struct alignas(128) large_callable {
+    const int context = 1729;
+
+    int operator()(const copy_counter& counter) {
+        assert((reinterpret_cast<uintptr_t>(this) & 0x7f) == 0);
```
StephanTLavavej, line 87: `<cstdint>`.

tests/std/tests/GH_005504_avoid_function_call_wrapping/test.cpp
```diff
@@ -0,0 +1,162 @@
...
+template <class Wrapper, class Callable>
+void test_plain_call(const int expected_copies) {
+    Wrapper fn{Callable{}};
+    assert(fn(copy_counter{}) == expected_copies);
+}
+
+template <class OuterWrapper, class InnerWrapper, class Callable>
+void test_wrapped_call(const int expected_copies) {
+    InnerWrapper inner{Callable{}};
+    OuterWrapper outer{move(inner)};
```
StephanTLavavej, line 102: `<utility>`.

tests/std/tests/GH_005504_avoid_function_call_wrapping/test.cpp
```diff
@@ -0,0 +1,162 @@
...
+    assert(!inner);
+    assert(outer(copy_counter{}) == expected_copies);
+}
+
+template <class Wrapper>
+void check_call_null(Wrapper& wrapper, const bool throws) {
+    if (throws) {
+        try {
+            wrapper(copy_counter{});
+            abort(); // should not reach
```
StephanTLavavej: I think it would be more consistent to `assert(false)` so we always use the same failure mechanism.

tests/std/tests/GH_005504_avoid_function_call_wrapping/test.cpp
```diff
@@ -0,0 +1,162 @@
...
+        } catch (bad_function_call&) {
```
StephanTLavavej: It doesn't actually matter, but we traditionally catch by const reference.

tests/std/tests/GH_005504_avoid_function_call_wrapping/test.cpp
```diff
@@ -0,0 +1,162 @@
...
+    InnerWrapper inner{};
+    OuterWrapper outer{move(inner)};
+    assert(!inner);
+    assert(!outer == outer_is_null);
+    check_call_null(outer, outer_throws);
+}
+
+int main() {
+    // Plain calls
+    alloc_checker(0), test_plain_call<function<fn_type>, small_callable>(0);
```
StephanTLavavej: `alloc_checker` is a struct and you're deliberately constructing a temporary, so we should use braces to make this unusual pattern clearer.

tests/std/tests/GH_005504_avoid_function_call_wrapping/test.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+template <class Wrapper>
+void test_plain_null(const bool throws) {
+    Wrapper fn{};
+    assert(!fn);
+    check_call_null(fn, throws);
+}
+
+template <class OuterWrapper, class InnerWrapper>
+void test_wrapped_call(const bool outer_is_null, const bool outer_throws) {
```
StephanTLavavej: This overloads `test_wrapped_call` for 2 arguments, but I'm virtually certain you meant to name this `test_wrapped_null`, especially because it's commented as `// wrapped nulls` below.

StephanTLavavej:
Thanks! :heart_eyes_cat: I renamed a test function, please double-check.

tests/std/tests/GH_005504_avoid_function_call_wrapping/test.cpp
AlexGuteniev: Yes, missed edit after copying, and unintended overload.

StephanTLavavej:

StephanTLavavej:
:package: :gift: :gift_heart:
===

stl/inc/functional
AlexGuteniev: Not doing different signatures here per #5862 resolution

PR #5809, opened 2025-10-25
`move_only_function`: remove workaround for broken `noexcept(param)`

AlexGuteniev:
This seems to be fixed. There used to be compilation error.

### DIFF
--- a/stl/inc/functional
+++ b/stl/inc/functional
@@ -1535,17 +1535,6 @@ _NODISCARD void* _Function_new_large(_CTypes&&... _Args) {
 template <class _Rx, bool _Noexcept, class... _Types>
 class _Move_only_function_base {
 public:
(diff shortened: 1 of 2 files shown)


AlexGuteniev:
I have not verified if that `noexcept` actually noexcepts.
And that would be hard to verify, because there's `noexcept` anyway in both the function that is assigned to that pointer, and the `operator()` from which it is called. It is just an optimization believed to be useful, but hard to obserbe.

StephanTLavavej:
Great catch! :heart_eyes_cat: The internal bug was resolved back in May 2025 when the test team verified that the repro had been fixed, although a particular PR wasn't linked.

StephanTLavavej:

StephanTLavavej:
:magic_wand: :broom: :smile_cat:
===

PR #5810, opened 2025-10-26
`move_only_function`: eliminate the repeated part in call specializations

AlexGuteniev:
This part is repeated, and can be moved into base class.

Also adjust the script to a newer preprocessor comment.

We have encapsulation by privately inheriting `_Move_only_function_call`, and enabling `result_type` via `using typename _Call::result_type;`, this seems unaffected.

### DIFF
--- a/stl/inc/functional
+++ b/stl/inc/functional
@@ -1537,17 +1537,6 @@ class _Move_only_function_base {
 public:
     using result_type = _Rx;
 
-    // TRANSITION, DevCom-1208330: use noexcept(_Noexcept) instead
-    template <bool>
-    struct _Invoke_t {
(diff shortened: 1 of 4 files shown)


StephanTLavavej:
Thanks! :heart_eyes_cat: Verified that the divergence between script and product code is now fixed.

StephanTLavavej:

StephanTLavavej (approved):

StephanTLavavej:
I've resolved a trivial adjacent add-delete conflict with #5809 in `<functional>`.

StephanTLavavej:
:cat: :cat: :cat:
===

Issue #813, opened 2020-05-08
Extend vectorized algorithms to ARM64/ARM64EC


cbezault:
This work is ongoing thanks to the Arm engineers @hazzlim and @jaykang10, with assistance from the Vector Lord @AlexGuteniev. Dashboard as of 2026-03-24:

https://github.com/microsoft/STL/blob/7b9a0f5d6fa70641a61408da21afec131b40f323/stl/inc/xutility#L76-L101

StephanTLavavej:
`C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\VC\Tools\MSVC\14.26.28801\include\isa_availability.h` defines:
* `__ISA_AVAILABLE_ARMNT` (no NEON support)
* `__ISA_AVAILABLE_NEON` (128-bit NEON support for 32-bit ARM)
* `__ISA_AVAILABLE_NEON_ARM64` (128-bit NEON support for ARM64; there's a comment about how the distinction is "temporary" but I suspect it's permanent)

sylveon:
We can assume NEON is present in both 32 bits and 64 bits. https://docs.microsoft.com/en-ca/cpp/build/overview-of-arm-abi-conventions specifies WoA requires it.

StephanTLavavej:
Updating this issue to no longer mention ARM32; at this time we still need to keep it compiling and working, but we no longer care about optimizing for it. Only ARM64 performance matters.

AlexGuteniev:
Will need to take care of alignment; see #5766

Edit: no, we don't. ARM64 handles misaligned well.

StephanTLavavej:
This work is now ongoing! The "dashboard" is in `<xutility>`:

https://github.com/microsoft/STL/blob/8784104ce0d272ebbf2307371cc653418df7cb4d/stl/inc/xutility#L81-L105

XuSai1998:
Can find neon optimization be used now?

AlexGuteniev:
> Can find neon optimization be used now?

There's no vectorization implemented yet, not even in a PR, the current state is this:

https://github.com/microsoft/STL/blob/88db4d5be11df4ec7c2374dc25b6fae00172d29b/stl/inc/xutility#L81-L105

I hope it will be there some time next year though, given the current @hazzlim 's pace.

---

Another way to address this problem is auto-vectorization. Currently it works well for simple algorithms without early-return, but `find` is an early return one, so the situation is not likely to improve soon.

[...]

PR #5819, opened 2025-10-31
Add Neon implementation of std::swap_ranges

hazzlim:
Add an implementation of std::swap_ranges using Neon intrinsics.

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -84,9 +84,9 @@ extern "C" {
 #ifdef _M_ARM64
 __declspec(noalias) void __cdecl __std_swap_ranges_trivially_swappable_noalias(
     void* _First1, void* const _Last1, void* _First2) noexcept {
-    constexpr size_t _Mask_64 = ~((static_cast<size_t>(1) << 6) - 1);
     if (_Byte_length(_First1, _Last1) >= 64) {
-        const void* _Stop_at = _First1;
+        constexpr size_t _Mask_64 = ~((static_cast<size_t>(1) << 6) - 1);
+        const void* _Stop_at      = _First1;
         _Advance_bytes(_Stop_at, _Byte_length(_First1, _Last1) & _Mask_64);
         do {
             const uint8x16_t _Left1  = vld1q_u8(static_cast<uint8_t*>(_First1) + 0);


hazzlim:
@microsoft-github-policy-service agree company="Arm"

hazzlim:
Hopefully I’ve done something reasonable here - please let me know if you would prefer a different approach when adding new implementations to vector_algorithms.cpp

The performance numbers are below - looks good apart from the case of size(1), which I think is due to the added overhead of the function call and the conditional checks now `std::swap_ranges` is not being inlined into the benchmark.

<head>

<link rel=File-List

</head>

  | MSVC Speedup
-- | --
std_swap_ranges<uint8_t,   highly_aligned_allocator>/1 | 0.5x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/5 | 0.9x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/15 | 1.9x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/26 | 4.1x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/38 | 4.8x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/60 | 7.3x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/125 | 12.3x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/800 | 21.9x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/3000 | 22.4x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/9000 | 22.5x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/1 | 0.5x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/5 | 1x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/15 | 1.9x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/26 | 3.9x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/38 | 4.9x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/60 | 7.5x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/125 | 11.8x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/800 | 14.6x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/3000 | 15.1x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/9000 | 14.7x
  |  
  |  
  | clang-cl Speedup
std_swap_ranges<uint8_t,   highly_aligned_allocator>/1 | 0.6x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/5 | 1.1x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/15 | 1.5x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/26 | 1.4x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/38 | 1.4x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/60 | 1.5x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/125 | 1.5x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/800 | 1x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/3000 | 1x
std_swap_ranges<uint8_t,   highly_aligned_allocator>/9000 | 1x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/1 | 0.6x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/5 | 1.2x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/15 | 1.5x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/26 | 1x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/38 | 1.2x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/60 | 1.4x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/125 | 1.4x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/800 | 1x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/3000 | 1x
std_swap_ranges<uint8_t,   not_highly_aligned_allocator>/9000 | 1.1x



</body>

</html>

stl/src/vector_algorithms.cpp
```diff
@@ -5,14 +5,14 @@
 #error _M_CEE_PURE should not be defined when compiling vector_algorithms.cpp.
 #endif
 
-#if defined(_M_IX86) || defined(_M_X64) // NB: includes _M_ARM64EC
+#if defined(_M_IX86) || defined(_M_X64) || defined(_M_ARM64) // NB: includes _M_ARM64EC
```
AlexGuteniev: This addition made the comment more confusing.
I think it deserves changing to `// NB: _M_X64 includes _M_ARM64EC`

stl/src/vector_algorithms.cpp
```diff
@@ -78,6 +82,107 @@ namespace {
 
 extern "C" {
 
+#ifdef _M_ARM64
+__declspec(noalias) void __cdecl __std_swap_ranges_trivially_swappable_noalias(
+    void* _First1, void* const _Last1, void* _First2) noexcept {
+
```
AlexGuteniev: Unnecessary newline.
We do separate parts of functions with empty lines, like you did below, but don't start the body with that.

stl/src/vector_algorithms.cpp
```diff
@@ -156,6 +261,7 @@ __declspec(noalias) void __cdecl __std_swap_ranges_trivially_swappable_noalias(
         *_First2c               = _Ch;
     }
 }
+#endif // ^^^ !defined(_M_ARM64)
```
AlexGuteniev: `// ^^^ !defined(_M_ARM64) ^^^` for consistency with other preprocessor comments

stl/src/vector_algorithms.cpp
```diff
@@ -166,6 +272,7 @@ void* __cdecl __std_swap_ranges_trivially_swappable(
 
 } // extern "C"
 
+#if !defined(_M_ARM64)
```
AlexGuteniev: Just `#ifndef`. Even if you anticipate some additions in the future - change to `#if` can be done when the said additions are done.

stl/src/vector_algorithms.cpp
```diff
@@ -7694,4 +7801,5 @@ __declspec(noalias) bool __stdcall __std_bitset_from_string_2(void* const _Dest,
 }
 
 } // extern "C"
-#endif // defined(_M_IX86) || defined(_M_X64)
+#endif // ^^^ !defined(_M_ARM64)
```
AlexGuteniev: Ditto `// ^^^ !defined(_M_ARM64) ^^^` for consistency with other preprocessor comments

hazzlim:
> I've made some nitpick about whitespace, preprocessor, and comments. (I'm something of preprocessor comments traditions keeper myself).
> 
> If you're addressing them, please add commits, don't force-push.

Nice, thanks for reviewing - should all be addressed in the follow-up commit :)

stl/src/vector_algorithms.cpp
```diff
@@ -78,6 +82,106 @@ namespace {
...
+    void* _First1, void* const _Last1, void* _First2) noexcept {
+    constexpr size_t _Mask_64 = ~((static_cast<size_t>(1) << 6) - 1);
+    if (_Byte_length(_First1, _Last1) >= 64) {
+        const void* _Stop_at = _First1;
+        _Advance_bytes(_Stop_at, _Byte_length(_First1, _Last1) & _Mask_64);
+        do {
+            const uint8x16_t _Left1  = vld1q_u8(static_cast<uint8_t*>(_First1) + 0);
+            const uint8x16_t _Left2  = vld1q_u8(static_cast<uint8_t*>(_First1) + 16);
+            const uint8x16_t _Left3  = vld1q_u8(static_cast<uint8_t*>(_First1) + 32);
+            const uint8x16_t _Left4  = vld1q_u8(static_cast<uint8_t*>(_First1) + 48);
```
AlexGuteniev, line 95: Question: why don't you use `vld1q_u8_x4` (and `vld1q_u8_x2` below)?
Are they not available or not always available? Do they have no advantages?

stl/src/vector_algorithms.cpp
```diff
@@ -5,14 +5,14 @@
 #error _M_CEE_PURE should not be defined when compiling vector_algorithms.cpp.
 #endif
 
-#if defined(_M_IX86) || defined(_M_X64) // NB: includes _M_ARM64EC
+#if defined(_M_IX86) || defined(_M_X64) || defined(_M_ARM64) // NB: _M_X64 includes _M_ARM64EC
```
cpplearner: IIUC this condition is always true and can be removed: at least one of `_M_IX86`, `_M_X64`, and `_M_ARM64` must be defined when compiling this file, because this STL only supports targeting x64 (which has `_M_X64` defined), x86 (`_M_IX86`), ARM64 (`_M_ARM64`), and ARM64EC (`_M_X64`).

hazzlim: Aha good point, I have removed this now

stl/src/vector_algorithms.cpp
hazzlim: So we generally prefer intrinsics that generate LDP of NEON registers (as I have done here) rather than those that generate LD1 for multiple NEON registers because these have a greater range of addressing modes, and are faster on some cores. The version here is also slightly faster on the microsoft/STL microbenchmarks than using vld1q_u8_x4/vld1q_u8_x2

Although the tradeoff there is that LD1 is better from a code-size perspective, because in len >= 64 case we have:
LD1 {v0, v1, v2, v3}
instead of
LDP v0, v1
LDP q2, q3
(And its also better from a codesize perspective in the case of len >= 32 at present, because MSVC is not quite as good as you might hope at generating LDPs)

AlexGuteniev: I'd definitely favor speed over size most of the time, thanks for explanation!

AlexGuteniev: I've seen DevCom-10992168

hazzlim: Ok great, I’ll leave as-is

stl/src/vector_algorithms.cpp
```diff
@@ -156,6 +259,7 @@ __declspec(noalias) void __cdecl __std_swap_ranges_trivially_swappable_noalias(
         *_First2c               = _Ch;
     }
 }
+#endif // ^^^ !defined(_M_ARM64) ^^^
 
 // TRANSITION, ABI: __std_swap_ranges_trivially_swappable() is preserved for binary compatibility
 void* __cdecl __std_swap_ranges_trivially_swappable(
```
StephanTLavavej: The legacy function `__std_swap_ranges_trivially_swappable` was never defined for ARM64, so we don't need to start providing it.

This exists because it was briefly defined and used for x64/x86. Theoretically, there could be x64/x86 OBJs/LIBs out there that were compiled by users during that brief window of releases, and if the final link is performed by an updated MSVC Build Tools, we need to provide this legacy function that the old x64/x86 OBJs/LIBs would be looking for.

But before your PR, ARM64 OBJs/LIBs have never looked for any definitions from `vector_algorithms.cpp`, so we can start from a clean slate.

stl/src/vector_algorithms.cpp
```diff
@@ -21,10 +20,14 @@ extern "C" long __isa_enabled;
 #ifndef _DEBUG
 #pragma optimize("t", on) // Override /Os with /Ot for this TU
 #endif // !defined(_DEBUG)
-#endif // ^^^ !defined(_M_ARM64EC) ^^^
+#endif // ^^^ !defined(_M_ARM64) && !defined(_M_ARM64EC) ^^^
```
StephanTLavavej: I believe we want to unconditionally override the STL's (unwise, probably accidental) use of [`/Os`](https://learn.microsoft.com/en-us/cpp/build/reference/os-ot-favor-small-code-favor-fast-code?view=msvc-170) for its separately compiled code (see #2108). This [`#pragma optimize`](https://learn.microsoft.com/en-us/cpp/preprocessor/optimize?view=msvc-170) should be moved out of the architecture-varying header inclusions. I think it was already a mistake to guard this for ARM64EC.

The docs say "It takes effect at the first function defined after the pragma is seen.", so I believe we should actually move this up.

stl/src/vector_algorithms.cpp
```diff
@@ -78,6 +81,106 @@ namespace {
...
+    if (_Byte_length(_First1, _Last1) >= 4) {
+        const void* _Stop_at = _First1;
+        _Advance_bytes(_Stop_at, _Byte_length(_First1, _Last1) & _Mask_4);
+        do {
+            uint32x2_t _Left  = vdup_n_u32(0);
+            uint32x2_t _Right = vdup_n_u32(0);
+            _Left             = vld1_lane_u32(static_cast<uint32_t*>(_First1), _Left, 0);
+            _Right            = vld1_lane_u32(static_cast<uint32_t*>(_First2), _Right, 0);
+            vst1_lane_u32(static_cast<uint32_t*>(_First1), _Right, 0);
+            vst1_lane_u32(static_cast<uint32_t*>(_First2), _Left, 0);
```
StephanTLavavej: Just confirming, no change requested: This is accessing `_First1` and `_First2` as `uint32_t*`, but they aren't necessarily 4-byte aligned. Is this cromulent?

We have test coverage for this scenario so I think we're good:

https://github.com/microsoft/STL/blob/1118f375859015034d221c6fdba73a9605c4a086/tests/std/tests/VSO_0000000_vector_algorithms/test.cpp#L1119-L1123

stl/src/vector_algorithms.cpp
```diff
@@ -78,6 +81,106 @@ namespace {
...
+            _Advance_bytes(_First1, 64);
+            _Advance_bytes(_First2, 64);
+        } while (_First1 != _Stop_at);
+    }
+
+    constexpr size_t _Mask_32 = ~((static_cast<size_t>(1) << 5) - 1);
+    if (_Byte_length(_First1, _Last1) >= 32) {
+        const void* _Stop_at = _First1;
+        _Advance_bytes(_Stop_at, _Byte_length(_First1, _Last1) & _Mask_32);
+        do {
```
StephanTLavavej: I see a minor efficiency improvement here. This structure is following the x64/x86 code (which is great; our codebase greatly values consistency which makes reviewing and avoiding bugs so much easier). However, that code has to deal with AVX2 and SSE4.2 potentially not being available, so each step might need to loop.

For ARM64, the structure is much more uniform. Only the initial 64-byte step needs to loop. After that, we have at most 63 bytes remaining, so if we can consume a 32-byte chunk, there will be only one. After that, we have at most 31 bytes remaining, so if we can consume a 16-byte chunk, there will be only one, etc.

This will allow us to avoid some branches, which should be a pure perf win.

stl/src/vector_algorithms.cpp
```diff
@@ -78,6 +81,106 @@ namespace {
 
 extern "C" {
 
+#ifdef _M_ARM64
+__declspec(noalias) void __cdecl __std_swap_ranges_trivially_swappable_noalias(
+    void* _First1, void* const _Last1, void* _First2) noexcept {
+    constexpr size_t _Mask_64 = ~((static_cast<size_t>(1) << 6) - 1);
+    if (_Byte_length(_First1, _Last1) >= 64) {
+        const void* _Stop_at = _First1;
+        _Advance_bytes(_Stop_at, _Byte_length(_First1, _Last1) & _Mask_64);
```
StephanTLavavej: `_Mask_64` can be scoped more tightly. (I'll apply this to the existing x64/x86 code in a followup PR to avoid distracting changes here.)

StephanTLavavej:
Thanks @hazzlim, this is awesome! :heart_eyes_cat: I pushed some commits, can you rerun your benchmark measurements? We finally have ARM64 runtime testing in PR checks, but I won't be able to gather perf measurements myself until Feb 2026-ish. The changes to override `/Os` for ARM64, and to eliminate unnecessary loops, should improve performance but perhaps by an unobservable amount.

StephanTLavavej:

stl/src/vector_algorithms.cpp
hazzlim: I can see this was discussed on the discord also, but for the record here - yes this is fine as the LD1 lane instructions this will produce allow unaligned loads.

hazzlim:
> Thanks @hazzlim, this is awesome! 😻 I pushed some commits, can you rerun your benchmark measurements? We finally have ARM64 runtime testing in PR checks, but I won't be able to gather perf measurements myself until Feb 2026-ish. The changes to override `/Os` for ARM64, and to eliminate unnecessary loops, should improve performance but perhaps by an unobservable amount.

Nice, thanks for doing this! Not sure how I missed that we could remove the loops for `len < 64`,  nice one 😺 

I will re-run the perf and report, it may well be unobservable but I think it should also improve the LDP generation so that's a win :)

StephanTLavavej:
Thanks for working towards #813 and congratulations on your first microsoft/STL commit! :heart_eyes_cat: :tada: :mechanical_arm:

This will ship in the MSVC Build Tools 14.51 in a future update to Visual Studio 2026.

PR #5828, opened 2025-11-03
`<regex>`: Skip stack unwinding when encountering a successful match of an ECMAScript regex

muellerj2:
In ECMAScript mode (and in those cases where leftmost-longest mode behaves the same), the matcher returns the first encountered successful match. This means that `_Match_pat()` can be left immediately when a match is found, skipping the usual unwinding of the stack frames.

This should result in some slight performance improvements for ECMAScript regexes, as the matcher will do strictly less work. At least on my machine, though, any actual performance difference vanishes in the noise when I run the regex_search benchmark. That's why I don't provide any benchmark results.

But the main point of this change is that it is a prerequisite for further changes that will speed up matching by reducing the number of allocations the matcher performs.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -1764,7 +1764,6 @@ public:
         _Matched = false;
 
         bool _Succeeded = _Match_pat(_Rep) || _Matched;
-        _STL_INTERNAL_CHECK(_Frames_count == 0);
 
         if (!_Succeeded) {
             return false;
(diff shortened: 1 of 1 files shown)


StephanTLavavej:

StephanTLavavej:
Why go to all the effort of stack unwinding when we can flop over like a kitty and look adorable? :cat2: :black_cat: :heart_eyes_cat:

Issue #5830, opened 2025-11-05
`<stacktrace>`: ARM64EC isn't working properly


StephanTLavavej:
While working on enabling ARM64EC runtime test coverage (on an ARM64 host), I'm observing that `<stacktrace>` doesn't work.

For example, we expect `"P0881R7_stacktrace!all_innermost"` here:

https://github.com/microsoft/STL/blob/defc75b4758bd03e075dc9d0de093572dc57c160/tests/std/tests/P0881R7_stacktrace/test.cpp#L193

This works for x64, x86, and ARM64. But for ARM64EC, we get `"P0881R7_stacktrace"`. The content of `all.at(0).description()` is `"P0881R7_stacktrace+0xB0E0"` and the `"+0xB0E0"` is being trimmed off:

https://github.com/microsoft/STL/blob/defc75b4758bd03e075dc9d0de093572dc57c160/tests/std/tests/P0881R7_stacktrace/test.cpp#L103-L108

Similarly, in the header units and modules test,

https://github.com/microsoft/STL/blob/defc75b4758bd03e075dc9d0de093572dc57c160/tests/std/include/test_header_units_and_modules.hpp#L851-L861

`desc` is `"P2465R3_standard_library_modules+0x95254"` before trimming and `"P2465R3_standard_library_modules"` after.

My understanding of ARM64EC is extremely limited, so I don't know if this is an issue in `dbgeng.dll`, or the debug info that the compiler is emitting, or what.

Issue #5685, opened 2025-08-19
ARM64/ARM64EC runtime assertions in `P0811R3_midpoint_lerp`


StephanTLavavej:
With the latest available Windows 11 SDK (10.0.26100.4188), `P0811R3_midpoint_lerp` fails `assert(check_feexcept(0))` at runtime on ARM64:

https://github.com/microsoft/STL/blob/a2686a066284a6475502730bdba87329bc7e167e/tests/std/tests/P0811R3_midpoint_lerp/test.cpp#L1048-L1053

<details><summary>Click to expand test logs for the 2 failing configurations:</summary>

```
C:\GitHub\STL\out\ARM64>python %STL%\tools\scripts\print_failures.py testing_ARM64.log
code: FAIL
name: std :: tests/P0811R3_midpoint_lerp:34
output: Build setup steps:
Build steps:
[...]

PR #5831, opened 2025-11-06
Add ARM64EC runtime test coverage

StephanTLavavej:
Followup to #5815.

* Runtime test coverage for ARM64EC.
  + Like the ARM64 runtime test coverage, I'm using 10 shards, which I've discovered is our current maximum (I am unsure how to increase it further).
* Compile with `-fuse-ld=link`, link with `/machine:arm64ec`.
  + Filed for followup: #5825
* Add ARM64EC workarounds to `P0811R3_midpoint_lerp`.
  + Updated for followup: #5685
* Fix `Dev09_158457_tr1_mem_fn_calling_conventions` for ARM64EC.
  + The expected output didn't consider ARM64EC, although the product code and the test code correctly avoided emitting `__vectorcall` for ARM64EC: https://github.com/microsoft/STL/blob/defc75b4758bd03e075dc9d0de093572dc57c160/stl/inc/type_traits#L396-L397 https://github.com/microsoft/STL/blob/defc75b4758bd03e075dc9d0de093572dc57c160/tests/std/tests/Dev09_158457_tr1_mem_fn_calling_conventions/test.cpp#L67-L68 https://github.com/microsoft/STL/blob/defc75b4758bd03e075dc9d0de093572dc57c160/tests/std/tests/Dev09_158457_tr1_mem_fn_calling_conventions/test.cpp#L118-L119
* Work around `<stacktrace>` bug.
  + Filed for followup: #5830
* Simplify `P0881R7_stacktrace` to a single thread.
  + I spent a lot of time being confused why the output was duplicated and sometimes interleaved. The internal implementation of `<stacktrace>` takes a lock, which is why this test was spinning up a second thread, but I think this isn't worth it. If we really wanted to stress the lock, it would have to look very different, and we haven't had issues with this (nor is the code rapidly churning). With @AlexGuteniev's blessing I'm changing this to be single-threaded.
* Add `/OPT:REF,NOICF` to `universal_prefix.lst`, fixing `P0896R4_P1614R2_comparisons`.
  + After `printf`-instrumenting this test, I found that the issue was attempting to verify that `&f1` and `&f2` (the addresses of empty functions) were non-equal, and these functions were getting ICFed for ARM64EC: https://github.com/microsoft/STL/blob/defc75b4758bd03e075dc9d0de093572dc57c160/tests/std/tests/P0896R4_P1614R2_comparisons/test.cpp#L397-L398 https://github.com/microsoft/STL/blob/defc75b4758bd03e075dc9d0de093572dc57c160/tests/std/tests/P0896R4_P1614R2_comparisons/test.cpp#L479
  + This affects all tests (std/tr1/libcxx) for all architectures. I'm actually surprised we haven't encountered this before, but the [linker docs](https://learn.microsoft.com/en-us/cpp/build/reference/opt-optimizations?view=msvc-170) clearly say "By default, `/OPT:ICF` is enabled by the linker unless `/OPT:NOICF` or `/DEBUG` is specified."
  + `/OPT:REF,NOICF` is what we use for debug builds of the product itself. There's no reason to emit test executables that are more bloated than necessary (`REF` should be very cheap in terms of CPU time, and possibly even saving I/O time). We never want `ICF` for test executables, as it impacts correctness in this corner case.
* Simplify names to 'Build ARCH' and 'Test ARCH'.
  + This is a cosmetic improvement. All of the Build stages just build, and all of the Test stages actually execute tests (the fact that they also build is implied, although they don't build quite as thoroughly - no `/analyze` for the STL's own sources and no benchmarks).
* Distinguish the Cross builds.
  + For clarity, since native is the default expectation. Will be helpful if we ever encounter errors that are specific to a host arch.
* Add an early ARM64-native build.
  + This is partially to make toolset updates easier for me, but also to detect any critical problems with the ARM64 pool that would doom the large test runs. I've structured the dependencies carefully to avoid making the critical path even longer (i.e. the total latency of all checks).
* Style: Explicitly mark Code Format as having no dependencies, even though it's first.
  + This was the only order-dependent one, and I think it's simpler to be explicit.

### DIFF
--- a/tests/std/tests/P0881R7_stacktrace/test.cpp
+++ b/tests/std/tests/P0881R7_stacktrace/test.cpp
@@ -13,7 +13,6 @@
 #include <stdexcept>
 #include <string>
 #include <string_view>
-#include <thread>
 
 #ifdef HAS_EXPORT
 #define MAYBE_EXPORT __declspec(dllexport)
@@ -310,11 +309,6 @@ void test_impl() {
 
 int main() {
 #if !((defined(__clang__) && defined(_M_ARM64)) || defined(_M_ARM64EC)) // TRANSITION, LLVM-74530, GH-5830
-    // First, run the test on a single thread, so that if anything fails, the output isn't interleaved.
-    test_impl();
-
-    // Then, run the test on multiple threads, as <stacktrace> takes internal locks that should be exercised.
-    jthread t{test_impl};
     test_impl();
 #endif // ^^^ no workaround ^^^
 }


tests/std/tests/P0881R7_stacktrace/test.cpp
```diff
@@ -309,8 +309,12 @@ void test_impl() {
 }
 
 int main() {
-#if !(defined(__clang__) && defined(_M_ARM64)) // TRANSITION, LLVM-74530
+#if !((defined(__clang__) && defined(_M_ARM64)) || defined(_M_ARM64EC)) // TRANSITION, LLVM-74530, GH-5830
+    // First, run the test on a single thread, so that if anything fails, the output isn't interleaved.
+    test_impl();
```
AlexGuteniev: I understand the intention of this, but this may reduce the usefulness of multi-threaded test. By the time interleaved test run, the DLLs will be loaded, the PDBs will be loaded, and the exercise would be trivial.

AlexGuteniev: On the other hand, I won't insist on the multi-threaded test presence at all.

StephanTLavavej: I think I'd rather drop the multithreading then. I'll push a commit.

StephanTLavavej:

PR #5832, opened 2025-11-07
`<__msvc_chrono.hpp>`: Optimize 24 MHz to nanoseconds conversion in `steady_clock::now`

statementreply:
1. Reduce the fraction (using `std::ratio`).
2. Swap `_Ctr` and `period::den` when evaluating the mul-div to make better use of the constants.


AlexGuteniev:
Is compiler unable to do this on its own?

AlexGuteniev:
> Is compiler unable to do this on its own?

I guess yes https://godbolt.org/z/beqxjEces

StephanTLavavej:

StephanTLavavej:
:two: :four: :stopwatch:
===

PR #5834, opened 2025-11-08
Remove non-Standard `basic_istream::ipfx()`/`isfx()`, `basic_ostream::opfx()`/`osfx()`, and `locale::empty()`

StephanTLavavej:
Let's finish cleaning house.

* `basic_istream::ipfx()`/`isfx()` and `basic_ostream::opfx()`/`osfx()` were deprecated since VS 2022 17.9 in February 2024 (for C++17 and later) and VS 2022 17.11 in August 2024 (unconditionally).
  + #4006
  + #4605
* `locale::empty()` was deprecated since VS 2022 17.14 in May 2025.
  + #5197

I've verified that the DLL's export surface is unchanged.

There appears to have been virtually zero usage; I see no occurrences of the silencing macros in our internal or Real World Code test suites (modulo preprocessed files), which is why I'm being a bit more aggressive here.

Although we shouldn't ever need to change them, I'm keeping the comments on the function definitions. However, the comment on the declaration of `locale::empty()` serves no purpose, so I'm removing it.

### DIFF
--- a/stl/inc/__msvc_ostream.hpp
+++ b/stl/inc/__msvc_ostream.hpp
@@ -130,12 +130,9 @@ public:
         bool _Ok; // true if stream state okay at construction
     };
 
-#pragma push_macro("opfx")
-#pragma push_macro("osfx")
-#undef opfx
-#undef osfx
+#ifdef _CRTBLD
     // TRANSITION, ABI: non-Standard opfx() is preserved for binary compatibility
-    _DEPRECATE_IO_PFX_SFX bool __CLR_OR_THIS_CALL opfx() { // test stream state and flush tie stream as needed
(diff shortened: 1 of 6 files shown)


zacklj89 (approved):

StephanTLavavej:

BillyONeal:
> There appears to have been virtually zero usage; I see no occurrences of the silencing macros in our internal or Real World Code test suites (modulo preprocessed files), which is why I'm being a bit more aggressive here.

I found one! :(

https://github.com/accellera-official/systemc/blob/3d500fe7b1d3b86bb6f016ba18fb20bd856fb504/src/sysc/datatypes/int/sc_int64_io.cpp#L154

PR #5836, opened 2025-11-10
The Astonishing xmath Overhaul

StephanTLavavej:
# Background

While thinking about #2108 and measuring the STL's size, which has increased as we've added new features (in satellite DLLs these days) and decreased as we've eliminated support for XP/Vista/7/8/8.1, I realized that I could mitigate the size increase of compiling with [`/O2`][] only (without [`/Os`][]) if I could find more code to chop out of the STL.

[`/O2`]: https://learn.microsoft.com/en-us/cpp/build/reference/o1-o2-minimize-size-maximize-speed?view=msvc-170
[`/Os`]: https://learn.microsoft.com/en-us/cpp/build/reference/os-ot-favor-small-code-favor-fast-code?view=msvc-170

I was able to replace the `_Stodx` family of retained-for-bincompat functions with lightweight wrappers around the `strtod` family, closely resembling the modern `_Stodx_v3` that we use in production today, eliminating a ton of old Dinkumware math machinery.

And because our `double` and `long double` have the same 64-bit representation, I simply unified our remaining functions for `long double` (e.g. for sine and cosine) to call the `double` implementations with appropriate casting.

Finally, I performed some less-consequential cleanups to make the remaining code quicker to build and easier to understand. Eliminating all of this code also reduces our static analysis burden in the future (as tools like CodeQL like to complain about various code patterns).

Ultimately, this reduces the STL's size by about 3%. It won't completely reverse the size increase from dropping `/Os`, but it will mitigate it. Here's the size of `msvcp140_oss.dll` in absolute bytes and relative terms:

| Bytes |    main |   xmath |     /O2 |    Both |
|------:|--------:|--------:|--------:|--------:|
|   x64 | 527,872 | 511,488 | 659,456 | 632,832 |
|   x86 | 424,448 | 409,600 | 609,280 | 588,800 |
| arm64 | 635,904 | 618,496 | 778,752 | 751,104 |

| Relative | main | xmath |  /O2 | Both |
|---------:|-----:|------:|-----:|-----:|
|      x64 | 1.00 |  0.97 | 1.25 | 1.20 |
|      x86 | 1.00 |  0.97 | 1.44 | 1.39 |
|    arm64 | 1.00 |  0.97 | 1.22 | 1.18 |

This also speeds up the STL's build. On my 5950X, I measured a decrease from 30.6 seconds to 27.5 seconds, for a speedup of 1.11x.

# Bincompat

I verified that this doesn't change the export surface of the DLL.

I'm removing separately compiled functions that were declared in `xmath.hpp` for use between different STL TUs, but none of them were marked with the all-important `_CRTIMP2_PURE` that indicates being dllexported.

Because there is Only One STL (i.e. one DLL, or one static LIB) involved, we never have to worry about mismatch between STL TUs. The only mismatch we need to worry about is between headers and separately compiled sources, but I'm not changing any headers or interfaces here. (And because this code is very old, it wasn't being injected into the import LIB; headers looking for import LIB functions is the one case where we can't remove stuff that wasn't dllexported.)

There is the potential for a slight runtime behavioral difference for VS 2015 RTM/Update 1 compiled code, discussed immediately below. I believe this is an acceptable risk.

# The `_Stodx` Overhaul

* Overhaul the old `_Stodx` family.

`_Stodx` was superseded by `_Stodx_v2` in "The Iostreams Floating-Point Correctness/Performance Fix To End All Fixes, WCFB01 Edition", [internal TFVC changeset 1572756][], merged 2016-02-04 for VS 2015 Update 2. This fixed a huge number of correctness issues, including round-trip failures, and made `stream >> dbl` 19.5x faster and `stream >> flt` 17.0x faster.

[internal TFVC changeset 1572756]: https://vstfdevdiv.corp.microsoft.com/DevDiv2/DevDiv/_versionControl/changeset/1572756

At the time, I was aggressive enough to release this behavioral change in an update, and was vindicated with no user complaints. However, I was very unfamiliar with floating-point, iostreams, and making significant changes to Dinkumware's legacy code. So I added header-only `_Stodx_v2` instead of changing separately compiled `_Stodx`.

Today, I still don't understand iostreams very well, but I am much more familiar with floating-point, and we have a decade of ABI-compat experience, including significant changes to Dinkumware legacy code (e.g. #5473).

The interface of the `_Stodx` family is simple: it primarily returns a floating-point value, reads a string, potentially updates an end pointer, takes a power of 10, and writes an error code as an out-param. The `_Stodx_v2` approach can be used as a drop-in replacement.

Currently, I expect that the amount of VS 2015 RTM/Update 1 separately compiled code is extremely low. Anything compiled with VS 2015 Update 2 or the final Update 3 will be completely unaffected by this. For bincompat with RTM/Update 1, if any such user binaries exist, this may be a minor behavioral change for them (when running against an updated VCRedist), as in we may return floating-point values that are different by an ULP or so. It won't totally backport the correctness fixes (which also involved iostreams preparing larger buffers for the conversions). In fact, because of the limited input that was sent to `_Stodx`, there potential for behavioral difference is further reduced. Due to the simplicity of the `_Stodx` interface, I am confident that we can replace its guts without breaking old user binaries. Unlike the more complicated iostreams code that calls this, `_Stodx` itself doesn't have unusual preconditions or postconditions.

Note that #3364 on 2023-05-04 (shipped in VS 2022 17.7) further replaced `_Stodx_v2` with `_Stodx_v3`, dropping the bogusly-designed "power of 10" logic. For this PR, I used the older `_Stodx_v2` approach to match the oldest `_Stodx` interface. (I also kept `_Stodx`'s use of `long` for the power of 10.)

I further suspect that the wide `_WStodx` family was dead even in VS 2015 RTM (I performed even older refactorings), but haven't done the archaeology to confirm that. Since it's very easy to just replace `strtod` with `wcstod` etc., I've chosen to do that.

* Remove dead functions, macros, and declarations.
* Delete dead files.
* Remove deleted files from the build system.
* Fuse `_Stoldx`/`_WStoldx` after `_Stodx`/`_WStodx`.
* Remove fused files from the build system.

# The `long double` Overhaul

* Fuse `long double` functions with their `double` counterparts.
  + They have the same representation for us. The STL is separately compiled so we don't need to worry about 80-bit here.
* Remove fused files from the build system.

# More Cleanups

* Fuse functions into the files that use them, in unnamed namespaces.
* Remove fused files from the build system.
* Move constants to be function-local.
  + Upgrade them to `constexpr`, but only the arrays should be `static constexpr`.
* Fuse `HUGE_EXP`/`FHUGE_EXP` into their only uses.
* Replace epsilons previously defined as unions with hexfloat constants instead.
  + See `meow.cpp` below for an equivalence proof.
* Replace usage of infinity unions with `numeric_limits` instead.
  + See `purr.cpp` below for an equivalence proof.
* Add missing `_EXTERN_C_UNLESS_PURE`.
  + This was present in xfvalues.cpp but not the others. It didn't affect anything (i.e. these "constants" were still exported as `extern "C"` unless pure) because xmath.hpp declared them within `_EXTERN_C_UNLESS_PURE`. However, in general we try to keep definitions matching declarations, which avoids unintentional changes during refactoring, and makes it easier to see the nature of a thing.
* Mark the unions as preserved for bincompat.
  + Move `_FXbig` and `_Xbig` up, because they're still used.
* Drop union declarations, fuse `_CRTIMP2_PURE` into definitions.
  + Drop the `/* const */` comments - I added those long ago when I thought we still needed these unions and just needed to make them const in vNext.
* Centralize and demacroize retained-for-bincompat unions in xvalues.cpp.
  + This avoids the need to undef macros between the `double` and `float` unions, and the bit patterns are much clearer with direct hexadecimal constants anyways.
  + I'm dropping the "TRANSITION, ABI: Twice as large as necessary." comment on `unsigned short _Word[8];`. It's still true, but the whole union is retained for bincompat, so it'll simply be removed completely instead of right-sized.
  + See `woof.cpp` below for an equivalence proof.
* Remove fused files from the build system.

# Equivalence Proofs

<details><summary>Click to expand pet noises:</summary>

```
C:\Temp>type meow.cpp
```
```cpp
#include <cassert>
#include <print>
using namespace std;

union _Dconst {
    unsigned short _Word[8];
    float _Float;
    double _Double;
    long double _Long_double;
};

#define NBITS    (48 + _DOFF)
#define INIT(w0) {0, 0, 0, w0}
extern const _Dconst _Eps   = {INIT((_DBIAS - NBITS - 1) << _DOFF)};
extern const _Dconst _Rteps = {INIT((_DBIAS - NBITS / 2) << _DOFF)};
#undef NBITS
#undef INIT

#define NBITS    (16 + _FOFF)
#define INIT(w0) {0, w0}
extern const _Dconst _FEps   = {INIT((_FBIAS - NBITS - 1) << _FOFF)};
extern const _Dconst _FRteps = {INIT((_FBIAS - NBITS / 2) << _FOFF)};
#undef NBITS
#undef INIT

int main() {
    assert(_Eps._Double == 0x1p-54);
    assert(_Rteps._Double == 0x1p-27);
    assert(_FEps._Float == 0x1p-25f);
    assert(_FRteps._Float == 0x1p-12f);

    println("  _Eps._Double: {0:a} {0}", _Eps._Double);
    println("_Rteps._Double: {0:a} {0}", _Rteps._Double);
    println("  _FEps._Float: {0:a} {0}", _FEps._Float);
    println("_FRteps._Float: {0:a} {0}", _FRteps._Float);
}
```
```
C:\Temp>cl /EHsc /nologo /W4 /std:c++latest /MTd /Od meow.cpp
meow.cpp

C:\Temp>meow
_Eps._Double: 1p-54 5.551115123125783e-17
_Rteps._Double: 1p-27 7.450580596923828e-09
_FEps._Float: 1p-25 2.9802322e-08
_FRteps._Float: 1p-12 0.00024414062
```

```
C:\Temp>type purr.cpp
```
```cpp
#include <cassert>
#include <limits>
#include <print>
using namespace std;

union _Dconst {
    unsigned short _Word[8];
    float _Float;
    double _Double;
    long double _Long_double;
};

#define INIT(w0) {0, 0, 0, w0}
extern /* const */ _Dconst _Inf = {INIT(_DMAX << _DOFF)};
#undef INIT

#define INIT(w0) {0, w0}
extern /* const */ _Dconst _FInf = {INIT(_FMAX << _FOFF)};
#undef INIT

int main() {
    assert(_Inf._Double == numeric_limits<double>::infinity());
    assert(_FInf._Float == numeric_limits<float>::infinity());

    println("_Inf._Double: {0:a} {0}", _Inf._Double);
    println("_FInf._Float: {0:a} {0}", _FInf._Float);
}
```
```
C:\Temp>cl /EHsc /nologo /W4 /std:c++latest /MTd /Od purr.cpp
purr.cpp

C:\Temp>purr
_Inf._Double: inf inf
_FInf._Float: inf inf
```

```
C:\Temp>type woof.cpp
```
```cpp
#include <print>
#include <string_view>
using namespace std;

union _Dconst {
    unsigned short _Word[8];
    float _Float;
    double _Double;
    long double _Long_double;
};

#define INIT(w0)      {0, 0, 0, w0}
#define INIT2(w0, w1) {w1, 0, 0, w0}
const _Dconst _Denorm = {INIT2(0, 1)};
const _Dconst _Inf    = {INIT(_DMAX << _DOFF)};
const _Dconst _Nan    = {INIT((_DMAX << _DOFF) | (1 << (_DOFF - 1)))};
const _Dconst _Snan   = {INIT2(_DMAX << _DOFF, 1)};
#undef INIT
#undef INIT2

#define INIT(w0)      {0, w0}
#define INIT2(w0, w1) {w1, w0}
const _Dconst _FDenorm = {INIT2(0, 1)};
const _Dconst _FInf    = {INIT(_FMAX << _FOFF)};
const _Dconst _FNan    = {INIT((_FMAX << _FOFF) | (1 << (_FOFF - 1)))};
const _Dconst _FSnan   = {INIT2(_FMAX << _FOFF, 1)};
#undef INIT
#undef INIT2

int main() {
    println("_Denorm: {::x}", _Denorm._Word);
    println("_Inf: {::x}", _Inf._Word);
    println("_Nan: {::x}", _Nan._Word);
    println("_Snan: {::x}", _Snan._Word);
    println("_FDenorm: {::x}", _FDenorm._Word);
    println("_FInf: {::x}", _FInf._Word);
    println("_FNan: {::x}", _FNan._Word);
    println("_FSnan: {::x}", _FSnan._Word);
}
```
```
C:\Temp>cl /EHsc /nologo /W4 /std:c++latest /MTd /Od woof.cpp
woof.cpp

C:\Temp>woof
_Denorm: [1, 0, 0, 0, 0, 0, 0, 0]
_Inf: [0, 0, 0, 7ff0, 0, 0, 0, 0]
_Nan: [0, 0, 0, 7ff8, 0, 0, 0, 0]
_Snan: [1, 0, 0, 7ff0, 0, 0, 0, 0]
_FDenorm: [1, 0, 0, 0, 0, 0, 0, 0]
_FInf: [0, 7f80, 0, 0, 0, 0, 0, 0]
_FNan: [0, 7fc0, 0, 0, 0, 0, 0, 0]
_FSnan: [1, 7f80, 0, 0, 0, 0, 0, 0]
```
</details>

### DIFF
--- a/stl/CMakeLists.txt
+++ b/stl/CMakeLists.txt
@@ -291,12 +291,10 @@ set(SOURCES
     ${CMAKE_CURRENT_LIST_DIR}/src/xferaise.cpp
     ${CMAKE_CURRENT_LIST_DIR}/src/xfexp.cpp
     ${CMAKE_CURRENT_LIST_DIR}/src/xfsinh.cpp
-    ${CMAKE_CURRENT_LIST_DIR}/src/xfvalues.cpp
     ${CMAKE_CURRENT_LIST_DIR}/src/xgetwctype.cpp
     ${CMAKE_CURRENT_LIST_DIR}/src/xlgamma.cpp
     ${CMAKE_CURRENT_LIST_DIR}/src/xlocale.cpp
     ${CMAKE_CURRENT_LIST_DIR}/src/xlock.cpp
-    ${CMAKE_CURRENT_LIST_DIR}/src/xlvalues.cpp
     ${CMAKE_CURRENT_LIST_DIR}/src/xmbtowc.cpp
     ${CMAKE_CURRENT_LIST_DIR}/src/xmtx.cpp
     ${CMAKE_CURRENT_LIST_DIR}/src/xnotify.cpp
--- a/stl/msbuild/stl_base/stl.files.settings.targets
+++ b/stl/msbuild/stl_base/stl.files.settings.targets
@@ -85,12 +85,10 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
         <ClCompile Include="$(CrtRoot)\github\stl\src\xferaise.cpp" />
         <ClCompile Include="$(CrtRoot)\github\stl\src\xfexp.cpp" />
         <ClCompile Include="$(CrtRoot)\github\stl\src\xfsinh.cpp" />
-        <ClCompile Include="$(CrtRoot)\github\stl\src\xfvalues.cpp" />
         <ClCompile Include="$(CrtRoot)\github\stl\src\xgetwctype.cpp" />
         <ClCompile Include="$(CrtRoot)\github\stl\src\xlgamma.cpp" />
         <ClCompile Include="$(CrtRoot)\github\stl\src\xlocale.cpp" />
         <ClCompile Include="$(CrtRoot)\github\stl\src\xlock.cpp" />
-        <ClCompile Include="$(CrtRoot)\github\stl\src\xlvalues.cpp" />
         <ClCompile Include="$(CrtRoot)\github\stl\src\xmbtowc.cpp" />
         <ClCompile Include="$(CrtRoot)\github\stl\src\xmtx.cpp" />
         <ClCompile Include="$(CrtRoot)\github\stl\src\xonce.cpp" />


zacklj89 (approved):

StephanTLavavej:

PR #5845, opened 2025-11-11
Enable vectorization of std::rotate on ARM64

hazzlim:
Add an implementation of _Swap_3_ranges using Neon intrinsics and enable vectorization of std::rotate on ARM64 targets.


StephanTLavavej:
Thanks! BTW, if you're having issues with code formatting, if you use VSCode and install Microsoft's C/C++ Extension, you can add `"C_Cpp.clang_format_path": "C:/Program Files/Microsoft Visual Studio/18/Insiders/VC/Tools/Llvm/x64/bin/clang-format.exe",` to your settings. Combined with how our `.vscode/settings.json` enables format-on-save for C++ files, that'll keep your code always-formatted with the exact version of clang-format that we use in the repo.

StephanTLavavej:

StephanTLavavej:
Thanks - when things spin, science happens! :arrows_clockwise: :arrows_counterclockwise: :scientist:

PR #5874, opened 2025-11-15
Make sure expected `__isa_enabled` bit was set before disabling it in feature-dependent tests

AlexGuteniev:
This makes sure that tests are running all code paths.

Currently it is not much needed, AVX2 is everywhere. 
But we may pick next x64 features someday, and also may pick up ARM64 not always available features.

We also need to be able to execute tests intentionally on downlevel host to see if we haven't used instructions that are not available. Here escape environment variable `STL_TEST_DOWNLEVEL_HOST` set to `1` will help.

### DIFF
--- a/tests/std/include/test_vector_algorithms_support.hpp
+++ b/tests/std/include/test_vector_algorithms_support.hpp
@@ -71,21 +71,26 @@ inline void disable_instructions(ISA_AVAILABILITY isa) {
 constexpr std::size_t dataCount = 1024;
 
 template <class TestFunc>
-void run_randomized_tests_with_different_isa_levels(TestFunc tests) {
-    std::mt19937_64 gen;
-    initialize_randomness(gen);
-
-    tests(gen);
+void run_tests_with_different_isa_levels(TestFunc tests) {
+    tests();
 
 #if (defined(_M_IX86) || (defined(_M_X64) && !defined(_M_ARM64EC))) && !defined(_M_CEE_PURE)
     const auto original_isa = __isa_enabled;
 
     disable_instructions(__ISA_AVAILABLE_AVX2);
-    tests(gen);
+    tests();
 
     disable_instructions(__ISA_AVAILABLE_SSE42);
-    tests(gen);
+    tests();
 
     __isa_enabled = original_isa;
 #endif // (defined(_M_IX86) || (defined(_M_X64) && !defined(_M_ARM64EC))) && !defined(_M_CEE_PURE)
+}
+
+template <class TestFunc>
+void run_randomized_tests_with_different_isa_levels(TestFunc tests) {
+    std::mt19937_64 gen;
+    initialize_randomness(gen);
+
+    run_tests_with_different_isa_levels([&] { tests(gen); });
 }
--- a/tests/std/tests/GH_002431_byte_range_find_with_unreachable_sentinel/test.cpp
+++ b/tests/std/tests/GH_002431_byte_range_find_with_unreachable_sentinel/test.cpp
@@ -4,7 +4,6 @@
 #include <algorithm>
 #include <cassert>
 #include <cstddef>
-#include <isa_availability.h>
 #include <ranges>
 
 #pragma warning(push) // TRANSITION, OS-23694920
@@ -52,13 +51,7 @@ int main() {
     void* p2 = VirtualAlloc(p, page, MEM_COMMIT, PAGE_READWRITE);
     assert(p2 != nullptr);
 
-    test_all_element_sizes(p, page);
-#if defined(_M_IX86) || (defined(_M_X64) && !defined(_M_ARM64EC))
-    disable_instructions(__ISA_AVAILABLE_AVX2);
-    test_all_element_sizes(p, page);
-    disable_instructions(__ISA_AVAILABLE_SSE42);
-    test_all_element_sizes(p, page);
-#endif // defined(_M_IX86) || (defined(_M_X64) && !defined(_M_ARM64EC))
+    run_tests_with_different_isa_levels([&] { test_all_element_sizes(p, page); });
 
     VirtualFree(p, 0, MEM_RELEASE);
 }
--- a/tests/std/tests/GH_003617_vectorized_meow_element/test.cpp
+++ b/tests/std/tests/GH_003617_vectorized_meow_element/test.cpp
@@ -6,7 +6,6 @@
 #ifdef _M_X64
 
 #include <cstddef>
-#include <isa_availability.h>
 #include <vector>
 
 #include "test_min_max_element_support.hpp"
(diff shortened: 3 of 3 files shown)


tests/std/include/test_vector_algorithms_support.hpp
```diff
@@ -41,7 +42,32 @@ inline void initialize_randomness(std::mt19937_64& gen) {
...
+#ifdef __clang__
+#pragma clang diagnostic pop
+#else // ^^^ defined(__clang__) / !defined(__clang__) vvv
+#pragma warning(pop)
+#endif // ^^^ !defined(__clang__) ^^^
+
+    if (env_val == nullptr || std::atoi(env_val) == 0) {
+        if ((__isa_enabled & (1UL << as_ulong)) == 0) {
+            std::printf("The feature %lu is not available, the test does not have full coverage!\n", as_ulong);
+            std::printf("You can set environment variable STL_TEST_DOWNLEVEL_HOST to 1,\n"
```
StephanTLavavej: I'm not a fan of this. We assume STL development (and CI machines) are using modern CPUs. I would rather emit an *error*.

AlexGuteniev: I think we need this.
1. We may intentionally find a potato machine to test downlevel codepaths so that it does not use unavailable instruction.
2. AVX-512 is controversial. When we will use it, we have to use proper CI machines, but as Intel having problems bringing AVX-512 to developer-grade processors, we may want to let some developers run tests without AVX-512.

AlexGuteniev: And by default, when `STL_TEST_DOWNLEVEL_HOST` is not set, we'll just abort.

StephanTLavavej: Ah, I think I missed the abort below.

tests/std/tests/GH_002431_byte_range_find_with_unreachable_sentinel/test.cpp
```diff
@@ -59,7 +53,7 @@ int main() {
     assert(p2 != nullptr);
 
     test_all_element_sizes(p, page);
-#if defined(_M_IX86) || defined(_M_X64)
+#if defined(_M_IX86) || defined(_M_X64) && !defined(_M_ARM64EC)
```
StephanTLavavej: Let's use parens to avoid compiler warnings.

tests/std/tests/GH_002431_byte_range_find_with_unreachable_sentinel/test.cpp
StephanTLavavej: Mismatched endif comment.

tests/std/include/test_vector_algorithms_support.hpp
```diff
@@ -54,7 +80,7 @@ void run_randomized_tests_with_different_isa_levels(TestFunc tests) {
 
     tests(gen);
 
-#if (defined(_M_IX86) || defined(_M_X64)) && !defined(_M_CEE_PURE)
+#if (defined(_M_IX86) || defined(_M_X64) && !defined(_M_ARM64EC)) && !defined(_M_CEE_PURE)
```
StephanTLavavej: We should also use parens here.

tests/std/include/test_vector_algorithms_support.hpp
StephanTLavavej: Also need to update comment here.

StephanTLavavej: This hasn't been an issue before, but now that we're exercising it more widely, it seems like a pitfall waiting to happen - we should restore the ISA level in case tests run anything after this.

tests/std/include/test_vector_algorithms_support.hpp
```diff
@@ -41,7 +42,32 @@ inline void initialize_randomness(std::mt19937_64& gen) {
...
+    const unsigned long as_ulong = static_cast<unsigned long>(isa);
+
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#else // ^^^ defined(__clang__) / !defined(__clang__) vvv
+#pragma warning(push)
+#pragma warning(disable : 4996)
+#endif // ^^^ !defined(__clang__) ^^^
+    const char* const env_val = std::getenv("STL_TEST_DOWNLEVEL_HOST");
```
StephanTLavavej: Let's just use `getenv_s` to avoid warning wrath.

StephanTLavavej: "host" is kind of confusing, because tests are run on the target machine. Let's just say "machine".

tests/std/include/test_vector_algorithms_support.hpp
```diff
@@ -41,7 +42,32 @@ inline void initialize_randomness(std::mt19937_64& gen) {
...
+#pragma clang diagnostic pop
+#else // ^^^ defined(__clang__) / !defined(__clang__) vvv
+#pragma warning(pop)
+#endif // ^^^ !defined(__clang__) ^^^
+
+    if (env_val == nullptr || std::atoi(env_val) == 0) {
+        if ((__isa_enabled & (1UL << as_ulong)) == 0) {
+            std::printf("The feature %lu is not available, the test does not have full coverage!\n", as_ulong);
+            std::printf("You can set environment variable STL_TEST_DOWNLEVEL_HOST to 1,\n"
+                        "if you intentionally test on a host with not all features available.");
```
StephanTLavavej: Missing newline.

tests/std/include/test_vector_algorithms_support.hpp
StephanTLavavej: These could really be a single printf. It's strange to use two printfs for three lines.

tests/std/include/test_vector_algorithms_support.hpp
```diff
@@ -41,7 +42,32 @@ inline void initialize_randomness(std::mt19937_64& gen) {
...
+            abort();
```
StephanTLavavej: This will look weird if it fails, although we did printf. I think I'd rather assert, which is more common for tests.

tests/std/tests/GH_003617_vectorized_meow_element/test.cpp
```diff
@@ -33,11 +28,13 @@ void test_gh_3617() {
 int main() {
     test_gh_3617();
 
+#ifndef _M_ARM64EC
     disable_instructions(__ISA_AVAILABLE_AVX2);
     test_gh_3617();
 
     disable_instructions(__ISA_AVAILABLE_SSE42);
     test_gh_3617();
```
StephanTLavavej: In fact, we could further centralize this with a helper that doesn't randomize tests.

StephanTLavavej:
Thanks! :smiley_cat: I pushed significant changes after being persuaded of this approach, please double-check.

StephanTLavavej:

AlexGuteniev:
All good

StephanTLavavej:
Thanks for ensuring that we aren't silently missing test coverage! :ghost: :detective: :mag_right:

PR #5878, opened 2025-11-17
Add stl-lit flag to show test run time histogram

AlexGuteniev:


AlexGuteniev:
@StephanTLavavej observed that `Dev11_1158803_regex_thread_safety` tests regex ref counting, and it should be removed, as there's no need to keep testing it.

StephanTLavavej:

StephanTLavavej:
:bar_chart: :stopwatch: :1st_place_medal:
===

PR #5905, opened 2025-11-24
`<chrono>`: Replace an elaborate table with a `switch`

StephanTLavavej:
`<chrono>` formatting validated modifier characters by searching an elaborate table containing bitflags of allowed modifiers. It tried to use `ranges::find` with a PMD projection, but we needed a product code workaround for VSO-1664341 "`/clr` C++20 System.NullReferenceException when calling ranges algorithms with PMD projections".

I initially tried to replace the `ranges::find` with a range-for loop as a perma-workaround, which was an improvement, but then I went further. We can replace all of this machinery with a `switch`.

By changing `void _Check_modifier` to `bool _Valid_modifier`, we can further simplify the control flow by immediately returning `true` or `false`, and having the caller take care of throwing any exception.

I'm adding `_STL_INTERNAL_CHECK(_Modifier == 'E' || _Modifier == 'O');` for a previously implicit assumption.

Finally, I'm adding a comment to cite the relevant Standardese for where this `switch`'s data comes from.

### DIFF
--- a/stl/inc/chrono
+++ b/stl/inc/chrono
@@ -5353,61 +5353,52 @@ namespace chrono {
                 if (_Spec._Type != '\0' && !_Is_valid_type<_Ty>(_Spec._Type)) {
                     _STD _Throw_format_error("Invalid type.");
                 }
-                _Check_modifier(_Spec._Type, _Spec._Modifier);
+
+                if (!_Valid_modifier(_Spec._Type, _Spec._Modifier)) {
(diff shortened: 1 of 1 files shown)


AlexGuteniev:
I think the jump table emitted by that `switch` would perform better than linear array lookup here. And maybe even better than the bitset emitted if it was written with `if`s.

zacklj89 (approved):

StephanTLavavej:

PR #5917, opened 2025-11-29
`<functional>`: unwrap `move_only_function` with abominable and noexcept types

AlexGuteniev:
Towards #5504.
Follow up on #5808.

# 📦 Newly unwrapped cases

 * `move_only_function<R(Args...) cv1-ref1-noex1>(move_only_function<R(Args...) cv2-ref2-noex2>(callable));` where `noexcept` and `const` may be present in only inner, and `&` or `&&` may be present in only outer
 * `move_only_function<R(Args...) cv-ref>(function<R(Args...)>(callable))` where cv-ref can have `const`, `&`, and `&&`.

In the tests make sure the type system is not broken, and the cases which are not supposed to compile, still do not compile.

# 🐿️ Squirrelly case

Inner `noexcept`, outer non-`noexcept` case is squirrelly.

To make sure we adhere to Strict Aliasing rule, I've made `_Function_base` not dependent on whether we have `noexcept` signature or not. so that noexcept base can be wrapped in non-noexcept base safely.

### DIFF
--- a/stl/inc/functional
+++ b/stl/inc/functional
@@ -1760,7 +1760,11 @@ public:
     template <bool _Noexcept>
     _NODISCARD auto _Get_invoke() const noexcept {
         if constexpr (_Noexcept) {
+#ifdef __cpp_noexcept_function_type
             return reinterpret_cast<_Rx(__stdcall*)(void*, _Types&&...) noexcept>(_Get_impl(_Data)->_Invoke);
+#else
+            static_assert(false); // when noexcept isn't in the type system, this should never be selected
+#endif
         } else {
             return _Get_impl(_Data)->_Invoke;
         }


stl/inc/functional
```diff
@@ -2060,6 +2061,19 @@ private:
         is_constructible_v<decay_t<_Fn>, initializer_list<_Ux>&, _CTypes...>
         && _Call::template _Is_callable_from<decay_t<_Fn>>;
 
+    template <class...>
+    friend class move_only_function;
+
+    template <class _Vt>
+    static constexpr bool _Is_move_only_fuction_varying_cv_ref_noex() {
```
StephanTLavavej: Typo.

tools/scripts/move_only_function_specializations.py
```diff
@@ -6,16 +6,18 @@
...
     using _VtInvQuals = {cv} _Vt {ref_inv};
 
     template <class _Vt>
     static constexpr bool _Is_callable_from = {callable};
 
+    static constexpr bool _Noexcept = {noex_val};
+
     _Rx operator()(_Types... _Args) {cv} {ref} {noex} {{
-        return this->_Get_invoke()({self}, _STD forward<_Types>(_Args)...);
+        return this->template _Get_invoke<{noex_val}>()({self}, _STD forward<_Types>(_Args)...);
```
StephanTLavavej: Although the generator makes this centralized, I believe it would be clearer to use the named `_Noexcept` here, instead of the literal `true` or `false` again.

stl/inc/functional
```diff
@@ -1105,9 +1105,8 @@ protected:
 
 private:
 #if _HAS_CXX23
-    friend _Function_base<_Ret, false, _Types...>;
+    friend _Function_base<_Ret, _Types...>;
 #endif // _HAS_CXX23
-
```
StephanTLavavej, line 1110: Newline being removed here.

stl/inc/functional
```diff
@@ -1756,8 +1756,13 @@ public:
                                                  || sizeof(_Vt) > _Function_data::_Buf_size<_Vt>
                                                  || !is_nothrow_move_constructible_v<_Vt>;
 
+    template <bool _Noexcept>
     _NODISCARD auto _Get_invoke() const noexcept {
-        return _Get_impl(_Data)->_Invoke;
+        if constexpr (_Noexcept) {
+            return reinterpret_cast<_Rx(__stdcall*)(void*, _Types&&...) noexcept>(_Get_impl(_Data)->_Invoke);
```
StephanTLavavej, line 1764: It's weird to emit this code when `noexcept` isn't in the type system, even if the compiler happens to tolerate the `noexcept` function type. I think it would be better to preprocessor-guard this with a `static_assert(false)`.

StephanTLavavej:

StephanTLavavej:
:package: :gift: :gift_heart:
===

PR #5919, opened 2025-11-30
Fix call unwrapping optimization

AlexGuteniev:
Fix regression in #5808.

We can't move out lvalue reference.

The optimization for lvalue references is still possible, but we have to handle this case slightly differently.
Now, if the function is passed as lvalue, we copy from the passed function. instead of moving it out.

----

Other cases of lvalue reference in call unwrapping:
 * Inner `move_only_function`: [`[func.wrap.move.ctor]/6`](https://eel.is/c++draft/func.wrap.move.ctor#6) and the definition of `move_only_function` enforces that attempt to use lvalue reference here won't compile. So no handling is needed.
 * Inner `copyable_function`: will need to handle lvalue reference  with this approach applied in one more place, in addition to these three places
 * Inner `function_ref`. There can be lvalue references. But I don't know if we want to nullify moved-out `function_ref` or not. More natural would be to leave moved out `function_ref` intact, so we will always do as if we have lvalue reference.

### DIFF
--- a/stl/inc/functional
+++ b/stl/inc/functional
@@ -1653,8 +1653,7 @@ public:
                     }
                 };
 
-                _Guard_type _Guard;
-                _Guard._Ptr = _Where;
+                _Guard_type _Guard{_Where};
                 _Old_fn_impl->_Copy(_Where);
                 _Guard._Ptr = nullptr;
             } else {


stl/inc/functional
```diff
@@ -1629,24 +1629,51 @@ public:
...
+                        // _Ptr is not nullptr only if an exception is thrown as a result of _Copy.
+                        // Check _Ptr before calling operator delete to save the call in the common case.
+                        if (_Ptr) {
+                            ::operator delete(_Ptr);
+                        }
+                    }
+                };
+
+                _Guard_type _Guard;
+                _Guard._Ptr = _Where;
```
StephanTLavavej: We can use aggregate init.

StephanTLavavej:

StephanTLavavej:
Thanks for noticing and fixing this! :house: :boom: :derelict_house: :hammer_and_wrench: :house_with_garden:

PR #5921, opened 2025-12-01
Clang support for LWG-3617 `function`/`packaged_task` deduction guides and deducing `this`

StephanTLavavej:
We merged #2966 on 2022-08-05, shipped in VS 2022 17.4, for MSVC and EDG. At that time, we were requiring Clang 14, which didn't support "deducing `this`".

In VS 2022 17.14, we began requiring Clang 19, which added support for "deducing `this`". I didn't notice this in #5247 merged on 2025-01-24. Nobody else noticed either, but it's my job! :joy_cat:

We should activate the product and test code for this LWG issue resolution for Clang now.


StephanTLavavej:

PR #5922, opened 2025-12-01
Simplify away `_STL_WIN32_WINNT` machinery for bincompat functions

StephanTLavavej:
Overview
===
We used to have a fairly elaborate scheme in the STL where we had the notion of a minimum Windows version being targeted, which would be determined by various platform macros. If the minimum Windows version was sufficiently new, we could unconditionally call various Windows APIs, otherwise we needed runtime detection. This was centralized in our macro `_STL_WIN32_WINNT`.

After a long series of PRs raising our minimum Windows version to Win10 and dropping code for XP/Vista/7/8/8.1, this scheme has become vestigial. The only thing it actually controls is whether the STL's DLL and static LIB provide wrappers retained for bincompat. It's simpler to directly express the conditions where we need to provide those wrappers. (The wrappers originally handled runtime detection of newer Windows APIs. Our modern code directly calls the Windows APIs so it doesn't use the wrappers at all. We have to retain them for bincompat in case old user binaries are looking for them, but we've dropped the runtime detection and they just directly call what they wrap.)

Commits
===
* `CMakeLists.txt`: We don't need to define `_VCRT_WIN32_WINNT`.
  + Internally, only `src/vctools/crt/vcruntime/src/internal/winapi_downlevel.cpp` and `src/vctools/crt/vcstartup/src/utility/initializers.cpp` inspect this, and they aren't headers, so they aren't picked up by the STL's build at all.
* Include `Windows.h` instead of `synchapi.h`, don't define weird platform macros in `CMakeLists.txt`.
  + We never inspect `_X86_`, `_AMD64_`, `_ARM64_`, or `_ARM64EC_` ourselves.
  + They're needed if we directly include `synchapi.h`, but not if we include `Windows.h` (which defines them). [`AcquireSRWLockExclusive`][] and [`InitOnceBeginInitialize`][] say "Header: `synchapi.h` (include `Windows.h` on Windows 7, Windows Server 2008 Windows Server 2008 R2)".
* Simplify away `_STL_WIN32_WINNT` machinery. (Need changes for the internal `src/vctools` files mentioned here.)
  + `src/vctools/crt/crt-common.settings.targets` has dead code mentioning `CrtOsVersion`. It defines `_STL_WIN32_WINNT` according to the following criteria:
    - `TargetEnclave` is never set for the STL, so we can ignore it.
    - For ARM64 (not ARM64EC, which is treated like x64), it uses `0x0A00`, the Win10 value.
    - Otherwise, for `TargetAppFamily` (indicated by `_CRT_APP` defined by `src/vctools/crt/crt_build.settings.targets`) or for `TargetCoreSystem` (indicated by `_CRT_ONECORE` defined by `src/vctools/crt/crt_build.settings.targets`, but we inspect `_ONECORE` defined by `src/vctools/VCTools.props`), it uses `0x0602`, the Win8 value.
    - Otherwise, it uses `0x0501`, the WinXP value.
  + We only ever compare `_STL_WIN32_WINNT` with " < Vista" and " < Win8". Therefore, the distinction between ARM64's Win10 value and `_CRT_APP`/`_ONECORE`'s Win8 value is not significant; both are "new", and otherwise we are "old". So, we can use `#if !(defined(_M_ARM64) || defined(_CRT_APP) || defined(_ONECORE))` to guard old bincompat code. (I think I prefer this over fully De Morganing it.)
  + Note: `CMakeLists.txt` had a simplified form (just "ARM64 is Win10") of the MSVC-internal machinery in `src/vctools/crt/crt-common.settings.targets`. We haven't ported the App/OneCore builds to GitHub, hence the difference.
  + Note: The fallback definition in `<yvals_core.h>` was irrelevant; it wasn't inspected by any of our headers (at least not now; I forget whether this was the case in the past), and for our separately compiled source files, it was defined by the build system as previously described.
  + `src/vctools/crt/vcstartup/build/msvcrt.settings.targets` defines `_STL_WIN32_WINNT` but shouldn't.
* Comment why ARM64, App, and OneCore are special.

Validation
===
Validated this by adding the following block to `winapisupp.cpp` before removing the legacy scheme; x64 and arm64 successfully built internally:

```cpp
#if defined(_M_ARM64) || defined(_CRT_APP) || defined(_ONECORE)
static_assert(!(_STL_WIN32_WINNT < _WIN32_WINNT_VISTA), "Modern scheme says we're new, but legacy scheme disagrees.");
static_assert(!(_STL_WIN32_WINNT < _WIN32_WINNT_WIN8), "Modern scheme says we're new, but legacy scheme disagrees.");
#else // ^^^ new, according to modern scheme / old, according to modern scheme vvv
static_assert(_STL_WIN32_WINNT < _WIN32_WINNT_VISTA, "Modern scheme says we're old, but legacy scheme disagrees.");
static_assert(_STL_WIN32_WINNT < _WIN32_WINNT_WIN8, "Modern scheme says we're old, but legacy scheme disagrees.");
#endif // ^^^ old, according to modern scheme ^^^
```

Verified that the dllexports are unchanged for x64 and arm64. We have internal dllexport validation for all architectures, but I like to manually check too. The files driving this are actually available on GitHub, see `stl/msbuild/stl_base` containing the files [`i386.exports`][], [`amd64.exports`][], and [`arm64.exports`][].

[`AcquireSRWLockExclusive`]: https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-acquiresrwlockexclusive
[`InitOnceBeginInitialize`]: https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-initoncebegininitialize
[`i386.exports`]: https://github.com/microsoft/STL/blob/2e27f1f05a155bb3637dad04b4d9c4c4590c8542/stl/msbuild/stl_base/i386.exports
[`amd64.exports`]: https://github.com/microsoft/STL/blob/2e27f1f05a155bb3637dad04b4d9c4c4590c8542/stl/msbuild/stl_base/amd64.exports
[`arm64.exports`]: https://github.com/microsoft/STL/blob/2e27f1f05a155bb3637dad04b4d9c4c4590c8542/stl/msbuild/stl_base/arm64.exports

### DIFF
--- a/stl/src/winapinls.cpp
+++ b/stl/src/winapinls.cpp
@@ -1,6 +1,10 @@
 // Copyright (c) Microsoft Corporation.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+// Previously, x64/x86 Desktop builds had to support targeting old versions of Windows, so we dllexported functions to
+// dynamically load new Windows APIs. Now, we directly call those Windows APIs, and the dllexported functions have
+// become simple wrappers, but they still need to be preserved for binary compatibility. The ARM64, App, and OneCore
+// builds never supported targeting old versions of Windows, so they never provided these dllexported functions.
 #if !(defined(_M_ARM64) || defined(_CRT_APP) || defined(_ONECORE))
 
 #include <yvals.h>
--- a/stl/src/winapisupp.cpp
+++ b/stl/src/winapisupp.cpp
@@ -35,6 +35,10 @@ extern "C" _CRTIMP2 BOOL __cdecl __crtIsPackagedApp() noexcept {
 
 #endif // !defined(_CRT_WINDOWS) && !defined(UNDOCKED_WINDOWS_UCRT)
 
+// Previously, x64/x86 Desktop builds had to support targeting old versions of Windows, so we dllexported functions to
+// dynamically load new Windows APIs. Now, we directly call those Windows APIs, and the dllexported functions have
+// become simple wrappers, but they still need to be preserved for binary compatibility. The ARM64, App, and OneCore
+// builds never supported targeting old versions of Windows, so they never provided these dllexported functions.
 #if !(defined(_M_ARM64) || defined(_CRT_APP) || defined(_ONECORE))
 
 // TRANSITION, ABI: preserved for binary compatibility


stl/src/winapinls.cpp
```diff
@@ -1,9 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-#include <yvals_core.h>
-
-#if _STL_WIN32_WINNT < _STL_WIN32_WINNT_VISTA
+#if !(defined(_M_ARM64) || defined(_CRT_APP) || defined(_ONECORE))
```
AlexGuteniev, line 8: Whereas I appreciate the headers and the make file cleanup, this is now more obscure.
Maybe comment? There are only two occurrences, and their number is not expected to grow.

cpplearner: Not sure what to comment other than "these functions are preserved for binary compatibility"

StephanTLavavej: Good idea, done.

StephanTLavavej: @cpplearner I tried my best to explain why ARM64, App, and OneCore are special and don't need these functions.

StephanTLavavej:

Issue #216, opened 2019-10-25
STL: When wchar_t is a real type, users shouldn't see unsigned short machinery


StephanTLavavej:
Because the STL grudgingly supports `/Zc:wchar_t-` (see #212), we occasionally need to detect `_NATIVE_WCHAR_T_DEFINED`.

For a correct (although weird) example, the STL always defines `numeric_limits<wchar_t>`:

https://github.com/microsoft/STL/blob/6b0238d7035236af0f81917262691db701e70fe3/stl/inc/limits#L460-L461

Then if `wchar_t` is a real type, it additionally defines `numeric_limits<unsigned short>`:

https://github.com/microsoft/STL/blob/6b0238d7035236af0f81917262691db701e70fe3/stl/inc/limits#L685-L688

This is backwards, but it works. Users with real `wchar_t` get both specializations, while users with fake `wchar_t` get only one.

However, the STL appears to have incorrect logic elsewhere. It follows the `numeric_limits` pattern for `char_traits`:

https://github.com/microsoft/STL/blob/6b0238d7035236af0f81917262691db701e70fe3/stl/inc/xstring#L265-L275

But when `wchar_t` is a real type, `char_traits<unsigned short>` doesn't make any sense!

`<fstream>` has many similar examples. `basic_ifstream` is always constructible from `const wchar_t*`:

https://github.com/microsoft/STL/blob/6b0238d7035236af0f81917262691db701e70fe3/stl/inc/fstream#L823-L824

And when `wchar_t` is real, it's also constructible from `const unsigned short*`:

https://github.com/microsoft/STL/blob/6b0238d7035236af0f81917262691db701e70fe3/stl/inc/fstream#L850-L852

I believe that all of our `_NATIVE_WCHAR_T_DEFINED` logic needs to be audited. When we're separately compiling the STL, `wchar_t` is always a real type, and we additionally need to provide `unsigned short` counterparts (so we should detect whether we're building the STL). But when headers are being used to compile user code, users with real `wchar_t` should see only `wchar_t` for character-ish things, never `unsigned short`. (Users with fake `wchar_t` will see only `wchar_t` being a typedef for `unsigned short`, as they should.)

In theory, this should not affect ABI compatibility, although it may affect source compatibility (if users with real `wchar_t` somehow took a dependency on the `unsigned short` machinery).

[...]

PR #5923, opened 2025-12-01
Unify locale helpers with `if constexpr`

StephanTLavavej:
Works towards #189.

Commits
===
* Drop `__CRTDECL` on internal header-only functions.
  + Replaced: `__CRTDECL (_LStr(coll|xfrm)|_Makloc(byte|chr|str)|_Getloctxt)` => `$1`
  + These functions aren't dllexported or address-taken, so their calling convention doesn't matter. (`<vcruntime.h>` defines `__CRTDECL` to be `__CLRCALL_PURE_OR_CDECL`.)
* Unify with `if constexpr`, part 1: `_LStrcoll()`
* Unify with `if constexpr`, part 2: `_LStrxfrm()`
* Unify with `if constexpr`, part 3: `_Maklocbyte`, `_Maklocchr`, `_Maklocstr`
* Fix bogus SAL annotations on `_Strxfrm()`, `_Wcsxfrm()`, and `_LStrxfrm()`, following #5444.
* Fix comment: `_XStrxfrm` (no such thing) => `_LStrxfrm`
* `_Maklocbyte`, `_Maklocchr`, `_Maklocstr`: Guard `unsigned short` with `#ifdef _CRTBLD`.
  + This follows `_LStrcoll` and `_LStrxfrm`, changed by #5361.
  + Followup to the very old #216. This leaves `codecvt<unsigned short, char, mbstate_t>` and `ctype<unsigned short>` as the only exceptions; I currently don't understand what's going on with them.
* Pass an explicit template arg to `_Maklocchr` and `_Maklocstr` instead of a typed null pointer.
  + This is fine because it can't be deduced now. These are header-only (and aren't `extern "C"`) and their signatures are changing, so we don't need to rename them.
* Unify `wchar_t` and `unsigned short`, part 1: Prepare `_Maklocstr`
* Unify `wchar_t` and `unsigned short`, part 2: Prepare `_Maklocbyte`, `_Maklocchr`
* Unify `wchar_t` and `unsigned short`, part 3: Extract `_Is_wchar_or_crtbld_ushort`.
  + `remove_cv_t<_Elem>` is currently unnecessary, but it makes me feel better to have type traits that behave like the primary type categories.
* `_Maklocstr`: Remove dead `_Count -= _Bytes`.
  + `_Count` is unused at this point (and there's no outer loop).
* `_Maklocstr`: Improve variable scoping, part 1.
  + `_Count` can be scoped to the for-loop.
  + `_Count1` can be const.
  + `_Wchars` and `_Ptr1` can be initialized when defined, instead of assigned at the beginning of the for-loop.
* `_Maklocstr`: Improve variable scoping, part 2.
  + `_Bytes` doesn't communicate information across loops or loop iterations. It's always assigned to in the loop body, then read from at the end of the iteration. Making it a const local in each loop iteration is significantly clearer.
* Guard `locale::_Locimp::_Makeushloc()` with `#ifdef _CRTBLD`.
  + Users with real `wchar_t` should never see this.

### DIFF
--- a/stl/inc/xlocale
+++ b/stl/inc/xlocale
@@ -193,10 +193,10 @@ public:
         static void __CLRCALL_OR_CDECL _Makewloc(
             const _Locinfo&, category, _Locimp*, const locale*); // make wchar_t facets
 
-#if defined(_NATIVE_WCHAR_T_DEFINED) && !_ENFORCE_FACET_SPECIALIZATIONS
+#ifdef _CRTBLD
         static void __CLRCALL_OR_CDECL _Makeushloc(
             const _Locinfo&, category, _Locimp*, const locale*); // make ushort facets
-#endif // defined(_NATIVE_WCHAR_T_DEFINED) && !_ENFORCE_FACET_SPECIALIZATIONS
+#endif // defined(_CRTBLD)
 
         static void __CLRCALL_OR_CDECL _Makexloc(
             const _Locinfo&, category, _Locimp*, const locale*); // make remaining facets


StephanTLavavej:

zacklj89 (approved):

PR #5817, opened 2025-10-30
`<iterator>`: Remove `stdext::checked_array_iterator` and `stdext::unchecked_array_iterator`

StephanTLavavej:
# Overview
* [VS 2022 17.8][] released [on 2023-11-14][17.8 rel] deprecated `stdext::checked_array_iterator` in C++17 mode.
  + #3818
* [VS 2022 17.11][] released [on 2024-08-13][17.11 rel] deprecated `stdext::checked_array_iterator` unconditionally.
  + #4605

One or two years of deprecation notice is enough. (VS 2022 17.14 and VS 2026 18.0 still support these non-Standard iterators and will be supported for a long time, giving customers plenty of notice and time to migrate.)

These non-Standard iterators make it more difficult to write portable code. They're no longer necessary to avoid obnoxious "output iterator is a raw pointer" warnings from STL algorithms (which we removed a long time ago). `gsl::span` (available downlevel to C++14) and `std::span` in C++20 are superior mechanisms, since they act as ranges instead of individual iterators.

[VS 2022 17.8]: https://github.com/microsoft/STL/releases/tag/vs-2022-17.8
[VS 2022 17.11]: https://github.com/microsoft/STL/releases/tag/vs-2022-17.11
[17.8 rel]: https://learn.microsoft.com/en-us/visualstudio/releases/2022/release-notes-v17.8
[17.11 rel]: https://learn.microsoft.com/en-us/visualstudio/releases/2022/release-notes-v17.11

# Commits
* Remove product code.
* Tombstone the deprecation warning.
* Delete entire tests.
* Remove test coverage.
* Ensure that we never mention these names again.
* Remove unnecessary test usage (originally to silence warnings).

PR #5924, opened 2025-12-01
Various cleanups

StephanTLavavej:
Overview
===
I notice a lot of things while reviewing code, not all of which I immediately address. I spent my Thanksgiving weekend draining some of my todo backlog (I still have ~160 lines).

As a general rule, we don't like "grab bag PRs" touching lots of code, because they increase risk and are more difficult to review. (We especially dislike mixing features/fixes with unrelated cleanups.) To mitigate this, I've separated out all of the changes with behavioral impact (#5921), higher risk and MSVC-internal impact (#5922), and complicated multi-step transformations (#5923). The commits here are low risk and cleanly organized for ease of reviewing. I believe that decomposing this further wouldn't be productive.

Commits
===
* `xtime.cpp` doesn't need `awint.hpp`, just `Windows.h`.
* `validate.cpp` doesn't use `assert()` anymore.
  + Followup to #650 merged 2020-04-01.
* `// namespace$` => `// unnamed namespace`
  + We had 11 occurrences of the former, now being added to 39 occurrences of the latter.
* Fuse `_HAS_CXX20` regions.
* Guard `_Supports_transparency` for `_HAS_CXX23`.
  + Followup to #3736 merged 2023-06-15.
* Exclude charconv tables and test cases from VSCode search.
  + They pollute search results when looking for numbers, and generally don't need to be searched. (There's a UI button to "Use Exclude Settings and Ignore Files", so this exclusion can be easily bypassed.)
* Drop unnecessary `inline` for function templates.
* `_STATIC_LAMBDA` => `_STATIC_CALL_OPERATOR`
  + Followup to #5580 merged 2025-06-13.
* Add `_STATIC_CALL_OPERATOR` to more stateless lambdas.
* `decltype(tzdb::leap_seconds)` => `vector<leap_second>`
  + This was a relic of an early attempt to give `tzdb::leap_seconds` a fancy type.
* Drop unnecessary newlines in do-while loops.
* `_CMPXCHG_MASK_OUT_PADDING_BITS` implies `_HAS_CXX20`, so `_INLINE_VAR` => `inline`
* `tests/utils/stl/util.py`: Remove dead code, part 1.
* `tests/utils/stl/util.py`: Remove dead code, part 2.
  + We're never AIX.
  + psutil 2.0.0 was released on 2014-03-10. The current version is 7.1.3.
* Unwrap "strengthened" comments.
* `_Vb_val::_Getal()` is a direct member function; all other calls omit `this->`.
* Use `is_null_pointer`.
  + These locations are happy to ignore cv-quals.
* Add newlines to indicate that we don't like to interleave includes and conditionals.
* Avoid interleaving includes and conditionals.
* Include test support headers with angle brackets.
* `int main()` => `void test()` in `test.compile.pass.cpp`.
  + We don't like to have `main()` functions in compile-only tests, to make it extra-clear that they aren't running any code.
* Drop `// COMPILE-ONLY` in `test.compile.pass.cpp`.
  + These comments are redundant.
* Make `VSO_0000000_type_traits` compile-only.
  + `test_all_function_types()` is a non-template, so we don't need to call it.
  + Rename `main()` to `test_VSO_707437()` and move it up. It's testing overload resolution.
* Delete `tests/tr1/env_single.lst`.
  + Followup to #4458 merged 2024-03-15.
* `_USE_NAMED_IDL_NAMESPACE` is unused.
  + IIRC this was intended for WCFB02 (the stranded vNext branch in Team Foundation Version Control).
* Consistently say "Unknown architecture".
  + And don't give string literals to `#error`.
* `vector_algorithms.cpp`: Use `#ifdef _WIN64` when deciding between x64 and x86.
* Qualify a `_STD _In_range` call.
* `<execution>`: `_In_range` => `_In_rng` to avoid shadowing.
* Consistently reduce code duplication when checking intrinsic availability.
* `stride_view` ctor: Mark `range_difference_t<_Vw> _Stride_` as const, to be consistent with others.
* `<random>`: Add parens when mixing left shifts with bitwise OR.
* `_Zip_fn` and `_Zip_transform_fn` should be `class`, not `struct`.
* Make ranges function objects `struct`s when possible.
* `steady_clock::now()` should use `_STL_INTERNAL_STATIC_ASSERT`.
  + We fully control the inputs here, so users don't need these `static_assert`s.
* `_STL_INTERNAL_STATIC_ASSERT(false)` => `static_assert(false)` for C++17 and later.
  + These don't impose any costs, as they occur on "can't happen" branches and we've eliminated the `_Always_false` machinery. We should just leave them in for user code.
* `_STL_INTERNAL_STATIC_ASSERT(false)` => `static_assert(false, "reason")` in C++14 code.
* `CMakeLists.txt`: Avoid inconsistently quote-expanding `VCLIBS_TARGET_ARCHITECTURE`.
* `iosptrs.cpp`: Test `_M_CEE_PURE` instead of `_M_CEE_MIXED`.
  + We don't need to worry about `/clr:safe`. We don't need to provide a rationale.
* Move `stdext` macros to `<exception>` and limit their scope.
  + Followup to #5764 merged 2025-10-10, #5817 merged 2025-11-05.
* `xalloc.cpp`: Include just `<yvals.h>`, we don't need `<stdexcept>`.
* `yvals_core.h`: Remove distinction between "directly controls" and "indirectly controls".
  + This was confusing for contributors to maintain, and didn't provide value.
* `yvals_core.h`: Omit "Other deprecation warnings".
  + Listing this isn't useful; the messages clearly explain what Standard deprecates stuff, and how to silence them.
* Avoid using `is_constant_evaluated()` to initialize const variables.
  + This is confusing and hazardous (witness the `elements[0]` hackaround - that's always 0). The squirrelliness happens because under certain circumstances, the compiler can perform constant initialization, making `is_constant_evaluated()` report that it was called at compile-time. For this reason, we should directly test it with `if`.
* Oops: `xkeycheck.h` doesn't have an "unnamed namespace".

### DIFF
--- a/stl/inc/xkeycheck.h
+++ b/stl/inc/xkeycheck.h
@@ -290,7 +290,7 @@ Enable warning C4005 to find the forbidden define.
 #define namespace EMIT WARNING C4005
 #error The C++ Standard Library forbids macroizing the keyword "namespace". \
 Enable warning C4005 to find the forbidden define.
-#endif // unnamed namespace
+#endif // namespace
 
 #if defined(new) && defined(_ENFORCE_BAN_OF_MACRO_NEW)
 #define new EMIT WARNING C4005


stl/inc/xkeycheck.h
```diff
@@ -290,7 +290,7 @@ Enable warning C4005 to find the forbidden define.
 #define namespace EMIT WARNING C4005
 #error The C++ Standard Library forbids macroizing the keyword "namespace". \
 Enable warning C4005 to find the forbidden define.
-#endif // namespace
+#endif // unnamed namespace
```
cpplearner: This is talking about the keyword `namespace`.
```suggestion
#endif // namespace
```

StephanTLavavej: Ha, good catch.

StephanTLavavej:

Issue #770, opened 2020-05-01
<xlocmon>: "drop 10 zeros" logic is imprecise


StephanTLavavej:
`money_put::do_put` contains:
https://github.com/microsoft/STL/blob/f0f0b55d57e2c9b9da5206b7dbdd71f17e7a0ee3/stl/inc/xlocmon#L676-L678
Like almost all of our pre-charconv floating-point formatting, I believe that this is incorrect - dividing by a power of 10 isn't an exact operation.

This probably isn't a significant issue in practice (as few people have more than 10^35 units of money, although one can dream).

PR #5925, opened 2025-12-01
`<xlocnum>`: Defend against `sprintf_s()` failure in `num_put::do_put()`

StephanTLavavej:
Prompted by a customer report (which I'm guessing involved some kind of static analysis tool), internal VSO-2644079.

The `num_put::do_put()` overloads prepare format strings at runtime (controlled by various flags), then they call `sprintf_s()` with appropriately sized buffers. This should always succeed (even if the logic to calculate the size of the floating-point buffers is looser than I'd prefer, with a "fudge factor", I believe the calculated size is never too small). However, we were unconditionally using `static_cast<size_t>` on the return value of [`sprintf_s()`][], which is documented to return "The number of characters written, or -1 if an error occurred." That looks bad, even if no badness happens at runtime.

To improve this, I was inspired by the otherwise-terrible code (see #770) in `money_put::do_put()`, which does handle `sprintf_s()` returning negative values:

https://github.com/microsoft/STL/blob/2e27f1f05a155bb3637dad04b4d9c4c4590c8542/stl/inc/xlocmon#L674-L678

As this should be a can't-happen case, I think that just returning `_Dest` (saying "we didn't write anything") is the simplest option. Throwing an exception would be much more impactful (and we shouldn't throw for STL logic errors). An alternative would be to fastfail in production, similar to what we do for hardening.

I'm additionally extracting the calls to `_Ffmt()` and `_Ifmt()`, which prepare the runtime format strings. There's no functional change, but it makes `num_put::do_put()` easier to understand when we avoid doing tons of things in a single expression.

[`sprintf_s()`]: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/sprintf-s-sprintf-s-l-swprintf-s-swprintf-s-l?view=msvc-170

### DIFF
--- a/stl/inc/xlocnum
+++ b/stl/inc/xlocnum
@@ -1313,7 +1313,9 @@ protected:
         char _Buf[2 * _Max_int_dig];
         char _Fmt[6];
 
-        const int _Count = _CSTD sprintf_s(_Buf, sizeof(_Buf), _Ifmt(_Fmt, "ld", _Iosbase.flags()), _Val);
+        (void) _Ifmt(_Fmt, "ld", _Iosbase.flags());
+
+        const int _Count = _CSTD sprintf_s(_Buf, sizeof(_Buf), _Fmt, _Val);
 
         if (_Count < 0) { // It should be impossible for sprintf_s() to fail here.
             _STL_INTERNAL_CHECK(false); // In STL tests, report an error.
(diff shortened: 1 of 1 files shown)


stl/inc/xlocnum
```diff
@@ -1313,35 +1313,67 @@ protected:
...
-        return _Iput(_Dest, _Iosbase, _Fill, _Buf,
-            static_cast<size_t>(_CSTD sprintf_s(_Buf, sizeof(_Buf), _Ifmt(_Fmt, "lu", _Iosbase.flags()), _Val)));
+        (void) _Ifmt(_Fmt, "lu", _Iosbase.flags());
+
+        const int _Count = _CSTD sprintf_s(_Buf, sizeof(_Buf), _Fmt, _Val);
+
+        if (_Count < 0) { // It should be impossible for sprintf_s() to fail here.
+            _STL_INTERNAL_CHECK(false); // In STL tests, report an error.
+            return _Dest; // In production, return early to avoid static_cast<size_t> of a negative _Count.
+        }
```
davidmrdavid, line 1340: It's a non-blocking concern, but my instinct would be to fastfail (like in hardening) in production if any 'impossible' intermediate results are detected. Other than for ease of implementation, is there a reason not to do that?

My instinct is to avoid encoding silent failure at all costs, so I'd prefer not to have this 'impossible' path in the first place. But I'll trust your intuition here on whether this is a reasonable concern.

StephanTLavavej: I'll think about that for a potential followup. My main concern would be whatever static analysis tool was involved here not being able to detect a noreturn codepath.

PR #5940, opened 2025-12-05
`GH_000685_condition_variable_any`: Revert test changes for LWG-4301

StephanTLavavej:
The compiler front-end devs have reported that the test coverage added by #5885 has been sporadically failing in the MSVC-internal test harness (which I believe uses weaker VMs than our mighty F32as_v6). If a 1 second timeout isn't enough, this test is too flaky to be added to automation.

Fortunately, the product code changes in #5885 to implement LWG-4301 were easy to verify by inspection. There's little risk of this fix regressing, so we should simply revert the test changes.


StephanTLavavej:

PR #5941, opened 2025-12-06
`<mutex>`: Use `if constexpr` for `lock()` and `try_lock()`

cpplearner:


StephanTLavavej:
Wow, great catch, I forgot we were using overloads here! :heart_eyes_cat:

StephanTLavavej:

StephanTLavavej:
Thanks again! Can't believe we used to have to live without `if constexpr`. :cat: :cat: :cat:

PR #5942, opened 2025-12-06
Fix off-by-one in format

SiliconA-Z:
The original loop stopped at _Last (condition _First != _Last; ++_First), so the upper bound of each lead-byte range from CPINFOEXW::LeadByte was never marked. The Windows lead-byte ranges are inclusive, so the final byte in each pair must be set.

### DIFF
--- a/tests/tr1/tests/cwchar2/test.cpp
+++ b/tests/tr1/tests/cwchar2/test.cpp
@@ -5,7 +5,9 @@
 #define TEST_NAME "<cwchar>, part 2"
 
 #include "tdefs.h"
+#include <clocale>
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
Can you add a test case that exercises this fix?

StephanTLavavej:
I am confused because the product code change affects `<format>` but the test doesn't exercise it. (Additionally, we avoid modifying the legacy `tr1` test suite, except when absolutely necessary.)

StephanTLavavej:
Closing due to no response. The bogus test case looks like poorly reviewed AI code to me.

PR #5943, opened 2025-12-06
Reject overlong forms and out-of-range code points in _Mbrtowc

SiliconA-Z:
Added UTF-8 validation in _Mbrtowc to reject overlong encodings and out-of-range code points, returning EILSEQ instead of accepting ill-formed sequences.

### DIFF
--- a/stl/src/xmbtowc.cpp
+++ b/stl/src/xmbtowc.cpp
@@ -126,6 +126,12 @@ _MRTIMP2 _Success_(return >= 0) int __cdecl _Mbrtowc(
             }
         }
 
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
Similarly to #5942, can you add a test case (or cases) to exercise this fix?

StephanTLavavej:
Without a test case, it is unclear whether this is fixing a real bug, or if it's AI-written code that's not doing anything productive.

StephanTLavavej:
Closing due to no response, superseded by #6060, thanks @cpplearner!

PR #5948, opened 2025-12-07
Test tooling: Use stricter regex for lines in `expected_results.txt` files, provide readable errors for format errors

vmichal:
When parsing `expected_results.txt` for some test suite, the python script was not ready to handle the following issues:
- Missing result code (only file name present) yielded generic "AttributeError: 'NoneType' object has no attribute 'group'" when attempting to access regex match groups.
- Typo in result code yielded "AttributeError: module 'stl.test.tests' has no attribute", which is arguably better, but still could be better.

Both issues are quite easy to trigger when experimenting with a larger number of tests, so user-friendly error messages seem like a nice touch.

### DIFF
--- a/tests/utils/stl/test/file_parsing.py
+++ b/tests/utils/stl/test/file_parsing.py
@@ -139,14 +139,14 @@ def parse_result_file(filename: Union[str, bytes, os.PathLike]) \
     for line in parse_commented_file(filename):
         m = _EXPECTED_RESULT_REGEX.match(line)
         if m is None:
-            raise Exception(f'Line "{line}" has incorrect format in {filename}.')
+            raise Exception(f'Incorrectly formatted line "{line}" in {filename}.')
         prefix = m.group("prefix")
         result = m.group("result")
         result_code = getattr(lit.Test, result, None)
         if result_code is None:
             result_code = getattr(stl.test.tests, result, None)
         if result_code is None:
-            raise Exception(f'Unknown result code "{result}" in "{line}" in {filename}.')
+            raise Exception(f'Unknown result code "{result}" in line "{line}" in {filename}.')
         res[prefix] = result_code
 
     _expected_result_entry_cache[str(filename)] = res


vmichal:
BTW I am more than open to alternative wording of error messages. I experimented with a few alternatives and this one I disliked the least.

tests/utils/stl/test/file_parsing.py
```diff
@@ -58,7 +58,7 @@ class _ParseCtx:
 _TAGS_REGEX = re.compile(r'^(?P<tags>(\*|\w+(,\w+)*))\t+(?P<remainder>.*)$')
 _ENV_VAR_MULTI_ITEM_REGEX = re.compile(r'(?P<name>\w+)="(?P<value>.*?)"')
 _CROSSLIST_REGEX = re.compile(r'^RUNALL_CROSSLIST$')
-_EXPECTED_RESULT_REGEX = re.compile(r'^(?P<prefix>.*) (?P<result>.*?)$')
+_EXPECTED_RESULT_REGEX = re.compile(r'^(?P<prefix>[^ ]*) (?P<result>.*?)$')
```
StephanTLavavej: This could be significantly stricter. We expect to have 1 or more characters in the prefix (so instead of `*` we should use `+`), we expect the result to be 1 or more word characters ("FAIL", "SKIPPED", etc.), and now that you're excluding spaces from the prefix, I don't think we need the non-greedy matching.

tests/utils/stl/test/file_parsing.py
```diff
@@ -138,11 +138,15 @@ def parse_result_file(filename: Union[str, bytes, os.PathLike]) \
     res = dict()
     for line in parse_commented_file(filename):
         m = _EXPECTED_RESULT_REGEX.match(line)
+        if m is None:
+            raise Exception(f"Line \"{line}\" has incorrect format in {filename}.")
```
StephanTLavavej: We can use single quotes to avoid needing to escape double quotes. (For backslashes, we can use fR to avoid the need for escaping.)

StephanTLavavej: These are good error messages but because the line can be long, I think we should front-load the description of what is bad about it.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: After my commits, the messages are:

```
Exception: Unknown result code "WOOF" in line "std/utilities/smartptr/unique.ptr/unique.ptr.create/make_unique.array.pass.cpp:0 WOOF" in D:/GitHub/STL/tests/libcxx/expected_results.txt.
Exception: Incorrectly formatted line "std/utilities/smartptr/unique.ptr/unique.ptr.create/make_unique.array.pass.cpp:0" in D:/GitHub/STL/tests/libcxx/expected_results.txt.
Exception: Incorrectly formatted line "std/utilities/smartptr/unique.ptr/unique.ptr.create/make_unique.array.pass.cpp:0 FA:IL" in D:/GitHub/STL/tests/libcxx/expected_results.txt.
```

We could put additional effort into special-casing an exception for a missing result code (by far my most common mistake) but this is already such an improvement over the status quo.

StephanTLavavej:

StephanTLavavej:
Thanks for improving this easily-overlooked pain point for contributors! :heart_eyes_cat: :white_check_mark: :tada:

PR #5949, opened 2025-12-08
Enable vectorized minmax_element using Neon on ARM64

hazzlim:
Implement the namespace _Sorting algorithms using Neon, and enable _VECTORIZED_MINMAX_ELEMENT on ARM64 targets.

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -3119,7 +3119,7 @@ namespace {
                 auto _Cur_vals = _Traits::_Load(_First);
 
                 // We don't have unsigned 64-bit stuff, so we'll use sign correction just for that case
-                constexpr bool _Sign_correction = sizeof(_Ty) == 8 && !_Sign && !_Traits::_Has_unsigned_cmp;
+                constexpr bool _Sign_correction = sizeof(_Ty) == 8 && !_Sign;
 
                 if constexpr (_Sign_correction) {
                     _Cur_vals = _Traits::_Sign_correction(_Cur_vals, false);


hazzlim:
I have only enabled _VECTORIZED_MINMAX_ELEMENT in the first instance, as it seemed to make some sense to enable the other _Sorting algorithms in separate PRs.

This PR does not vectorize (u)int64_t on ARM64 as this was not faster than the scalar code.

The benchmark results are below:

<head>

<link rel=File-List

</head>

Name | MSVC Speedup | Clang Speedup
-- | -- | --
bm<uint8_t, Op::Min>/8021 | 24.735 | 9.268
bm<uint8_t, Op::Min>/63 | 5.182 | 2.995
bm<uint8_t, Op::Max>/8021 | 24.695 | 9.561
bm<uint8_t, Op::Max>/63 | 4.896 | 2.976
bm<uint8_t, Op::Both>/8021 | 19.184 | 7.811
bm<uint8_t, Op::Both>/63 | 1.977 | 1.841
bm<uint16_t, Op::Min>/8021 | 12.053 | 4.524
bm<uint16_t, Op::Min>/31 | 3.052 | 2.089
bm<uint16_t, Op::Max>/8021 | 11.808 | 4.756
bm<uint16_t, Op::Max>/31 | 2.933 | 2.047
bm<uint16_t, Op::Both>/8021 | 5.426 | 4.052
bm<uint16_t, Op::Both>/31 | 1.413 | 1.521
bm<uint32_t, Op::Min>/8021 | 6.133 | 1.908
bm<uint32_t, Op::Min>/15 | 1.544 | 1.094
bm<uint32_t, Op::Max>/8021 | 6.074 | 1.92
bm<uint32_t, Op::Max>/15 | 1.53 | 1.132
bm<uint32_t, Op::Both>/8021 | 3.146 | 2.877
bm<uint32_t, Op::Both>/15 | 0.869 | 1.195
bm<int8_t, Op::Min>/8021 | 24.735 | 9.211
bm<int8_t, Op::Min>/63 | 5.222 | 2.778
bm<int8_t, Op::Max>/8021 | 25.244 | 9.286
bm<int8_t, Op::Max>/63 | 5.417 | 2.889
bm<int8_t, Op::Both>/8021 | 11.538 | 11.25
bm<int8_t, Op::Both>/63 | 1.989 | 1.76
bm<int16_t, Op::Min>/8021 | 11.953 | 4.667
bm<int16_t, Op::Min>/31 | 3.029 | 1.872
bm<int16_t, Op::Max>/8021 | 11.808 | 4.571
bm<int16_t, Op::Max>/31 | 3.123 | 1.882
bm<int16_t, Op::Both>/8021 | 6.582 | 5.729
bm<int16_t, Op::Both>/31 | 1.414 | 1.541
bm<int32_t, Op::Min>/8021 | 6.25 | 1.88
bm<int32_t, Op::Min>/15 | 1.6 | 1.135
bm<int32_t, Op::Max>/8021 | 6.133 | 1.867
bm<int32_t, Op::Max>/15 | 1.674 | 1.094
bm<int32_t, Op::Both>/8021 | 3.222 | 1.784
bm<int32_t, Op::Both>/15 | 0.877 | 0.903
bm<float, Op::Min>/8021 | 8.928 | 4.364
bm<float, Op::Min>/15 | 1.87 | 1.358
bm<float, Op::Max>/8021 | 9.111 | 4.267
bm<float, Op::Max>/15 | 2.062 | 1.371
bm<float, Op::Both>/8021 | 5.227 | 1.626
bm<float, Op::Both>/15 | 0.913 | 0.7
bm<double, Op::Min>/8021 | 4.426 | 2.029
bm<double, Op::Min>/7 | 0.929 | 0.731
bm<double, Op::Max>/8021 | 4.563 | 2.133
bm<double, Op::Max>/7 | 0.977 | 0.725
bm<double, Op::Both>/8021 | 2.583 | 0.786
bm<double, Op::Both>/7 | 0.445 | 0.402



</body>

</html>

stl/inc/xutility
```diff
@@ -124,6 +124,16 @@ _STL_DISABLE_CLANG_WARNINGS
 #endif // ^^^ !defined(__clang__) || __clang_major__ < 19 ^^^
 #endif // ^^^ !defined(_USE_BUILTIN_IS_TRIVIALLY_EQUALITY_COMPARABLE) ^^^
 
+#if _VECTORIZED_MINMAX_ELEMENT
+#ifdef _M_ARM64
+template <class _Ty>
+_INLINE_VAR constexpr bool _Is_64bit_int_on_arm64_v = _STD is_integral_v<_Ty> && sizeof(_Ty) == 8;
```
AlexGuteniev: I think we need to embed this in `_Is_min_max_optimization_safe`. See `Vector_alg_in_replace_is_safe` precedent:
https://github.com/microsoft/STL/blob/c370e0570403672773c1e54e83317086299f9c2a/stl/inc/algorithm#L436-L438

I also think we should avoid `is_integral_v` checking. We optimize for pointers too, treating them as unsigned integers, and do this intentionally.

stl/src/vector_algorithms.cpp
```diff
@@ -1631,6 +1966,119 @@ namespace {
 #endif // ^^^ !defined(_M_ARM64EC) ^^^
         };
 
+#ifdef _M_ARM64
+        struct _Traits_8_neon : _Traits_8_base, _Traits_neon_base {
```
AlexGuteniev: I think we should not be providing`_Traits_8_neon`, and also should not provide `_8` functions.

When/if `minmax` or `is_sorted_until` are vectorized. 8 traits can be added with only function needed there.

Because we should strive to avoid dead code.

stl/src/vector_algorithms.cpp
```diff
@@ -934,17 +933,47 @@ namespace {
...
+
+            static int8x16_t _All_ones() noexcept {
+                return vdupq_n_s8(static_cast<int8_t>(0xFF));
+            }
+
+            static int8x16_t _Blend(const int8x16_t _Px1, const int8x16_t _Px2, const int8x16_t _Msk) noexcept {
+                return vbslq_s8(vreinterpretq_u8_s8(_Msk), _Px2, _Px1);
+            }
+
+            static int8x16_t _Sign_correction(const int8x16_t _Val, const bool) noexcept {
```
AlexGuteniev: We don't say `const` on unnamed parameters.

stl/inc/xutility
YexuanXiao: Also include enum?

StephanTLavavej: `GH_005421_vector_algorithms_integer_class_type_iterator` is failing. MSVC's error isn't great, but Clang is somewhat clearer. I don't immediately see where the `void` is coming from, though.

```
In file included from C:\a\_work\1\s\tests\std\tests\GH_005421_vector_algorithms_integer_class_type_iterator\test.cpp:5:
In file included from C:\stlBuild\out\inc\algorithm:10:
In file included from C:\stlBuild\out\inc\__msvc_heap_algorithms.hpp:11:
C:\stlBuild\out\inc\xutility(130,82): error: invalid application of 'sizeof' to an incomplete type 'void'
  130 | _INLINE_VAR constexpr bool _Is_64bit_int_on_arm64_v = _STD is_integral_v<_Ty> && sizeof(_Ty) == 8;
      |                                                                                  ^~~~~~~~~~~
C:\stlBuild\out\inc\xutility(7543,66): note: in instantiation of variable template specialization '_Is_64bit_int_on_arm64_v<void>' requested here
 7543 |     if constexpr (_Is_min_max_optimization_safe<_FwdIt, _Pr> && !_Is_64bit_int_on_arm64_v<_Ty>) {
      |                                                                  ^
C:\stlBuild\out\inc\xutility(7572,14): note: in instantiation of function template specialization 'std::_Min_element_unchecked<test::redifference_iterator<std::_Signed128, const int *>, std::less<void>>' requested here
 7572 |         _STD _Min_element_unchecked(_STD _Get_unwrapped(_First), _STD _Get_unwrapped(_Last), _STD _Pass_fn(_Pred)));
      |              ^
C:\stlBuild\out\inc\xutility(7578,17): note: in instantiation of function template specialization 'std::min_element<test::redifference_iterator<std::_Signed128, const int *>, std::less<void>>' requested here
 7578 |     return _STD min_element(_First, _Last, less<>{});
      |                 ^
C:\a\_work\1\s\tests\std\tests\GH_005421_vector_algorithms_integer_class_type_iterator\test.cpp(83,12): note: in instantiation of function template specialization 'std::min_element<test::redifference_iterator<std::_Signed128, const int *>>' requested here
   83 |     assert(min_element(arr_begin, arr_end) == arr_begin);
      |            ^
1 error generated.
```

StephanTLavavej: I agree that:

* Baking this into `_Is_min_max_optimization_safe` is ideal.
* Because the rationale was "This PR does not vectorize (u)int64_t on ARM64 as this was not faster than the scalar code.", `is_integral_v` shouldn't be checked here. Presumably, pointers (which are also 8-byte on ARM64) and enums with 8-byte underlying types, would have the same undesirable performance characteristics, and should also be uniformly excluded from vectorization.

I would like to request that this rationale be captured in a comment; we should distinguish "we can't physically do this" from "we could do this but it wouldn't be faster".

AlexGuteniev: > Also include enum?

No.

muellerj2: The cause is that `test::redifference_iterator<std::_Signed128, const int *>` fails the `_Cpp17_input_iterator` concept because `signed_integral<_Signed128>` is false.

Edit: No, wait, I'm wrong about this because I misread the code.

Edit 2: But I wasn't far off. `signed_integral<_Signed128>` is false, so `redifference_iterator<_Signed128, const int*>` doesn't get a member `iterator_category` here:

https://github.com/microsoft/STL/blob/c4d3fbaa389a56f63bb1253ea3c1b088bc49ee95/tests/std/include/range_algorithm_support.hpp#L963-L967

I think this means we don't take this case in `iterator_traits`:

https://github.com/microsoft/STL/blob/c4d3fbaa389a56f63bb1253ea3c1b088bc49ee95/stl/inc/__msvc_iter_core.hpp#L194-L201

But this one:
https://github.com/microsoft/STL/blob/c4d3fbaa389a56f63bb1253ea3c1b088bc49ee95/stl/inc/__msvc_iter_core.hpp#L231-L240

StephanTLavavej: Is this related to WG21-N5014 \[iterator.iterators\]/2.2 "A type `X` meets the *Cpp17Iterator* requirements if [...] `iterator_traits<X>::difference_type` is a signed integer type or `void`"?

I don't understand what `redifference_iterator` and especially `redifference_iterator_category_base` were trying to do, so it's not obvious to me how to repair this without damaging whatever this was trying to exercise in ranges.

hazzlim: I'm not sure I completely understand the error here, but FWIW it seems to have gone away now I've rolled it into `_Is_min_max_optimization_safe`...

I'll push this along with the rest of the review changes tomorrow!

AlexGuteniev: > Also include enum?

enums may redefine operators, so are not manually vectorizable for most algorithms, unless it can be proven that no user defined operators are present. User defined operators might mask out bits, use lookup arrays, or throw exceptions. Not very plausible scenarios, but still possible.

----

Potentially the compiler can let us inspect if there are no user defined operators that may make the comparison unusual. For Clang `__is_trivially_equality_comparable` could have worked, see #5479, but it still does not work for enums as for now. There's no MSVC equivalent, as far as I'm aware. 

For sorting algorithms, like this one, we actually need `operator <` to be trivial, I'm not aware even for Clang trait that would have checked for that. 

Note that the goal of providing  `__is_trivially_equality_comparable` was different from the optimizations: it was to make atomic ops `compare_exchange_strong` and `compare_exchange_weak` possible. And that problem has already been approached from the other end, by providing library access to padding bits. For optimizations, `__is_trivially_equality_comparable` helped to achieve something (in this STL as #5479 and in libc++), but not really too much. So probably this direction isn't worth pursuing further, taking into account it would consume precious front end compiler resources.

AlexGuteniev: > I don't understand what `redifference_iterator` and especially `redifference_iterator_category_base` were trying to do, so it's not obvious to me how to repair this without damaging whatever this was trying to exercise in ranges.

I think it is trying to set odd `difference_type` to make sure algorithms and views handle that properly. 

In this particular test, vectorized algorithms are tested to have proper casts when they need to pass the difference as `size_t` or `ptrdiff_t`.

hazzlim: I have rolled this into `_Is_min_max_optimization_safe`, and added a comment that hopefully makes the rationale for this clearer. 

Regarding the `is_integral_v` check, I was trying to avoid catching `double` here, as `sizeof(_Ty) == 8` would also be true. Hopefully using `!_STD is_floating_point_v<_Ty>` is better!

stl/src/vector_algorithms.cpp
hazzlim: I've remove `_Traits_8_neon`, and guarded the definitions of the `_8` functions with `#ifndef _M_ARM64`.

Related: I also realize it didn't make sense to define `minmax` and `is_sorted_until` for the time being, so I have guarded those too.

I've left declarations of the `_8` functions as-is in `xutility` / `algorithm`, as I figured we don't mind unused and undefined declarations - but let me know if we want to wrap those in guards also!

stl/src/vector_algorithms.cpp
hazzlim: Fixed :)

hazzlim:
Aha I see that `VSO_0000000_vector_algorithms_floats are failing`... I will take a look there

(Stupidly I didn't realize that floating point functions were not exercised under VSO_0000000_vector_algorithms tests, sorry!)

stl/src/vector_algorithms.cpp
```diff
@@ -2382,37 +2944,39 @@ namespace {
...
-                                _BitScanForward(&_H_pos, _Mask);
+                                unsigned long _H_pos = _Traits::_Get_first_h_pos(_Mask);
 
                                 // Extract its vertical index
+
+#ifdef _M_ARM64
+                                const auto _V_pos = _Traits::_Get_v_pos(_Idx_min);
+#else
                                 const auto _V_pos = _Traits::_Get_v_pos(_Cur_idx_min, _H_pos);
+#endif
```
AlexGuteniev: I think we can simplify and always use `_Idx_min`.

Not sure if we need to do this here or as a follow up.

AlexGuteniev: Ditto below.

hazzlim:
> Aha I see that `VSO_0000000_vector_algorithms_floats are failing`... I will take a look there
> 
> (Stupidly I didn't realize that floating point functions were not exercised under VSO_0000000_vector_algorithms tests, sorry!)

Should be fixed - it's a shame we don't have `-flax-vector-conversions=false` for Neon on MSVC 😢

stl/src/vector_algorithms.cpp
hazzlim: Happy to change here if we think it makes more sense than doing it separately?

stl/src/vector_algorithms.cpp
```diff
@@ -990,6 +1034,20 @@ namespace {
             static void _Exit_vectorized() noexcept {
                 _mm256_zeroupper();
             }
+
+            static unsigned long _Get_first_h_pos(unsigned long _Mask) {
+                unsigned long _H_pos;
+                // CodeQL [SM02313] _H_pos is always initialized: element exists, so _Mask != 0.
+                _BitScanForward(&_H_pos, _Mask);
```
AlexGuteniev: This can be `_tzcnt_u32`.
We assume that AVX2 implies BMI and BMI2.

I decided not to bother for uncommon code path back when I added AVX2 here, but since we have to, we can take advantage of it.

Note that SSE should stay `_BitScanForward`, with SSE4.2 we only assume `popcnt` from bit manipulations.

stl/src/vector_algorithms.cpp
```diff
@@ -990,6 +1034,20 @@ namespace {
...
+                unsigned long _H_pos;
+                // CodeQL [SM02313] _H_pos is always initialized: element exists, so _Mask != 0.
+                _BitScanForward(&_H_pos, _Mask);
+                return _H_pos;
+            }
+
+            static unsigned long _Get_last_h_pos(unsigned long _Mask) {
+                unsigned long _H_pos;
+                // CodeQL [SM02313] _H_pos is always initialized: element exists, so _Mask != 0.
+                _BitScanReverse(&_H_pos, _Mask);
```
AlexGuteniev: And this can be `31 - _lzcnt_u32`.
We assume that AVX2 implies BMI and BMI2.

~~And we can bring `_H_pos -= sizeof(_Cur_max_val) - 1; // Correct from highest val bit to lowest` inside `_Get_last_h_pos`, so that for `_lzcnt_u32` and for ARM64 the negations would cancel out.~~ no, this one isn't good.

stl/src/vector_algorithms.cpp
AlexGuteniev: I think let's do this here.

stl/src/vector_algorithms.cpp
StephanTLavavej: I think it's confusing to have declarations of functions that are never defined.

stl/src/vector_algorithms.cpp
```diff
@@ -2016,6 +2409,89 @@ namespace {
...
+            }
+
+            static _Vec_t _H_max(const _Vec_t _Cur) noexcept {
+                return vdupq_n_f64(vmaxvq_f64(_Cur));
+            }
+
+            static _Idx_t _H_min_u(const _Idx_t _Cur) noexcept {
+                const uint64x2_t _Cur_u = vreinterpretq_u64_s64(_Cur);
+                uint64x2_t _Swapped     = vextq_u64(_Cur_u, _Cur_u, 1);
+                uint64x2_t _Mask_lt     = vcltq_u64(_Swapped, _Cur_u);
```
AlexGuteniev: `const` maybe.

We generally try to add `const` for local such variables to aid understanding that variables are not modified, so that non-`const` stand out.

stl/src/vector_algorithms.cpp
```diff
@@ -2016,6 +2409,89 @@ namespace {
...
+                return vreinterpretq_s64_u64(vbslq_u64(_Mask_lt, _Swapped, _Cur_u));
+            }
+
+            static _Idx_t _H_max_u(const _Idx_t _Cur) noexcept {
+                const uint64x2_t _Cur_u = vreinterpretq_u64_s64(_Cur);
+                uint64x2_t _Swapped     = vextq_u64(_Cur_u, _Cur_u, 1);
+                uint64x2_t _Mask_gt     = vcgtq_u64(_Swapped, _Cur_u);
```
AlexGuteniev: And `const` here

stl/src/vector_algorithms.cpp
```diff
@@ -2382,37 +2926,34 @@ namespace {
...
                                 auto _Idx_min = _Traits::_H_min_u(_Idx_min_val); // The smallest indices
                                 // Select the smallest vertical indices from the smallest element mask
                                 _Mask &= _Traits::_Mask(_Traits::_Cmp_eq_idx(_Idx_min, _Idx_min_val));
-                                unsigned long _H_pos;
 
                                 // Find the smallest horizontal index
-
-                                // CodeQL [SM02313] _H_pos is always initialized: element exists, so _Mask != 0.
-                                _BitScanForward(&_H_pos, _Mask);
+                                unsigned long _H_pos = _Traits::_Get_first_h_pos(_Mask);
```
AlexGuteniev: This wasn't `const` because of the inconvenient `_BitScanForward`, but now it is not used right here, we can add `const`.

stl/src/vector_algorithms.cpp
```diff
@@ -2770,13 +3332,10 @@ namespace {
...
                     const auto _Is_less = _Traits::_Cmp_gt(_Right, _Left);
-                    unsigned long _Mask = _Traits::_Mask(_Traits::_Mask_cast(_Is_less));
+                    auto _Mask          = _Traits::_Mask(_Traits::_Mask_cast(_Is_less));
 
                     if (_Mask != 0) {
-                        unsigned long _H_pos;
-
-                        // CodeQL [SM02313] _H_pos is always initialized: we just tested `if (_Mask != 0)`.
-                        _BitScanForward(&_H_pos, _Mask);
+                        unsigned long _H_pos = _Traits::_Get_first_h_pos(_Mask);
```
AlexGuteniev: This wasn't `const` because of the inconvenient `_BitScanForward`, but now it is not used right here, we can add `const`.

stl/src/vector_algorithms.cpp
```diff
@@ -2770,13 +3332,10 @@ namespace {
                     }
 
                     const auto _Is_less = _Traits::_Cmp_gt(_Right, _Left);
-                    unsigned long _Mask = _Traits::_Mask(_Traits::_Mask_cast(_Is_less));
+                    auto _Mask          = _Traits::_Mask(_Traits::_Mask_cast(_Is_less));
```
AlexGuteniev: And this is pre-existing, there should have been `const` before this change.
Presumably, was copied from another occurrence this way, where `_Mask` is potentially modified.

stl/src/vector_algorithms.cpp
```diff
@@ -2798,14 +3357,10 @@ namespace {
...
-                        unsigned long _Mask =
-                            _Traits::_Mask(_mm256_and_si256(_Traits::_Mask_cast(_Is_less), _Tail_mask));
+                        auto _Mask = _Traits::_Mask(_mm256_and_si256(_Traits::_Mask_cast(_Is_less), _Tail_mask));
 
                         if (_Mask != 0) {
-                            unsigned long _H_pos;
-
-                            // CodeQL [SM02313] _H_pos is always initialized: we just tested `if (_Mask != 0)`.
-                            _BitScanForward(&_H_pos, _Mask);
+                            unsigned long _H_pos = _Traits::_Get_first_h_pos(_Mask);
```
AlexGuteniev: ditto `const`

stl/src/vector_algorithms.cpp
```diff
@@ -2798,14 +3357,10 @@ namespace {
                         }
 
                         const auto _Is_less = _Traits::_Cmp_gt(_Right, _Left);
-                        unsigned long _Mask =
-                            _Traits::_Mask(_mm256_and_si256(_Traits::_Mask_cast(_Is_less), _Tail_mask));
+                        auto _Mask = _Traits::_Mask(_mm256_and_si256(_Traits::_Mask_cast(_Is_less), _Tail_mask));
```
AlexGuteniev: Ditto `const`

stl/src/vector_algorithms.cpp
```diff
@@ -2713,19 +3271,23 @@ namespace {
...
+#endif // ^^^ !defined(_M_ARM64) && !defined(_M_ARM64EC) ^^^
 
         template <_Min_max_mode _Mode, class _Traits, bool _Sign>
         auto __stdcall _Minmax_disp(const void* const _First, const void* const _Last) noexcept {
-#ifndef _M_ARM64EC
+#ifdef _M_ARM64
+            if (_Byte_length(_First, _Last) >= 16) {
+                return _Minmax_impl<_Mode, typename _Traits::_Neon, _Sign>(_First, _Last);
+            }
+#elif !defined(_M_ARM64EC)
```
AlexGuteniev: I also observe that providing ARM64 in `minmax` and `is_sorted_until` dispatches looks premature, as this PR does not try to enable them. But I don't see any problem with that, as the functions seen by the linker, like `__std_minmax_1` for ARM64, are not provided.

hazzlim:
> Now it looks good to me, though I haven't looked up what these intrinsics do. Let's spam even more `const` though.

Nice - should have added all of these const-qualifiers :)

stl/src/vector_algorithms.cpp
hazzlim: Ah yes good point - I have removed these all the same and added macro guards around the `minmax` and `is_sorted_until` dispatches, I agree these should get added later. I think I was originally trying to reduce the number of macro guards, but there's already quite a few now!

AlexGuteniev:
Curious how Clang gets only modest speedup, but still gets speedup.
Does Clang auto-vectorize somehow? Does MSVC do something dumb here?

> Name	|MSVC Speedup|	Clang Speedup
>--|---|--
> bm<uint8_t, Op::Min>/8021  | 24.735 | 9.268

hazzlim:
> Curious how Clang gets only modest speedup, but still gets speedup. Does Clang auto-vectorize somehow? Does MSVC do something dumb here?
> 
> > Name
> > MSVC Speedup
> > Clang Speedup
> > 
> > 
> > 
> > 
> > bm<uint8_t, Op::Min>/8021
> > 24.735
> > 9.268

Clang does not auto-vectorize, both are scalar code - but Clang keeps the current minimum in a register whereas MSVC reloads it every iteration of the main loop. The extra load on the critical path makes MSVC a lot slower.

AlexGuteniev:
> MSVC reloads it every iteration of the main loop

Oh, the same problem it has on x86 and x64 too!

May worth reporting on DevCom though, as this occurrence causes ridiculous slowdown.

hazzlim:
> > MSVC reloads it every iteration of the main loop
> 
> Oh, the same problem it has on x86 and x64 too!
> 
> May worth reporting on DevCom though, as this occurrence causes ridiculous slowdown.

Sure, I will open a ticket on DevCom :)

stl/inc/algorithm
```diff
@@ -211,8 +213,10 @@ pair<_Ty*, _Ty*> _Minmax_element_vectorized(_Ty* const _First, _Ty* const _Last)
         _Res = ::__std_minmax_element_2(_First, _Last, _Signed);
     } else if constexpr (sizeof(_Ty) == 4) {
         _Res = ::__std_minmax_element_4(_First, _Last, _Signed);
+#ifndef _M_ARM64
     } else if constexpr (sizeof(_Ty) == 8) {
         _Res = ::__std_minmax_element_8(_First, _Last, _Signed);
+#endif // ^^^ !defined(_M_ARM64) ^^^
```
StephanTLavavej: Although this results in a nice-looking diff, we avoid splitting up braces like this. Example: https://github.com/microsoft/STL/blob/cbe2ee99caf122555a305d18f69e57c75b3fe5ec/stl/inc/filesystem#L149-L159

After thinking about this, I think that a special `static_assert(false)` would be ideal.

stl/inc/xutility
```diff
@@ -7281,9 +7289,19 @@ constexpr bool _Is_predicate_greater = _Is_any_of_v<_Pr,
...
     greater<>, greater<_Iter_value_t<_Iter>>>;
 
+#ifdef _M_ARM64
+// We choose not to vectorize minmax_element on ARM64 as this does not improve performance over the scalar code.
+template <class _Ty>
+_INLINE_VAR constexpr bool _Is_64bit_int_on_arm64_v = sizeof(_Ty) == 8 && !_STD is_floating_point_v<_Ty>;
+#else // ^^^ defined(_M_ARM64) / !defined(_M_ARM64) vvv
+template <class _Ty>
+_INLINE_VAR constexpr bool _Is_64bit_int_on_arm64_v = false;
+#endif // ^^^ defined(_M_ARM64) ^^^
```
StephanTLavavej: Missing `!` in comment.

stl/inc/xutility
```diff
@@ -7281,9 +7289,19 @@ constexpr bool _Is_predicate_greater = _Is_any_of_v<_Pr,
 #endif // _HAS_CXX20
     greater<>, greater<_Iter_value_t<_Iter>>>;
 
+#ifdef _M_ARM64
+// We choose not to vectorize minmax_element on ARM64 as this does not improve performance over the scalar code.
+template <class _Ty>
+_INLINE_VAR constexpr bool _Is_64bit_int_on_arm64_v = sizeof(_Ty) == 8 && !_STD is_floating_point_v<_Ty>;
```
StephanTLavavej: `_STD` qualification is unnecessary for `is_floating_point_v`; we're in `namespace std` here, and variable templates can't be hijacked by ADL.

stl/src/vector_algorithms.cpp
```diff
@@ -1842,6 +2159,81 @@ namespace {
...
+        struct _Traits_f_neon : _Traits_f_base, _Traits_neon_base {
+            using _Vec_t                            = float32x4_t;
+            using _Idx_t                            = int32x4_t;
+            static constexpr bool _Has_unsigned_cmp = false;
+
+            static uint64_t _Mask(const _Idx_t _Val) noexcept {
+                return _Traits_4_neon::_Mask(_Val);
+            }
+
+            static unsigned long _Get_first_h_pos(uint64_t _Mask) {
```
StephanTLavavej: Missing `noexcept`s.

stl/src/vector_algorithms.cpp
```diff
@@ -2262,8 +2751,13 @@ namespace {
             return _Res;
         }
 
+#ifdef _M_ARM64
+        template <_Min_max_mode _Mode, class _Traits, const bool _Sign>
```
StephanTLavavej: `const bool` is unnecessary in a template parameter.

stl/src/vector_algorithms.cpp
```diff
@@ -1023,6 +1076,97 @@ namespace {
...
+
+            static _Vec_t _H_min_u(const _Vec_t _Cur) noexcept {
+                const uint8_t m = vminvq_u8(vreinterpretq_u8_s8(_Cur));
+                return vreinterpretq_s8_u8(vdupq_n_u8(m));
+            }
+
+            static _Vec_t _H_max_u(const _Vec_t _Cur) noexcept {
+                const uint8_t M = vmaxvq_u8(vreinterpretq_u8_s8(_Cur));
+                return vreinterpretq_s8_u8(vdupq_n_u8(M));
+            }
```
StephanTLavavej, line 1117: While this is a source file so we don't *need* to adhere to `_Ugly` discipline, we've chosen to follow it here. These variables are non-ugly. Plain `r` occurs below too.

stl/src/vector_algorithms.cpp
```diff
@@ -1230,6 +1373,97 @@ namespace {
...
+        struct _Traits_2_neon : _Traits_2_base, _Traits_neon_base {
+            using _Vec_t = int16x8_t;
+
+            // Compresses a 128-bit Mask of 8 16-bit values into a 64-bit Mask of 8 8-bit values.
+            static uint64_t _Mask(const _Vec_t _Val) noexcept {
+                const uint16x4_t _Res = vshrn_n_u32(vreinterpretq_u32_s16(_Val), 8);
+                return vget_lane_u64(vreinterpret_u64_u16(_Res), 0);
+            }
+
+            static unsigned long _Get_first_h_pos(uint64_t _Mask) {
```
StephanTLavavej: These params could be const.

stl/src/vector_algorithms.cpp
```diff
@@ -934,17 +933,47 @@ namespace {
...
+            }
+
+            static int8x16_t _Sign_correction(const int8x16_t _Val, bool) noexcept {
+                return _Val;
+            }
+
+            static void _Exit_vectorized() noexcept {}
+        };
+#else // ^^^ defined(_M_ARM64EC) / !defined(_M_ARM64EC) vvv
 #ifndef _M_ARM64EC
```
StephanTLavavej: This `#else` is for the newly introduced `#ifdef _M_ARM64` above, but it's commented as being about `_M_ARM64EC`.

Actually, if we chain these together, I think it becomes easier to read and more consistent with other occurrences. That is, changing from:

```cpp
#ifdef _M_ARM64
cool neon stuff
#else // ^^^ defined(_M_ARM64) / !defined(_M_ARM64) vvv
#ifndef _M_ARM64EC
cool sse/avx stuff
#endif // ^^^ !defined(_M_ARM64EC) ^^^
#endif // ^^^ !defined(_M_ARM64) ^^^
```

to:

```cpp
#ifdef _M_ARM64
cool neon stuff
#elif !defined(_M_ARM64EC)
cool sse/avx stuff
#endif // ^^^ !defined(_M_ARM64EC) ^^^
```

stl/src/vector_algorithms.cpp
```diff
@@ -2574,7 +3133,7 @@ namespace {
                 auto _Cur_vals = _Traits::_Load(_First);
 
                 // We don't have unsigned 64-bit stuff, so we'll use sign correction just for that case
-                constexpr bool _Sign_correction = sizeof(_Ty) == 8 && !_Sign;
+                constexpr bool _Sign_correction = sizeof(_Ty) == 8 && !_Sign && !_Traits::_Has_unsigned_cmp;
```
StephanTLavavej: This appears to be an editing relic. `#ifndef _M_ARM64` is guarding `_Minmax_impl` immediately above, and only `_Traits_neon_base` sets `_Has_unsigned_cmp` to `true`. Since we aren't ARM64 here, `_Has_unsigned_cmp` is always `false` here.

If this needs to be added when vectorizing `_Minmax_impl` for ARM64, we should do that in a future PR, not this PR.

StephanTLavavej:

StephanTLavavej:
:chart_with_downwards_trend: :chart_with_upwards_trend: :mechanical_arm:
===

Issue #5944, opened 2025-12-06
`<regex>`: Revising the stack and complexity limits


muellerj2:
With #5939, the non-recursive `<regex>` rewrite has reached the point where we should decide how we adjust the stack and complexity limits in the implementation.

While rewriting the matcher, I made sure that the existing counters have kept working roughly the same (except that the counts are sometimes a bit lower) so that the set of allowed input strings should not have shrunk at least.

But while these counters were natural for the recursive parser (counting recursive calls), the counters are now adjusted in lots of different places, creating quite a mess. I think we should change the counters and limits to something more natural and maintainable now.

## Why are such limits needed?

In theoretical computer science courses, it is taught that a regular expression can be matched by a deterministic finite automaton (DFA), which basically translates into a linear-time algorithm. But (a) a DFA might have exponential size compared to the non-deterministic finite automaton (NFA) that the regular expression was parsed into, and (b) practical regex grammars can be used to accept non-regular languages. For example, matching a regex grammar supporting backreferences is (probably) not possible in linear time, because this problem is NP-hard. The limits are required to defend against super-linear running times when matching regexes against input strings.

## The current limits are too strict for long input strings
The limits are supposed to defend against super-linear running time, but the limits are currently constants, so they even defend against linear running time. This means that the current implementation can only successfully match each regex grammar against input strings up to some maximum constant length. This is a pointlessly strict limit, because users can easily achieve this same restriction by just explicitly limiting the length of the input string they pass to the regex algorithms.

This is particularly bad for the stack limit, which is currently set to 1000. A constant like that made sense for the recursive parser, because it was supposed to limit the usage of the thread's stack with a constant size of usually 1 MB. But now this stack lives on the heap and can grow.

## The stack counter
After #5939, we have reached the point that the stack counter is bounded from above by what is essentially `_Frames_count` times some factor (I would have to do the exact math, but the factor should be roughly $$3\cdot \texttt{\\_Ncap} + 1$$ for ECMAScript, which in turn is at most 3001 because of the capture group limit in the parser). So I propose to do away with the current stack counter and replace it by `_Frames_count`, if we want to keep applying a limit on the stack size as well.

## The stack limit
[...]

PR #5954, opened 2025-12-09
`<regex>`: Revise the stack limit

### LINKED ISSUE #997: <regex>: Grouping within repetition causes regex stack error

muellerj2:
This PR revises the stack limit based on the discussion in #5944. Resolves #997.

I might have been a bit too paranoid about negative values and overflows.

I store the input length in the matcher because I think I will probably use this member for the revision of the complexity limit as well.

This removes the stack counting in `_Matcher3::_Find_first_inner_capture_group()`. The number of recursive calls is bounded by the recursion limit of 1000 in `_Parser2::_Disjunction()`, but the function uses less stack than the equivalent recursion in the parser does, so there is not much point anymore to an explicit recursion limit check in this function (though the explicit limit mitigated the impact of #5792, but I don't think there is another bug like this here).

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -1722,10 +1722,6 @@ public:
         }
 
         _Input_length = _STD distance(_Pfirst, _Plast);
-        if (_Input_length < 0) {
-            _Input_length = 0;
-        }
-
         _Frames_limit = _Calculate_frames_limit();
 
 // sanitize multiline mode setting
@@ -3410,7 +3406,6 @@ void _Matcher3<_BidIt, _Elem, _RxTraits, _It, _Alloc>::_Pop_frame(size_t _Idx) {
     _STL_INTERNAL_CHECK(_Idx + 1 == _Frames_count);
     _Frames_count = _Idx;
 }
-
 
 template <class _BidIt, class _Elem, class _RxTraits, class _It, class _Alloc>
 size_t _Matcher3<_BidIt, _Elem, _RxTraits, _It, _Alloc>::_Calculate_frames_limit() {


AlexGuteniev:
I'm surprised that there are no `test_gh_997()` changes. If #997 is resolved, there should be no exception expected in its test.

muellerj2:
> If https://github.com/microsoft/STL/issues/997 is resolved, there should be no exception expected in its test.

The test already checks in its current form that no exception is thrown. #5889 already fixed the example regex in #997 and adjusted the test, I had just not declared the bug fixed because something like `(a|bc)*` would still result in an exception.

But fine, I will add the regex `(a|bc)*` as an example to `test_gh_997()` as well.

muellerj2:
I reworked the limit calculation a bit. It's slightly more lenient now (by about 7 % for pointer-sized input iterators, so more like 10.7 than 10 bytes for stack frames per character), but in exchange it behaves better when the number of characters is between `PTRDIFF_MAX/10` and `PTRDIFF_MAX`.

stl/inc/regex
```diff
@@ -1725,6 +1721,13 @@ public:
             _Char_class_d = _Lookup_char_class(static_cast<_Elem>('D'));
         }
 
+        _Input_length = _STD distance(_Pfirst, _Plast);
+        if (_Input_length < 0) {
+            _Input_length = 0;
+        }
```
StephanTLavavej: Some caution is often good, but this is overly paranoid. We've already performed `_Adl_verify_range(_Pfirst, _Plast);` to rule out transposed ranges (in debug mode). If the iterators are transposed, we have bigger problems.

StephanTLavavej:

StephanTLavavej:
:heart_eyes_cat: :green_heart: :gift: :gift_heart: :heart_eyes_cat:
===

PR #5959, opened 2025-12-12
The Uncanny xmath Overhaul

StephanTLavavej:
Followup to #5836.

I believe that there are no functional changes here. I've verified that the dllexport surface is unaffected.

Commits
===
* Fuse xfcosh.cpp into xcosh.cpp.
* Fuse xfsinh.cpp into xsinh.cpp.
* Fuse xfdtest.cpp into xdtest.cpp.
* Fuse xfexp.cpp into xexp.cpp.
* Fuse xfdscale.cpp into xexp.cpp.
  + `_FDscale()` wasn't dllexported, so we can move it into an unnamed namespace.
* Fuse xfdnorm.cpp into xexp.cpp.
  + `_FDnorm()` wasn't dllexported, so we can move it into an unnamed namespace.
* Replace `_Feraise()` with `errno = ERANGE;`.
  + It had no other effect. I'm adding `// underflow` versus `// overflow` comments to preserve clarity. `_Feraise()` wasn't dllexported. It was marked `__CLRCALL_PURE_OR_CDECL` but not `_CRTIMP2_PURE`. Remove now-unused macros.
* Fuse underflow/overflow cases, avoid missing `break;`.
* Preserve `_Dtest()` for bincompat, replaced with `fpclassify()`.
  + They return the same values, see below.
* `_FINITE` => `FP_NORMAL`
* `_INFCODE` => `FP_INFINITE`
* `_NANCODE` => `FP_NAN`
* `0` => `FP_ZERO`, carefully.
* Remove now-unused `_Dtest()` macros.
* Replace `DSIGN` and `FSIGN` with `signbit()`.
  + See below for an equivalence proof.
* Move `_Dval`, `_Fval`, and their index macros into xexp.cpp's unnamed namespace.
  + It's okay to move these unions out of `_EXTERN_C_UNLESS_PURE`, because nothing else uses them.
* Change index macros to `constexpr int`, cleanup comments.
* Move `_Xfe_overflow()`, `_Xfe_underflow()` into xexp.cpp's unnamed namespace.
  + They obviously aren't dllexported.
* Replace `extern const` `_Xbig` and `_FXbig` with replicated `constexpr`.
  + They weren't dllexported. Although this results in a bit of code duplication, the result is more optimizer-friendly, and will allow us to eliminate xmath.hpp.
* Delete xmath.hpp, include only relevant headers.
* Fix dangerous `_Dval`, `_Fval` union sizes.
  + These are within an unnamed namespace, so we can alter their sizes. There was no actual harm here, since we never indexed into the nonexistent memory, but it was still risky, and likely to annoy static analysis someday.

Equivalence Proofs
===

`_Dtest` is `fpclassify`
---
```
C:\Temp>type meow.cpp
```
```cpp
#include <cmath>
#include <limits>
#include <print>
using namespace std;

static_assert(FP_SUBNORMAL == -2); // _DENORM
static_assert(FP_NORMAL == -1); // _FINITE
static_assert(FP_ZERO == 0);
static_assert(FP_INFINITE == 1); // _INFCODE
static_assert(FP_NAN == 2); // _NANCODE

extern "C" _CRTIMP2_PURE short __CLRCALL_PURE_OR_CDECL _Dtest(double*) noexcept;

void classify(double dbl) {
    const short stl = _Dtest(&dbl);
    const int crt   = fpclassify(dbl);
    println("dbl: {:6}; stl: {:2}; crt: {:2}; Equal: {}", dbl, stl, crt, stl == crt);
}

int main() {
    classify(numeric_limits<double>::denorm_min());
    classify(3.14);
    classify(0.0);
    classify(-0.0);
    classify(numeric_limits<double>::infinity());
    classify(numeric_limits<double>::quiet_NaN());
}
```
```
C:\Temp>cl /EHsc /nologo /W4 /std:c++latest /MTd /Od meow.cpp
meow.cpp

C:\Temp>meow
dbl: 5e-324; stl: -2; crt: -2; Equal: true
dbl:   3.14; stl: -1; crt: -1; Equal: true
dbl:      0; stl:  0; crt:  0; Equal: true
dbl:     -0; stl:  0; crt:  0; Equal: true
dbl:    inf; stl:  1; crt:  1; Equal: true
dbl:    nan; stl:  2; crt:  2; Equal: true
```

`DSIGN` is `signbit`
---
```
C:\Temp>type meow.cpp
```
```cpp
#include <cmath>
#include <limits>
#include <print>
using namespace std;

union _Dval {
    unsigned short _Sh[8];
    double _Val;
};
union _Fval {
    unsigned short _Sh[8];
    float _Val;
};

#define _D0 3
#define _F0 1

#define DSIGN(x) (reinterpret_cast<_Dval*>(&(x))->_Sh[_D0] & _DSIGN)
#define FSIGN(x) (reinterpret_cast<_Fval*>(&(x))->_Sh[_F0] & _FSIGN)

void test_dbl(double dbl) {
    const bool macro = static_cast<bool>(DSIGN(dbl));
    println("dbl: {:4}; macro: {:5}; signbit: {:5}; Equal: {}", dbl, macro, signbit(dbl), macro == signbit(dbl));
}

void test_flt(float flt) {
    const bool macro = static_cast<bool>(FSIGN(flt));
    println("flt: {:4}; macro: {:5}; signbit: {:5}; Equal: {}", flt, macro, signbit(flt), macro == signbit(flt));
}

int main() {
    test_dbl(numeric_limits<double>::infinity());
    test_dbl(-numeric_limits<double>::infinity());
    test_flt(numeric_limits<float>::infinity());
    test_flt(-numeric_limits<float>::infinity());
}
```
```
C:\Temp>cl /EHsc /nologo /W4 /std:c++latest /MTd /Od meow.cpp
meow.cpp

C:\Temp>meow
dbl:  inf; macro: false; signbit: false; Equal: true
dbl: -inf; macro: true ; signbit: true ; Equal: true
flt:  inf; macro: false; signbit: false; Equal: true
flt: -inf; macro: true ; signbit: true ; Equal: true
```

### DIFF
--- a/stl/src/xexp.cpp
+++ b/stl/src/xexp.cpp
@@ -11,9 +11,10 @@ namespace {
     constexpr int _D3 = 0;
 
     union _Dval { // pun floating type as integer array
-        unsigned short _Sh[8];
+        unsigned short _Sh[4];
         double _Val;
     };
+    static_assert(sizeof(_Dval::_Sh) == sizeof(_Dval::_Val), "_Dval members should be the same size");
 
     short _Dnorm(_Dval* ps) noexcept { // normalize double fraction
         short xchar         = 1;
@@ -112,9 +113,10 @@ namespace {
     constexpr int _F1 = 0;
 
     union _Fval { // pun floating type as integer array
-        unsigned short _Sh[8];
+        unsigned short _Sh[2];
         float _Val;
     };
+    static_assert(sizeof(_Fval::_Sh) == sizeof(_Fval::_Val), "_Fval members should be the same size");
 
     short _FDnorm(_Fval* ps) noexcept { // normalize float fraction
         short xchar         = 1;


stl/src/xvalues.cpp
```diff
@@ -9,13 +9,10 @@
 #endif
 #endif
 
-#include "xmath.hpp"
+#include <yvals.h>
 
 _EXTERN_C_UNLESS_PURE
 
-extern const double _Xbig = ((48 + _DOFF) + 2) * 0.347;
-extern const float _FXbig = ((16 + _FOFF) + 2) * 0.347f;
```
AlexGuteniev, line 17: The removal of these values seems the trickiest part to prove safety for me.

I do see that they are used from separately compiled main DLL code, so we shouldn't have any stale references to them.

StephanTLavavej: They weren't marked as `_CRTIMP2_PURE`, and they don't appear in `dumpbin /exports`. They were used only between STL TUs, but There Is Only One STL (DLL or LIB) so we never need to worry about mismatch.

Dynamically linking users couldn't access these at all. Statically linking users would have to manually declare them with their `_Ugly` names, forfeiting any claim to be righteous (and the values are meaningless to them, so nobody would try).

In contrast, the variables below are dllexported:

https://github.com/microsoft/STL/blob/7c0b1c40c48449f25ec7ad2195d61c394133123c/stl/src/xvalues.cpp#L26

stl/src/xcosh.cpp
```diff
@@ -1,16 +1,18 @@
...
 
 _CRTIMP2_PURE double __CLRCALL_PURE_OR_CDECL _Cosh(double x, double y) noexcept { // compute y * cosh(x), |y| <= 1
-    switch (_Dtest(&x)) { // test for special codes
-    case _NANCODE:
-    case _INFCODE:
+    switch (_STD fpclassify(x)) { // test for special codes
+    case FP_NAN:
+    case FP_INFINITE:
         return x;
-    case 0:
```
davidmrdavid, line 13: no changes requested, just sharing an ultra-subjective nit: when reviewing the granular commits, it would have been nice to combine the changes to the switch-case (like `_NANCODE` -> `FP_NAN`) with the change from `_Dtest` to `fpclassify`. It took me a minute to understand that the changes were related, though admittedly I was probably not fully awake by then.

Alas, that's probably on me. Thanks as always for the pristine git history :)

StephanTLavavej:

PR #5961, opened 2025-12-12
`<regex>`: Simplify unwinding

muellerj2:
Following #5828 and #5835, general stack unwinding is skipped when a regex (or an asserted pattern) matches successfully and matching does not follow the leftmost-longest rule. As a consequence, conditions based on `_Failed` or `_Longest` in the general unwinding loop have become superfluous, since they are always true. (For some opcodes, only `_Failed` was tested, but this was because they could not be generated in leftmost-longest mode.)

This removes all of these unnecessary conditions, replacing them by an initial assert to check that `_Failed || _Longest` is always true. Subsequently, we can use the same handler for `_After_assert` and `_Do_nothing` because no work is performed when backtracking over them. Similarly, the handler for `_Disjunction_eval_alt_on_failure` would immediately fall through to the handler of `_Disjunction_eval_alt_always`, so we can just fuse these two opcodes.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -1671,8 +1671,7 @@ public:
 enum class _Rx_unwind_ops {
     _After_assert = 1,
     _After_neg_assert,
-    _Disjunction_eval_alt_on_failure,
-    _Disjunction_eval_alt_always,
+    _Disjunction_eval_alternative,
     _Do_nothing,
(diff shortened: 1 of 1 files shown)


StephanTLavavej:

StephanTLavavej:
:yarn: :arrow_right: :thread:
===

PR #5970, opened 2025-12-20
`filesystem.cpp`: `__std_fs_change_permissions()` and `__std_fs_get_temp_path()` should avoid TOCTOU

SiliconA-Z:
Also GetFileAttributesW is redundant. We don't need it.


StephanTLavavej:
Can you please provide a more detailed explanation of the transformations you're making here? I'd like to believe that this is all behavior-preserving (except for avoiding the time-of-check / time-of-use issues), but with nontrivial transformations this can be hard to see. For example, I observe that `__std_fs_get_temp_path()` was doing something with reparse points. I think it was just trying to verify that if a directory is a reparse point, it can be opened, which is superseded by directly trying to open the temp directory, but an actual explanation from the author of the PR would help avoid problems during review (e.g. where your intent is X and bogus, I look at the code and think your intent was Y and fine, but if I knew the intent was X then I would realize it was bogus). Thanks!

StephanTLavavej:
Closing due to no response. TOCTOU is not a correctness issue for `<filesystem>` and the risk of destabilization is too great.

PR #5973, opened 2025-12-22
`<atomic>`: drop `_Atomic_reinterpret_as`

AlexGuteniev:
The goal of the change is mostly clarity. Bit cast is a vocabulary function.

`_Bit_cast` was invented too late, otherwise `_Atomic_reinterpret_as` should have been `_Bit_cast` all along.

In addition to size checks,  `_Atomic_reinterpret_as` also check for integrality, but that's an arbitrary requirement we don't really need to enforce.

Bit cast has some triviality check, which spotted unwrapped `_Storage_for` on lines 927 and 1062 (944 and 1080 before the change), fixed as a separate commit.

### DIFF
--- a/stl/inc/atomic
+++ b/stl/inc/atomic
@@ -718,8 +718,7 @@ struct _Atomic_storage<_Ty, 1> { // lock-free using 1-byte intrinsics
 
             for (;;) {
                 _ATOMIC_CHOOSE_INTRINSIC(static_cast<unsigned int>(_Order), _Prev_bytes, _InterlockedCompareExchange8,
-                    _STD _Atomic_address_as<char>(_Storage), _STD _Bit_cast<char>(_Desired),
-                    _Expected_bytes);
+                    _STD _Atomic_address_as<char>(_Storage), _STD _Bit_cast<char>(_Desired), _Expected_bytes);
                 if (_Prev_bytes == _Expected_bytes) {
                     return true;
                 }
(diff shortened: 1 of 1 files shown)


stl/inc/atomic
```diff
@@ -341,21 +341,6 @@ _NODISCARD inline memory_order _Combine_cas_memory_orders(
...
-template <class _Integral, class _Ty>
-_NODISCARD _Integral _Atomic_reinterpret_as(const _Ty& _Source) noexcept {
-    // interprets _Source as the supplied integral type
-    static_assert(is_integral_v<_Integral>, "Tried to reinterpret memory as non-integral");
-    if constexpr (is_integral_v<_Ty> && sizeof(_Integral) == sizeof(_Ty)) {
-        return static_cast<_Integral>(_Source);
-    } else if constexpr (is_pointer_v<_Ty> && sizeof(_Integral) == sizeof(_Ty)) {
-        return reinterpret_cast<_Integral>(_Source);
-    } else {
-        _Integral _Result{}; // zero padding bits
```
AlexGuteniev, line 353: This was misleading.

If `_Ty` has padding bits, this does zero the corresponding bits in `_Integral`, along with other bits.

But `memcpy` just below copies the values of these padding bits anyway.

AlexGuteniev: Hm, after thinking about it a bit I can see it.

The intent was to make some non-power-of-two sized struct handled here. Like `struct s { char a; char b; char c; };`.

"Padding bits" is a misnomer; these are actually bytes to complete to integral type.

For now and until vNext we don't really have lock-free atomics for like `struct s { char a; char b; char c; }`. We may reconsider the current change to preserve this part, or defer this to vNext.

BillyONeal: I admit to not having the best understanding on getting the compiler to zero out the correct bits but I believe I wrote tests that tried to introduce the infinite loop problem on bogus padding

StephanTLavavej: I am very happy to be switching to the requires-same-size `bit_cast` here, since this logic was wacky.

StephanTLavavej:

BillyONeal:
> `_Bit_cast` was invented too late, otherwise `_Atomic_reinterpret_as` should have been `_Bit_cast` all along.

I keep yelling at the engineers to invent me a time machine but I'm guessing they never succeed as they didn't deliver it to me yesterday

StephanTLavavej:
:zero: :one: :zero:
===

PR #5976, opened 2025-12-27
`filesystem.cpp`: `__std_fs_write_reparse_data_buffer()` should use the size of the buffer, not buffer pointer

SiliconA-Z:


StephanTLavavej:
This certainly appears to be a bug, after looking into how `FSCTL_GET_REPARSE_POINT` and `FSCTL_SET_REPARSE_POINT` work. This codepath is used in a single place, when copying a junction. I'll need to manually exercise this scenario (automated test coverage could potentially be problematic).

StephanTLavavej:
We don't think this fix is correct, `sizeof(*_Buffer)` contains a "fake" array element.

Morddin:
After a quick search on the internet, it seems that `sizeof(_Buffer)` should be replaced with `REPARSE_DATA_BUFFER_HEADER_SIZE`.

`REPARSE_DATA_BUFFER_HEADER_SIZE` is defined in the Windows Driver Kit as:
```c
#define REPARSE_DATA_BUFFER_HEADER_SIZE   UFIELD_OFFSET(REPARSE_DATA_BUFFER, GenericReparseBuffer)
```
Its value is 8.

StephanTLavavej:
Superseded by #6064, thanks everyone.

PR #5977, opened 2025-12-27
format: we need to include the last byte

SiliconA-Z:
We skip the last byte, which I guess happens to work for us, but is still improper.

### DIFF
--- a/stl/src/format.cpp
+++ b/stl/src/format.cpp
@@ -31,7 +31,7 @@ extern "C" [[nodiscard]] __std_win_error __stdcall __std_get_cvt(
             break;
         }
 
-        for (unsigned char _First = _Info.LeadByte[_Idx], _Last = _Info.LeadByte[_Idx + 1]; _First != _Last; ++_First) {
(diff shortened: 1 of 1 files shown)


cpplearner:
You already opened #5942. Why do you create this PR, which fixes the same problem in a different way?

StephanTLavavej:
Yes, can you please explain what's going on here?

Additionally, we will need test coverage that actually exercises this fix.

PRs are challenging to review when they're dropped on our doorstep with little explanation or context. We want to welcome PRs but we really need more info.

StephanTLavavej:
This does seem to address a real bug, according to [`CPINFOEXW` structure](https://learn.microsoft.com/en-us/windows/win32/api/winnls/ns-winnls-cpinfoexw) "Ranges are inclusive", and the structure of the fix looks reasonable at first glance, but we still need to figure out how to exercise this.

StephanTLavavej:
Superseded by #6059, thanks @cpplearner.

PR #5978, opened 2025-12-27
ios_base::_Addstd: need to check last byte

SiliconA-Z:
wclog is not checked, it seems to be forgotten about when _Addstd is called.


stl/src/ios.cpp
```diff
@@ -36,7 +36,7 @@ void __CLRCALL_PURE_OR_CDECL ios_base::_Ios_base_dtor(ios_base* _This) { // dest
 
 void __CLRCALL_PURE_OR_CDECL ios_base::_Addstd(ios_base* _This) { // add standard stream to destructor list
     _BEGIN_LOCK(_LOCK_STREAM)
-    for (_This->_Stdstr = 1; _This->_Stdstr < _Nstdstr; ++_This->_Stdstr) {
+    for (_This->_Stdstr = 1; _This->_Stdstr <= _Nstdstr; ++_This->_Stdstr) {
```
cpplearner, line 39: This change seems unnecessary and even misleading. `_This->_Stdstr` should never be larger than `_Nstdstr` after the loop, but with `<=` the loop structure suggests that it could be.

PR #5980, opened 2025-12-28
Vectorize `replace_copy` for x64 / x86 manually

### LINKED ISSUE #5908: Vectorize `replace_copy` for x64 / x86
Yes we have a low-hanging fruit in vectorization 🍎 

Some time ago I discovered that the compiler only needs some help to auto vectorize `replace_copy` and `replace_copy_if`. The help was very minor, compared to the couple of other places where we also help the compiler auto-vectorize.

So the PR #4431 was open, and it was merged with only mild reluctance. 

Shortly thereafter, the optimization ceased to work. I've filed DevCom-10895463. Unfortunately, it hasn't been fixed so far. The final VS 2022 ended up having `replace_copy` and `replace_copy_if` non-vectorized.

We cannot manually vectorize `replace_copy_if`, as it takes user's predicate, and there isn't a reasonable standard predicate to query against. (Well, actually there might be a way, but even if what I'm thinking of would work, you will not like it). Anyway, we can still vectorize `replace_copy`, and it is pretty easy thing to do.

Note that unlike assisted auto-vectorization, the manual vectorization condition has to be strict. Only contiguous iterators! And be sure to get the pointers properly, see #5683 and the linked issue and paper. Also a new approach to control macro should be used, but this part is hard to miss.

Unlike `replace`, it doesn't need to rely on AVX2 masked stores, it is classic byte blending, so it can work on any element size.

AlexGuteniev:
Fix #5908

The auto vectorization recipe applied in #4431 did not work in 17.14. 
Another recipe would have worked, but it does not work in the latest Insiders.
So chasing the compiler auto vectorization here appears to be unproductive.

Do not help the compiler :tm:

**Benchmark results**

Benchmark             | Before  |  After 
----------------------|---------|---------
`rc<std::uint8_t>`    |  917 ns | 58.4 ns 
`rc<std::uint16_t>`   | 2074 ns |  109 ns 
`rc<std::uint32_t>`   | 1299 ns |  207 ns 
`rc<std::uint64_t>`   |  930 ns |  548 ns

### DIFF
--- a/tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
+++ b/tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
@@ -765,8 +765,11 @@ void test_case_replace_copy(const vector<T>& input, vector<T>& out_expected, vec
 #endif // ^^^ !_HAS_CXX20 ^^^
 }
 
-template <class T, bool in_place>
+template <class T>
 void test_replace(mt19937_64& gen) {
+    // replace() is vectorized for 4 and 8 bytes only.
+    constexpr bool replace_is_vectorized = sizeof(T) >= 4;
+
     using TD = conditional_t<sizeof(T) == 1, int, T>;
     uniform_int_distribution<TD> dis(0, 9);
 
@@ -782,7 +785,7 @@ void test_replace(mt19937_64& gen) {
         v->reserve(dataCount);
     }
 
-    if constexpr (in_place) {
+    if constexpr (replace_is_vectorized) {
         for (const auto& v : {&in_out_expected, &in_out_actual, &in_out_actual_r}) {
             v->reserve(dataCount);
         }
@@ -792,7 +795,7 @@ void test_replace(mt19937_64& gen) {
         const T old_val = static_cast<T>(dis(gen));
         const T new_val = static_cast<T>(dis(gen));
 
-        if constexpr (in_place) {
+        if constexpr (replace_is_vectorized) {
             test_case_replace(in_out_expected, in_out_actual, in_out_actual_r, old_val, new_val);
         }
         test_case_replace_copy(source, out_expected, out_actual, out_actual_r, old_val, new_val);
@@ -811,7 +814,7 @@ void test_replace(mt19937_64& gen) {
             v->assign(source.size(), T{0});
         }
 
-        if constexpr (in_place) {
+        if constexpr (replace_is_vectorized) {
             test_case_replace(in_out_expected, in_out_actual, in_out_actual_r, old_val, new_val);
         }
         test_case_replace_copy(source, out_expected, out_actual, out_actual_r, old_val, new_val);
@@ -1345,16 +1348,15 @@ void test_vector_algorithms(mt19937_64& gen) {
     test_includes<unsigned long long>(gen);
 #endif // _HAS_CXX17
 
-    // replace() is vectorized for 4 and 8 bytes only.
-    test_replace<char, false>(gen);
-    test_replace<signed char, false>(gen);
-    test_replace<unsigned char, false>(gen);
-    test_replace<short, false>(gen);
-    test_replace<unsigned short, false>(gen);
-    test_replace<int, true>(gen);
-    test_replace<unsigned int, true>(gen);
-    test_replace<long long, true>(gen);
-    test_replace<unsigned long long, true>(gen);
+    test_replace<char>(gen);
+    test_replace<signed char>(gen);
+    test_replace<unsigned char>(gen);
+    test_replace<short>(gen);
+    test_replace<unsigned short>(gen);
+    test_replace<int>(gen);
+    test_replace<unsigned int>(gen);
+    test_replace<long long>(gen);
+    test_replace<unsigned long long>(gen);
 
     test_reverse<char>(gen);
     test_reverse<signed char>(gen);


stl/inc/algorithm
```diff
@@ -358,6 +369,28 @@ __declspec(noalias) void _Replace_vectorized(
 }
 #endif // ^^^ _VECTORIZED_REPLACE ^^^
 
+#if _VECTORIZED_REPLACE_COPY
+template <class _Ty, class _TVal1, class _TVal2>
+__declspec(noalias) void _Replace_copy_vectorized(const _Ty* const _First, const _Ty* const _Last, _Ty* _Dest,
```
StephanTLavavej: `_Dest` can be top-level const.

stl/inc/algorithm
```diff
@@ -476,6 +509,17 @@ constexpr bool _Output_iterator_for_vector_alg_is_safe() {
...
+#if _VECTORIZED_REPLACE_COPY
+template <class _Out, class _In>
+constexpr bool _Output_iterator_for_known_size_vector_alg_is_safe() {
+    if constexpr (_Iterator_is_contiguous<_Out>) {
+        return is_same_v<_Iter_value_t<_Out>, remove_const_t<_Iter_value_t<_In>>>;
+    } else {
+        return false;
+    }
+}
+#endif // ^^^ _VECTORIZED_REPLACE_COPY
```
StephanTLavavej: Missing arrows.

stl/inc/algorithm
```diff
@@ -4579,15 +4614,36 @@ _CONSTEXPR20 _OutIt replace_copy(_InIt _First, _InIt _Last, _OutIt _Dest, const
...
-            if (*_UFirst == _Oldval) {
-                *_UDest = _Newval;
+
+#if _VECTORIZED_REPLACE_COPY
+    if constexpr (_Vector_alg_in_find_is_safe<decltype(_UFirst), _Ty>
+                  && _Output_iterator_for_known_size_vector_alg_is_safe<decltype(_UDest), decltype(_UFirst)>()) {
+        if (!_STD _Is_constant_evaluated()) {
+            const auto _First_ptr = _STD _To_address(_UFirst);
+            const auto _Dest_ptr  = _STD _To_address(_UDest);
+            const auto _Last_ptr  = _STD _To_address(_ULast);
```
StephanTLavavej, line 4626: Style: The order first, dest, last is inconsistent and surprising.

stl/inc/algorithm
```diff
@@ -4579,15 +4614,36 @@ _CONSTEXPR20 _OutIt replace_copy(_InIt _First, _InIt _Last, _OutIt _Dest, const
...
+            const auto _Count     = _ULast - _UFirst;
+
+            _UDest += static_cast<_Iter_diff_t<decltype(_UDest)>>(_Count);
+            (void) _STD _To_address(_UDest);
```
StephanTLavavej: This doesn't follow the precedent of `reverse_copy`.

stl/src/vector_algorithms.cpp
```diff
@@ -6486,6 +6486,80 @@ __declspec(noalias) size_t __stdcall __std_mismatch_8(
     return _Mismatching::_Mismatch_impl<uint64_t>(_First1, _First2, _Count);
 }
 
+} // extern "C"
+
+namespace {
+    namespace _Replacing {
+        template <class _Traits, class _Ty>
+        void __stdcall _Replace_copy_impl(
```
StephanTLavavej: Missing `__declspec(noalias)`.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -1288,10 +1346,15 @@ void test_vector_algorithms(mt19937_64& gen) {
 #endif // _HAS_CXX17
 
     // replace() is vectorized for 4 and 8 bytes only.
-    test_replace<int>(gen);
-    test_replace<unsigned int>(gen);
-    test_replace<long long>(gen);
-    test_replace<unsigned long long>(gen);
+    test_replace<char, false>(gen);
```
StephanTLavavej: Instead of passing a bool, let's just determine it from the size.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -723,40 +723,98 @@ void last_known_good_replace(FwdIt first, FwdIt last, const T old_val, const T n
...
+    replace_copy(input.begin(), input.end(), out_actual.begin(), old_val, new_val);
+    last_known_good_replace_copy(input.begin(), input.end(), out_expected.begin(), old_val, new_val);
+    assert(out_expected == out_actual);
+
+#if _HAS_CXX20
+    ranges::replace_copy(input, out_actual_r.begin(), old_val, new_val);
+    assert(out_expected == out_actual_r);
+#else // ^^^ _HAS_CXX20 / !_HAS_CXX20 vvv
+    (void) out_actual_r;
+#endif // ^^^ !_HAS_CXX20 ^^
```
StephanTLavavej: Missing arrow.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -723,40 +723,98 @@ void last_known_good_replace(FwdIt first, FwdIt last, const T old_val, const T n
...
+}
+
+template <class T, bool in_place>
```
StephanTLavavej: This shadows `std::in_place`.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: I pushed significant changes, please double-check.

AlexGuteniev:
All good

StephanTLavavej:

StephanTLavavej:
:cat: :heart_eyes_cat: :smile_cat:
===

PR #5986, opened 2026-01-04
Add test coverage for `<flat_set>` to `test_header_units_and_modules.hpp`

vmichal:
Add `<flat_set>` test coverage to `test/std/include/test_header_units_and_modules.hpp`.
Test is consistent with previously existing `<flat_map>` test.

Fixes (see https://discord.com/channels/737189251069771789/1120873380615295027/1450952672982204468)
>[...]. One of FIXMEs is in tests/std/include/test_header_units_and_modules.hpp where we need some coverage of the adaptors (if you look at the other coverage, it intentionally doesn't attempt to exercise everything, just a reasonable test of basic usage so we'll know if modules catastrophically fail to work with the code).

### DIFF
--- a/tests/std/include/test_header_units_and_modules.hpp
+++ b/tests/std/include/test_header_units_and_modules.hpp
@@ -281,12 +281,12 @@ void test_flat_set() {
     assert(erase_if(fs, simple_truth) == 1);
     assert(fs.empty());
 
-    flat_multiset<int> fss;
-    fss.emplace(42);
-    fss.emplace(42);
-    assert(fss.size() == 2);
-    assert(erase_if(fss, simple_truth) == 2);
-    assert(fss.empty());
+    flat_multiset<int> fms;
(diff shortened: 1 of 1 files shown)


tests/std/include/test_header_units_and_modules.hpp
```diff
@@ -269,7 +269,24 @@ void test_flat_set() {
...
+    constexpr auto simple_truth = [](const auto&) { return true; };
+
+    flat_set<int> fs;
+    fs.emplace(42);
+    fs.emplace(42);
+    assert(fs.size() == 1);
+    assert(erase_if(fs, simple_truth) == 1);
+    assert(fs.empty());
+
+    flat_multiset<int> fss;
```
StephanTLavavej: Naming nitpick: `fss` doesn't make sense, it should be `fms` for Flat Multi Set.

PR #5987, opened 2026-01-04
`<flat_map>`: Simplify duplicate definitions of lookup functions using explicit object parameter

vmichal:
Deduplicate definitions of _Ugly-named lookup functions in `<flat_map>` by deducing `this`.

- Reduce duplication in implementations of `_Lower_bound`, `_Upper_bound`, `_Equal_range` and `_Find` in `_Flat_map_base`.
- Deduplicate implementations of `flat_map::_At` (not present in `_Flat_map_base`).
- Merge implementation of `_Lower_bound` and `_Upper_bound` into a single generic function and passing `_RANGES lower_bound` or `_RANGES upper_bound` as arguments.
- Add sanity check tests that user-facing functions (`lower_bound` etc) can still be called on a temporary object (even though the returned iterator would dangle).

- [ ] Maybe reconsider whether it is appropriate to only use lvalue ref instead of forwarding ref for the deduced this. It should be OK since these _Ugly-named member functions are only called from other member functions and never on the object directly, so this breaks no user code. This is demonstrated by tests.

Followup on https://discord.com/channels/737189251069771789/1120873380615295027/1457169557033849009 , feel free to reject if too bold.

### DIFF
--- a/tests/std/tests/P0429R9_flat_map/test.cpp
+++ b/tests/std/tests/P0429R9_flat_map/test.cpp
@@ -766,7 +766,7 @@ void test_throwing_compare_swap() {
     test_throwing_compare_swap_single<flat_multimap, deque, deque>();
 }
 
-// Test that changes in GH-5987 did not break calls of lookup member functions by using deduced this.
+// Test that changes in GH-5987 did not break calls of lookup member functions by using deducing this.
 template <typename T>
 void test_lookup_call_on_temporaries_single() {
     (void) T{}.lower_bound(42);


stl/inc/flat_map
```diff
@@ -1133,74 +1122,38 @@ private:
...
-        const auto _Key_unchecked_begin = _STD _Get_unwrapped(_STD cbegin(_Data.keys));
-        const auto _Key_unchecked_it    = _STD upper_bound(
-               _Key_unchecked_begin, _STD _Get_unwrapped(_STD cend(_Data.keys)), _Key_val, _Pass_key_comp());
-        const auto _Dist = _Key_unchecked_it - _Key_unchecked_begin;
-
-        auto _Key_it = _STD cbegin(_Data.keys);
-        _STD _Seek_wrapped(_Key_it, _Key_unchecked_it);
-        auto _Val_it = _Data.values.begin() + static_cast<_RANGES range_difference_t<_MappedContainer>>(_Dist);
-        return iterator{_STD move(_Key_it), _STD move(_Val_it)};
+        using return_type =
```
StephanTLavavej: `return_type` isn't ugly.

stl/inc/flat_map
```diff
@@ -1133,74 +1122,38 @@ private:
...
+            _STD conditional_t<_STD is_const_v<_STD remove_reference_t<_SelfTy>>, const_iterator, iterator>;
```
StephanTLavavej: `_STD` qualification is unnecessary for these type aliases and variable templates, which can't be hijacked by ADL.

stl/inc/flat_map
```diff
@@ -1133,74 +1122,38 @@ private:
...
+    _NODISCARD auto _Upper_bound(this _SelfTy& _Self, const _KeyTy& _Key_val) {
+        return _Self._Find_bound(_Key_val, _RANGES upper_bound);
     }
 
     template <class _KeyTy>
-    _NODISCARD pair<const_iterator, const_iterator> _Equal_range(const _KeyTy& _Key_val) const {
+    _NODISCARD auto _Equal_range(this auto& _Self, const _KeyTy& _Key_val) {
         _STL_INTERNAL_STATIC_ASSERT(is_same_v<_KeyTy, key_type> || _Transparent<key_compare>);
-        return {lower_bound(_Key_val), upper_bound(_Key_val)};
+        return _STD make_pair(_Self.lower_bound(_Key_val), _Self.upper_bound(_Key_val));
```
StephanTLavavej: `make_pair()` is an old-style helper that's almost always unnecessary in the modern age of CTAD. (It introduces a function call in debug mode so it's worth avoiding.) In fact, not even our old product code ever calls `make_pair()`.

(There's an interaction with `reference_wrapper` which is not relevant to the iterators involved here.)

stl/inc/flat_map
```diff
@@ -1100,24 +1100,13 @@ private:
     }
 
     template <class _KeyTy>
-    _NODISCARD iterator _Find(const _KeyTy& _Key_val) {
+    _NODISCARD auto _Find(this auto& _Self, const _KeyTy& _Key_val) {
```
StephanTLavavej: This is an abbreviated function template, which we have (so far) avoided using in product code. Compared to "deducing this" which saves a ton of code, using `auto` to avoid naming a template parameter isn't a major savings, and conceals the template-ness somewhat. I don't think we should begin using this feature as part of this change.

stl/inc/flat_map
```diff
@@ -1133,74 +1122,38 @@ private:
...
-        auto _Key_it = _STD cbegin(_Data.keys);
-        _STD _Seek_wrapped(_Key_it, _Key_unchecked_it);
-        auto _Val_it = _Data.values.begin() + static_cast<_RANGES range_difference_t<_MappedContainer>>(_Dist);
-        return iterator{_STD move(_Key_it), _STD move(_Val_it)};
-    }
-
-    template <class _KeyTy>
-    _NODISCARD const_iterator _Lower_bound(const _KeyTy& _Key_val) const {
+    template <class _KeyTy, class _SelfTy, class _BoundFn>
+    _NODISCARD auto _Find_bound(this _SelfTy& _Self, const _KeyTy& _Key_val, _BoundFn _Bound_fun) {
```
StephanTLavavej: We generally list template parameters in the order of their appearance, unless there's a reason (e.g. explicit template args passed elsewhere) to do otherwise.

stl/inc/flat_map
StephanTLavavej: We're taking `this _SelfTy& _Self` which isn't perfect forwarding, it's just a plain lvalue reference. Therefore `_SelfTy` cannot itself be a reference, so we don't need to `remove_reference_t`.

tests/std/tests/P0429R9_flat_map/test.cpp
```diff
@@ -766,6 +766,29 @@ void test_throwing_compare_swap() {
     test_throwing_compare_swap_single<flat_multimap, deque, deque>();
 }
 
+// Test that changes in GH-5987 did not break calls of lookup member functions by using deduced this.
```
StephanTLavavej: The feature is conventionally called "deducing this".

PR #5989, opened 2026-01-04
`<flat_map>`: Unify implementation of `emplace` and `insert` between `flat_map` and `flat_multimap`

vmichal:
- Unify implementation of `emplace` and `insert` family of functions, unify helper function.
- Implement `flat_(multi)map::operator=(initializer_list)` consistently with flat_set, i.e. in parent class.
- Perform some small cleanups in `flat_map`, `flat_multimap`, `flat_set` and `flat_multiset` - remove unnecessary `using _Mybase::xxx` declarations and unify structure of four classes.

Towards #2910 #2912

### DIFF
--- a/stl/inc/flat_map
+++ b/stl/inc/flat_map
@@ -33,30 +33,6 @@ _STL_DISABLE_CLANG_WARNINGS
 #undef msvc
 
 _STD_BEGIN
-template <class _Key, class _Mapped, class _KeyCompare>
-struct _Flat_map_value_compare_provider {
-    struct value_compare {
-    public:
-        _NODISCARD bool operator()(
-            pair<const _Key&, const _Mapped&> _Left, pair<const _Key&, const _Mapped&> _Right) const {
-            return _Key_comparator(_Left.first, _Right.first);
-        }
-
-        value_compare(_KeyCompare _Comp) : _Key_comparator(_Comp) {}
-
-    private:
-        _KeyCompare _Key_comparator;
-    };
-};
-
-template <class _KeyContainer, class _MappedContainer>
-struct _Flat_map_container_provider {
-    struct containers {
-        _KeyContainer keys;
-        _MappedContainer values;
-    };
-};
-
 // Implementation
 
 template <bool _IsUnique, class _Key, class _Mapped, class _Compare, class _KeyContainer, class _MappedContainer>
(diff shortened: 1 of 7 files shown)


stl/inc/flat_map
```diff
@@ -1209,21 +1262,18 @@ class flat_map : public _Flat_map_base<true, _Key, _Mapped, _Compare, _KeyContai
 private:
     using _Mybase = _Flat_map_base<true, _Key, _Mapped, _Compare, _KeyContainer, _MappedContainer>;
 
-    using _Mybase::_Data;
-    using _Mybase::_Key_compare;
```
frederick-vs-ja, line 1155: These were intentional. They were originally protected in the base classes and I intentionally made them private in the derived classes. Do you think this is a bad idea?

vmichal: After merging the bodies of `flat_map::_Try_emplace` and `flat_multimap::_Emplace_key_mapped` and moving the result into the base class, I've found no use for either of these names in either derived class.

What was your original intent with these lines, except for the obvious "bring them to scope" for previously existing functions? I strived to eliminate as much code as reasonable 😅 .

frederick-vs-ja: > What was your original intent with these lines, except for the obvious "bring them to scope" for previously existing functions?

I hope I made the previous comment clear enough such that it was not talking about functions. For `_Data` and `_Key_compare`, the only intent was making them private.

vmichal: > I hope I made the previous comment clear enough such that it was not talking about functions. For `_Data` and `_Key_compare`, the only intent was making them private.

Yes, you did; I have not expressed myself clearly.
I meant to say that I assumed those two `using` declarations were present only to bring names `_Data` and `_Key_compare` to scope for use in member functions `flat_map::_Try_emplace` and `flat_multimap::_Emplace_key_mapped`, hence I removed them after the refactoring, when they were no longer needed.

So the intent is to prevent users from misusing the container adaptors through inheritance?
I will revert those changes. if you'd like.

frederick-vs-ja: > So the intent is to prevent users from misusing the container adaptors through inheritance?

Yes. But I'm not sure whether it would be fine or conventional to expose protected _Ugly members. I'd like leave this to maintainer(s).

StephanTLavavej: Ideally, things should be `private`. If something needs to be `protected` for our derived classes, I'm not going to worry about user inheritance being able to see it, it's still `_Ugly`. The main thing is to avoid `public` which will lead to IDEs suggesting it to ordinary users.

StephanTLavavej (approved):

StephanTLavavej:
Resolved trivial adjacent-edit merge conflict, thanks!

PR #5990, opened 2026-01-04
`<flat_meow>`: Misc changes: `const` locals and parameters, more `_STL_ASSERT` that functions with `_Sorted` tag are called with sorted sequences

### LINKED ISSUE #5991: `<flat_meow>`: Should move constructors use the `extract() &&` member function?
@StephanTLavavej  mentioned at https://discord.com/channels/737189251069771789/1120873380615295027/1450952672982204468 
> My note to self: "If we change flat_set's move assign to use extract() like flat_map does, libcxx\test\std\containers\container.adaptors\flat.multiset\flat.multiset.cons\move_assign.pass.cpp fails because they have a squirrelly container that has (implicitly) deleted copy/move ctors. I think this fails the sequence container requirements so the test is bogus." We could leave this as-is, but it might be worth investigating upstream.

I'm not sure that such change is even desirable. Using the pattern `_STD move(_Other).extract();` seems reasonable when constructing a new object due to guaranteed copy elision, but it constructs a unnecessary temporary when used in operator=. Although extract solves the self-assignment problem, I believe `if (this != _STD addressof(_Other))` is at least equally appropriate.

Please, see a minimal example https://godbolt.org/z/7744YMxGb

If you agree, this is already speculatively implemented in #5990 .

vmichal:
- Audit local variables and function arguments in `<flat_map>` and `<flat_set>`, add `const` wherever possible. An exception is made for functions touched by #5989 and #5987 to prevent conflicts during merging.
- Guard `_Flat_map_base` against self-move-assignment (consistent with `_Flat_set_base` implementation) by `if (this != _STD addressof(_Other) { ... }`. Removes previous solution with `extract()` , see #5991 
- Sprinkle the `_Flat_map_base` with `_STL_ASSERT`s to check that functions with `_Sorted` tag type are invoked with sorted sequences, as per [flat.map.syn]/10 and similar.
- Minor updates to `<flat_meow>` tests

fixes #5991

### DIFF
--- a/stl/inc/flat_map
+++ b/stl/inc/flat_map
@@ -1135,9 +1135,10 @@ private:
             _Sorted_new_elements.advance(_Old_distance);
             _RANGES sort(_Sorted_new_elements, value_compare{_Key_compare});
         } else {
-            [[maybe_unused]] auto const _New_elements_begin =
-                _Data.keys.begin() + static_cast<key_container_type::difference_type>(_Old_distance);
-            _STL_ASSERT(_Is_sorted_and_unique(_New_elements_begin, _Data.keys.end()), _Msg_not_sorted);
+            _STL_ASSERT(_Is_sorted_and_unique(
+                            _Data.keys.begin() + static_cast<key_container_type::difference_type>(_Old_distance),
+                            _Data.keys.end()),
+                _Msg_not_sorted);
         }
 
         // Merge the newly inserted elements with the existing elements


stl/inc/flat_map
```diff
@@ -1080,13 +1091,17 @@ private:
             auto _Sorted_new_elements = _Sorted_view;
             _Sorted_new_elements.advance(_Old_distance);
             _RANGES sort(_Sorted_new_elements, value_compare{_Key_compare});
+        } else {
+            [[maybe_unused]] auto const _New_elements_begin =
```
StephanTLavavej: Should be west const. But actually, this is being thrown away in release mode; the named variable isn't worth it.

StephanTLavavej (approved):

PR #5993, opened 2026-01-06
Fix no-match handling when calling `memchr`/`wmemchr` on ARM64EC

hazzlim:


hazzlim:
Looks like https://github.com/microsoft/STL/pull/5597 was slightly incorrect.

I haven't tested on ARM64EC, as it requires quite a lot of changes to switch these paths on - but I have verified that this produces the correct result on ARM64.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for noticing and fixing this bug! :lady_beetle: :eagle: :eye:

PR #5994, opened 2026-01-07
`<flat_meow>`: Don't do extra work in `count()`, `contains()`, and `equal_range()`

### LINKED ISSUE #5992: `<flat_map>`: `count()`, `contains()`, and `equal_range()` are doing too much work
(@vmichal's #5987 is changing this code, but the issue is pre-existing and orthogonal. Update these citations after merging that PR.)

https://github.com/microsoft/STL/blob/18e1eb5360e9179104f6a03d9b663c2a38083af4/stl/inc/flat_map#L1124-L1134
https://github.com/microsoft/STL/blob/18e1eb5360e9179104f6a03d9b663c2a38083af4/stl/inc/flat_map#L1194-L1198

First, the implementations of `count()` and `contains()` are forming full key-value iterators, but they only need to care about the keys.

Second, `count()` and `equal_range()` are calling `lower_bound()` and `upper_bound()` independently, which performs two binary searches to get to the same general region. Going through `ranges::equal_range` would perform less work.

Third, `count()` should probably be optimized for non-multi `flat_map`, as it becomes morally equivalent to `contains()`.

vmichal:
Please note that there is some ongoing conversation and unanswered questions in #5992, so this is likely not the final state of this PR.

Optimize lookups in non-multi containers:
- `count()` is the same as `contains()`  when searched key is not generic
- Optimize `equal_range()` when its length is at most one and searched key is not generic
- Don't construct full key-value iterators in `contains()`
- Delay construction of full key-value iterators in `_Flat_map_base::_Find()`
- Shrink bounds for binary search in `_Flat_map_base::_Emplace_hint`
- Introduce helper `_Flat_map_base::_Iterator_from_key_iterator`  for simpler conversion `key_container_type::(const_)iterator` -> `flat_map::(const_)iterator`

- Replace all calls to `_RANGES::lower_bound` and similar with `_STD` versions (see discussion https://discord.com/channels/737189251069771789/832305953562427513/1458509280184369246)

Resolves #5992

### DIFF
--- a/stl/inc/flat_map
+++ b/stl/inc/flat_map
@@ -1204,7 +1204,7 @@ private:
     _NODISCARD auto _Equal_range(this _SelfTy& _Self, const _KeyTy& _Key_val) {
         _STL_INTERNAL_STATIC_ASSERT(is_same_v<_KeyTy, key_type> || _Transparent<key_compare>);
         if constexpr (_IsUnique && is_same_v<_KeyTy, key_type>) { // Optimization restricted due to GH-5992
-            // In a non-multi container, equal range can have size at most 1
+            // In a non-multi container, equal_range can have size at most 1
             const auto _First =
                 _STD lower_bound(_Self._Data.keys.begin(), _Self._Data.keys.end(), _Key_val, _Self._Pass_key_comp());
             const bool _Missing = _First == _STD end(_Self._Data.keys) || _Self._Key_compare(_Key_val, *_First);
--- a/stl/inc/flat_set
+++ b/stl/inc/flat_set
@@ -692,7 +692,7 @@ private:
     _NODISCARD pair<const_iterator, const_iterator> _Equal_range(const _KeyTy& _Key_val) const {
         _STL_INTERNAL_STATIC_ASSERT(is_same_v<_KeyTy, key_type> || _Transparent<key_compare>);
         if constexpr (_IsUnique && is_same_v<_KeyTy, key_type>) { // Optimization restricted due to GH-5992
-            // In a non-multi container, equal range can have size at most 1
+            // In a non-multi container, equal_range can have size at most 1
             const auto _First   = _STD lower_bound(_Mycont.begin(), _Mycont.end(), _Key_val, _Pass_comp());
             const bool _Missing = _First == _STD end(_Mycont) || _Mycomp(_Key_val, *_First);
             return pair{_First, _Missing ? _First : _STD next(_First)};


stl/inc/flat_map
```diff
@@ -1172,46 +1173,55 @@ private:
     template <class _KeyTy>
     _NODISCARD size_type _Count(const _KeyTy& _Key_val) const {
         _STL_INTERNAL_STATIC_ASSERT(is_same_v<_KeyTy, key_type> || _Transparent<key_compare>);
-        return upper_bound(_Key_val) - lower_bound(_Key_val);
+        if constexpr (_IsUnique && is_same_v<_KeyTy, key_type>) { // Optimization restricted due to GH-5592
```
StephanTLavavej: The citation of #5592 is incorrect; you meant #5992.

stl/inc/flat_map
```diff
@@ -1172,46 +1173,55 @@ private:
...
+            _STD upper_bound(_Self._Data.keys.begin(), _Self._Data.keys.end(), _Key_val, _Self._Pass_key_comp()));
     }
 
     template <class _SelfTy, class _KeyTy>
     _NODISCARD auto _Equal_range(this _SelfTy& _Self, const _KeyTy& _Key_val) {
         _STL_INTERNAL_STATIC_ASSERT(is_same_v<_KeyTy, key_type> || _Transparent<key_compare>);
-        return pair{_Self.lower_bound(_Key_val), _Self.upper_bound(_Key_val)};
+        if constexpr (_IsUnique && is_same_v<_KeyTy, key_type>) { // Optimization restricted due to GH-5592
+            // In a non-multi container, equal range can have size at most 1
+            auto const _First =
```
StephanTLavavej: Style: We use west const.

stl/inc/flat_map
```diff
@@ -1172,46 +1173,55 @@ private:
...
+        } else {
+            auto const [_First, _Last] =
+                _STD equal_range(_Self._Data.keys.begin(), _Self._Data.keys.end(), _Key_val, _Self._Pass_key_comp());
+            return pair{_Self._Iterator_from_key_iterator(_First), _Self._Iterator_from_key_iterator(_Last)};
+        }
+    }
+
+    template <class _SelfTy>
+    _NODISCARD conditional_t<is_const_v<_SelfTy>, const_iterator, iterator> _Iterator_from_key_iterator(
+        this _SelfTy& _Self, const typename key_container_type::const_iterator& _Iter) {
```
StephanTLavavej: Member function parameters are a Down With Typename context.

StephanTLavavej (approved):

StephanTLavavej:
:tada:
===

PR #5995, opened 2026-01-07
`<flat_map>`: Minor cleanups of iterator operations in `flat_map` members `_Erase_if` etc.

vmichal:
Minor cleanups without changing meaning - removing single-use local variables and cleanups in iterator arithmetic.


vmichal:
Ah, incorrect target branch 😢

StephanTLavavej:
I've made that mistake too! :joy_cat:

PR #5997, opened 2026-01-07
STL.natvis: Add `c_str` intrinsic function

gregg-miskelly:
This PR adds a `c_str` intrinsic function to std::basic_string so that it is easier to compare the value of an std::string in a conditional breakpoint. Ideally, we would also include `operator==` and `operator!=` implementations, but CppDebug doesn't have a good way of implementing that currently, but this is still a useful start.

Example of comparing an std::string too a string literal in a conditional breakpoints with this change: `strcmp(myStdStringVariable.c_str(), "some_literal") == 0`


StephanTLavavej:
@gregg-miskelly Thanks for triple-mirroring this as VS-PR-698144! I'll get this merged to GitHub and MSVC in my next merge batch.

StephanTLavavej:

StephanTLavavej:
Thanks for improving one of our most important visualizers, and congratulations on your first microsoft/STL commit! :smile_cat:

:red_square::green_square:
:blue_square::yellow_square:
===

PR #6001, opened 2026-01-09
`<flat_meow>`: Minor cleanups - Use member `begin`/`end` when possible, use arithmetic for random access iterators

vmichal:
Implement changes discussed here https://discord.com/channels/737189251069771789/832305953562427513/1458503204546543739

- Since `<flat_meow>` adaptors require underlying containers to provide random access iterators, we don't need calls to `distance`, `next`, `prev` and similar - replace them with arithmetic operators.
- Since key and mapped containers are required to fulfill container requirements, use member functions for `begin`, `end` etc.

- At the moment, some libcxx tests have been FAILed since they use `MinSequenceContainer` that lacks member `cend()`. I have opened PR to LLVM-175114 that will fix it.


tests/libcxx/expected_results.txt
```diff
@@ -204,6 +204,19 @@ std/containers/views/mdspan/layout_right/ctor.layout_stride.pass.cpp FAIL
 # If any feature-test macro test is failing, this consolidated test will also fail.
 std/language.support/support.limits/support.limits.general/version.version.compile.pass.cpp FAIL
 
+# Test uses MinSequenceContainer that does not provide `cend()` member function (LLVM-175114)
```
frederick-vs-ja: I think it's better to move these to the "LIKELY BOGUS TESTS" section.

StephanTLavavej: This is actually the correct section as an upstream PR has been created (and now merged, https://github.com/llvm/llvm-project/pull/175114 ). It should actually appear in sorted order though.

stl/inc/flat_map
```diff
@@ -526,21 +526,21 @@ public:
     // [container.reqmts] iterators
     _NODISCARD iterator begin() noexcept {
         _STL_INTERNAL_STATIC_ASSERT(random_access_iterator<iterator>);
-        return iterator{_STD cbegin(_Data.keys), _Data.values.begin()};
+        return iterator{_Data.keys.begin(), _Data.values.begin()};
```
StephanTLavavej: This is changing `cbegin` to `begin`.

StephanTLavavej:
Blocked on merging #6006 to `main` and then `feature/flat_map` which I should be able to do by Tue/Wed. I would prefer to not lose significant test coverage.

StephanTLavavej:
:smile_cat:
===

PR #6002, opened 2026-01-09
Improve test diagnostics when `expected_results.txt` contains syntax errors

vmichal:
When a formatting error is encountered while parsing `expected_results.txt`, report the line number as well.

The situation notably improved in #5948, but there is still a tiny improvement to be made.

### DIFF
--- a/tools/scripts/check_libcxx_paths.py
+++ b/tools/scripts/check_libcxx_paths.py
@@ -21,16 +21,49 @@ if __name__ == "__main__":
         stripped_lines = [line.strip() for line in file]
 
     filtered_lines = [s for s in stripped_lines if s and not s.startswith("#")] # Ignore empty lines and comments.
-    config_rgx = re.compile(r"(:\d+)? (FAIL|SKIPPED)$")
-    unique_tests = {config_rgx.sub("", line) for line in filtered_lines}
+
+    damaged_lines = set()
+    unknown_results = set()
+    unique_tests = set()
+
+    syntax_rgx = re.compile(r"(?P<test>[^: ]+)(?P<config>:\d+)? (?P<result>\w+)")
+    for line in filtered_lines:
+        m = syntax_rgx.fullmatch(line)
+        if m is None:
+            damaged_lines.add(line)
+        else:
+            result = m.group("result")
+            if result == "FAIL" or result == "SKIPPED":
+                unique_tests.add(m.group("test"))
+            else:
+                unknown_results.add(line)
 
     # Build up a list of nonexistent tests so they can be printed in sorted order.
     absolute_libcxx_test = absolute_repo_path / "llvm-project/libcxx/test"
     nonexistent_tests = [s for s in unique_tests if not (absolute_libcxx_test / s).is_file()]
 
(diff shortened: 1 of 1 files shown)


tests/utils/stl/test/file_parsing.py
```diff
@@ -101,7 +101,7 @@ def _do_crosslist(ctx: _ParseCtx):
 
 
 def _parse_env_lst(env_lst: Path, ctx: _ParseCtx):
-    for line in parse_commented_file(env_lst):
+    for line_number, line in _parse_commented_file(env_lst):
```
StephanTLavavej: Pylance reports that `line_number` is unused. Using `_` avoids this.

tests/utils/stl/test/file_parsing.py
```diff
@@ -136,17 +136,17 @@ def parse_result_file(filename: Union[str, bytes, os.PathLike]) \
         return _expected_result_entry_cache[str(filename)]
 
     res = dict()
-    for line in parse_commented_file(filename):
+    for line_number, line in _parse_commented_file(filename):
         m = _EXPECTED_RESULT_REGEX.match(line)
         if m is None:
-            raise Exception(f'Incorrectly formatted line "{line}" in {filename}.')
+            raise Exception(f'Incorrectly formatted line {line_number}: "{line}" in {filename}.')
```
StephanTLavavej, line 142: I have no objection to these changes, but how are they taking effect without my recently added `check_libcxx_paths.py` finding problems earlier during configuration? (It implicitly expects properly-formatted lines and is strict about seeing `FAIL` or `SKIPPED` so if it sees something unexpected like `WOOF` or `SKIP` then it thinks the whole line is a filename, and reports a missing file.)

vmichal: Easy answer (I think obvious to you) - I did not really pay attention to the broader context of all involved python utilities, so I did not see this interaction.

Now, after review, I can see what you're hinting at. I think we should provide my proposed diagnostics in both cases - when running cmake and when running `stl-lit`. That suggests some degree of merging (at least partially) `check_libcxx_paths` and `file_parsing.py`; or at least one calling functionality from the other.

I will try to implement something, probably on Thursday. I think this has a rather low priority since it only affects us. Until then I converted this PR to draft. Feel free to close, if you prefer.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: I hulked out `check_libcxx_paths.py` to more intelligently complain about damaged lines (those whose syntax appears to be very malformed), unknown results (the common mistake of saying SKIP instead of SKIPPED), and nonexistent tests only if the syntax is otherwise valid. This is a vast improvement over the earlier behavior, thanks for prompting me to do this.

`check_libcxx_paths.py` doesn't give line numbers but I believe its output is easy enough to track down. This is a nice belt-and-suspenders approach.

StephanTLavavej:

StephanTLavavej:
Thanks for making it easier to understand test harness failures! :brain: :bulb: :smile_cat:

Issue #5971, opened 2025-12-21
`<regex>`: Consider special-casing some simple loop structures


muellerj2:
Currently, loops like `a*` or `\d*` go through the whole NFA state transition processing for each repetition, which slows down the handling of such loops a lot.

I think it's probably worth it to recognize and special-case such simple and comonly used loop structures in the matcher. (But we should not add such special-case handling to non-simple loops and instead resolve #5957 to effectively extend the simple loop handling to branchless loops.)

StephanTLavavej:
Agreed. I'd expect zero-or-more / one-or-more repeats of space / whitespace being the most common, followed by similar loops to absorb digits (as you mentioned), word characters (I use `\w+` all the time), or handwritten classes like `[a-zA-Z]+`.

PR #6005, opened 2026-01-10
`<regex>`: Revise complexity limit

### LINKED ISSUE #5944: `<regex>`: Revising the stack and complexity limits

muellerj2:
Resolves #5944.

This limits the number of state transitions by a complexity limit that grows linearly in the length of the input string. Additionally, this also takes into account the length of string comparisons in `_N_str` and `_N_back` nodes (though divided by 64, as comparing one character is much cheaper than a single state transition).

I implemented a complexity limit of `300000 + 256 * input string length`. I think that should be enough for reasonable uses (and it's likely this limit will become more generous in practice when #5971 is addressed): Matching the regex `a*` currently performs about 3 state transitions per matched character, so 85 such loops can be tried before a complexity exception is thrown.

The limit might be larger than what can fit into `_Iter_diff_t<_It>` (or `ptrdiff_t`). For this reason, two counter variables are used to implement the limit.

Drive-by change: I somehow missed that `_Matcher3::_Pop_frame()` is no longer in use, so this PR removes this function as well.

### DIFF
--- a/tests/std/tests/VSO_0000000_regex_use/test.cpp
+++ b/tests/std/tests/VSO_0000000_regex_use/test.cpp
@@ -2403,7 +2403,7 @@ void test_gh_5944() {
         regex re("a*[^b]*a*[^b]*a*[^b]*a*[^b]*a*[^b]*a*[^b]*");
         (void) regex_match("aaaaaaaaaaaaaaaaaaaaaaaaaaaaab", re);
         assert(false);
-    } catch (regex_error& ex) {
+    } catch (const regex_error& ex) {
         assert(ex.code() == error_complexity);
     }
 }


StephanTLavavej:
> The limit might be larger than what can fit into `_Iter_diff_t<_It>` (or `ptrdiff_t`). For this reason, two counter variables are used to implement the limit.

I wonder, would it be simpler to use `_Unsigned128`?

muellerj2:
I guess we could also use `long long` and clamp the limit at `LLONG_MAX`. A computation should still run for decades until it reaches `LLONG_MAX`.

The current solution just avoids this clamping.

muellerj2:
I pushed an alternative implementation of the complexity limit using a `long long` counter, which is simpler but not completely future-proof. Feel free to choose the solution you prefer.

stl/inc/regex
```diff
@@ -3421,9 +3410,22 @@ size_t _Matcher3<_BidIt, _Elem, _RxTraits, _It, _Alloc>::_Calculate_frames_limit
 }
 
 template <class _BidIt, class _Elem, class _RxTraits, class _It, class _Alloc>
-void _Matcher3<_BidIt, _Elem, _RxTraits, _It, _Alloc>::_Increase_complexity_count() {
-    if (0 < _Max_complexity_count && --_Max_complexity_count <= 0) {
+long long _Matcher3<_BidIt, _Elem, _RxTraits, _It, _Alloc>::_Calculate_complexity_limit(
+    _Iter_diff_t<_It> _Input_length) {
+    constexpr long long _Intercept = 300000LL;
+    constexpr long long _Slope     = 256LL;
```
StephanTLavavej: This is, annoyingly and unfortunately, an extremely forbidden word according to our internal codebase validation tools. (I wouldn't expect anyone to know; look it up on Wikipedia if you care.) We'll need to choose another synonym.

stl/inc/regex
```diff
@@ -3890,13 +3893,15 @@ bool _Matcher3<_BidIt, _Elem, _RxTraits, _It, _Alloc>::_Match_pat(_Node_base* _N
...
+                    _Iter_diff_t<_It> _Count{};
                     if ((_Res0 = _STD _Compare_translate_left(_Tgt_state._Cur, _End, _Node->_Data._Str(),
-                             _Node->_Data._Str() + _Node->_Data._Size(), _Traits, _Sflags))
+                             _Node->_Data._Str() + _Node->_Data._Size(), _Traits, _Sflags, _Count))
                         != _Tgt_state._Cur) {
                         _Tgt_state._Cur = _Res0;
                     } else {
                         _Failed = true;
                     }
+                    _Increase_complexity_count(static_cast<_Iter_diff_t<_It>>(_Count / 64));
```
StephanTLavavej, line 3906: Let's add comments about why we sometimes divide by 64, as this is a mysteriously magic number.

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -2392,6 +2392,22 @@ void test_gh_5939() {
...
+
+    // long strings should be matched successfully if the regex is simple
+    g_regexTester.should_match(string(20000000, 'a'), "a+");
+
+    // too much backtracking in complex regex expressions must result in a complexity exception
+    try {
+        regex re("a*[^b]*a*[^b]*a*[^b]*a*[^b]*a*[^b]*a*[^b]*");
+        (void) regex_match("aaaaaaaaaaaaaaaaaaaaaaaaaaaaab", re);
+        assert(false);
+    } catch (regex_error& ex) {
```
StephanTLavavej: We conventionally catch by const reference.

StephanTLavavej (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: The single `long long` is much easier to reason about.

StephanTLavavej:

StephanTLavavej:
:detective: :chart_with_upwards_trend: :stop_sign:
===

PR #6006, opened 2026-01-10
Update llvm-project

StephanTLavavej:
* Update llvm-project.
  + Picks up @vmichal's llvm/llvm-project#175114.
* WG21-P2404R3 was implemented by llvm/llvm-project#99420.
* Mark new flat_meow tests as FAIL in `main`.
* More "unevaluable pointer value" constexpr errors.
* More `-Wself-move` warnings.
* Mark static_asserts as FAIL, not yet analyzed.
* constexpr step limits.
* Properly categorize failures related to llvm/llvm-project#171438, thanks @frederick-vs-ja!
* Perma-workaround DevCom-11026685 by extracting an alias template, thanks @frederick-vs-ja!
  + This is superseded by Simplification 3 below.
* Simplification 1: Move up aliases in `cartesian_product_view::_Iterator`, no other changes.
  + This enables the next commit.
* Simplification 2: Use `cartesian_product_view::_Iterator::difference_type`.
  + We already provide this for the outside world, and it's simpler to use it internally too.
* Simplification 3: Use `_Iterator::difference_type` consistently.
  + Instead of having to form the difference type again, and sometimes needing to work around DevCom-10253131 / DevCom-11026685 in MSVC, we can just ask the iterator what its difference type is.
  + I reported VSO-2680018 "EDG doesn't implement C++20 Down With Typename for return types of friend functions" which was also bothering us in `flat_meow`.
  + I removed some unnecessary manual wrapping (the empty `//` comments controlling clang-format behavior).
  + In `zip_view::_Sentinel`, I extract a `_Difference_type` alias. This is because the body of the `friend` `operator-()` needs to refer to the type, but only the `_Sentinel` has granted it friendship; it can't see `zip_view::_Iterator`. This ends up being cleaner anyways.
* Mark C++26 tests as FAIL for the internal harness.

### DIFF
--- a/tests/libcxx/expected_results.txt
+++ b/tests/libcxx/expected_results.txt
@@ -582,6 +582,11 @@ std/containers/container.adaptors/flat.set/robust_against_nonbool.compile.pass.c
 std/containers/container.adaptors/flat.set/types.compile.pass.cpp FAIL
 std/language.support/support.limits/support.limits.general/flat_set.version.compile.pass.cpp FAIL
 
+# P1789R3 Library Support For Expansion Statements
+# Add 'std-at-least-c++26' to tests/utils/stl/test/tests.py when beginning work on C++26.
+std/utilities/intseq/intseq.binding/structured_binding.pass.cpp FAIL
+std/utilities/intseq/intseq.binding/tuple_interface.compile.pass.cpp FAIL
+
 # P2255R2 "Type Traits To Detect References Binding To Temporaries"
 # (type_traits.version.compile.pass.cpp still fails for Clang due to LLVM-48860)
 std/language.support/support.limits/support.limits.general/type_traits.version.compile.pass.cpp FAIL
@@ -589,6 +594,10 @@ std/utilities/meta/meta.unary/meta.unary.prop/reference_constructs_from_temporar
 std/utilities/meta/meta.unary/meta.unary.prop/reference_constructs_from_temporary.pass.cpp:1 FAIL
 std/utilities/meta/meta.unary/meta.unary.prop/reference_converts_from_temporary.pass.cpp:0 FAIL
 std/utilities/meta/meta.unary/meta.unary.prop/reference_converts_from_temporary.pass.cpp:1 FAIL
+
+# P2592R3 Hashing Support For chrono Value Classes
+# Add 'std-at-least-c++26' to tests/utils/stl/test/tests.py when beginning work on C++26.
+std/time/time.hash/time.hash_enabled.pass.cpp FAIL
 
 # P2835R7 atomic_ref::address()
 # Add 'std-at-least-c++26' to tests/utils/stl/test/tests.py when beginning work on C++26.


tests/libcxx/expected_results.txt
```diff
@@ -1517,6 +1526,16 @@ std/algorithms/alg.modifying.operations/alg.copy/copy.pass.cpp:1 FAIL
...
 
+# Not analyzed. MSVC emits error C2338: static assertion failed: 'SentinelHasMinus<View>'
+std/ranges/range.adaptors/range.adjacent.transform/sentinel/minus.pass.cpp:0 FAIL
+std/ranges/range.adaptors/range.adjacent.transform/sentinel/minus.pass.cpp:1 FAIL
+
+# Not analyzed. Clang emits error: static assertion failed due to requirement 'check_equality_comparable_with<nonmovable_equality_with_int, int>()'
+std/concepts/concepts.compare/concept.equalitycomparable/equality_comparable_with.compile.pass.cpp:2 FAIL
+
+# Not analyzed. Clang emits error: static assertion failed due to requirement 'check_three_way_comparable_with<user_defined::NonMovableIntComparable, int>()'
+std/language.support/cmp/cmp.concept/three_way_comparable_with.compile.pass.cpp:2 FAIL
```
frederick-vs-ja: These were due to incomplete workaround conditions for LLVM-171438.

Currently, Clang's `__builtin_common_type` is buggy and affects libc++ which uses it. And these 2 tests only use the compiler as the condition. The implementation can be correct when using Clang but not `__builtin_common_type`.

StephanTLavavej: Thanks! Yet another example of why asserting buggy results is a mistake (as we've discovered time and time again in our repo). I've recategorized these failures.

tests/libcxx/expected_results.txt
```diff
@@ -1517,6 +1532,10 @@ std/algorithms/alg.modifying.operations/alg.copy/copy.pass.cpp:1 FAIL
 std/input.output/file.streams/fstreams/filebuf.virtuals/xsputn.pass.cpp:0 FAIL
 std/input.output/file.streams/fstreams/filebuf.virtuals/xsputn.pass.cpp:1 FAIL
 
+# Not analyzed. MSVC emits error C2338: static assertion failed: 'SentinelHasMinus<View>'
+std/ranges/range.adaptors/range.adjacent.transform/sentinel/minus.pass.cpp:0 FAIL
+std/ranges/range.adaptors/range.adjacent.transform/sentinel/minus.pass.cpp:1 FAIL
```
frederick-vs-ja: This is a bug of MSVC. Reduced and reported DevCom-11026685.

As a workaround, it seems that we can extract a member alias template in `adjacent_transform_view::_Sentinel`

```C++
            template <bool _OtherConst>
            using _Inner_view_diff = range_difference_t<_Maybe_const<_OtherConst, _Inner_view>>;
```

and then use `_Inner_view_diff<_OtherConst>` as the return type of `operator-` overloads for `_Sentinel`.

StephanTLavavej: I think this is a duplicate of DevCom-10253131 which we have another workaround for.

StephanTLavavej: I've pushed a perma-workaround and I'm exploring further simplifications.

davidmrdavid (approved):
LGTM, happy to re-approve if /when there's another push due to simplifications.

davidmrdavid (approved):
`typename` is witchcraft, and the types are long, but the spell is sound.

StephanTLavavej:

StephanTLavavej:
I had to push a commit to mark C++26 tests as FAIL for the internal harness.

PR #6007, opened 2026-01-11
Optimize integer `to_chars` with division by 100

AlexGuteniev:
Keep repeating that part, as inlining fails

### DIFF
--- a/benchmarks/src/integer_to_string.cpp
+++ b/benchmarks/src/integer_to_string.cpp
@@ -41,7 +41,7 @@ auto generate_array() {
 
 template <class T, double M, double S>
 void integer_to_chars(benchmark::State& state) {
-    auto a = generate_array<char, M, S>();
+    auto a = generate_array<T, M, S>();
     char buff[20]; // can hold -2^63 and 2^64 - 1
 
     auto it = a.begin();


stl/inc/charconv
```diff
@@ -64,31 +64,64 @@ _NODISCARD _CONSTEXPR23 to_chars_result _Integer_to_chars(
...
-                    _Value               = static_cast<_Unsigned>(_Value / 1'000'000'000);
-
-                    for (int _Idx = 0; _Idx != 9; ++_Idx) {
-                        *--_RNext = static_cast<char>('0' + _Chunk % 10);
-                        _Chunk /= 10;
+#ifndef _WIN64
+            auto _Trunc = _Value;
+#else // ^^^ defined(_WIN64) / !defined(_WIN64) vvv
+
+            if constexpr (sizeof(_Unsigned) > 4) { // For 64-bit numbers, work in chunks to avoid 64-bit divisions.
```
StephanTLavavej: This is damaged (both the logic and the comments), and seriously affecting 64-bit perf on 64-bit architectures. I recommend testing the macro positively, and using our 64-bit / 32-bit comments (one of the few cases where we prefer to not just repeat the condition):

```cpp
#ifdef _WIN64

#else // ^^^ 64-bit / 32-bit vvv

#endif // ^^^ 32-bit ^^^
```

AlexGuteniev: I wanted to do better by using preprocessor instead of just TMP and did worse! 

Fixed.

I observe we still use `^^^ defined(_WIN64) / !defined(_WIN64) vvv` elsewhere.

benchmarks/src/integer_to_string.cpp
```diff
@@ -38,6 +39,26 @@ auto generate_array() {
     return a;
 }
 
+template <class T, double M, double S>
+void integer_to_chars(benchmark::State& state) {
+    auto a = generate_array<char, M, S>();
+    char d[20];
```
StephanTLavavej: Let's copy the comment below about why 20 is correct. We can also use the `buff` name, as this function has several single-letter variables flying around.

benchmarks/src/integer_to_string.cpp
```diff
@@ -38,6 +39,26 @@ auto generate_array() {
     return a;
 }
 
+template <class T, double M, double S>
+void integer_to_chars(benchmark::State& state) {
+    auto a = generate_array<char, M, S>();
```
StephanTLavavej: Bug: `T` is totally unused in `integer_to_chars()`; you're always benchmarking `char`.

StephanTLavavej:
Thanks! :heart_eyes_cat: After fixing the bug that was invalidating the benchmark, my 5950X results are good:

Benchmark                                |  Before | After   | Speedup
-----------------------------------------|---------|---------|--------
`integer_to_chars<uint8_t, 2.5, 1.5>`    | 6.48 ns | 6.23 ns | 1.04
`integer_to_chars<uint16_t, 5.0, 3.0>`   | 7.56 ns | 6.90 ns | 1.10
`integer_to_chars<uint32_t, 10.0, 6.0>`  | 9.50 ns | 8.30 ns | 1.14
`integer_to_chars<uint64_t, 20.0, 12.0>` | 12.5 ns | 9.37 ns | 1.33

StephanTLavavej:

StephanTLavavej:
Thanks for improving my favorite function! :tada: :rocket: :cat2:

PR #3164, opened 2022-10-19
Re-enable ASAN string annotations

strega-nil-ms:
tested with the bugs with the previous implementation!

Fixes #3205 
internal PR for testing purposes: ~~MSVC-PR-433542~~ ~~MSVC-PR-440101~~ MSVC-PR-440279
Fixes the following internal bugs: VSO-1664238, VSO-1586016, VSO-1543191

PR #6010, opened 2026-01-14
`<optional>`: Add Address Sanitizer annotations

### LINKED ISSUE #5974: `<optional>`: address sanitizer annotation
We can annotate empty `optional`.

This may help finding dangling references to objects in `optional`s that have been reset to empty state.

ozguronsoy:
When `optional` is empty the internal storage is poisoned, it's unpoisoned when a value is assigned. We need to unpoison in the destructor, so the annotations must be restricted to non-trivially destructible value types.

- Added `_ANNOTATE_OPTIONAL`, `_DISABLE_OPTIONAL_ANNOTATION`, etc. to `__msvc_sanitizer_annotate_container.hpp`.
- Added poisoning storage in `_Optional_destruct_base` on empty construction and `reset`.
- Added unpoisoning of storage in `_Optional_construct_base::_Construct` and `~_Optional_destruct_base()`.
- Added poisoning and unpoisoning tests.

Resolves #5974

### DIFF
--- a/tests/std/tests/GH_005974_asan_annotate_optional/test.cpp
+++ b/tests/std/tests/GH_005974_asan_annotate_optional/test.cpp
@@ -10,72 +10,120 @@ int main() {}
 
 #include <cassert>
 #include <optional>
+#include <string>
+#include <string_view>
+#include <type_traits>
+#include <variant>
+using namespace std;
+
+#if _HAS_CXX20
+#define CONSTEXPR20 constexpr
+#else // ^^^ _HAS_CXX20 / !_HAS_CXX20 vvv
+#define CONSTEXPR20 inline
+#endif // ^^^ !_HAS_CXX20 ^^^
 
 #ifdef __SANITIZE_ADDRESS__
 extern "C" int __cdecl __asan_address_is_poisoned(const volatile void* addr);
 #define ASAN_VERIFY_POISONED(addr)   assert(__asan_address_is_poisoned((addr)) != 0)
 #define ASAN_VERIFY_UNPOISONED(addr) assert(__asan_address_is_poisoned((addr)) == 0)
-#else
+#else // ^^^ defined(__SANITIZE_ADDRESS__) / !defined(__SANITIZE_ADDRESS__) vvv
 #define ASAN_VERIFY_POISONED(addr)   ((void) (addr))
 #define ASAN_VERIFY_UNPOISONED(addr) ((void) (addr))
-#endif
+#endif // ^^^ !defined(__SANITIZE_ADDRESS__) ^^^
 
-struct Payload {
-    long long x;
-    long long y;
-    long long z;
-    long long w;
-};
+template <class StrType>
+void test_poisoning() {
+    // std::optional<T> is ASan-annotated only for non-trivially destructible T.
+    static_assert(!is_trivially_destructible_v<StrType>);
+    {
+        // Verify layout assumption, specific to our implementation but not a guarantee provided to users.
+        // This test assumes that a std::optional's contained value is stored at offset 0, so it has the same address.
+        const optional<StrType> opt{"cats"};
+        assert(static_cast<const void*>(&opt) == static_cast<const void*>(&opt.value()));
+    }
+    {
+        // Same layout assumption for std::variant.
+        const variant<optional<StrType>, int> var{"cats"};
+        assert(static_cast<const void*>(&var) == static_cast<const void*>(&get<optional<StrType>>(var).value()));
+    }
+    {
+        optional<StrType> opt;
+        ASAN_VERIFY_POISONED(&opt);
+        assert(!opt.has_value());
 
-void test_poison_on_empty_access() {
-    [[maybe_unused]] std::optional<Payload> opt;
-    ASAN_VERIFY_POISONED(reinterpret_cast<Payload*>(&opt));
+        opt.emplace("cats");
+        ASAN_VERIFY_UNPOISONED(&opt);
+        assert(opt.value() == "cats");
+    }
+    {
+        optional<StrType> opt{"cats"};
+        ASAN_VERIFY_UNPOISONED(&opt);
+        assert(opt.value() == "cats");
+
+        opt.reset();
+        ASAN_VERIFY_POISONED(&opt);
+        assert(!opt.has_value());
+    }
+    {
+        // Verify that std::optional's destructor unpoisons its storage, so the bytes can be reused for another object.
+        variant<optional<StrType>, int> var{"cats"};
+        ASAN_VERIFY_UNPOISONED(&var);
+        assert(get<optional<StrType>>(var).value() == "cats");
+
+        get<optional<StrType>>(var).reset();
+        ASAN_VERIFY_POISONED(&var);
+        assert(!get<optional<StrType>>(var).has_value());
+
+        var = 1729;
+        ASAN_VERIFY_UNPOISONED(&var);
+        assert(get<int>(var) == 1729);
+    }
 }
 
-void test_emplace_unpoisoning() {
-    std::optional<Payload> opt;
-    opt.emplace();
-    ASAN_VERIFY_UNPOISONED(reinterpret_cast<Payload*>(&opt));
-}
-
-void test_assignment_unpoisoning() {
-    std::optional<Payload> opt = std::nullopt;
-    opt                        = Payload{};
-    ASAN_VERIFY_UNPOISONED(reinterpret_cast<Payload*>(&opt));
-}
-
-void test_repoison_after_reset() {
-    std::optional<Payload> opt = Payload{};
-    ASAN_VERIFY_UNPOISONED(reinterpret_cast<Payload*>(&opt));
-    opt.reset();
-    ASAN_VERIFY_POISONED(reinterpret_cast<Payload*>(&opt));
-}
-
-constexpr bool test_constexpr() {
 #if _HAS_CXX20
-    bool res                   = true;
-    std::optional<Payload> opt = std::nullopt;
-    opt                        = Payload{};
-    opt.reset();
-    opt = Payload{86, 0, 0, 0};
-    res = opt->x == 86;
-    opt.emplace(42, 0, 0, 0);
-    res = res && (opt->x == 42);
-    return res;
-#else
-    std::optional<Payload> opt{Payload{86, 0, 0, 0}};
-    return opt->x == 86;
-#endif
+template <class StrType>
+constexpr bool test_constexpr() {
+    {
+        optional<StrType> opt;
+        assert(!opt.has_value());
+        opt.emplace("cats");
+        assert(opt.value() == "cats");
+    }
+    {
+        optional<StrType> opt{"cats"};
+        assert(opt.value() == "cats");
+        opt.reset();
+        assert(!opt.has_value());
+    }
+    return true;
 }
+#endif // _HAS_CXX20
+
+class MyString {
+public:
+    constexpr MyString(const char* const ptr) : m_sv{ptr} {}
+
+    CONSTEXPR20 ~MyString() {} // non-trivially destructible
+
+    constexpr bool operator==(const char* const ptr) const {
+        return m_sv == ptr;
+    }
+
+private:
+    string_view m_sv{};
+};
+static_assert(!is_trivially_destructible_v<MyString>);
 
 int main() {
-    test_poison_on_empty_access();
-    test_emplace_unpoisoning();
-    test_assignment_unpoisoning();
-    test_repoison_after_reset();
-    static_assert(test_constexpr(), "constexpr test failed");
-
-    return 0;
+    // Test std::optional with both std::string and MyString as value types.
+    // std::string is the realistic scenario.
+    // MyString ensures that we're exercising std::optional's ASan annotations instead of std::string's.
+    test_poisoning<string>();
+    test_poisoning<MyString>();
+#if _HAS_CXX20
+    static_assert(test_constexpr<string>());
+    static_assert(test_constexpr<MyString>());
+#endif // _HAS_CXX20
 }
 
 #endif // ^^^ no workaround ^^^


ozguronsoy:

StephanTLavavej:
Thank you! I need to focus on landing `<flat_map>` and `<flat_set>`, but I'll review this when I get a chance.

stl/inc/__msvc_sanitizer_annotate_container.hpp
```diff
@@ -125,10 +153,17 @@ extern const bool _Asan_vector_should_annotate;
 #ifdef _INSERT_STRING_ANNOTATION
 extern const bool _Asan_string_should_annotate;
 #endif
+
+#ifdef _INSERT_OPTIONAL_ANNOTATION
+extern const bool _Asan_optional_should_annotate;
```
frederick-vs-ja, line 158: This variable isn't used. Is it really necessary?

stl/inc/__msvc_sanitizer_annotate_container.hpp
```diff
@@ -125,10 +153,17 @@ extern const bool _Asan_vector_should_annotate;
 #ifdef _INSERT_STRING_ANNOTATION
 extern const bool _Asan_string_should_annotate;
 #endif
+
+#ifdef _INSERT_OPTIONAL_ANNOTATION
+extern const bool _Asan_optional_should_annotate;
+#endif
 } // extern "C"
 
 #if defined(_INSERT_VECTOR_ANNOTATION) || defined(_INSERT_STRING_ANNOTATION)
```
frederick-vs-ja: I think there should be a concentrated controlling macro for these.

stl/inc/optional
```diff
@@ -74,7 +75,11 @@ struct _Optional_destruct_base { // either contains a value of _Ty or is empty (
     };
     bool _Has_value;
 
-    constexpr _Optional_destruct_base() noexcept : _Dummy{}, _Has_value{false} {} // initialize an empty optional
+    constexpr _Optional_destruct_base() noexcept : _Dummy{}, _Has_value{false} {
+#ifdef _INSERT_OPTIONAL_ANNOTATION
+        __asan_poison_memory_region(_STD addressof(_Value), sizeof(_Value));
```
frederick-vs-ja: This invocations probably break constant evaluation. We should wrap them in `if (!_STD _Is_constant_evaluated()) { }`, IMO.

Also, why not consistently write `sizeof(_Ty)`?

tests/std/tests/GH_005974_optional_asan/test.cpp
```diff
@@ -0,0 +1,55 @@
...
+};
+
+void test_poison_on_empty_access() {
+    [[maybe_unused]] std::optional<Payload> opt;
+    ASAN_VERIFY_POISONED(reinterpret_cast<Payload*>(&opt));
+}
+
+void test_emplace_unpoisoning() {
+    std::optional<Payload> opt;
+    opt.emplace(Payload());
```
frederick-vs-ja: For clarity:
```suggestion
    opt.emplace();
```

tests/std/tests/GH_005974_optional_asan/test.cpp
```diff
@@ -0,0 +1,55 @@
...
+    ASAN_VERIFY_UNPOISONED(reinterpret_cast<Payload*>(&opt));
+}
+
+void test_assignment_unpoisoning() {
+    std::optional<Payload> opt = std::nullopt;
+    opt                        = Payload();
```
frederick-vs-ja: For clarity, to avoid confusion with type-id:
```suggestion
    opt                        = Payload{};
```
Ditto below.

tests/std/tests/GH_005974_asan_annotate_optional/test.cpp
```diff
@@ -0,0 +1,55 @@
...
+    ASAN_VERIFY_UNPOISONED(reinterpret_cast<Payload*>(&opt));
+    opt.reset();
+    ASAN_VERIFY_POISONED(reinterpret_cast<Payload*>(&opt));
+}
+
+int main() {
+    test_poison_on_empty_access();
+    test_emplace_unpoisoning();
+    test_assignment_unpoisoning();
+    test_repoison_after_reset();
```
frederick-vs-ja: I think we can make these cases constexpr-friendly and test them in `static_assert` since C++20.

tests/std/tests/GH_005974_optional_asan/test.cpp
ozguronsoy: Done.

tests/std/tests/GH_005974_optional_asan/test.cpp
ozguronsoy: Done.

stl/inc/__msvc_sanitizer_annotate_container.hpp
ozguronsoy: I don't think it's necessary since this check is only done here. I also added `defined(_INSERT_OPTIONAL_ANNOTATION)` to the check.

stl/inc/optional
ozguronsoy: Done.

> Also, why not consistently write sizeof(_Ty)?

That's how I usually write it. Changed it to `sizeof(_Ty)` to match the existing style.

stl/inc/__msvc_sanitizer_annotate_container.hpp
ozguronsoy: Added a check for this variable inside the `if(!_STD _Is_constant_evaluated())` block.

tests/std/tests/GH_005974_asan_annotate_optional/test.cpp
ozguronsoy: Added a separate function to test constexpr functionality in `static_assert`.

ozguronsoy:
Hi @StephanTLavavej, since the `<flat_map>` and `<flat_set>` have landed, can we move on with the review when you are available?

StephanTLavavej:
Yes, I'm trying to get through my PR review backlog. I know you've been waiting a month, for which I apologize. I'm the only maintainer flying this jumbo jet at the moment, so I ask for your patience!

ozguronsoy:
Wow that must be tough. It's no problem, I can wait.

tests/std/tests/GH_005974_optional_asan/env.lst
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+RUNALL_INCLUDE ..\usual_17_matrix.lst
```
davidmrdavid, line 4: Following in the precedent set by [`GH_002030_asan_annotate_string`](https://github.com/microsoft/STL/blob/main/tests/std/tests/GH_002030_asan_annotate_string/env.lst) and it's vector variant, I'd imagine we'd have the same test matrix as them, which is larger.

Any reason not to reuse the same test matrix as in the linked example?

tests/std/tests/GH_005974_optional_asan/env.lst
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) Microsoft Corporation.
```
davidmrdavid, line 1: ultra subjective nit - can we name this folder `GH_005974_asan_annotate_optional`? That way, we'll have a common prefix (_asan_annotate_) to search all asan annotation tests :)

tests/std/tests/GH_005974_asan_annotate_optional/test.cpp
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
```
AlexGuteniev, line 3: I think a comment `// REQUIRES: x64 || x86` like in other ASan tests would prevent ARM64EC failure

StephanTLavavej: Yes, it's always a good idea to follow precedent.

ozguronsoy: Done, thanks. I also added a `notarget.lst`.

tests/std/tests/GH_005974_optional_asan/env.lst
ozguronsoy: Done.

tests/std/tests/GH_005974_optional_asan/env.lst
ozguronsoy: Done.

StephanTLavavej:
I'll resolve the stealth merge conflicts with #6157 and #6163 and try to get this reviewed soon, in the hopes of getting it into 14.51.

stl/inc/__msvc_sanitizer_annotate_container.hpp
StephanTLavavej: After a quick glance, I agree that another macro doesn't appear to be necessary.

stl/src/asan_noop.cpp
```diff
@@ -2,8 +2,12 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 extern "C" {
-extern const bool _Asan_string_should_annotate_default = false;
-extern const bool _Asan_vector_should_annotate_default = false;
+extern const bool _Asan_string_should_annotate_default   = false;
+extern const bool _Asan_vector_should_annotate_default   = false;
+extern const bool _Asan_optional_should_annotate_default = false;
+
+void __cdecl __asan_poison_memory_region_default(void const volatile*, size_t) {}
```
StephanTLavavej: We conventionally use west const.

stl/inc/optional
```diff
@@ -93,6 +100,11 @@ struct _Optional_destruct_base { // either contains a value of _Ty or is empty (
 
     _CONSTEXPR20 void reset() noexcept {
         _Has_value = false;
+#ifdef _INSERT_OPTIONAL_ANNOTATION
+        if (!_STD _Is_constant_evaluated() && _Asan_optional_should_annotate) {
+            __asan_poison_memory_region(_STD addressof(_Value), sizeof(_Ty));
+        }
+#endif
```
StephanTLavavej: Style: Elsewhere, we poison before assigning `_Has_value = false;`.

tests/std/tests/GH_005974_asan_annotate_optional/test.cpp
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+// REQUIRES: x64 || x86
```
StephanTLavavej: Stealth merge conflicts with #6157 and #6163.

StephanTLavavej:
I reviewed the product code and pushed changes to merge `main` and address nitpicks. However, before getting to the test code, I realized that the product code is doomed in its current form. Consider the following conforming code, which is rejected by these changes:

```
```
```cpp
#include <optional>
#include <print>
#include <string>
#include <variant>
using namespace std;

int main() {
    variant<optional<string>, int> var{"cats"};
    println("{}", get<0>(var).value());
    var = 1729;
    println("{}", get<1>(var));
}
```
```
D:\GitHub\STL\out\x64>cl /EHsc /nologo /W4 /std:c++latest /MT /Od meow.cpp && meow
meow.cpp
cats
1729

D:\GitHub\STL\out\x64>cl /EHsc /nologo /W4 /std:c++latest /MT /Od /fsanitize=address /Zi meow.cpp && meow
meow.cpp
cats
=================================================================
==21032==ERROR: AddressSanitizer: use-after-poison on address 0x006358cffcb0 at pc 0x7ff700521fe3 bp 0x006358cffa10 sp 0x006358cffa18
WRITE of size 4 at 0x006358cffcb0 thread T0
    #0 0x7ff700521fe2 in std::_Variant_storage_<1, int>::_Variant_storage_<1, int><int>(struct std::integral_constant<unsigned __int64, 0>, int &&) D:\GitHub\STL\out\x64\out\inc\variant:357
[...]
```

<details><summary>Click to expand full ASan diagnostic:</summary>

```
==21032==ERROR: AddressSanitizer: use-after-poison on address 0x006358cffcb0 at pc 0x7ff700521fe3 bp 0x006358cffa10 sp 0x006358cffa18
WRITE of size 4 at 0x006358cffcb0 thread T0
    #0 0x7ff700521fe2 in std::_Variant_storage_<1, int>::_Variant_storage_<1, int><int>(struct std::integral_constant<unsigned __int64, 0>, int &&) D:\GitHub\STL\out\x64\out\inc\variant:357
    #1 0x7ff7005214ea in std::_Variant_storage_<0, class std::optional<class std::basic_string<char, struct std::char_traits<char>, class std::allocator<char>>>, int>::_Variant_storage_<0, class std::optional<class std::basic_string<char, struct std::char_traits<char>, class std::allocator<char>>>, int><1, int, 0>(struct std::integral_constant<unsigned __int64, 1>, int &&) D:\GitHub\STL\out\x64\out\inc\variant:406
    #2 0x7ff700527fe0 in std::_Construct_in_place<class std::_Variant_storage_<0, class std::optional<class std::basic_string<char, struct std::char_traits<char>, class std::allocator<char>>>, int>, struct std::integral_constant<unsigned __int64, 1>, int>(class std::_Variant_storage_<0, class std::optional<class std::basic_string<char, struct std::char_traits<char>, class std::allocator<char>>>, int> &, struct std::integral_constant<unsigned __int64, 1> &&, int &&) D:\GitHub\STL\out\x64\out\inc\xutility:621
    #3 0x7ff700528b3e in std::variant<class std::optional<class std::basic_string<char, struct std::char_traits<char>, class std::allocator<char>>>, int>::_Emplace_valueless<1, int>(int &&) D:\GitHub\STL\out\x64\out\inc\variant:1136
    #4 0x7ff70052269c in std::variant<class std::optional<class std::basic_string<char, struct std::char_traits<char>, class std::allocator<char>>>, int>::operator=<int, 0, 0>(int &&) D:\GitHub\STL\out\x64\out\inc\variant:1013
    #5 0x7ff700521306 in main D:\GitHub\STL\out\x64\meow.cpp:10
    #6 0x7ff7005faefb in invoke_main D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:78
    #7 0x7ff7005faefb in __scrt_common_main_seh D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288
    #8 0x7ff8d89ae8d6  (C:\Windows\System32\KERNEL32.DLL+0x18002e8d6)
    #9 0x7ff8da24c53b  (C:\Windows\SYSTEM32\ntdll.dll+0x18008c53b)

Address 0x006358cffcb0 is located in stack of thread T0 at offset 32 in frame
    #0 0x7ff70052110f in main D:\GitHub\STL\out\x64\meow.cpp:7

  This frame has 2 object(s):
    [32, 80) 'var' <== Memory access at offset 32 is inside this variable
    [48, 52) 'compiler temporary'
HINT: this may be a false positive if your program uses some custom stack unwind mechanism, swapcontext or vfork
      (longjmp, SEH and C++ exceptions *are* supported)
SUMMARY: AddressSanitizer: use-after-poison D:\GitHub\STL\out\x64\out\inc\variant:357 in std::_Variant_storage_<1, int>::_Variant_storage_<1, int><int>(struct std::integral_constant<unsigned __int64, 0>, int &&)
Shadow bytes around the buggy address:
  0x006358cffa00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x006358cffa80: 00 00 f1 f1 f1 f1 01 f3 f3 f3 f3 00 00 00 00 00
  0x006358cffb00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x006358cffb80: f1 f1 f1 f1 01 f3 f3 f3 f3 00 00 00 00 00 00 00
  0x006358cffc00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x006358cffc80: 00 00 f1 f1 f1 f1[f7]f7 f7 f7 00 00 f2 f2 f2 f2
  0x006358cffd00: 04 f3 f3 f3 f3 00 00 00 00 00 00 00 00 00 00 00
  0x006358cffd80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x006358cffe00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x006358cffe80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x006358cfff00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb

NOTE: the stack trace above identifies the code that *accessed* the poisoned memory.
To identify the code that *poisoned* the memory, try the experimental setting ASAN_OPTIONS=poison_history_size=<size>.
==21032==ABORTING
```
</details>

I arrived at this realization by noticing the asymmetry between the trivial dtor and non-trivial dtor cases, which got me thinking about the poisoned/unpoisoned state of the memory after the `optional` object's lifetime has ended.

The problem is that if an `optional` object (whether non-trivial like the `optional<string>` in the repro above, or trivial like an `optional<double>` would be) leaves its memory in a poisoned state (e.g. as the non-trivial dtor does; a default-constructed or `reset()` trivial `optional` will also be poisoned), then any attempted reuse of that memory will lead to an ASan false positive failure. I used `variant` in my repro, but other scenarios are possible.

`optional` and `vector` are fundamentally different because `vector` owns remote storage that nobody else can reuse. `string` is `vector`-like except for the Small String Optimization, but notice that we specifically avoid using ASan annotations for the Small String Optimization (IIRC this happened in the very-complicated #3164):

https://github.com/microsoft/STL/blob/2626cf1ee9d26be701b9bdbd2fb30db240456456/stl/inc/xstring#L701-L704

I'm going to need to run this past our ASan experts to see if some part of these changes can be salvaged. I expect that the trivial dtor case can't be salvaged. However, it might be possible to annotate the non-trivial dtor case; for the lifetime of the `optional` object, we do control its memory, and I think we can poison it if we're default-constructed or `reset()`, as long as we (perhaps counterintuitively) unpoison it in the destructor.

Please hold off on making changes to this PR while I go check. I'll also want to add the `variant` scenario to the test coverage, of course.

stl/inc/optional
StephanTLavavej: After dropping the trivial case, I think the non-trivial case should actually follow the other order, heh.

tests/std/tests/GH_005974_asan_annotate_optional/env.lst
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# This test matrix is the usual test matrix, with all currently unsupported options removed, crossed with the ASan flags.
```
StephanTLavavej: This isn't your fault at all, but the annotate_vector matrix was derived from `usual_matrix.lst` through a *far* more elaborate series of transformations than this comment suggests. We should derive the annotate_optional matrix from `usual_17_matrix.lst` through the same (almost completely undocumented) process, instead of starting with the annotate_vector matrix.

tests/std/tests/GH_005974_asan_annotate_optional/test.cpp
```diff
@@ -0,0 +1,81 @@
...
+#define ASAN_VERIFY_POISONED(addr)   ((void) (addr))
+#define ASAN_VERIFY_UNPOISONED(addr) ((void) (addr))
+#endif
+
+struct Payload {
+    long long x;
+    long long y;
+    long long z;
+    long long w;
+};
```
StephanTLavavej: This test will need to be completely overhauled to exercise non-trivially-destructible `optional`s.

StephanTLavavej:

StephanTLavavej:
Thanks for this major safety improvement - it was surprisingly tricky to get merged on the MSVC-internal side but with the ASan team's help we did it! :heart_eyes_cat: :tada: :cat2:

PR #6011, opened 2026-01-14
Toolset update: MSVC Compiler 19.50.35722

StephanTLavavej:
# Changelog
  * Updated dependencies.
    + Updated MSVC Compiler to 19.50.35722 (now required).
    + Updated CMake to 4.1.2 (now required).
    + Updated Python to 3.14.2 (now required).

# Commits
* Enable [native NVMe][].
  + This is unlikely to observably improve throughput, but I figured it would be nice to do. I manually verified that this takes effect in our prototype VM.
* Cycle locations.
* Python 3.14.2.
* MSVC Compiler 19.50.35722.
* CMake 4.1.2.
* New pools.
* common_reference.compile.pass.cpp is now passing.
* Rename `str` => `s` in Python scripts.
  + Followup to #5888, where internal AI code review noted that `str` has builtin meaning to Python. Python lets us use `str` as a variable name anyways, but that's unnecessarily confusing. We already use `s` when line length is a concern, so we can just follow that consistently. I've manually exercised each altered codepath.
* Azure Pipelines: Consistently use `pwsh` instead of `cmd`
  + See below.

# Using `pwsh` in Azure Pipelines
This has no immediate benefits, but will allow us to do more complicated things in the scripts in the future. We already use `pwsh` in a couple of places for this reason.

In config.yml, replace `VsDevCmd.bat` with `Launch-VsDevShell.ps1`, now that it supports 'arm64' (and 'x64' as a synonym of 'amd64').

In general, this transformation involves:

* Replace [`steps.script`][], which is a shortcut for the [`CmdLine@2` task][], with the [`PowerShell@2` task][].
* Configure it with `pwsh: true` to use new `pwsh` instead of old `powershell`.
* Immediately set [`$PSNativeCommandUseErrorActionPreference`][] to `$true`, which will cause `pwsh` to properly react to failures of native commands like `cmake` and `ninja`.
* Replace `cmd`'s `if exist` and `rmdir` with `pwsh`'s [`Test-Path`][] and [`Remove-Item`][].
* Replace `call "$(vsDevCmdBat)"` with `& "$(launchVsDevShell)"`, using the [call operator `&`][].
* Note that `Launch-VsDevShell.ps1` lacks an analogue of `-no_logo`, so the CI logs will be slightly more verbose.
* Replace the line continuations from `cmd`'s caret `^` to `pwsh`'s backtick `` ` ``.
* In checkout-self.yml, [`mkdir`][] is an alias for `New-Item -Type Directory`. Add `-Force | Out-Null`, as we already do in: https://github.com/microsoft/STL/blob/4634ad466dfdb706ba65e197ec8257d6c7bce25a/azure-devops/format-validation.yml#L41
* In cmake-configure-build.yml, add quotes around the value of `LIT_FLAGS`. This is because it contains double dashes, which `pwsh` will interpret as an operator if they aren't quoted: https://github.com/microsoft/STL/blob/4634ad466dfdb706ba65e197ec8257d6c7bce25a/azure-devops/cmake-configure-build.yml#L21-L27
* In format-validation.yml, add `$PSNativeCommandUseErrorActionPreference = $true` to 'Create Diff' which was already using `pwsh`.
* In run-tests.yml, we need to move `workingDirectory` within `task.inputs`; we can't send it to the `task` itself.

[native NVMe]: https://techcommunity.microsoft.com/blog/windowsservernewsandbestpractices/announcing-native-nvme-in-windows-server-2025-ushering-in-a-new-era-of-storage-p/4477353
[`steps.script`]: https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-script?view=azure-pipelines
[`CmdLine@2` task]: https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/cmd-line-v2?view=azure-pipelines
[`PowerShell@2` task]: https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/powershell-v2?view=azure-pipelines
[`$PSNativeCommandUseErrorActionPreference`]: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.5#psnativecommanduseerroractionpreference
[`Test-Path`]: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/test-path?view=powershell-7.5
[`Remove-Item`]: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/remove-item?view=powershell-7.5
[call operator `&`]: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_operators?view=powershell-7.5#call-operator-
[`mkdir`]: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_built-in_functions?view=powershell-7.5#mkdir-alias-md

### DIFF
--- a/azure-devops/build-benchmarks.yml
+++ b/azure-devops/build-benchmarks.yml
@@ -14,18 +14,23 @@ parameters:
   type: string
 
 steps:
-- script: |
-    if exist "$(benchmarkBuildOutputLocation)\${{ parameters.compiler }}" (
-      rmdir /S /Q "$(benchmarkBuildOutputLocation)\${{ parameters.compiler }}"
-    )
-    call "$(vsDevCmdBat)" -host_arch=${{ parameters.hostArch }} -arch=${{ parameters.targetArch }} -no_logo
-    cmake -G Ninja ^
-    -DCMAKE_CXX_COMPILER=${{ parameters.compiler }} ^
-    -DCMAKE_BUILD_TYPE=Release ^
-    -DSTL_BINARY_DIR="$(buildOutputLocation)" ^
-    -DVCLIBS_TARGET_ARCHITECTURE=${{ parameters.targetPlatform }} ^
-    -S $(Build.SourcesDirectory)/benchmarks -B "$(benchmarkBuildOutputLocation)\${{ parameters.compiler }}"
+- task: PowerShell@2
   displayName: 'Configure Benchmarks for ${{ parameters.compiler }}'
+  inputs:
+    pwsh: true
+    targetType: inline
+    script: |
+      $PSNativeCommandUseErrorActionPreference = $true
+      if (Test-Path -LiteralPath "$(benchmarkBuildOutputLocation)\${{ parameters.compiler }}") {
+          Remove-Item -LiteralPath "$(benchmarkBuildOutputLocation)\${{ parameters.compiler }}" -Recurse -Force
+      }
+      & "$(launchVsDevShell)" -HostArch ${{ parameters.hostArch }} -Arch ${{ parameters.targetArch }}
+      cmake -G Ninja `
+      -DCMAKE_CXX_COMPILER=${{ parameters.compiler }} `
+      -DCMAKE_BUILD_TYPE=Release `
+      -DSTL_BINARY_DIR="$(buildOutputLocation)" `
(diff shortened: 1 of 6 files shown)


StephanTLavavej:

PR #6012, opened 2026-01-14
Remove workarounds for ancient Clang versions

StephanTLavavej:
MSVC's STL requires the version of Clang that's shipped in VS (currently Clang 20). However, a couple of projects in our Real World Code test suite (RWC) were defining our escape hatch macro (which we provide as a last resort and haven't documented) to get away with using egregiously ancient Clang versions. This is a headache for development, since we don't otherwise test those configurations, and we continually need to take advantage of newer compiler features and fixes, which is why we have the version requirement in the first place.

One of those RWC projects was recently disabled (Taichi, by MSVC-PR-695358, merged 2025-12-17) until we can begin running it against our supported Clang version. I'm aware of one or two more (e.g. Root) that embed a "Cling" interpreter, based on a typically older version of Clang. If necessary, these projects out there in the actual real world can use our older supported versions while they migrate to modern versions of Clang. However, we should stop attempting to preserve workarounds for these increasingly outdated versions of Clang, in our current development sources.

While we don't support or test MSVC's STL with Intel's Clang-based compiler, our unofficial policy is that we avoid egregiously breaking it when avoiding that is easy. I checked, and [Intel's release notes][] say:

> Intel® oneAPI DPC++/C++ Compiler version 2025.0 supports the C/C++ standards through the Clang 19 front end. 

Therefore, removing these workarounds for Clang 18 and older won't affect Intel 2025.0. (I have absolutely no interest in avoiding breaking older versions of Intel's compiler with the latest versions of MSVC's STL. If users are pinned to Intel 2024.0, they can also pin to MSVC Build Tools 14.50.)

[Intel's release notes]: https://www.intel.com/content/www/us/en/developer/articles/release-notes/oneapi-dpcpp/2025.html

* Remove workarounds for Clang 15 and older.
  + Clang 16 was released 2023-03-17. We began requiring it in the MSVC Build Tools 14.37 (VS 2022 17.7).
  + For `__is_scoped_enum`, don't attempt to detect EDG-imitating-Clang, which we historically haven't special-cased or tested.
* Remove workarounds for Clang 18 and older.
  + Clang 19 was released 2024-09-17. We began requiring it in the MSVC Build Tools 14.44 (VS 2022 17.14).
  + Fuse away `_USE_BUILTIN_IS_TRIVIALLY_EQUALITY_COMPARABLE`, we don't use escape hatches elsewhere.
  + For `is_layout_compatible`, the condition previously said "if we're not Clang at all (i.e. we're MSVC or EDG-imitating-MSVC), or we're EDG-imitating-Clang, or we're Clang 19+". This is always true now (regardless of the fact that we don't special-case EDG-imitating-Clang).
  + For `is_pointer_interconvertible_base_of`, don't special-case EDG-imitating-Clang when working around llvm/llvm-project#135273. Again, we don't special-case or test this configuration, and in this case the workaround isn't especially harmful anyways.

### DIFF
--- a/stl/inc/type_traits
+++ b/stl/inc/type_traits
@@ -2013,15 +2013,13 @@ _NO_SPECIALIZATIONS_OF_TYPE_TRAITS constexpr bool is_nothrow_invocable_r_v =
 #endif // _HAS_CXX17
 
 #if _HAS_CXX20
-#if !defined(__clang__) || defined(__EDG__) \
-    || __clang_major__ >= 19 // TRANSITION, VSO-2397560 (RWC relying on ancient Clang)
 _EXPORT_STD template <class _Ty1, class _Ty2>
 struct _NO_SPECIALIZATIONS_OF_TYPE_TRAITS is_layout_compatible : bool_constant<__is_layout_compatible(_Ty1, _Ty2)> {};
 
 _EXPORT_STD template <class _Ty1, class _Ty2>
 _NO_SPECIALIZATIONS_OF_TYPE_TRAITS constexpr bool is_layout_compatible_v = __is_layout_compatible(_Ty1, _Ty2);
 
-#if !defined(__clang__) || defined(__EDG__) // TRANSITION, LLVM-135273
+#ifndef __clang__ // TRANSITION, LLVM-135273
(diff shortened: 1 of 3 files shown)


StephanTLavavej:

PR #6013, opened 2026-01-14
`<vector>`: Avoid off-by-default warning C4365 in `vector<bool>` copy algorithms with `/J`

StephanTLavavej:
Reported by a major customer.

Our policy is that we attempt to be `/W4` clean but not `/Wall` clean, as `/Wall` contains a large number of extremely noisy warnings, many of which aren't suitable for using in production (some are intended for focused investigations, or were added for one unusual scenario, or were just way noisier than their value proved to be).

However, we generally do attempt to keep the codebase clean with respect to the [off-by-default warning C4365](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4365?view=msvc-170), as we build our tests with `/w14365`: https://github.com/microsoft/STL/blob/4634ad466dfdb706ba65e197ec8257d6c7bce25a/tests/std/tests/prefix.lst#L6

Under the extremely obscure compiler option [`/J`](https://learn.microsoft.com/en-us/cpp/build/reference/j-default-char-type-is-unsigned?view=msvc-170), `char` is treated as an unsigned type. This triggers C4365 signed/unsigned mismatch warnings when assigning promoted `int` to `char` in the `vector<bool>` optimization for `copy()` implemented by #3353 (merged 2023-09-21, shipped in MSVC Build Tools 14.39 / VS 2022 17.9).

Avoiding those warnings is easy, so even though `/J` is rarely used, and we don't officially support being C4365 clean, we may as well silence them and add test coverage.

Enabling the test coverage is significantly more obnoxious but all of the issues were in the test itself and are avoidable with repetitive casts.

### DIFF
--- a/tests/std/tests/GH_000625_vector_bool_optimization/env.lst
+++ b/tests/std/tests/GH_000625_vector_bool_optimization/env.lst
@@ -1,7 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-RUNALL_INCLUDE ..\usual_matrix.lst
+RUNALL_INCLUDE ..\impure_matrix.lst
 RUNALL_CROSSLIST
 *	PM_CL=""
 *	PM_CL="/J"


stl/inc/vector
```diff
@@ -3908,7 +3908,7 @@ _CONSTEXPR20 _OutIt _Copy_vbool(_VbIt _First, _VbIt _Last, _OutIt _Dest) {
             if (_UnalignedFirstBits != 0) {
                 const auto _SourceBitMask = static_cast<char>(UCHAR_MAX << _UnalignedFirstBits);
                 const auto _DestBitMask   = static_cast<char>(UCHAR_MAX >> (8 - _UnalignedFirstBits));
-                *_VbDest_ch               = (*_VbDest_ch & _DestBitMask) | (*_VbFirst_ch & _SourceBitMask);
+                *_VbDest_ch = static_cast<char>((*_VbDest_ch & _DestBitMask) | (*_VbFirst_ch & _SourceBitMask));
```
frederick-vs-ja: Would it be better to use `unsigned char` thoroughly to avoid the signedness issue?

AlexGuteniev: I'm much in favor of this idea, especially given that thoroughness here means about ten lines.

StephanTLavavej: I love `unsigned char` and wouldn't object to using it, but I don't think it would help with the signed/unsigned warnings here. The problem is that the RHS is promoted to `int`, to assigning it to an actual `unsigned char` or a `/J` behavior `char` will trigger a sign conversion.

AlexGuteniev: No, make this cast to `unsigned char` **and** adjust the other casts to `unsigned char`.

StephanTLavavej: Since I haven't queued this up in a merge batch, I guess I can make that change.

davidmrdavid (approved):
LGTM, but will re-approve if/when this is applied: https://github.com/microsoft/STL/pull/6013/changes/655e813d8a22cda3f142e0d3764e45ce66bd576b#r2695480146

StephanTLavavej:
@davidmrdavid I've pushed a commit to use `unsigned char` in these functions. Even if it doesn't help with the warning, I agree that it's conceptually cleaner to remain in the unsigned domain, and to avoid `char` squirrelliness. (I also changed the comments since it was weird to say plain "char", but left the `_ch` naming.)

StephanTLavavej:

StephanTLavavej:
I had to push an additional commit because `/clr:pure` is incompatible with ~~life~~ `/J`.

Issue #5950, opened 2025-12-09
`STL.natvis`: Need visualizers for `flat_multimap` and `flat_multiset`


StephanTLavavej:
The feature branch is currently missing visualizers for `flat_multimap` and `flat_multiset`. We might be able to use "alternative type", or additional work might be required.

Also, I am not super fond of the special cases that have accumulated for `deque`. I consider this to be unusual to the point of not special-casing; `flat_meow` defaults to using `vector` which is way better.

vmichal:
I have fixed problems listed you have listed @StephanTLavavej, but a new strange thing appeared: The visualization of vector iterators and deque iterators is not the same - `deque::iterator` is visualized as a "normal type" whereas `vector::iterator` is visualized like a smart pointer, so its visualization contains braces. Check the attached screenshot visualizing `flat_multimap<int, int, *,*,*>::iterator` for various underlying containers.

<img width="1130" height="177" alt="Image" src="https://github.com/user-attachments/assets/992953bd-90c3-4f18-9321-151c060231f2" />

Is this desirable?

StephanTLavavej:
Those braces are definitely weird and confusing.

vmichal:
Unfortunately, those braces seem consistent with the visualization of `std::span::iterator` and `std::array` and possibly more.
A larger change was required, so I separated it to PR #6019 since it is not tightly related to `flat_map`. However I have tested both changes together and the visualization of `flat_(multi)map` works as expected.

PR #6015, opened 2026-01-16
`<flat_meow>`: Implement natvis visualizers for `flat_multimeow`, simplify visualizer for `flat_map`

vmichal:
Revisit natvis definitions for `flat_meow`: remove unnecessary special cases for `flat_map` without changing the visualization.

- Use the same visualizer for `flat_(multi)set`
- Remove all special cases for `flat_map` and merge them into a single definition.
- Use the same visualizer for `flat_(multi)map`

Towards #5950

### DIFF
--- a/stl/debugger/STL.natvis
+++ b/stl/debugger/STL.natvis
@@ -2325,7 +2325,9 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
   <!-- fallback when the primary visualization (above) fails to parse, e.g. some custom containers -->
   <Type Name="std::flat_map&lt;*&gt;" Priority="MediumLow">
-    <DisplayString>{_Data}</DisplayString>
+    <AlternativeType Name="std::flat_multimap&lt;*&gt;"/>
+    <Intrinsic Name="size" Expression="_Data.keys.size()"/>
+    <DisplayString>{{ size={size()} }}</DisplayString>
     <Expand>
       <Item Name="[comparator]" ExcludeView="simple">_Key_compare</Item>
       <Item Name="[keys]">_Data.keys</Item>


stl/debugger/STL.natvis
```diff
@@ -2404,8 +2323,8 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
     </Expand>
   </Type>
 
-  <!-- fallback for a custom container -->
-  <Type Name="std::flat_map&lt;*&gt;">
+  <!-- fallback when the primary visualization (above) fails to parse, e.g. some custom containers -->
+  <Type Name="std::flat_map&lt;*&gt;" Priority="MediumLow">
```
StephanTLavavej, line 2327: Doesn't this also need an alternative type for `flat_multimap`?

vmichal: Good catch, fixed.

vmichal:
I have updated the fallback visualization of `std::flat_(multi)map` to at least show `size`, which calls `_Data.keys.size()` (there should be no problem parsing that, as we use it in our implementation).

Should the main visualization for `std::flat_(multi)map` fail, it would be more likely because of the use of `operator[]` on key and mapped containers. We don't actively use these functions in `<flat_map>`, so I can imagine a world where the user specializes `std::flat_(multi)map` with a container that lacks `operator[]` and has random-access iterators.

vmichal:
Just to clarify the intent of multiple PRs: The visualization of `<flat_meow>` data structures will additionally improve by accepting two other loosely related changes
- Consistent visualization of comparison objects proposed in #6016 
- Harmonization of other iterators (mostly `vector`, but for consistency it also changes very similar `array` and `span`) and the visualization of their `end()` state #6019

These changes are self-contained, so I extracted them from feature branch and opened a PR to main. There are no adjacent-edit conflicts.

StephanTLavavej:
Wonderful, thank you! :heart_eyes_cat:

PR #6017, opened 2026-01-16
`<flat_map>`: Fix new libcxx test failure due to argument type mismatch when calling `ranges::remove_if` from `flat_map::erase_if`

vmichal:
The newly failing libcxx test `std\containers\container.adaptors\flat.multimap\robust_against_nonbool.compile.pass.cpp` contains a template lambda written in such a way that template argument deduction fails unless it is called precisely with `flat_map::const_reference`
```
[]<class T>(std::pair<const StrictComparable<T>&, const StrictComparable<T>&>) -> BooleanTestable const& { return yes; };
```

Since the predicate given to `flat_map::erase_if` is specified to be callable with `flat_map::const_reference` in [flat.map.erasure]/2, I think the libcxx test is standards compliant.

Adjust the code of `flat_map::erase_if` to satisfy the test.

### DIFF
--- a/stl/inc/flat_map
+++ b/stl/inc/flat_map
@@ -1025,10 +1025,9 @@ private:
     _NODISCARD size_type _Erase_if(_Predicate _Pred) {
         _Clear_guard _Guard{this};
 
-        const auto _New_last =
-            _RANGES remove_if(_View_to_mutate(), _Pred, [](const auto& _Element) _STATIC_CALL_OPERATOR {
-                return const_reference{_Element}; // N5032 [flat.map.erasure]/2 and [flat.multimap.erasure]/2
-            }).begin();
+        // N5032 [flat.map.erasure]/2 and [flat.multimap.erasure]/2
+        const auto _Proj     = [](const auto& _Element) _STATIC_CALL_OPERATOR { return const_reference{_Element}; };
+        const auto _New_last = _RANGES remove_if(_View_to_mutate(), _Pred, _Proj).begin();
         const auto _Old_size = size();
 
         _Data.keys.erase(_New_last._Key_it, _Data.keys.end());


stl/inc/flat_map
```diff
@@ -1025,7 +1025,9 @@ private:
     _NODISCARD size_type _Erase_if(_Predicate _Pred) {
         _Clear_guard _Guard{this};
 
-        const auto _New_last = _RANGES remove_if(_View_to_mutate(), _Pred).begin();
+        const auto _New_last = _RANGES remove_if(_View_to_mutate(), [_Pred](auto const& ref) {
+            return bool(_Pred(const_reference{ref})); // N5032 [flat.map.erasure]/2 and [flat.multimap.erasure]/2
```
frederick-vs-ja: We can make the transformation a projection argument for `ranges::remove_if`. I guess this would be a bit clearer.

stl/inc/flat_map
```diff
@@ -1025,7 +1025,9 @@ private:
     _NODISCARD size_type _Erase_if(_Predicate _Pred) {
         _Clear_guard _Guard{this};
 
-        const auto _New_last = _RANGES remove_if(_View_to_mutate(), _Pred).begin();
+        const auto _New_last = _RANGES remove_if(_View_to_mutate(), [_Pred](auto const& ref) {
```
StephanTLavavej: We conventionally use west const, i.e. `const auto&`.

StephanTLavavej: `ref` is not `_Ugly`.

stl/inc/flat_map
StephanTLavavej: We conventionally avoid functional-style casts like `bool(STUFF)`, because they have C-style cast semantics (perhaps surprisingly). Should this be a `static_cast<bool>`?

stl/inc/flat_map
AlexGuteniev: Actually, it doesn't need to be so, as there's `std::ref`, it is just a tradition I guess.

frederick-vs-ja: Yeah. This was explained in https://github.com/microsoft/STL/pull/3535#discussion_r1125484198.

StephanTLavavej: Yep. And I consider shadowing and quasi-shadowing to be utterly intolerable, especially because they interfere with searching.

stl/inc/flat_map
vmichal: I implemented it that way at first, but then decided that the current version could simplify the compiler's job a bit. I will change it as suggested.

stl/inc/flat_map
```diff
@@ -1025,7 +1025,9 @@ private:
     _NODISCARD size_type _Erase_if(_Predicate _Pred) {
         _Clear_guard _Guard{this};
 
-        const auto _New_last = _RANGES remove_if(_View_to_mutate(), _Pred).begin();
+        const auto _New_last = _RANGES remove_if(_View_to_mutate(), _Pred, [](const auto& _Element) {
```
StephanTLavavej: We conventionally apply `_STATIC_CALL_OPERATOR` to stateless lambdas now.

StephanTLavavej: Style: This is fine, but would be clearer if the projection lambda were extracted.

StephanTLavavej:
Thanks for figuring out how to fix this! :heart_eyes_cat:

PR #6019, opened 2026-01-16
`STL.natvis`: Improve iterator visualizations

vmichal:
Changes to iterator visualization motivated by visualization of `flat_meow`:
The visualizer for `_Deque_iterator` can always detect `== end()` state since the iterator always has access to the underlying container. This is not always the case for `_Span_iterator` and `_Vector_iterator`, but only at non-zero `_ITERATOR_DEBUG_LEVEL`.

This PR adds a branch to the visualizer parsing logic to show the `end` condition at least in the debug build (which is most relevant for debugging after all).

before:
image

after:
image

Towards #5950 (specifically https://github.com/microsoft/STL/issues/5950#issuecomment-3757674279)

### DIFF
--- a/stl/debugger/STL.natvis
+++ b/stl/debugger/STL.natvis
@@ -1851,11 +1851,23 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
     </Expand>
   </Type>
 
-  <Type Name="std::_Span_iterator&lt;*&gt;">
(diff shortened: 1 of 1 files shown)


StephanTLavavej:

StephanTLavavej:
:chart_with_upwards_trend: :smile_cat: :tada:
===

PR #6021, opened 2026-01-17
`<cmath>`: Tolerate 80-bit `long double`

statementreply:
Use the same code for `long double` `lerp` overloads as other math functions. This avoids calling `long double` `_Common_lerp`, which inspects `long double`'s bit representation.


StephanTLavavej:
I don't care about adding support for 80-bit `long double` where it didn't previously work, but this increases consistency with 3-arg `hypot` etc. and is therefore worthwhile. Thanks!

StephanTLavavej:

StephanTLavavej:
:abacus: :bar_chart: :chart_with_upwards_trend:
===

Issue #5957, opened 2025-12-11
`<regex>`: Extend some optimizations for simple loops to branchless but reentrant and branching but non-reentrant loops


muellerj2:
A simple loop is non-reentrant and branchless, but each implemented optimization for simple loops only depends on one of these properties. We should therefore extend these optimizations to such loops that satisfy one property but not the other.

This means that the parser has to detect non-reentrancy and branchlessness independently from each other and set appropriate flags on the loop nodes. Afterwards, the matcher has to be changed to apply an optimization not just to simple loops but also when the flag for the required property is set.

It should certainly be worth it to implement this for branchless-but-reentrant loops, as this extends #5939's stack growth optimization to many more loops such as `[^a]+` in the regex `(a[^a]+)+`. This might greatly reduce stack usage and thus time spent on reallocating the stack frame buffer, but also makes it less likely that regex matching hits the stack limit.

muellerj2:
I'm working on this.

PR #6022, opened 2026-01-17
`<regex>`: Optimize matching of branchless loops

muellerj2:
Towards #5957. This implements recognition of branchless but non-simple (=reentrant) loops and expands the stack growth optimization in the matcher to such loops. The main change is to recognize such loops in the parser and make it set a new flag on such loops; the changes to the matcher are minimal, just requiring two line changes: One to teach it about the flag and the other one to teach the code in the `_N_end_rep` handler for simple/branchless loops about a different opcode used in the `_N_rep` handler for reentrant greedy loops.

This PR does not implement the corresponding change for non-reentrant loops. It would be easy to do so, since the updated `_Compute_loop_simplicity()` already recognizes such loops. But I'm skeptical that the optimization for non-reentrant loops is actually worth it: It avoids a few assignments in exchange for a branch and more code. For this reason, I would like to do some benchmarking first before I allocate a flag bit to non-reentrancy.

Benchmark:

-- | -- | -- | --
bm_match_sequence_of_9a1b/"(?:a\*b)\*"/100 | 24553.6 | 15346 | 1.60
bm_match_sequence_of_9a1b/"(?:a\*b)\*"/200 | 73242.2 | 32470.6 | 2.26
bm_match_sequence_of_9a1b/"(?:a\*b)\*"/400 | 111607 | 73242.2 | 1.52

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -1901,7 +1901,7 @@ private:
     void _Quantifier();
     bool _Alternative();
     void _Disjunction();
-    void _Calculate_loop_simplicity(_Node_base* _Nx, _Node_base* _Ne, _Node_rep* _Outer_rep, const bool _Nonreentrant);
+    void _Calculate_loop_simplicity(_Node_base* _Nx, _Node_base* _Ne, _Node_rep* _Outer_rep, bool _Nonreentrant);
 
     _FwdIt _Pat;
     _FwdIt _End;


stl/inc/regex
```diff
@@ -1900,7 +1901,7 @@ private:
     void _Quantifier();
     bool _Alternative();
     void _Disjunction();
-    void _Calculate_loop_simplicity(_Node_base* _Nx, _Node_base* _Ne, _Node_rep* _Outer_rep);
+    void _Calculate_loop_simplicity(_Node_base* _Nx, _Node_base* _Ne, _Node_rep* _Outer_rep, const bool _Nonreentrant);
```
StephanTLavavej: This declaration shouldn't mark the value parameter as `const`.

stl/inc/regex
```diff
@@ -1279,7 +1279,8 @@ enum _Node_flags : int { // flags for nfa nodes with special properties
     _Fl_class_cl_all_bits = 0x800, // TRANSITION, ABI: GH-5242
     _Fl_begin_needs_w     = 0x100,
     _Fl_begin_needs_s     = 0x200,
-    _Fl_begin_needs_d     = 0x400
+    _Fl_begin_needs_d     = 0x400,
+    _Fl_rep_branchless    = 0x800,
```
StephanTLavavej, line 1283: No change requested: I'm assuming this is ABI-safe (and doesn't even require a `_Parser2` rename) because if old code doesn't set the branchless bit during parsing, then the new matcher will just run the slower but still correct general code, and conversely if the new code sets the branchless bit during parsing, the old matcher will just run the slower but correct general code it had.

StephanTLavavej:

StephanTLavavej:
:palm_tree: :no_entry: :repeat:
===

PR #6024, opened 2026-01-18
`<flat_set>`, `<flat_map>`: compare for equality instead of equivalence for usual types

### LINKED ISSUE #6020: `<flat_set>`: optimize `unique` algorithm call
https://github.com/microsoft/STL/blob/631495006131ec371214f966f034e23197eb831d/stl/inc/flat_set#L708-L716

This could have been vectorized `unique`, but the predicate is on the way.

For `less<>`, `less<T>`, `ranges::less` predicate and for usual types, the equivalence can be replaced with binary equality, and the predicate can be omitted. 

In the case of predicate-less `unique` call for normal underlying container (like, not `deque`) the `unique` will be vectorized using already existing code path.

---

For `flat_map` this could work in principle too, but would effectively require implementing vector algorithms in headers, and would be way more tedious, so that's probably not what we would do, at least in a near future.

AlexGuteniev:
Resolves #6020

# ⚙️ Optimization

For usual types, compare for equality, instead of equivalence:
 - For integers this allows using vectorization path for `<flat_set>`
 - For strings this allows way cheaper comparison, as suggested by @StephanTLavavej 
 - For anything if `ranges::less` or `ranges::greater` are used, as @cpplearner suggested, other types with string-like situation may benefit too

@frederick-vs-ja suggested to take advantage of sorted range, and always perform only one comparison. But we need other `unique` algorithm for that.

# ⏱️ Benchmark results

Benchmark         |  Before |  After  
------------------|---------|---------
flat_map_strings  |  773 ns |  639 ns 
flat_set_strings  |  450 ns |  358 ns 
flat_map_integers | 99.6 ns |  100 ns 
flat_set_integers | 48.0 ns | 47.5 ns 

# 🥈 Results interpretation

Most of the time is spent on sorting the inserted data, yet some is spent on making it unique, so the improvement can be observed, but it is limited.

For strings this played out well.

For integers there's no improvement seen with the current benchmark. `flat_map` of integers is not improved, because the compiler is able to replace equivalence with equality on its own; and even if it wouldn't, the difference would be very minor anyway. 

`flat_set` of integers goes the vectorized path, but the benchmark needs more pieces to make vectorization beneficial. And with more pieces, the sorting would consume even greater portion of time, due to above-linear complexity.

So looks like this is useless for integers, but strings still benefit from this.

### DIFF
--- a/benchmarks/src/flat_meow_assign.cpp
+++ b/benchmarks/src/flat_meow_assign.cpp
@@ -4,6 +4,7 @@
 #include <benchmark/benchmark.h>
 #include <flat_map>
 #include <flat_set>
+#include <string>
 
 using namespace std;
 
@@ -11,8 +12,8 @@ void flat_map_strings(benchmark::State& state) {
     flat_map<string, int> pieces;
     for (auto _ : state) {
         pieces = {{"soldier"s, 1}, {"soldier"s, 2}, {"soldier"s, 3}, {"soldier"s, 4}, {"soldier"s, 5}, {"soldier"s, 6},
-            {"soldier"s, 7}, {"soldier"s, 8}, {"tour"s, 9}, {"horse"s, 10}, {"elephant"s, 11}, {"vizier"s, 12},
-            {"king"s, 13}, {"elephant"s, 14}, {"horse"s, 15}, {"tour"s, 16}};
+            {"soldier"s, 7}, {"soldier"s, 8}, {"tower"s, 9}, {"horse"s, 10}, {"elephant"s, 11}, {"vizier"s, 12},
+            {"king"s, 13}, {"elephant"s, 14}, {"horse"s, 15}, {"tower"s, 16}};
         benchmark::DoNotOptimize(pieces);
     }
 }
@@ -21,7 +22,7 @@ void flat_set_strings(benchmark::State& state) {
     flat_set<string> pieces;
     for (auto _ : state) {
         pieces = {"soldier"s, "soldier"s, "soldier"s, "soldier"s, "soldier"s, "soldier"s, "soldier"s, "soldier"s,
-            "tour"s, "horse"s, "elephant"s, "vizier"s, "king"s, "elephant"s, "horse"s, "tour"s};
+            "tower"s, "horse"s, "elephant"s, "vizier"s, "king"s, "elephant"s, "horse"s, "tower"s};
         benchmark::DoNotOptimize(pieces);
     }
 }


cpplearner:
If the comparator is `ranges::less` or `ranges::greater`, I think we can assume that `_Equivalence_is_equality` is true.

(This is because the `ranges` comparators require [the argument types to model `totally_ordered_with`](https://eel.is/c++draft/range.cmp#9.sentence-1), which requires that [exactly one of `bool(a < b)`, `bool(a > b)`, or `bool(a == b)` is true](https://eel.is/c++draft/concept.totallyordered#1.1) and that [`bool(t < u)` is equal to `bool(u > t)`](https://eel.is/c++draft/cmp.concept#1.2).)

frederick-vs-ja:
> @frederick-vs-ja suggested to take advantage of sorted range, and always perform only one comparison. But we need other `unique` algorithm for that.

Opened #6027 to track this.

stl/inc/flat_map
```diff
@@ -1084,12 +1084,22 @@ private:
         _STL_INTERNAL_CHECK(_Is_sorted_and_unique());
     }
 
+    auto _Erase_dupes_if_not_multi_pred() {
+        if constexpr (_Is_any_of_v<_Compare, _RANGES less, _RANGES greater>
+                      || (_Is_any_of_v<_Compare, less<>, less<_Key>, greater<>, greater<_Key>>
+                          && _Equivalence_is_equality<_Key>) ) {
```
StephanTLavavej: This occurs twice and should be extracted. Also, the `ranges::less` / `ranges::greater` special case needs to be commented, as the reason is not obvious.

stl/inc/flat_set
```diff
@@ -705,13 +705,21 @@ private:
         }
     }
 
+    auto _Erase_dupes_if_not_multi_pred() {
```
StephanTLavavej, line 708: Let's add an internal static assert that we're unique.

stl/inc/flat_map
```diff
@@ -1084,12 +1084,22 @@ private:
         _STL_INTERNAL_CHECK(_Is_sorted_and_unique());
     }
 
+    auto _Erase_dupes_if_not_multi_pred() {
+        if constexpr (_Is_any_of_v<_Compare, _RANGES less, _RANGES greater>
+                      || (_Is_any_of_v<_Compare, less<>, less<_Key>, greater<>, greater<_Key>>
+                          && _Equivalence_is_equality<_Key>) ) {
+            return [](const_reference _Left, const_reference _Right) { return _Left.first == _Right.first; };
```
StephanTLavavej: Stateless lambdas in product code should be marked `_STATIC_CALL_OPERATOR`.

stl/inc/flat_map
```diff
@@ -1084,12 +1084,22 @@ private:
...
     }
 
+    auto _Erase_dupes_if_not_multi_pred() {
+        if constexpr (_Is_any_of_v<_Compare, _RANGES less, _RANGES greater>
+                      || (_Is_any_of_v<_Compare, less<>, less<_Key>, greater<>, greater<_Key>>
+                          && _Equivalence_is_equality<_Key>) ) {
+            return [](const_reference _Left, const_reference _Right) { return _Left.first == _Right.first; };
+        } else {
+            return [this](const_reference _Left, const_reference _Right) {
+                return this->_Key_equal(_Left.first, _Right.first);
```
StephanTLavavej: Pre-existing: `_Key_equal` is a direct member function.

benchmarks/src/flat_meow_assign.cpp
```diff
@@ -0,0 +1,51 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <benchmark/benchmark.h>
+#include <flat_map>
+#include <flat_set>
+
+using namespace std;
+
+void flat_map_strings(benchmark::State& state) {
+    flat_map<string, int> pieces;
```
StephanTLavavej, line 12: Super duper need `<string>`.

StephanTLavavej:
:chess_pawn: :elephant: :prince:
===

PR #6035, opened 2026-01-21
`<flat_map>`: Make `_Pairing_iterator_provider::_Iterator` debugging and unwrapping more consistent with the rest of standard library

### LINKED ISSUE #6036: `<flat_map>`: Shouldn't the non-default `_Pairing_iterator_provider::_Iterator` ctor be private?
Followup on #6034 :
I believe we can (and should) make `_Pairing_iterator_provider::_Iterator` constructor private in the same way as that of `value_compare`.

The following code currently compiles, and I'm not convinced this is intended (note the + 1 added to `values` iterator).
```cpp
using C = flat_map<int, char>;
typename C::key_container_type keys{42, 137, 137, 3337, 0};
typename C::mapped_container_type values{'a', 'g', 'f', 'f', 'r'};
typename C::iterator iter{keys.begin(), values.begin() + 1};
```

In #6035 I have added assertions that would prevent operations with `iter` in the example above. But making the constructor private seems even more robust as it prevents creating such iterators completely (unless we had a bug in `flat_map`).

vmichal:
I extended tests in `Dev10_709168_marking_iterators_as_checked` to include `flat_meow`. This led to test failures that motivated some digging with the following results: The `flat_map` iterator's debug functionality currently looks a bit different from that of other iterators.
- Definition of `_Verify_range` should only exist in `_ITERATOR_DEBUG_LEVEL != 0` case. I have fixed this inconsistency.
- `flat_map<T, U>::iterator` fulfilled `_Unwrappable_v` but not `_Unwrappable_for_unverified_v` even though all underlying iterators (vector or deque iterators) did. I believe this is not intended, so I fixed it by adding member typedef `_Unwrap_when_unverified`.

There is a second slight inconsistency I'm not sure how to handle: The member function `_Verify_offset`:
- is present in `_Array_iterator` only when `_ITERATOR_DEBUG_LEVEL != 0`
- is always present in `_Deque_iterator`, `_Span_iterator`, `_Vector_const_iterator` and `_String_const_iterator` but it is non-empty only when `_ITERATOR_DEBUG_LEVEL != 0`, otherwise it is a no-op
- is always present in `_Pairing_iterator_provider::_Iterator` (for flat_map).

This is not a problem per se, just an inconsistency.

Apart from that, add assertions that `_Key_it - _Right._Key_it == _Mapped_it - _Right._Mapped_it`. I have implemented death tests for these assertions, but breaking invariants of a class requires ugly code, so I'm open to suggestions.

Make the non-default constructor of pairing iterator private to prevent construction outside of `<flat_map>`  code (resolves #6036 ).

### DIFF
--- a/stl/inc/flat_map
+++ b/stl/inc/flat_map
@@ -234,7 +234,11 @@ struct _Pairing_iterator_provider {
             _STD _Seek_wrapped(_Mapped_it, _Dst._Mapped_it);
         }
 
+#ifdef _ENABLE_STL_INTERNAL_CHECK
+    public:
+#else
     private:
+#endif
         _KeyIter _Key_it;
         _MappedIter _Mapped_it;
     };
--- a/tests/std/tests/P0429R9_flat_map/test.cpp
+++ b/tests/std/tests/P0429R9_flat_map/test.cpp
@@ -1503,16 +1503,9 @@ void test_death_different_size_replace() {
 }
 
 // Reach into the pairing iterator and modify one of the underlying iterators to break invariants
-template <class Cont, class Iter>
+template <class Iter>
 void vandalize_pairing_iterator(Iter& pairing_iterator) {
-    using key_iterator_type = Cont::key_container_type::const_iterator;
-    // FIXME static_assert fails for _Vector_const_iterator. Do we have a more elegant way
-    //  to break invariants? Droping these tests is the last option..
-    //  static_assert(is_trivially_copyable_v<key_iterator_type>);
-    key_iterator_type key_iterator_copy;
-    memcpy(static_cast<void*>(&key_iterator_copy), static_cast<void*>(&pairing_iterator), sizeof(key_iterator_type));
-    ++key_iterator_copy;
-    memcpy(static_cast<void*>(&pairing_iterator), static_cast<void*>(&key_iterator_copy), sizeof(key_iterator_type));
+    ++pairing_iterator._Key_it;
 }
 
 enum class iter_test { op_equals, op_spaceship, op_minus };
@@ -1522,7 +1515,7 @@ void test_death_iterator_at_different_offset() {
     using C = conditional_t<type == cont_type::unique, flat_map<int, char>, flat_multimap<int, char>>;
     C cont{{42, 'a'}, {137, 'g'}, {137, 'r'}, {3337, 'f'}, {7755, 'z'}};
     auto broken_iter = cont.begin();
-    vandalize_pairing_iterator<C>(broken_iter);
+    vandalize_pairing_iterator(broken_iter);
     if constexpr (mode == iter_test::op_equals) {
         (void) (cont.end() == broken_iter);
     } else if constexpr (mode == iter_test::op_spaceship) {


stl/inc/flat_map
```diff
@@ -58,18 +58,25 @@ struct _Pairing_iterator_provider {
...
         using difference_type   = ptrdiff_t;
         using value_type        = pair<iter_value_t<_KeyIter>, iter_value_t<_MappedIter>>;
         using reference         = pair<iter_reference_t<_KeyIter>, iter_reference_t<_MappedIter>>;
 
     private:
-        using _Const_iterator = _Pairing_iterator_provider<_KeyIter, _MappedConvIter, _MappedConvIter>::_Iterator;
+        using _Const_iterator     = _Pairing_iterator_provider<_KeyIter, _MappedConvIter, _MappedConvIter>::_Iterator;
+        using _Unwrapped_iterator = _Unwrapped_pairing_iterator<_KeyIter, _MappedIter, _MappedConvIter>;
+        template <class _Key, class _Mapped, class _MappedConv>
+        friend struct _Pairing_iterator_provider;
```
StephanTLavavej, line 69: Ah, this is necessary for `_Unwrapped()`. (Granting friendship to the outer class allows the inner class to benefit, since the inner class is a member and can see everything that the outer class can.) Let's group these friendships next to the constructor for clarity.

stl/inc/flat_map
```diff
@@ -213,10 +237,13 @@ struct _Pairing_iterator_provider {
...
         {
-            using _Unwrapped_iterator = _Unwrapped_pairing_iterator<_KeyIter, _MappedIter, _MappedConvIter>;
             return _Unwrapped_iterator{_STD _Get_unwrapped(_Key_it), _STD _Get_unwrapped(_Mapped_it)};
         }
 
+        static constexpr bool _Unwrap_when_unverified = _Do_unwrap_when_unverified_v<_KeyIter>
+                                                     && _Do_unwrap_when_unverified_v<_MappedIter>
+                                                     && _Do_unwrap_when_unverified_v<_MappedConvIter>;
+
         void _Seek_to(const _Unwrapped_pairing_iterator<_KeyIter, _MappedIter, _MappedConvIter>& _Dst)
```
StephanTLavavej: Now that you're extracting `_Unwrapped_iterator` from `_Unwrapped()`, we should use it in `_Seek_to()`.

stl/inc/flat_map
```diff
@@ -160,6 +173,9 @@ struct _Pairing_iterator_provider {
         }
 
         _NODISCARD difference_type operator-(const _Iterator& _Right) const {
+#if _ITERATOR_DEBUG_LEVEL != 0
+            _Compat(_Right);
+#endif // _ITERATOR_DEBUG_LEVEL != 0
```
StephanTLavavej: If we're always going to define `_Compat`, we should follow `vector`'s example and avoid guarding the calls (which will be optimized out in release mode).

stl/inc/flat_map
StephanTLavavej: Now that you're granting friendship between the `_Iterator`s, we don't need the public `_Key_iterator()` and `_Mapped_iterator()` members which are used only here.

tests/std/tests/P0429R9_flat_map/test.cpp
```diff
@@ -1540,8 +1581,21 @@ int main(int argc, char* argv[]) {
         test_death_different_size_ctor_sorted<cont_type::multi>,
         test_death_different_size_replace<cont_type::unique>,
         test_death_different_size_replace<cont_type::multi>,
+
```
StephanTLavavej: Stray newline.

tests/std/tests/P0429R9_flat_map/test.cpp
```diff
@@ -1502,6 +1502,47 @@ void test_death_different_size_replace() {
...
+void test_death_iterator_at_different_offset_operator_equals() {
+    using C = conditional_t<type == cont_type::unique, flat_map<int, char>, flat_multimap<int, char>>;
+    C cont{{42, 'a'}, {137, 'g'}, {137, 'r'}, {3337, 'f'}, {7755, 'z'}};
+    auto broken_iter = cont.begin();
+    vandalize_pairing_iterator<C>(broken_iter);
+    (void) (cont.end() == broken_iter);
+}
+
+template <cont_type type>
+void test_death_iterator_at_different_offset_three_way() {
```
StephanTLavavej: `operator_spaceship` would be more systematic IMO. Actually, we can just fuse these with `if constexpr` and an `enum class`.

tests/std/tests/P0429R9_flat_map/test.cpp
```diff
@@ -1502,6 +1502,47 @@ void test_death_different_size_replace() {
     cont.replace({42, 137, 137, 3337, 0}, {'a'});
 }
 
+// Reach into the pairing iterator and modify one of the underlying iterators to break invariants
+template <class Cont, class Iter>
+void vandalize_pairing_iterator(Iter& pairing_iterator) {
+    using key_iterator_type = Cont::key_container_type::const_iterator;
+    // FIXME static_assert fails for _Vector_const_iterator. Do we have a more elegant way
+    //  to break invariants? Droping these tests is the last option..
```
StephanTLavavej: `_ENABLE_STL_INTERNAL_CHECK` is probably the way to do this.

PR #6038, opened 2026-01-22
`<flat_map>`: Pass a tinier `value_compare`-like comparator to algorithms

frederick-vs-ja:
The `_Value_compare_to_pass_in_flat_map` wrapper will be returned as-is when passed to `_Pass_fn`, and thus reduces calling layers in some circumstances.

Also, `_Value_compare_to_pass_in_flat_map` is as least templated as possible, which possibly saves throughput.


StephanTLavavej:
Excellent, thank you!!!

PR #6049, opened 2026-01-26
Add Neon implementation of `count`

hazzlim:
This PR adds a manually vectorized implementation of `count` using Neon intrinsics.

<p dir="auto">Performance numbers (values are speedup figures relative 
to existing code; values greater than 1 indicate that the new code is 
faster):</p>

<markdown-accessiblity-table data-catalyst="">

</markdown-accessiblity-table>

<head>

<link rel=File-List

</head>

  | MSVC | Clang
-- | -- | --
bm<uint8_t,   not_highly_aligned_allocator, Op::Count>/8021/3056 | 23.523 | 21.429
bm<uint8_t,   not_highly_aligned_allocator, Op::Count>/63/62 | 3.07 | 3.122
bm<uint8_t,   not_highly_aligned_allocator, Op::Count>/31/30 | 1.745 | 1.472
bm<uint8_t,   not_highly_aligned_allocator, Op::Count>/15/14 | 0.939 | 0.848
bm<uint8_t,   not_highly_aligned_allocator, Op::Count>/7/6 | 0.777 | 0.925
bm<uint16_t,   not_highly_aligned_allocator, Op::Count>/8021/3056 | 11.579 | 10.233
bm<uint16_t,   not_highly_aligned_allocator, Op::Count>/63/62 | 4.062 | 3.826
bm<uint16_t,   not_highly_aligned_allocator, Op::Count>/31/30 | 2.286 | 2.367
bm<uint16_t,   not_highly_aligned_allocator, Op::Count>/15/14 | 1.46 | 1.103
bm<uint16_t,   not_highly_aligned_allocator, Op::Count>/7/6 | 0.8 | 0.909
bm<uint32_t,   not_highly_aligned_allocator, Op::Count>/8021/3056 | 6 | 4.107
bm<uint32_t,   not_highly_aligned_allocator, Op::Count>/63/62 | 3.769 | 2.4
bm<uint32_t,   not_highly_aligned_allocator, Op::Count>/31/30 | 2.595 | 1.593
bm<uint32_t,   not_highly_aligned_allocator, Op::Count>/15/14 | 1.549 | 1.277
bm<uint32_t,   not_highly_aligned_allocator, Op::Count>/7/6 | 1.142 | 1.295
bm<uint64_t,   not_highly_aligned_allocator, Op::Count>/8021/3056 | 1.042 | 1
bm<uint64_t,   not_highly_aligned_allocator, Op::Count>/63/62 | 1 | 0.932
bm<uint64_t,   not_highly_aligned_allocator, Op::Count>/31/30 | 1 | 0.886
bm<uint64_t,   not_highly_aligned_allocator, Op::Count>/15/14 | 0.93 | 0.875
bm<uint64_t,   not_highly_aligned_allocator, Op::Count>/7/6 | 0.905 | 0.88

</body>

</html>

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -5270,7 +5270,7 @@ namespace {
                 }
             }
 
-            if ((_Size_bytes & size_t{0x10}) != 0) {
+            if ((_Size_bytes & size_t{0x10}) != 0) { // use original _Size_bytes; we've read only 32-byte chunks
                 const auto _Comparand = _Traits::_Set_neon_q(_Val);
                 auto _Count_vector    = _Traits::_Set_neon_q(0);
 


AlexGuteniev:
I'm wondering about 64-bit case. I think the compiler auto-vectorization kicked in there, because element size equals to counter size. And for the other bit widths it didn't work.

It might be a hint that we may want to skip manually vectorizing this.
However, it is not slower, so doesn't hurt much either.

The same situation is actually observed for 64-bit case on x64 and for 32-bit case on x86. But for these cases manual vectorization may still be more efficient due to AVX2 dispatch. Also, we don't want to introduce x86 vs x64 difference. So, I think having all four element sizes is more justified for x64 and x86.

hazzlim:
> I'm wondering about 64-bit case. I think the compiler auto-vectorization kicked in there, because element size equals to counter size. And for the other bit widths it didn't work.
> 
> It might be a hint that we may want to skip manually vectorizing this. However, it is not slower, so doesn't hurt much either.
> 
> The same situation is actually observed for 64-bit case on x64 and for 32-bit case on x86. But for these cases manual vectorization may still be more efficient due to AVX2 dispatch. Also, we don't want to introduce x86 vs x64 difference. So, I think having all four element sizes is more justified for x64 and x86.

Yep it's definitely autovec-ed in the same way as the manual code for 64-bit. I don't know if we prefer to keep the manual given that they are ~the same speed?

AlexGuteniev:
I had a PR that helps the compiler auto vectorizing all sizes - see #4673. As you see, it was closed without merging.

Currently my opinion shifted towards manual vectorization preference. One of the factors is that auto vectorization may work for some compiler version, and fall apart for a subsequent one,

Let's wait for maintainer's decision on this.

StephanTLavavej:
In general I would prefer to avoid blends of manual vectorization and autovectorization, for the reason that Alex mentioned - the compiler has displayed varying ability to autovectorize over time. As long as it doesn't involve extreme implementation complexity, uniformly manually vectorizing all sizes seems the most predictable.

stl/src/vector_algorithms.cpp
```diff
@@ -5152,6 +5223,75 @@ namespace {
...
+            const void* _First, const void* const _Last, const _Ty _Val) noexcept {
+            size_t _Result           = 0;
+            const size_t _Size_bytes = _Byte_length(_First, _Last);
+
+            if (size_t _Size = _Size_bytes & ~size_t{0x1F}; _Size != 0) {
+                const auto _Comparand = _Traits::_Set_neon_q(_Val);
+                const void* _Stop_at  = _First;
+
+                for (;;) {
+                    if constexpr (sizeof(_Ty) >= sizeof(size_t)) {
```
StephanTLavavej, line 5238: No change requested: This looked strange since we know `sizeof(size_t)` is always 8 for ARM64, but I see that this is consistent with the x64/x86 codepath where it varies, and keeping the consistency seems more valuable than boiling this down to `sizeof(_Ty) == 8`.

stl/src/vector_algorithms.cpp
```diff
@@ -5152,6 +5223,75 @@ namespace {
...
+                        break;
+                    } else {
+                        if (_Size == 0) {
+                            break;
+                        }
+                    }
+                }
+            }
+
+            if ((_Size_bytes & size_t{0x10}) != 0) { // use original _Size_bytes; we've read only 32-byte chunks
```
StephanTLavavej, line 5273: No change requested: Appreciate the consistent comment! :heart_eyes_cat:

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:mechanical_arm: :vampire_man: :one: :two: :three: :four:
===

PR #6055, opened 2026-01-28
`<regex>`: Fix reentrant loops containing backreferences

muellerj2:
I introduced a subtle bug in #6022: I missed that there are loops that get marked as branchless but can match strings of different lengths, so the value of `_Loop_length` in the loop state can change. This applies to loops with backreferences, since the length of the captured string might have changed when the loop is reentered.

The problem is that `_Loop_length` is used during unwinding of branchless/simple loops, but its value is not restored during backtracking.

I see two ways to fix this:

* The matcher restores `_Loop_length`  like `_Loop_frame_idx` and `_Loop_idx` during backtracking. But there is no unused memory to store this value on the stack, so we would have to add a new frame to the stack or increase the size of each frame. 
* The parser does not mark loops containing backreferences as branchless if they are reentrant (even though they are branchless, strictly speaking).

Given that this is about relatively rare regexes that contain subpatterns like `(prefix(\1)*suffix)*`, the first option seems like a waste of memory pessimizing more common regexes. For this reason, I went for the latter option.

`_N_rep`, `_N_if`, `_N_assert` and `_N_class` (if collating elements are contained) are other node types that can cause a loop to match strings or capturing groups of different lengths. but they are already handled. The remaining node types always match strings of the same length.

While debugging this, I was a bit annoyed that the debugger showed the `_Fl_rep_branchless` flag under the name `_Fl_class_cl_all_bits`, so I moved the flag to another free bit for the `_N_rep` node type in this PR. (We could also reuse mask `0x0004`, since the old parser never set this bit on nodes of type `_N_rep`, but this is probably unnecessarily subtle while we still have plenty of bits available.)

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -1269,18 +1269,18 @@ _INLINE_VAR constexpr unsigned int _Bmp_size  = (_Bmp_max + _Bmp_chrs - 1U) / _B
 _INLINE_VAR constexpr unsigned int _ARRAY_THRESHOLD = 4U;
 
 enum _Node_flags : int { // flags for nfa nodes with special properties
-    _Fl_none              = 0x000,
-    _Fl_negate            = 0x001,
-    _Fl_greedy            = 0x002,
-    _Fl_longest           = 0x008, // TRANSITION, ABI: 0x004 is unused; the parser previously marked some nodes with it
-    _Fl_class_negated_w   = 0x100,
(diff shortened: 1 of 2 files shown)


StephanTLavavej:

StephanTLavavej:
Thanks for breaking and then fixing this! :derelict_house: :hammer_and_wrench: :house_with_garden:

PR #6056, opened 2026-01-28
`<flat_meow>`: Update older tests

JMazurkiewicz:
This PR updates older tests, notably:
* Formatting ranges tests.
  * Some tests were rewritten to use `flat_(multi)meow` instead of `(unordered_)(multi)meow`, since the latter were already used at least once.
* Range machinery tests.
  * Currently, when the underlying range models `contiguous_range`, then `flat_set` does too. Is this intended behaviour, or should we add an extra wrapping iterator ([like libc++](https://github.com/llvm/llvm-project/blob/main/libcxx/include/__flat_set/ra_iterator.h))?

### DIFF
--- a/tests/std/tests/P0896R4_ranges_range_machinery/test.cpp
+++ b/tests/std/tests/P0896R4_ranges_range_machinery/test.cpp
@@ -841,7 +841,11 @@ constexpr bool test_std_container() {
     }
     static_assert(test_empty<T&, true>());
     if constexpr (std::contiguous_iterator<I>) {
-        static_assert(test_data<T&, V*>());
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
> Currently, when the underlying range models `contiguous_range`, then `flat_set` does too.

I'm fine with this. Adding an extra layer doesn't seem to be worth the complexity cost.

StephanTLavavej:
Thanks, great catches! I pushed minor changes:

* Comment implementation-specific iterator strengths.
* Add braces. Pre-existing: Guard `<string_view>`.
* Teach `P0896R4_ranges_range_machinery` about contiguous set-like containers.

PR #6059, opened 2026-01-30
Fix off-by-one error in `__std_get_cvt`

cpplearner:
The issue described in #5942 is genuine, though the test that it adds doesn't seem so.

This PR picks up the necessary change in #5942, and adds a test case that demonstrates the issue under code page 932, where the string `"\xFC\x4B"` got converted to `"\\u{fc}K"` when it should map to `"\u9ED1"` (the Han character "黑").

Closes #5942 
Closes #5977

### DIFF
--- a/stl/src/format.cpp
+++ b/stl/src/format.cpp
@@ -31,7 +31,7 @@ extern "C" [[nodiscard]] __std_win_error __stdcall __std_get_cvt(
             break;
         }
 
-        for (unsigned char _First = _Info.LeadByte[_Idx], _Last = _Info.LeadByte[_Idx + 1]; _First <= _Last; ++_First) {
+        for (unsigned int _First = _Info.LeadByte[_Idx], _Last = _Info.LeadByte[_Idx + 1]; _First <= _Last; ++_First) {
             _Pcvt->_Isleadbyte[_First >> 3] |= 1u << (_First & 0b111u);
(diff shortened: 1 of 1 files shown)


stl/src/format.cpp
```diff
@@ -31,7 +31,7 @@ extern "C" [[nodiscard]] __std_win_error __stdcall __std_get_cvt(
             break;
         }
 
-        for (unsigned char _First = _Info.LeadByte[_Idx], _Last = _Info.LeadByte[_Idx + 1]; _First != _Last; ++_First) {
+        for (unsigned char _First = _Info.LeadByte[_Idx], _Last = _Info.LeadByte[_Idx + 1]; _First <= _Last; ++_First) {
```
StephanTLavavej: Question: Do we have to worry about `_Last` being 255?

cpplearner: None of the code pages I know can have 255 as a lead byte. But I think we should make it safe against wrap around, just in case.

```suggestion
        for (unsigned int _First = _Info.LeadByte[_Idx], _Last = _Info.LeadByte[_Idx + 1]; _First <= _Last; ++_First) {
```

StephanTLavavej:

StephanTLavavej:
:one: :beetle: :smile_cat:
===

PR #6060, opened 2026-01-31
Reject overlong UTF-8 forms in `_Mbrtowc` and add test case

cpplearner:
This PR adds test case on top of #5943, and removes the check for out-of-range code points.

While `_Mbrtowc` is used by both `<locale>` and `<format>` facilities, only the former are affected because the `<format>` functions call the constexpr function `_Decode_utf` if the encoding is UTF-8.

I removed the check for out-of-range values because I think `wch` can't exceed 0xFFFF: there's an early return if the result won't fit in the Basic Multilingual Plane,

https://github.com/microsoft/STL/blob/3f76681e18fa51e1441c9150099b8010c1d18e1c/stl/src/xmbtowc.cpp#L110-L115

### DIFF
--- a/tests/std/tests/VSO_0644691_utf_8_codecvt/test.cpp
+++ b/tests/std/tests/VSO_0644691_utf_8_codecvt/test.cpp
@@ -32,7 +32,7 @@ const encoding_test_char utf_test_cases[] = {
     {"\xED\x9F\xBF", L"\uD7FF"}, // U+D7FF unencoded from the Hangul Jamo Extended-B block,
                                  // last character before UTF-16 illegal region
     {"\xED\xA0\x80", nullptr}, // U+D800 beginning of surrogate range
-    {nullptr, L"\xD800"},
+    {nullptr, L"\xD800"}, //
     {"\xED\xBF\xBF", nullptr}, // U+DFFF end of surrogate range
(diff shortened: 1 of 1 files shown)


tests/std/tests/VSO_0644691_utf_8_codecvt/test.cpp
```diff
@@ -48,6 +48,9 @@ const encoding_test_char utf_test_cases[] = {
     // ^^^ TRANSITION, VSO-653059 ^^^
     {"\xC2\x61", nullptr}, // Wrong number of trailing bytes
     {"\xE0\xA0\x61", nullptr},
+
+    {"\xC0\x80", nullptr}, // overlong form
+    {"\xC0\x81", nullptr},
```
StephanTLavavej: We should also exercise the "needed 2 bytes but saw 3 or 4" codepath.

StephanTLavavej:
Thank you so much for figuring out how to salvage this! :heart_eyes_cat: Now I finally understand what the issue is. I pushed additional test coverage.

StephanTLavavej:

StephanTLavavej:
:heart_eyes_cat: :tada: :rocket:
===

PR #6067, opened 2026-02-03
Added Arm64 feature detections in STL

jaykang10:
Added Arm64 feature detections in STL

FEAT_DotProd (PF_ARM_V82_DP_INSTRUCTIONS_AVAILABLE)
FEAT_I8MM (PF_ARM_V82_I8MM_INSTRUCTIONS_AVAILABLE)
FEAT_SHA3 (PF_ARM_SHA3_INSTRUCTIONS_AVAILABLE)
FEAT_SVE (PF_ARM_SVE_INSTRUCTIONS_AVAILABLE)
FEAT_SVE2 (PF_ARM_SVE2_INSTRUCTIONS_AVAILABLE)
FEAT_SVE2p1 (PF_ARM_SVE2_1_INSTRUCTIONS_AVAILABLE)
FEAT_SVE_SHA3 (PF_ARM_SVE_SHA3_INSTRUCTIONS_AVAILABLE)
FEAT_SVE_AES (PF_ARM_SVE_AES_INSTRUCTIONS_AVAILABLE)
FEAT_SVE_BitPerm (PF_ARM_SVE_BITPERM_INSTRUCTIONS_AVAILABLE)

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -92,7 +92,7 @@ namespace {
     bool _Use_FEAT_BitPerm() noexcept {
         return IsProcessorFeaturePresent(PF_ARM_SVE_BITPERM_INSTRUCTIONS_AVAILABLE);
     }
-#endif //^^^ defined(_M_ARM64) ^^^
+#endif // ^^^ defined(_M_ARM64) ^^^
 
     size_t _Byte_length(const void* const _First, const void* const _Last) noexcept {
         return static_cast<const unsigned char*>(_Last) - static_cast<const unsigned char*>(_First);


jaykang10:
> @jaykang10 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

@microsoft-github-policy-service agree company="Arm"

frederick-vs-ja:
I think we should add them together with certain improvement of vectorization.

jaykang10:
> I think we should add them together with certain improvement of vectorization.

Yep, I agree with it.
@hazzlim I think we don’t need vcruntime support for Arm64 feature detection, and IsProcessorFeaturePresent should be sufficient.
Could you please add explicit feature detection for each case where the algorithms are vectorized for a specific feature?
I think this PR could be used as a reference.

hazzlim:
> > I think we should add them together with certain improvement of vectorization.
> 
> Yep, I agree with it. @hazzlim I think we don’t need vcruntime support for Arm64 feature detection, and IsProcessorFeaturePresent should be sufficient. Could you please add explicit feature detection for each case where the algorithms are vectorized for a specific feature? I think this PR could be used as a reference.

Seems reasonable, I will take this approach. 

From the initial discussion on the STL discord, it sounded like the use of `IsProcessorFeaturePresent` was considered a suitable temporary workaround until runtime ISA detection was added to vcruntime. @StephanTLavavej are you happy with this being the "final-form"?

jaykang10:
> `IsProcessorFeaturePresent`

IsProcessorFeaturePresent is supported by the Windows OS. As you know, SVE requires OS support because its stack manipulation differs from NEON, even if the CPU itself supports the feature. From my perspective, using IsProcessorFeaturePresent would be safer.

AlexGuteniev:
I'm wondering why crypto stuff (FEAT_SHA3, FEAT_SVE_SHA3, FEAT_SVE_AES) would be needed.
The algorithms are general purpose the STL currently doesn't provide any crypto features, as far as I know (it has random and hash, but both are general purpose, not crypto purpose).
Do crypo features include some general purpose instructions?

hazzlim:
I believe that this list comes from the set of features proposed by my team to add to the `vcruntime` (not necessarily _only_ for use by the STL). We were advised that there would be quite a high latency between this being added and being available for use, so we were keen to cover a range of features.

I don't think the crypto features are going to be used in the STL.

StephanTLavavej:
> @StephanTLavavej are you happy with this being the "final-form"?

If you're happy with the perf then I think this is perfectly maintainable, thanks.

AlexGuteniev:
I would like `__isa_available` from VSRuntime solution more, as it will allow testing with features disabled

stl/src/vector_algorithms.cpp
StephanTLavavej, line 1: When creating a new PR, please ensure that your `main` is up to date, and that your feature branch is properly synced. (Ideally, clean commits rebased on top of `main`.) Your branch was relative to `main` as of 2025-10-10 which was pretty old for such a fast-moving file as `vector_algorithms.cpp`. In this case it made no difference, just something to remember for the future!

stl/src/vector_algorithms.cpp
```diff
@@ -5,27 +5,30 @@
...
 
 extern "C" long __isa_enabled;
+
+#ifndef _DEBUG
+#pragma optimize("t", on) // Override /Os with /Ot for this TU
+#endif // !defined(_DEBUG)
 #endif // ^^^ !defined(_M_ARM64EC) ^^^
 
+#if defined(_M_ARM64)
+#include <windows.h>
```
StephanTLavavej: Nitpick: We capitalize `Windows.h` to be consistent with its true casing.

stl/src/vector_algorithms.cpp
```diff
@@ -5,27 +5,30 @@
...
 #include <isa_availability.h>
 
 extern "C" long __isa_enabled;
+
+#ifndef _DEBUG
+#pragma optimize("t", on) // Override /Os with /Ot for this TU
+#endif // !defined(_DEBUG)
 #endif // ^^^ !defined(_M_ARM64EC) ^^^
 
+#if defined(_M_ARM64)
```
StephanTLavavej: There's an ARM64 header section immediately above, where we include `<arm64_neon.h>`. This should be consolidated there.

stl/src/vector_algorithms.cpp
```diff
@@ -52,7 +55,45 @@ namespace {
...
+    }
+
+    bool _Use_FEAT_AES() noexcept {
+        return IsProcessorFeaturePresent(PF_ARM_SVE_AES_INSTRUCTIONS_AVAILABLE);
+    }
+
+    bool _Use_FEAT_BitPerm() noexcept {
+        return IsProcessorFeaturePresent(PF_ARM_SVE_BITPERM_INSTRUCTIONS_AVAILABLE);
+    }
+#endif //^^^ defined(_M_ARM64) ^^^
```
StephanTLavavej: Nitpick: Missing space.

stl/src/vector_algorithms.cpp
```diff
@@ -52,7 +55,45 @@ namespace {
...
+        return IsProcessorFeaturePresent(PF_ARM_SVE2_1_INSTRUCTIONS_AVAILABLE);
+    }
+
+    bool _Use_FEAT_SVE_SHA3() noexcept {
+        return IsProcessorFeaturePresent(PF_ARM_SVE_SHA3_INSTRUCTIONS_AVAILABLE);
+    }
+
+    bool _Use_FEAT_AES() noexcept {
+        return IsProcessorFeaturePresent(PF_ARM_SVE_AES_INSTRUCTIONS_AVAILABLE);
+    }
```
StephanTLavavej, line 90: As mentioned in the discussion, the cryptography stuff is going to be unused by the STL, but I am fine with merging all of this for now, then we can garbage-collect the unused functions when we have the full suite of STL vectorized algorithms for ARM64 completed. These are helpers in an unnamed namespace, so they can be removed without fear of breaking the import lib's compatibility.

StephanTLavavej:
> I would like `__isa_available` from VSRuntime solution more, as it will allow testing with features disabled

It should be possible to add a "Vulcan nerve pinch" function, present in the import lib only when building the STL's tests, that allows us to simulate the features being disabled. (We had a worse version of this for ConcRT ages ago, that one was bad because it was exposed to users.) This can be retrofitted in the future.

hazzlim:
Nit: imperative present tense for commit message + slight typo fix:
```
- Added Arm64 feature detections in STL
+ Add Arm64 feature detection in STL
```

StephanTLavavej:

StephanTLavavej:
Thanks for adding this feature detection and congratulations on your first microsoft/STL commit! :mechanical_arm: :six: :four:

Issue #6069, opened 2026-02-04
`<flat_map>`, `<flat_set>`: Unify `_Emplace_hint()`


StephanTLavavej:
`<flat_map>` and `<flat_set>` have very different implementations of `_Emplace_hint()`. They should be unified (not literally into the same function, but into very similar code patterns modulo the key/mapped difference).

There may be ways to further simplify them (I think there's too much special-casing of `_IsUnique` but I'm not certain).

Issue #6045, opened 2026-01-23
`<flat_map>`: Missing test coverage


StephanTLavavej:
`P0429R9_flat_map` is missing some test coverage, especially for `flat_map`-specific members that aren't present for `flat_multimap`, although the latter is slightly affected.

* `operator[]` (extremely important, obviously)
* `at()` (both success and exceptional cases)
* `emplace()`
* `try_emplace()` (including hint and transparent forms)
* `replace()` (when successful, not just death)

In some cases there's a tiny bit of coverage but not focused testing.

Since we have decent libcxx coverage, I'm not too worried, but `tests/std` is what exercises our more exotic compiler options, and also EDG, so we should ideally have coverage there.

mahtab04:
Hi, I’d like to work on improving test coverage for flat_map as described in this issue. I’ll focus on adding targeted tests for operator[], at(), emplace, try_emplace, and replace, following the existing test patterns.
Please let me know if you have any preferences before I proceed.

StephanTLavavej:
Go for it. Please read [the pull request template](https://github.com/microsoft/STL/blob/da06be8fa38418772506328177b703b28ca75699/.github/pull_request_template.md?plain=1) first.

PR #6000, opened 2026-01-08
`<flat_meow>`: Add tests for all `flat_map` constructors, deduction guides and then some

vmichal:
Changed functionality:
- Fix optimization of heterogeneous `flat_map::erase(OtherKey&&)` when `OtherKey` is not `key_type` .

Tests:
- Test that `flat_meow::insert`, `emplace_hint` and similar functions accepting hint correctly handle correct hint (insertion takes place as close as possible to given hint)
- Test all (I believe) constructors and deduction guides of `flat_map`
- Add death tests for all `_STL_ASSERT`s in both `<flat_map>` and `<flat_set>` - unsorted/duplicate elements in ranges and size mismatch of key and mapped container.
- Test all heterogeneous lookup and erase operations (related to previous work in #5994, see discussion in #5992 ).
- Extract common tests of container requirements and invariants from flat_set test and also apply them to flat_map objects.

Progress towards https://discord.com/channels/737189251069771789/1120873380615295027/1450952672982204468
> Also to exercise every deduction guide. [...] I want to make sure that every way we construct the container adaptors, we establish the invariants (so we actually-sort and actually-unique input when necessary).

PR #6071, opened 2026-02-04
P0429R9 `<flat_map>` and P1222R4 `<flat_set>`

### LINKED ISSUE #2910: P0429R9 `<flat_map>`

### LINKED ISSUE #2912: P1222R4 `<flat_set>`

StephanTLavavej:
* Fixes #2910.
* Fixes #2912.

This is the final merge of our extremely long-running `feature/flat_map` branch (which absorbed `feature/flat_set`) into `main`.

Thanks to all of our amazing contributors who worked on this over the years. I believe the exhaustive list is: @achabense @AlexGuteniev @CaseyCarter @duanqn @duckdoom5 @frederick-vs-ja @JMazurkiewicz @Nerixyz @Saalvage @seishun @vmichal

I counted 85 nontrivial PRs that went into this, starting in June 2023. There were an additional 57 PRs that were marked as https://github.com/microsoft/STL/labels/uncharted because they were merging `main` into the feature branch, generally with unremarkable merge conflict resolutions.

<details><summary>Click to expand significant PRs:</summary>

* #3817
* #3833
* #3896
* #3902
* #3993
* #4018
* #4019
* #4034
* #4050
* #4064
* #4079
* #4084
* #4112
* #4132
* #4148
* #4156
* #4315
* #4337
* #4344
* #4368
* #4369
* #4386
* #4468
* #4566
* #4590
* #4650
* #4778
* #4828
* #4887
* #4967
* #4969
* #5045
* #5306
* #5353
* #5357
* #5414
* #5419
* #5422
* #5451
* #5454
* #5455
* #5460
* #5465
* #5474
* #5499
* #5508
* #5514
* #5516
* #5520
* #5522
* #5702
* #5882
* #5926
* #5935
* #5938
* #5952
* #5953
* #5972
* #5986
* #5987
* #5988
* #5989
* #5990
* #5994
* #5996
* #5999
* #6000
* #6001
* #6008
* #6015
* #6017
* #6024
* #6033
* #6034
* #6035
* #6038
* #6040
* #6042
* #6044
* #6046
* #6051
* #6056
* #6057
* #6068
* #6070
</details>

<details><summary>Click to expand boring merge PRs:</summary>

* #3876
* #3947
* #4030
* #4044
* #4089
* #4205
* #4207
* #4262
* #4266
* #4269
* #4307
* #4325
* #4326
* #4352
* #4353
* #4396
* #4397
* #4482
* #4483
* #4580
* #4581
* #4601
* #4602
* #4690
* #4691
* #4735
* #4736
* #4774
* #4832
* #4833
* #4891
* #4892
* #4948
* #4949
* #5021
* #5052
* #5055
* #5085
* #5189
* #5250
* #5258
* #5305
* #5339
* #5426
* #5486
* #5515
* #5585
* #5648
* #5680
* #5697
* #5722
* #5789
* #5884
* #5915
* #5958
* #6014
* #6065
</details>

I've filed a couple of issues for followup, which shouldn't block the merge:

* #6045
* #6069

### DIFF
--- a/stl/debugger/STL.natvis
+++ b/stl/debugger/STL.natvis
@@ -2256,6 +2256,49 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
     </Expand>
   </Type>
 
+  <Type Name="std::extents&lt;*&gt;">
+    <DisplayString>{{ rank={_Rank}, rank_dynamic={_Rank_dynamic} }}</DisplayString>
+    <Expand>
+      <!-- Mixed static and dynamic extents -->
+      <CustomListItems Condition="_Rank != _Rank_dynamic" Optional="true">
+        <Variable Name="idx" InitialValue="0" />
+        <Variable Name="statptr" InitialValue="(const rank_type*) _Static_extents._Elems" />
+        <Variable Name="dynptr" InitialValue="($T1*) _Array._Elems" />
+        <Loop>
+          <Break Condition="idx == _Rank" />
+          <If Condition="*statptr == -1">
+            <Item Name="[{idx}] dynamic extent">*dynptr</Item>
+            <Exec>++dynptr</Exec>
+          </If>
+          <Item Condition="*statptr != -1" Name="[{idx}] static extent">($T1) *statptr</Item>
+          <Exec>++idx, ++statptr</Exec>
+        </Loop>
+      </CustomListItems>
+
+    <!-- Fully dynamic extents -->
+    <CustomListItems Condition="_Rank == _Rank_dynamic" Optional="true">
+      <Variable Name="idx" InitialValue="0" />
+      <Variable Name="dynptr" InitialValue="($T1*) _Array._Elems" />
+      <Loop>
+        <Break Condition="idx == _Rank_dynamic" />
+        <Item Name="[{idx}] dynamic extent">*dynptr</Item>
+        <Exec>++idx, ++dynptr</Exec>
+      </Loop>
+    </CustomListItems>
+
+    <!-- Fully static extents -->
+    <CustomListItems Condition="_Rank_dynamic == 0" Optional="true">
+      <Variable Name="idx" InitialValue="0" />
+      <Variable Name="statptr" InitialValue="(const rank_type*) _Static_extents._Elems" />
+      <Loop>
+        <Break Condition="idx == _Rank" />
+        <Item Name="[{idx}] static extent">($T1) *statptr</Item>
+        <Exec>++idx, ++statptr</Exec>
+      </Loop>
+    </CustomListItems>
+    </Expand>
+  </Type>
+
   <Type Name="std::flat_map&lt;*&gt;">
    <AlternativeType Name="std::flat_multimap&lt;*&gt;"/>
     <Intrinsic Name="size" Expression="_Data.keys.size()"/>


benchmarks/src/flat_meow_assign.cpp
```diff
@@ -0,0 +1,52 @@
...
+        benchmark::DoNotOptimize(pieces);
+    }
+}
+
+BENCHMARK(flat_map_strings);
+BENCHMARK(flat_set_strings);
+BENCHMARK(flat_map_integers);
+BENCHMARK(flat_set_integers);
+
+BENCHMARK_MAIN();
```
zacklj89, line 63: Nonblocking - Looking at #6024 , is this benchmark just used as a micro-sanity-check?

Maybe I'm misinterpreting because I didn't step into it or just haven't reviewed a benchmark in a while, but this seems like in the string case we might be benchmarking initializer list construction as well?

stl/inc/flat_map
```diff
@@ -0,0 +1,1612 @@
...
+    _NODISCARD const_iterator begin() const noexcept {
+        _STL_INTERNAL_STATIC_ASSERT(random_access_iterator<const_iterator>);
+        _STL_INTERNAL_STATIC_ASSERT(convertible_to<iterator, const_iterator>);
+        return const_iterator{_Data.keys.cbegin(), _Data.values.begin()};
+    }
+    _NODISCARD iterator end() noexcept {
+        return iterator{_Data.keys.cend(), _Data.values.end()};
+    }
+    _NODISCARD const_iterator end() const noexcept {
+        return const_iterator{_Data.keys.cend(), _Data.values.end()};
```
zacklj89: Should we be using `cend` here and `cbegin` on line 571?

stl/inc/flat_map
```diff
@@ -0,0 +1,1612 @@
...
+    _NODISCARD conditional_t<_IsUnique, pair<iterator, bool>, iterator> _Try_emplace(
+        _OtherKey&& _Key_val, _MappedArgTypes&&... _Mapped_args) {
+        _STL_INTERNAL_STATIC_ASSERT(is_constructible_v<mapped_type, _MappedArgTypes...>);
+        _STL_INTERNAL_STATIC_ASSERT(is_same_v<remove_cvref_t<_OtherKey>, key_type>
+                                    || (is_constructible_v<key_type, _OtherKey> && _Transparent<key_compare>) );
+
+        const auto _Key_it = _STD upper_bound(_Data.keys.begin(), _Data.keys.end(), _Key_val, _Pass_key_comp());
+        const auto _Index  = _Key_it - _Data.keys.begin();
+        if constexpr (_IsUnique) {
+            if (_Key_it != _Data.keys.begin() && !_Compare_keys(*(_Key_it - 1), _STD forward<_OtherKey>(_Key_val))) {
```
zacklj89: Does `forward` have any effect here today with the current `_Compare_keys` signature?

stl/inc/flat_map
```diff
@@ -0,0 +1,1612 @@
...
+
+protected:
+    template <class _OtherKey, class... _MappedArgTypes>
+    _NODISCARD conditional_t<_IsUnique, pair<iterator, bool>, iterator> _Try_emplace(
+        _OtherKey&& _Key_val, _MappedArgTypes&&... _Mapped_args) {
+        _STL_INTERNAL_STATIC_ASSERT(is_constructible_v<mapped_type, _MappedArgTypes...>);
+        _STL_INTERNAL_STATIC_ASSERT(is_same_v<remove_cvref_t<_OtherKey>, key_type>
+                                    || (is_constructible_v<key_type, _OtherKey> && _Transparent<key_compare>) );
+
+        const auto _Key_it = _STD upper_bound(_Data.keys.begin(), _Data.keys.end(), _Key_val, _Pass_key_comp());
```
zacklj89, line 891: Maybe a learning comment for me - this looks correct at a review glance, but I think the other data structures use `lower_bound`. Am I missing something with respect to this?

stl/inc/flat_map
```diff
@@ -0,0 +1,1612 @@
...
+public:
+#else
+private:
+#endif
+    _NODISCARD bool _Is_sorted_and_unique() const {
+        return _Is_sorted_and_unique(_Data.keys.begin(), _Data.keys.end());
+    }
+
+private:
+    _NODISCARD bool _Is_sorted_and_unique(
```
zacklj89: Not requesting a change (and I can't load the papers or any reference website right now?) but this function name was just taken as a merge of the implementation meaning

"In a unique container, this checks sorted+unique, but in a non-unique one, uniqueness is irrelevant"?

zacklj89 (approved):
Super great job by everyone involved, and STL as always thanks for making the reviews easy regardless of the massive amount of code. 

I'm sure I missed something. I'm approving for now as it seems like this is solid enough to get out of the feature branch.

stl/inc/flat_map
StephanTLavavej: Correct. I can see how this naming is slightly confusing. In the test code, we say "is sorted maybe unique" which is probably better:

https://github.com/microsoft/STL/blob/9cb3a5c0bb115eb054866eda65f25c5ea338d66b/tests/std/include/test_container_requirements.hpp#L124

I can rename this.

stl/inc/flat_map
StephanTLavavej: Great catch!! It has no effect because we take `const T&`. I'll drop this `forward`, which is the only occurrence I see.

stl/inc/xutility
```diff
@@ -1677,6 +1677,9 @@ concept _Allocator_for_container = requires(_Ty& _Alloc) {
     _Alloc.deallocate(_Alloc.allocate(size_t{1}), size_t{1});
 };
 
+template <class _Ty>
+concept _Not_allocator_for_container = !_Allocator_for_container<_Ty>;
```
davidmrdavid, line 1681: is this used? don't see it in this file (or CRTL+F fails me), though it's possibly in a collapsed segment of the PR

tests/std/include/test_container_requirements.hpp
```diff
@@ -0,0 +1,207 @@
...
+#include <type_traits>
+#include <utility>
+
+// Extracted common functionality from flat_map and flat_set tests.
+// May be extended for other containers if needed.
+
+template <class T>
+void assert_container_requirements(const T& s) {
+    T m = s;
+    assert(m == s);
```
davidmrdavid, line 21: non-blocking observation - I wish we named these with something more evocative than `m` and `s`, they make my reading a bit more difficult, though this is just test code so it's not important.

benchmarks/src/flat_meow_assign.cpp
StephanTLavavej: Great catch! Yes, this makes an observable difference:

Benchmark             |    Time
----------------------|---------
flat_map_strings_ORIG |  950 ns
flat_map_strings      |  716 ns
flat_set_strings_ORIG |  583 ns
flat_set_strings      |  506 ns
flat_map_integers     |  114 ns
flat_set_integers     | 76.8 ns

I'll refactor. We don't want to pollute the results with the temporary `std::string` constructions. (Actually constructing an `initializer_list` is basically free once we have the values, so this isn't a concern for the integers IMO.)

stl/inc/flat_map
StephanTLavavej: It doesn't matter because these `const` overloads will select the `const` overloads for the `values` container, so `_Data.values.meow()` and `_Data.values.cmeow()` are identical. I can see how this is visually confusing, though, so I can change it.

StephanTLavavej: Changing them makes them exactly match `flat_map`'s `cbegin()` and `cend()` which is a nice bonus.

tests/std/tests/P0429R9_flat_map_ms_specific/test.compile.pass.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include <deque>
+#include <flat_map>
+#include <functional>
+#include <memory>
+#include <type_traits>
+#include <vector>
+
+using namespace std;
+
+// Test MSVC STL-specific SCARY-ness
```
davidmrdavid, line 14: TIL about "scary iterators" :-)  -> https://devblogs.microsoft.com/cppblog/what-are-scary-iterators/

nit - it took me a minute, and some luck, to figure out what this comment was trying to say. I'd recommend a simple link or even just the expanded acronym as a hint to future maintainers.

stl/inc/flat_map
StephanTLavavej: This is an excellent question. It's what the Standardese depicts, WG21-N5032 \[flat.map.modifiers\]/17:

> *Effects:* If the map already contains an element whose key is equivalent to `k`, `*this` and `args...` are unchanged. Otherwise equivalent to:
> ```cpp
> auto key_it = ranges::upper_bound(c.keys, k, compare);
> auto value_it = c.values.begin() + distance(c.keys.begin(), key_it);
> c.keys.insert(key_it, std::forward<decltype(k)>(k));
> c.values.emplace(value_it, std::forward<Args>(args)...);
> ```

But this is `_Try_emplace()`, and *only* `flat_map` has `try_emplace()`. `flat_multimap` doesn't, because there's no need to "try" - it handles multiple equivalent keys, so it can unconditionally emplace.

So why is `flat_map::try_emplace()` depicted as inserting before the `upper_bound` when no equivalent key is present, when the `lower_bound` would be literally identical? (When there are no equivalent keys, `equal_range` is empty and so `lower_bound` and `upper_bound` are identical, i.e. the first place and the last place an insertion could happen without violating the ordering are the same.)

Observe that `flat_map::emplace()` is specified with very similar wording, including the `upper_bound`. \[flat.map.modifiers\]/2:

> `template<class... Args> constexpr pair<iterator, bool> emplace(Args&&... args);`
> *Effects:* Initializes an object `t` of type `pair<key_type, mapped_type>` with `std::forward<Args>(args)...`; if the map already contains an element whose key is equivalent to `t.first`, `*this` is unchanged. Otherwise, equivalent to:
> ```cpp
> auto key_it = ranges::upper_bound(c.keys, t.first, compare);
> auto value_it = c.values.begin() + distance(c.keys.begin(), key_it);
> c.keys.insert(key_it, std::move(t.first));
> c.values.insert(value_it, std::move(t.second));
> ```

Again this makes no difference for `flat_map` (when there's an equivalent key, the container is unchanged; otherwise the bounds are equal again).

But this *does* matter for `flat_multimap`, which shares the Standardese for `emplace`. \[flat.multimap.overview\]/4:

> Except as otherwise noted, operations on `flat_multimap` are equivalent to those of `flat_map`, except that `flat_multimap` operations do not remove or replace elements with equal keys.

This means that `flat_map::emplace`'s wording applies to `flat_multimap::emplace` too, where it makes a difference. (I believe there's an editorial oopsy here, the wording isn't quite sufficient to say "actually, the bit about skipping any work when an equivalent key is present should be ignored for `flat_multimap`", but it is clearly 100% intended.)

Aside: But why `upper_bound` for multi-emplace? \[associative.reqmts.general\]/54: "If a range containing elements equivalent to `t` exists in `a_eq`, `t` is inserted at the end of that range." This means that repeated emplacement (and insertion, /68) will insert elements in the order that they're added (instead of reverse order), which is intuitively expected.

So:

* `flat_map::emplace` is specified with `upper_bound` because it's needed for `flat_multimap::emplace`.
* `flat_map::try_emplace` is specified with `upper_bound` for consistency, even though `flat_multimap` lacks it.

As this is what the Standardese depicts, I think that no comment is warranted here. Thanks for leading me to look this up, it reinforced something I was vaguely aware of.

stl/inc/flat_map
```diff
@@ -0,0 +1,1612 @@
...
+            _NODISCARD const reference* operator->() const noexcept {
+                return _STD addressof(_Ref);
+            }
+
+        private:
+            friend _Iterator;
+
+            explicit _Arrow_proxy(const _Iterator& _Iter) : _Ref{*_Iter} {}
+
+            reference _Ref;
```
davidmrdavid, line 84: not important, but these newlines seem accidental, are they?

```suggestion
            friend _Iterator;
            explicit _Arrow_proxy(const _Iterator& _Iter) : _Ref{*_Iter} {}
            reference _Ref;
```

stl/inc/xutility
StephanTLavavej: `<xutility>` is one of the STL's central internal headers. `_Not_allocator_for_container` is defined in this central location for use by both `flat_map` and `flat_set`:

https://github.com/microsoft/STL/blob/9cb3a5c0bb115eb054866eda65f25c5ea338d66b/stl/inc/flat_map#L1442
https://github.com/microsoft/STL/blob/9cb3a5c0bb115eb054866eda65f25c5ea338d66b/stl/inc/flat_set#L882

Generally, you should rely on repo-wide search (Ctrl+Shift+F) instead of single-file search for anything that might be non-locally used.

tests/std/tests/P0429R9_flat_map_ms_specific/test.compile.pass.cpp
StephanTLavavej: This is extensively pre-existing:
https://github.com/microsoft/STL/blob/a690f8442de28e1bd1c461ad27279c591954482f/stl/inc/vector#L608
https://github.com/microsoft/STL/blob/a690f8442de28e1bd1c461ad27279c591954482f/stl/inc/xmemory#L795-L797
Maybe `<xmemory>` would be a good place for a comment.

stl/inc/xutility
StephanTLavavej: This inspired me to go use `_Not_allocator_for_container` in existing code for consistency.

tests/std/tests/P0429R9_flat_map_ms_specific/test.compile.pass.cpp
StephanTLavavej: Added a comment to `<xmemory>`.

stl/inc/flat_map
StephanTLavavej: I guess it seems kinda bloated because it's such a small (nested) class, but they are conceptually separate "sections": Friendship, constructor, data member. We generally separate these things with newlines as they're pretty different. I would prefer to leave this as-is.

tests/std/include/test_container_requirements.hpp
StephanTLavavej: I'll agree with this but I don't have better ideas and I agree it's not really important. I would accept a followup to rename if anyone has better ideas for more distinctive names.

StephanTLavavej:

StephanTLavavej:
I pushed a commit to work around VSO-2714784 and VSO-2714798 (ICEs in the internal chk builds of the compiler).

StephanTLavavej:
Resolved a trivial adjacent-add conflict with #6053 in `STL.natvis`.

psiha:
Hmm ~2 and half years...
[~2 and a half _days_ with Claude](https://github.com/psiha/vm/pull/74) ... scary stuff.

And IMO 'almost strictly superior' with 50% less code even including extensions (like merge, reserve, shrink_to_fit...), almost all logic for all 4 of the containers handled by one CRTP base (the modern deducing-this version), with maximum both source and binary code reuse (i.e. bloat minimization, with low level tricks for library-level SROA optimizations)...(even bigger difference if you only count 'actual' code, i.e. skip the onliner forwarders). You may want to check it out before committing to an ABI.

pps83:
Hi, any idea when these are going to be released?

I use vs2026 18.4.1.
Release notes list this: https://github.com/microsoft/STL/wiki/Changelog#msvc-build-tools-1451 flat_map/flat_set, but these aren't available even if I use `/std:c++23preview` or `/std:c++latest` compiler options (with preview MSVC tools).

AlexGuteniev:
Unfortunately there's no longer tracking of individual previews, so some of listed in changelog features may be available in currently latest preview, whereas other wouldn't.  This feature isn't available yet. The latest preview has `_MSVC_STL_UPDATE` set to  202601L, and the feature is expected between 202602L and 202603L.

Also please observe that you need to use Preview toolset for early features access. See Note 1 in  [Visual Studio Installer Prerequisites
](https://github.com/microsoft/STL?tab=readme-ov-file#visual-studio-installer-prerequisites) section of README.md. Or, if you are waiting for production feature availability, you don't need any special steps, but that would also take way longer to pop out.

pps83:
> you need to use Preview toolset

Yes, I know, I used it and I saw there is a new property in settings to select it.
I also do `dir /s /b flat_map` regularly to see if it's there :)

StephanTLavavej:
Please note that if you use the stable IDE to install the Preview Build Tools, that may result in an older version than the Insiders IDE installing the Preview Build Tools. (Experimentally verified a few days ago; I don't know if this will continue to be a possibility moving forward.) I recommend using the Insiders IDE to be certain you're getting the newest Preview Build Tools.

I merged `<flat_map>` into the MSVC repo at the same time as updating `_MSVC_STL_UPDATE` to `202602` (this was internal MSVC-PR-707265), on Feb 11. I swear I've seen compiler fixes ship in Preview that were merged after Feb 11 (the current cutoff was slightly before Feb 24, I believe), so I myself am confused as to why the STL is still identifying itself as January. I've sent a message to our internal engineering systems channel, asking if perhaps the compiler front-end is somehow shipping without updated library machinery.

StephanTLavavej:
Ok, I got to the bottom of this mystery. What's currently shipping in Preview are sources from Feb 4, predating the `<flat_map>` merge by a week. I thought we were shipping newer sources because two compiler bugs had been fixed (on Feb 10 and Feb 17) and I was able to remove workarounds for them, but those bugs were assertions in the internal "chk" builds of the compiler. So while their fixes hadn't shipped in Preview, that didn't matter (Preview is the retail "ret" compiler, unaffected by such assertions).

We've verified that `<flat_map>` is present (and `_MSVC_STL_UPDATE` is `202603`, March) in an upcoming build of the Preview Build Tools which should be appearing in VS Insiders in the near future. Thanks for your patience.

pps83:
> We've verified that `<flat_map>` is present (and `_MSVC_STL_UPDATE` is `202603`, March) in an upcoming build of the Preview Build Tools which should be appearing in VS Insiders in the near future. Thanks for your patience.

I think there were already 2-3 updates in vs2026 in March, still not flat stuffs.

Separately, while reading updates notes I've see some compiler fixes related to for-range handling of some new sematic (but cannot recall/find it anymore). I've seen [possibly related](https://developercommunity.visualstudio.com/t/VS2026-compiler-has-weird-bugs-in-range/11064925) similar problems: quite strange bugs from compiler, were basic for-range loop triggers `error C4723: potential divide by 0` even though no division is used, or telling me that i cannot use `size_t` in for range and should use `unsinged int`, while actual element of the for-range is actually `unsigned`.

StephanTLavavej:
> I think there were already 2-3 updates in vs2026 in March, still not flat stuffs.

`<flat_map>` just shipped in the latest Insiders / Preview update:

```
C:\Temp>"%ProgramFiles%\Microsoft Visual Studio\18\Insiders\VC\Auxiliary\Build\vcvarsall.bat" x64 -vcvars_ver=preview
** Visual Studio 2026 Developer Command Prompt v18.5.0-insiders
** Copyright (c) 2026 Microsoft Corporation
[vcvarsall.bat] Environment initialized for: 'x64'

C:\Temp>cl
Microsoft (R) C/C++ Optimizing Compiler Version 19.51.36223.2 for x64 (PREVIEW)

usage: cl [ option... ] filename... [ /link linkoption... ]

C:\Temp>type meow.cpp
```
```cpp
#include <flat_map>
#include <print>
#include <string_view>
using namespace std;

int main() {
    flat_map<int, string_view> fm{{11, "juliette"sv}, {22, "andromeda"sv}, {33, "meow"sv}};
    println("{}", fm);
}
```
```
meow.cpp
{11: "juliette", 22: "andromeda", 33: "meow"}
```

> Separately, while reading updates notes I've see some compiler fixes

This is off-topic. You can join the STL Discord and chat in the off-topic channels about MSVC Compiler issues.

PR #6072, opened 2026-02-05
`<iomanip>`: Add missing Mandates to `get_money`/`put_money`

frederick-vs-ja:
MSVC STL currently accepts the following example which violates [[ext.manip]/2](https://eel.is/c++draft/ext.manip#2) and [[ext.manip]/5](https://eel.is/c++draft/ext.manip#5). [Godbolt link](https://godbolt.org/z/aTdoKWMnG).

```C++
#include <functional>
#include <iomanip>
#include <istream>
#include <ostream>
#include <string>

void test(std::istream& is, std::ostream& os) {
    struct xstr : std:: string {};
    xstr x;

    is >> std::get_money(x);
    os << std::put_money(x);

    long double y{};
    auto ref_y = std::ref(y);
    is >> std::get_money(ref_y);
    os << std::put_money(ref_y);
}
```

We need to require that `_Money` is exactly `long double` or `basic_string`, not another convertible/deducible type.

### DIFF
--- a/stl/inc/iomanip
+++ b/stl/inc/iomanip
@@ -95,11 +95,15 @@ struct _Monobj { // store reference to monetary amount
 
 _EXPORT_STD template <class _Money>
 _NODISCARD _Monobj<_Money> get_money(_Money& _Val_arg, bool _Intl_arg = false) {
+    static_assert(is_same_v<_Money, long double> || _Is_specialization_v<_Money, basic_string>,
+        "_Money must be either long double or a basic_string specialization (N5032 [ext.manip]/2).");
     return _Monobj<_Money>(_Val_arg, _Intl_arg);
 }
 
 _EXPORT_STD template <class _Money>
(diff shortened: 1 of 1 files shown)


achabense:
> Mandates: The type moneyT is either long double or a specialization of the basic_string template ([[strings]](https://eel.is/c++draft/strings))[.](https://eel.is/c++draft/ext.manip#5.sentence-1)

(No change requested) Though this is mandated by the standard, I'd say the change will fail a lot of code (which should have failed in the past). A lot of non-serious projects use `const char*`, `double` and integers etc.
https://github.com/search?q=put_money%28+lang%3AC%2B%2B&type=code

StephanTLavavej:

StephanTLavavej:
Thanks for noticing and fixing this missing enforcement! :money_mouth_face: :money_with_wings: :dollar:

PR #6074, opened 2026-02-05
Use `extern "C++"` scopes for explicit specializations after P2615R1 Meaningful Exports

StephanTLavavej:
WG21-P2615R1 Meaningful Exports forbids `extern "C++"` from being directly applied to an explicit specialization. Now, every *declaration* is either a *name-declaration* or a *special-declaration*. An *explicit-specialization* is considered a *special-declaration*. But when directly applying `extern "C++"`, the grammar for *linkage-specialization* requires that the `extern` *string-literal* *name-declaration* syntax cannot accept a *special-declaration*.

Clang 17 [implemented](https://clang.llvm.org/cxx_status.html#cxx20) P2615R1 as a DR in C++20 mode, but as of Clang 20 they don't enforce this rule (yet?). @xiangfan-ms brought this to my attention as he was implementing P2615R1 in MSVC.

The fix is simple: use `extern "C++" { ... }` scopes. 

I am allowing the `_Codecvt_guard` helper to permanently remain in this scope, as it's not harmful, and I didn't think it was worth the effort to pull outside of the scope. (Indeed, right now the entire STL is marked with `extern "C++"` as a workaround.)

### DIFF
--- a/stl/inc/xlocale
+++ b/stl/inc/xlocale
@@ -803,7 +803,9 @@ _NODISCARD int _Codecvt_do_length(
 
 enum _Codecvt_mode { _Consume_header = 4, _Generate_header = 2 };
 
-extern "C++" template <>
+extern "C++" {
+
+template <>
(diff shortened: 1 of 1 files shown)


StephanTLavavej:

Issue #2713, opened 2022-05-11
STL: lack of test coverage for differing `/Zc:alignedNew`


CaseyCarter:
In what language modes does the STL support `/Zc:alignedNew` and/or `/Zc:alignedNew-`? 

| language mode | `/Zc:alignedNew` | `/Zc:alignedNew-` | Note |
| :--- | --- | --- | :-- |
| C++14 | ??? | Yes |`clang-cl` (but not `cl`) allows `/std:c++14 /Zc:alignedNew`|
| C++17 | Yes | Maybe ||
| C++20 | Yes | Maybe ||
| C++23 | Yes | Maybe ||

Once this is decided, we should:
1. Enforce that `__cpp_aligned_new` has a supported value for the current language mode if there are unsupported combinations. 
1. Add test coverage if we decide to support any of the non-Yes combinations in the table above (we already cover all of the `Yes`es). At least "C++14 with aligned new" (if we choose to support that) and "C++latest without aligned new".

StephanTLavavej:
We talked about this at the weekly maintainer meeting and the plan is:
* Don't support enabling `/Zc:alignedNew` in C++14 mode (which only clang-cl allows) - reject it with an error
* Support disabling this (with `/Zc:alignedNew-`) in C++17/20/23 mode.
* Test this, but only bother with testing C++23 `/Zc:alignedNew-` since the behavior almost certainly wouldn't vary in C++17/20 (when machinery exists in all modes, typically the only behavior variation is `constexpr` and very occasionally return types).

PR #6087, opened 2026-02-15
Make most tests pass with `/Zc:alignedNew-`

AlexGuteniev:
Towards #2713 whatever it now means.

Make sure tests pass with `/Zc:alignedNew-`, and then revert matrices back.

The following C++17 features do not pass:
 * P0035R4_over_aligned_allocation. Well, that's it.
 * P0220R1_any. I don't know if the feature itself works. Test relies on libc++ which does not assume any escape hatches.
 * P0220R1_polymorphic_memory_resources. It appears to be partially broken both as a feature and the tests.
   - Additionally, a part of `<generator>` test that uses pmr had to be disabled.

### DIFF
--- a/tests/std/tests/P0035R4_over_aligned_allocation/test.cpp
+++ b/tests/std/tests/P0035R4_over_aligned_allocation/test.cpp
@@ -3,7 +3,7 @@
 
 #if !defined(__cpp_aligned_new)
 #error overaligned allocation does not work with /Zc:alignedNew-
-#endif 
+#endif
 
 #define _ENABLE_EXTENDED_ALIGNED_STORAGE
 #define _HAS_DEPRECATED_TEMPORARY_BUFFER 1


tests/std/tests/GH_005504_avoid_function_call_wrapping/test.cpp
```diff
@@ -47,6 +48,7 @@ void operator delete(void* const mem, align_val_t) noexcept {
     ++dealloc_count;
     _aligned_free(mem);
 }
+#endif // ^^^ !defined(__cpp_aligned_new) ^^^
```
AlexGuteniev: Whoops, these comments are wrong.
Let's finish CI run to see if there are also failures though.

StephanTLavavej:

StephanTLavavej:
Thanks for investigating this! :cat2: :black_cat: :cat2:

PR #6084, opened 2026-02-14
Enable ARM64EC vector algorithms

AlexGuteniev:
I think it will make all things simpler, except skipping `uint64_t` for `minmax_element`, and that part isn't too bad.

Works towards #2635, which will now be completed at the same time as #813.

PR #6089, opened 2026-02-16
Vector algorithms cleanups

AlexGuteniev:
Easy cleanups:
 * More constants. (Ideally, make everything constant, except loop counter and loop-carried dependency).
 * `_M_IX86` / `_M_X64` to `_WIN64`. That's the last occurrence after #6084.
 * `_mm_cvtsi64x_si128` -> `_mm_cvtsi64_si128`. Standardize spelling.

Standardizing for `_mm_cvtsi64_si128` was preferred over the other way around because:
 - We already consistently prefer `_mm_cvtsi128_si64` over `_mm_cvtsi128_si64x`
 - Clang does not understand the si64x form of either
 - The form without x is more common (searched on GitHub and elsewhere)
 - The x character doesn't seem to convey any meaning. Seems that it is supposed to mean that general-purpose 64-bit integer is used, as opposed to `__m64` MMX vector, this can be seen in `_epi64x` ending intrinsics, which have `_epi64` counterparts, But this is no longer ever necessary, as no point in using MMX.

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -2083,28 +2083,28 @@ namespace {
             }
 
             static _Vec_t _H_min(const _Vec_t _Cur) noexcept {
-                int64x2_t _Swapped  = vextq_s64(_Cur, _Cur, 1);
-                uint64x2_t _Mask_lt = vcltq_s64(_Swapped, _Cur);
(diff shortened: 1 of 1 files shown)


StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thank you for your ongoing dedication to code quality and consistency! :magic_wand: :smiley_cat: :100:

Issue #5969, opened 2025-12-19
`<regex>`: Avoid heap allocations for few capturing groups and stack frames


muellerj2:
Even if string matching only involves a few capturing groups or repetitions, a few buffers are still allocated on the heap by the matcher:
* The `vector<bool> _Grp_valid` in `_Matcher3::_Tgt_state` to store which capturing groups are matched.
* The `vector<_Grp_t> _Grps` in `_Matcher3::_Tgt_state` to store the extents of capturing groups.
* If matching follows the leftmost-longest rule, vectors of the same sizes in `_Matcher3::_Res`.
* The stack frames in `vector<_Rx_state_frame_t<_It>> _Matcher3::_Frames`.

Only one of these vectors, `_Matcher3::_Frames`, can grow beyond its initial size during matching.

These heap allocations are comparatively costly especially when matching short inputs, so we should avoid performing these allocations and instead use some stack-allocated buffer(s) if the required amount of memory is small enough.

muellerj2:
I think a natural solution is to allocate a single stack-based buffer, use this buffer to supply the memory for the capturing group vectors (if the memory required by each of them is small enough) and assign the remaining memory to the stack frame vector.

PR #6092, opened 2026-02-18
`<regex>`: Implement small buffer optimization to store loop state

muellerj2:
Disclaimer: Low-level memory management is not my forte, because I mostly write Java code in my day-to-day job.

Towards #5969.

This implements a minimal fixed-size buffer to support some small buffer optimizations during regex matching. In the first step, this fixed-size buffer is only used for the loop state storage (because this requires minimal changes in the matcher), but I plan to use it to store the state of capturing groups as well in follow-up PRs. For frames, we rather need a vector-like data structure that can grow but uses stack storage when the number of frames is few.

The fixed-size buffer takes an allocator in preparation for #174.

Since I want to use some common storage space on the stack for all of these buffers, the stack storage is not an internal member of the buffers but supplied as a pointer to them from the outside. The fixed-size buffer reports back to what extent it has made use of the stack storage. (We don't need this in this PR yet, because the stack storage is only by a single fixed-size buffer. But this will change when more buffers make use of the stack storage.)

The stack storage is also not a member of `_Matcher3` so that we can adjust its size in the future without having to rename the class.

I wasn't sure what exception to throw if too much memory is requested from the fixed-size buffer, but then went with `regex_error(error_stack)` because I would like to avoid exceptions not of type `regex_error` and the description of `error_stack` is "There was insufficient memory to determine whether the regular expression could match the specified character sequence." (The other candidate is `error_space`. But it is described as "There was insufficient memory to convert the expression into a finite state machine.", so it appears to be intended for parsing and not matching.)

I do wonder whether we should call `std::launder()` when repurposing the stack storage (here, specifically, in `_Rx_fixed_size_buffer::_Use_external_buf()`). But it seems this isn't done elsewhere in the STL, including `_Optimistic_temporary_buffer`.

## Benchmarks

I highlighted the relevant lines in `regex_search` where the matcher uses some loop state. (The `regex_match` benchmark is inconclusive, the difference vanishes in the noise on my machine.)

-- | -- | -- | --
bm_lorem_search/"^bibe"/2 | 57.20 | 59.99 | 0.95
bm_lorem_search/"^bibe"/3 | 58.59 | 59.38 | 0.99
bm_lorem_search/"^bibe"/4 | 61.38 | 58.59 | 1.05
bm_lorem_search/"bibe"/2 | 3048.28 | 2849.48 | 1.07
bm_lorem_search/"bibe"/3 | 6093.75 | 5719.87 | 1.07
bm_lorem_search/"bibe"/4 | 11439.70 | 11474.60 | 1.00
bm_lorem_search/"bibe".collate/2 | 3076.18 | 3138.95 | 0.98
bm_lorem_search/"bibe".collate/3 | 5937.50 | 5781.25 | 1.03
bm_lorem_search/"bibe".collate/4 | 11718.80 | 11160.70 | 1.05
bm_lorem_search/"(bibe)"/2 | 6975.45 | 7324.22 | 0.95
bm_lorem_search/"(bibe)"/3 | 14299.70 | 14229.90 | 1.00
bm_lorem_search/"(bibe)"/4 | 28250.40 | 28250.40 | 1.00
**bm_lorem_search/"(bibe)+"/2** | **12207.00** | **11090.90** | **1.10**
**bm_lorem_search/"(bibe)+"/3** | **23541.90** | **22216.50** | **1.06**
**bm_lorem_search/"(bibe)+"/4** | **43946.30** | **43526.20** | **1.01**
**bm_lorem_search/"(?:bibe)+"/2** | **6975.45** | **5937.50** | **1.17**
**bm_lorem_search/"(?:bibe)+"/3** | **13497.40** | **11230.50** | **1.20**
**bm_lorem_search/"(?:bibe)+"/4** | **26681.00** | **22949.20** | **1.16**
bm_lorem_search/R"(\bbibe)"/2 | 83705.40 | 81961.50 | 1.02
bm_lorem_search/R"(\bbibe)"/3 | 168795.00 | 167411.00 | 1.01
bm_lorem_search/R"(\bbibe)"/4 | 336967.00 | 314991.00 | 1.07
bm_lorem_search/R"(\Bibe)"/2 | 184168.00 | 187976.00 | 0.98
bm_lorem_search/R"(\Bibe)"/3 | 360695.00 | 385010.00 | 0.94
bm_lorem_search/R"(\Bibe)"/4 | 749860.00 | 739397.00 | 1.01
bm_lorem_search/R"((?=....)bibe)"/2 | 4865.38 | 4865.38 | 1.00
bm_lorem_search/R"((?=....)bibe)"/3 | 10044.60 | 9626.07 | 1.04
bm_lorem_search/R"((?=....)bibe)"/4 | 19252.40 | 19043.00 | 1.01
bm_lorem_search/R"((?=bibe)....)"/2 | 4854.90 | 4603.80 | 1.05
bm_lorem_search/R"((?=bibe)....)"/3 | 9207.55 | 9207.55 | 1.00
bm_lorem_search/R"((?=bibe)....)"/4 | 17996.80 | 17578.30 | 1.02
bm_lorem_search/R"((?!lorem)bibe)"/2 | 4589.84 | 4589.84 | 1.00
bm_lorem_search/R"((?!lorem)bibe)"/3 | 8719.31 | 8893.69 | 0.98
bm_lorem_search/R"((?!lorem)bibe)"/4 | 17089.80 | 17089.80 | 1.00

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -1649,7 +1649,7 @@ private:
     using _Al_pointer   = _Alloc_ptr_t<_Alloc>;
     using _Al_size_type = _Alloc_size_t<_Alloc>;
 
-    struct _Val {
+    struct _Buffer_range {
         _Al_pointer _First{};
         _Al_pointer _Last{};
         bool _Allocated{false};
@@ -1745,7 +1745,7 @@ private:
         _Guard._Ptr      = nullptr;
     }
 
-    _Compressed_pair<_Alloc, _Val> _Mypair;
+    _Compressed_pair<_Alloc, _Buffer_range> _Mypair;
 };
 
 template <class _BidIt>


stl/inc/regex
```diff
@@ -1643,6 +1643,106 @@ public:
...
+
+    struct _NODISCARD _Allocation_guard {
+        _Alloc& _Al;
+        _Ty* _Ptr;
+        _Al_size_type _Size;
+
+        _Allocation_guard& operator=(const _Allocation_guard&) = delete;
+        _Allocation_guard& operator=(_Allocation_guard&&)      = delete;
+
+        ~_Allocation_guard() noexcept {
```
AlexGuteniev, line 1666: Isn't `noexcept` by default here in this case?

stl/inc/regex
```diff
@@ -1816,7 +1918,7 @@ public:
 private:
     _Tgt_state_t<_It> _Tgt_state;
     _Tgt_state_t<_It> _Res;
-    vector<_Loop_vals_v3_t<_Iter_diff_t<_It>>> _Loop_vals;
+    _Rx_fixed_size_buffer<_Loop_vals_v3_t<_Iter_diff_t<_It>>> _Loop_vals;
```
AlexGuteniev, line 1926: I assume we're not between releases and matcher doesn't need to be renamed again

stl/inc/regex
```diff
@@ -1643,6 +1643,106 @@ public:
...
+        auto _Ty_first = _Al.allocate(_Al_size);
+        _Allocation_guard _Guard{_Al, _Ty_first, _Al_size};
+        auto& _Data      = _Mypair._Myval2;
+        _Data._Last      = _STD _Uninitialized_fill_n(_Ty_first, _Ty_size, _Ty{}, _Al);
+        _Data._Allocated = true;
+        _Data._First     = _Ty_first;
+        _Guard._Ptr      = nullptr;
+    }
+
+    _Compressed_pair<_Alloc, _Val> _Mypair;
```
AlexGuteniev: It may be the time to use no unique address

stl/inc/regex
```diff
@@ -1643,6 +1643,106 @@ public:
...
+    _Rx_fixed_size_buffer(const _Rx_fixed_size_buffer&) = delete;
+    _Rx_fixed_size_buffer(_Rx_fixed_size_buffer&&)      = delete;
+
+    _NODISCARD size_t _Initialize(unsigned char* _Stack_buf, size_t _Stack_size, size_t _Size) {
+        if (_Stack_size / sizeof(_Ty) >= _Size) {
+            _Use_external_buf(_Stack_buf, _Size);
+            return _Size * sizeof(_Ty);
+        } else {
+            _Allocate_buf(_Size);
+            return 0U;
```
AlexGuteniev, line 1687: The suffix is misleading. Just 0.

(There's a pointer-size-dependent suffix I think, but we can't use that here due to C++14 at least)

stl/inc/regex
```diff
@@ -2330,8 +2432,9 @@ bool _Regex_match1(_It _First, _It _Last, match_results<_BidIt, _Alloc>* _Matche
         return false;
     }
 
-    _Matcher3<_BidIt, _Elem, _RxTraits, _It, void> _Mx(
-        _First, _Last, _Re._Get_traits(), _Re._Get(), _Re.mark_count() + 1, _Re.flags(), _Flgs);
+    alignas(_Loop_vals_v3_t<_Iter_diff_t<_It>>) unsigned char _Stackbuf[4096];
```
AlexGuteniev, line 2440: What is this alignment?

muellerj2: Usually it's the same as a pointer, except if some user passes an iterator where `_Iter_diff_t<_It>` requires some stricter alignment (although I don't see why anyone would want to do that).

stl/inc/regex
muellerj2: no_unique_address is C++20. I think `[[msvc::no_unique_address]]` is supported in older C++ dialects as well as an extension, but then we need a decision that we will make use of this extension.

stl/inc/regex
muellerj2: I think I just copied this from some guard in `<vector>`.

stl/inc/regex
muellerj2: That`s what I took from STL's remarks on Discord that he would like to audit no_unique_address for the 14.51 release: The branch-off is imminent, but hasn't happened yet.

stl/inc/regex
AlexGuteniev: Then that's right probably.

stl/inc/regex
StephanTLavavej: That's correct. As long as we land this soon-ish, we're still covered by the `_Matcher3` rename.

stl/inc/regex
StephanTLavavej: I don't know if it's supported downlevel by MSVC, Clang, and EDG. (`_MSVC_NO_UNIQUE_ADDRESS` is currently not set up to activate downlevel.) There's also a concern with CUDA.

I think messing with it here is too much risk for too little value. I'd rather stick with `_Compressed_pair`.

StephanTLavavej:
> I do wonder whether we should call `std::launder()`

As far as I'm concerned, `launder()` is silly and should never be used. If it would ever make a difference, that's a compiler bug. We've been implementing `vector` et al. creating elements via True Placement New for decades, and compilers simply aren't allowed to trash our code just because they feel like it, regardless of what the Standard might think.

stl/inc/regex
StephanTLavavej: I would have written just `0`, but `<regex>` has already accumulated a bunch of `0U` and I don't care enough to change 'em all. Ok as-is.

stl/inc/regex
```diff
@@ -1643,6 +1643,106 @@ public:
...
+#endif
+        return _Data._First[_Pos];
+    }
+
+private:
+    void _Use_external_buf(unsigned char* _Buf, size_t _Ty_size) {
+        auto _Ty_first = reinterpret_cast<_Ty*>(_Buf);
+        auto _Ty_last  = _Ty_first + _Ty_size;
+        auto& _Data    = _Mypair._Myval2;
+        _STD uninitialized_fill(_Ty_first, _Ty_last, _Ty{});
```
StephanTLavavej: `uninitialized_fill()` with `_Ty{}` should be more directly expressed as `uninitialized_value_construct()`.

Edit: Too bad it's C++17.

stl/inc/regex
```diff
@@ -1643,6 +1643,106 @@ public:
...
+
+        if (_Ty_size > size_t{PTRDIFF_MAX} || _Ty_size > allocator_traits<_Alloc>::max_size(_Al)) {
+            _Xregex_error(regex_constants::error_stack);
+        }
+
+        auto _Al_size  = static_cast<_Al_size_type>(_Ty_size);
+        auto _Ty_first = _Al.allocate(_Al_size);
+        _Allocation_guard _Guard{_Al, _Ty_first, _Al_size};
+        auto& _Data      = _Mypair._Myval2;
+        _Data._Last      = _STD _Uninitialized_fill_n(_Ty_first, _Ty_size, _Ty{}, _Al);
```
StephanTLavavej: Similarly, we have an `_Uninitialized_value_construct_n()`.

stl/inc/regex
```diff
@@ -1643,6 +1643,106 @@ public:
...
+
+        ~_Allocation_guard() noexcept {
+            if (_Ptr) {
+                _Al.deallocate(_Ptr, _Size);
+            }
+        }
+    };
+
+public:
+    _Rx_fixed_size_buffer(const _Alloc& _Al = _Alloc{}) : _Mypair(_One_then_variadic_args_t{}, _Al) {}
```
StephanTLavavej: This should be `explicit`. We should also add a newline as it's not repetitive with the deleted ctors.

stl/inc/regex
```diff
@@ -1643,6 +1643,106 @@ public:
     _Builder2& operator=(const _Builder2&) = delete;
 };
 
+template <class _Ty, class _Alloc = allocator<_Ty>>
+class _Rx_fixed_size_buffer {
+    using _Al_size_type = _Alloc_size_t<_Alloc>;
+
+    struct _Val {
+        _Ty* _First;
```
StephanTLavavej: This doesn't respect fancy pointers. (Although I must say that thinking about fancy pointers combined with `<regex>` makes me not want to live on this planet anymore.)

stl/inc/regex
```diff
@@ -1643,6 +1643,106 @@ public:
     _Builder2& operator=(const _Builder2&) = delete;
 };
 
+template <class _Ty, class _Alloc = allocator<_Ty>>
+class _Rx_fixed_size_buffer {
+    using _Al_size_type = _Alloc_size_t<_Alloc>;
```
StephanTLavavej, line 1650: Style: We always say `private:` at the beginning of a `class` even though that's the default (as it's easy to forget).

stl/inc/regex
```diff
@@ -1643,6 +1643,106 @@ public:
...
+            if (_Ptr) {
+                _Al.deallocate(_Ptr, _Size);
+            }
+        }
+    };
+
+public:
+    _Rx_fixed_size_buffer(const _Alloc& _Al = _Alloc{}) : _Mypair(_One_then_variadic_args_t{}, _Al) {}
+    _Rx_fixed_size_buffer(const _Rx_fixed_size_buffer&) = delete;
+    _Rx_fixed_size_buffer(_Rx_fixed_size_buffer&&)      = delete;
```
StephanTLavavej: We should delete the assignment operators too.

stl/inc/regex
```diff
@@ -1643,6 +1643,106 @@ public:
...
 
+template <class _Ty, class _Alloc = allocator<_Ty>>
+class _Rx_fixed_size_buffer {
+    using _Al_size_type = _Alloc_size_t<_Alloc>;
+
+    struct _Val {
+        _Ty* _First;
+        _Ty* _Last;
+        bool _Allocated;
+    };
```
StephanTLavavej: I think they're unnecessary, but I would feel better with DMIs here.

stl/inc/regex
```diff
@@ -1643,6 +1643,106 @@ public:
     _Builder2& operator=(const _Builder2&) = delete;
 };
 
+template <class _Ty, class _Alloc = allocator<_Ty>>
+class _Rx_fixed_size_buffer {
+    using _Al_size_type = _Alloc_size_t<_Alloc>;
+
+    struct _Val {
```
StephanTLavavej: Naming: This is a really short variable-like name (that we use everywhere for everything) for something that isn't repeatedly mentioned. We should be more verbose.

StephanTLavavej:

StephanTLavavej (approved):

StephanTLavavej:
:green_heart: :heart_eyes_cat: :gift_heart:
===

PR #6093, opened 2026-02-19
`<filesystem>`: Fix `weakly_canonical` for drive-relative paths like `"C:bar"`

### LINKED ISSUE #6054: `<filesystem>`: `weakly_canonical` doesn't make absolute path for relative path with only correct drive.
# Describe the bug
Say I have a directory `Temp` in C drive and my current directory is `C:\Temp`; `Temp` contains a subdirectory `foo`. There is no `bar` in `Temp` and `foo`.
And given the program below:
```c++
#include <filesystem>

int main()
{
    fs::current_directory("C:\\Temp");
    // Should be (and currently is) C:\Temp.
    std::cout << fs::weakly_canonical("C:") << "\n";
    // Should be (and currently is) C:\Temp\foo\bar, though bar doesn't exist.
    std::cout << fs::weakly_canonical("C:foo\\bar") << "\n";
    // Should be (**but currently is not**) C:\Temp\bar, though bar doesn't exist.
    std::cout << fs::weakly_canonical("C:bar") << "\n";
}
```

# Command-line test case
```
C:\Temp>type repro.cpp
#include <filesystem>
#include <iostream>

int main()
{
    // Prepare reproduction environment.
    fs::path tempDir{ "C:\\Temp" };
    fs::create_directories(tempDir / "foo");
    // Test
    fs::current_path(tempDir);
    std::cout << fs::weakly_canonical("C:") << "\n";
    std::cout << fs::weakly_canonical("C:foo\\bar") << "\n";
    std::cout << fs::weakly_canonical("C:bar") << "\n";
    // Clean reproduction environment.
    fs::remove_all(tempDir / "foo");
}

C:\Temp>cl /EHsc /W4 /WX /std:c++latest .\repro.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.44.35214 for x64

repro.cpp
Microsoft (R) Incremental Linker Version 14.44.35214.0

/out:repro.exe
repro.obj

C:\Temp>.\repro.exe
"C:\\Temp"
"C:\\Temp\\foo\\bar"
"C:bar"
```

# Expected behavior
The last output should be `"C:\\Temp\\bar"`.

# STL version
* Option 1: MSVC Compiler version: 19.44.35214
* Option 2: `_MSVC_STL_UPDATE` value: 202503
* Option 3: git commit hash: this problem still exists in the [current latest commit] (https://github.com/microsoft/STL/blob/3f76681e18fa51e1441c9150099b8010c1d18e1c/stl/inc/filesystem).

# Additional context
In the standard, `weakly_canonical` is [specified as](https://eel.is/c++draft/filesystems#fs.op.weakly.canonical-1):

> Effects: Using `status(p)` or `status(p, ec)`, respectively, to determine existence, return a path composed by `operator/=` from the result of calling canonical with a path argument composed of the leading elements of `p` that exist, if any, followed by the elements of `p` that do not exist, if any.

And the elements of path are [specifed as](https://eel.is/c++draft/filesystems#fs.class.path.general-3):

> A path is a sequence of elements that identify the location of a file within a filesystem. The elements are the root-name, root-directory, and an optional sequence of filename ([fs.path.generic]).

So `C:` as the root-name is a legal leading element in the path and thus the last line in test case should be `"C:\\Temp\\bar"` instead.

The problem is that MS-STL skips initial value of the element for this case:
https://github.com/microsoft/STL/blob/3f76681e18fa51e1441c9150099b8010c1d18e1c/stl/inc/filesystem#L4132-L4159

Here in the loop, for `C:bar`, the first `_Result` to call `_Canonical` is `C:bar` instead of `C:`. I'd like to submit a PR if this is considered as a bug :-).

MarkedMuichiro:
Fixes #6054

## Problem

On Windows, drive-relative paths such as "C:bar" are relative to the
current directory on that drive. However, `std::filesystem::weakly_canonical`
currently returns "C:bar" instead of anchoring the root-name ("C:")
to the drive's current directory (e.g. "C:\Temp\bar").

This happens because the implementation appends path elements before
canonicalizing the drive-relative root-name.

## Fix

When the normalized path has a root-name but no root-directory
(i.e. drive-relative paths like "C:bar"), canonicalize the root-name
("C:") first before appending remaining elements.

This aligns behavior with Windows drive-relative path semantics, where
"C:foo" is relative to the current directory on drive C rather than the
drive root.

Previously, weakly_canonical appended path elements before resolving
the drive-relative root, which produced incorrect results such as
returning "C:bar" instead of an absolute path.

## Test

Added Windows-only regression coverage in
`P0218R1_filesystem::test_weakly_canonical()` verifying:

- "C:" resolves to the drive's current directory
- "C:foo\\bar" resolves relative to that directory
- "C:bar" resolves to `<cwd>\\bar`

The test restores `current_path` to avoid leaking process state.

### DIFF
--- a/stl/inc/filesystem
+++ b/stl/inc/filesystem
@@ -4137,6 +4137,24 @@ namespace filesystem {
 
         bool _Call_canonical = true;
 
+        // GH-6054: Fix drive-relative paths like "C:bar".
+        // For such paths, root_path() is just the root-name "C:" (no root-directory).
+        // Canonicalize "C:" first (to the current directory on that drive), then append relative elements.
+        if (_Call_canonical && _Normalized.has_root_name() && !_Normalized.has_root_directory()) {
+            _Temp.clear();
+
+            const auto _Err = _Canonical(_Temp, _Result.native());
+
+            if (_Err == __std_win_error::_Success) {
+                _Result = _STD move(_Temp);
+            } else if (__std_is_file_not_found(_Err)) {
+                _Call_canonical = false; // match the loop's behavior
+            } else {
+                _Ec = _Make_ec(_Err);
+                return {};
+            }
+        }
+
         for (const auto& _Elem : _Normalized_relative) {
             _Result /= _Elem;
 
--- a/tests/std/tests/P0218R1_filesystem/test.cpp
+++ b/tests/std/tests/P0218R1_filesystem/test.cpp
@@ -3191,6 +3191,46 @@ void test_weakly_canonical() {
     const path p2 = weakly_canonical(meow_txt);
     EXPECT(p1.native() == p2.native());
 
+    {
+        const path old_cwd = current_path(ec);
+        EXPECT(good(ec));
+
+        const path abs_dir = p1.parent_path();
+        EXPECT(abs_dir.is_absolute());
+
+        // Mirror the repro structure: ensure 'foo' exists under the cwd we use.
+        create_directories(abs_dir / L"foo"sv, ec);
+        EXPECT(good(ec));
+
+        current_path(abs_dir, ec);
+        EXPECT(good(ec));
+
+        const path cwd = current_path(ec);
+        EXPECT(good(ec));
+        EXPECT(cwd.is_absolute());
+
+        const wstring drive = cwd.root_name().native(); // e.g. L"C:"
+        EXPECT(!drive.empty());
+        EXPECT(drive.back() == L':');
+
+        const path got_drive_only = weakly_canonical(drive, ec);
+        EXPECT(good(ec));
+        EXPECT(got_drive_only == cwd);
+
+        // Intentionally drive-relative (no backslash after colon).
+        const path got_foo_bar = weakly_canonical(drive + L"foo\\bar", ec);
+        EXPECT(good(ec));
+        EXPECT(got_foo_bar == (cwd / L"foo"sv / L"bar"sv));
+
+        const path got_bar = weakly_canonical(drive + L"bar", ec);
+        EXPECT(good(ec));
+        EXPECT(got_bar == (cwd / L"bar"sv));
+
+        // Restore cwd so this test doesn't leak process state.
+        current_path(old_cwd, ec);
+        EXPECT(good(ec));
+    }
+
     const path woof_txt(L"test_weakly_canonical/a/b/x/y/woof.txt"sv);
     const path p3 = weakly_canonical(woof_txt, ec);
     EXPECT(good(ec));


MarkedMuichiro:

tests/std/tests/P0218R1_filesystem/test.cpp
```diff
@@ -3191,6 +3191,52 @@ void test_weakly_canonical() {
     const path p2 = weakly_canonical(meow_txt);
     EXPECT(p1.native() == p2.native());
 
+    #ifdef _WIN32
```
AlexGuteniev: This library doesn't target anything that is not `_WIN32`, remove this preprocessor guard

MarkedMuichiro: Good catch — thank you!

I've removed the unnecessary _WIN32 guard since this implementation already targets Windows exclusively. Updated in latest push.

StephanTLavavej: Is this an AI-generated response? Is this PR AI-generated? Please disclose if so.

tests/std/tests/P0218R1_filesystem/test.cpp
```diff
@@ -1,4073 +1,4115 @@
-// Copyright (c) Microsoft Corporation.
-// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-
```
StephanTLavavej, line 3: This PR is creating a diff of over 4,000 lines, which is not acceptable. Please self-review your diffs (e.g. via VSCode) before pushing changes.

MarkedMuichiro: Yeah that’s my bad — I pushed something that blew up the diff way beyond the intended changes. I’m going to fix the branch so it only contains the minimal hunks for GH-6054 (filesystem change + small test addition) and re-push.

tests/std/tests/P0218R1_filesystem/test.cpp
MarkedMuichiro: Yeah — I did use AI assistance while drafting parts of this, then I went through and validated/adjusted it against the repo + behavior. I’ll keep the PR tightly scoped (and fix the oversized diff).

StephanTLavavej:
I'm not a fan of AI-generated PRs here. Our product code demands the highest level of understanding; if you have to ask an LLM how to operate a nuclear reactor, you should get out of the control room. That said, since this addresses a reported bug, we'll put some effort into reviewing it when we have time.

stl/inc/filesystem
```diff
@@ -4137,6 +4137,24 @@ namespace filesystem {
 
         bool _Call_canonical = true;
 
+        // GH-6054: Fix drive-relative paths like "C:bar".
+        // For such paths, root_path() is just the root-name "C:" (no root-directory).
+        // Canonicalize "C:" first (to the current directory on that drive), then append relative elements.
+        if (_Call_canonical && _Normalized.has_root_name() && !_Normalized.has_root_directory()) {
```
StephanTLavavej, line 4143: `_Call_canonical` was *just* initialized to `true`.

StephanTLavavej:
Well, that was fast. I don't think any significant effort was put into reviewing this PR before creation, since literally the first line I looked at had a howler.

I'm going to close this, sorry. What we're lacking is human attention and this is not helping.

Issue #6096, opened 2026-02-21
`<chrono>`: Incorrect time conversion for zones where DST was observed in the past


daksh-goyal:
# Describe the bug
There are certain time zones that used to observe Daylight Savings Time sometime in the past but have since abolished the practice. When trying to convert the time on one of these dates from the past when DST was observed, STL produces incorrect output.

# Command-line test case
```
C:\Temp>type repro.cpp
#include <iostream>
#include <chrono>
#include <format>
using namespace std;
using namespace std::chrono;

int main() {
    // Brazil (America/Sao_Paulo): Observed DST from ~Oct/Nov to ~Feb each year.
    // Last DST period: Nov 4, 2018 → Feb 17, 2019 (UTC-2 BRST).
    auto sp = locate_zone("America/Sao_Paulo");
    auto sp1 = local_days{year{2019}/month{1}/day{1}} + hours{13};
    cout << "=== America/Sao_Paulo (abolished DST Feb 2019) ===" << endl;
    cout << "2019-01-01 13:00 local -> " << format("{:%H:%M}", sp->to_sys(sp1)) << " UTC (expect 15:00, DST UTC-2)" << endl;
    auto si = sp->get_info(sp->to_sys(sp1));
    cout << "  offset=" << si.offset.count() << "s save=" << si.save.count() << "s" << endl;

    // Turkey: abolished DST in Sep 2016, permanent UTC+3
    // Summer 2016 had DST (UTC+3), standard was UTC+2
    auto istanbul = locate_zone("Europe/Istanbul");
    auto tur1 = local_days{year{2016}/month{7}/day{1}} + hours{12};
    auto tur2 = local_days{year{2015}/month{7}/day{1}} + hours{12};
    cout << "\n=== Europe/Istanbul (abolished DST Sep 2016) ===" << endl;
    cout << "2016-07-01 12:00 local -> " << format("{:%H:%M}", istanbul->to_sys(tur1)) << " UTC (expect 09:00, DST UTC+3)" << endl;
    cout << "2015-07-01 12:00 local -> " << format("{:%H:%M}", istanbul->to_sys(tur2)) << " UTC (expect 09:00, DST UTC+3)" << endl;
    auto ti = istanbul->get_info(istanbul->to_sys(tur1));
    cout << "  offset=" << ti.offset.count() << "s save=" << ti.save.count() << "s" << endl;

    // Russia: Summer 2010 had DST UTC+4, standard UTC+3
[...]

PR #6097, opened 2026-02-21
fix: improper use of ucal_inDaylightTime

daksh-goyal:
`ucal_inDaylightTime` is used to check whether the timezone is currently observing DST. STL (likely in an attempt to be defensive) tries to use it as a way to tell whether the timezone observed DST on the date in question (which could be in the past), which is wrong.

Fixing that by omitting use of that guard before obtaining DST offset.

fixes: #6096

### DIFF
--- a/tests/std/tests/P0355R7_calendars_and_time_zones_time_zones/test.cpp
+++ b/tests/std/tests/P0355R7_calendars_and_time_zones_time_zones/test.cpp
@@ -403,28 +403,28 @@ void timezone_sorted_vectors_test() {
 }
 
 void timezone_historical_dst_abolished_test() {
-   // Regression test: timezones that abolished DST must still report
-   // correct historical offsets for dates when DST was active.
-   const auto& my_tzdb = get_tzdb();
+    // Regression test: timezones that abolished DST must still report
+    // correct historical offsets for dates when DST was active.
+    const auto& my_tzdb = get_tzdb();
 
-   { // America/Sao_Paulo - abolished DST Feb 2019
-       auto tz = my_tzdb.locate_zone("America/Sao_Paulo");
-       auto info = tz->get_info(sys_days{year{2019} / January / day{1}});
-       assert(info.offset == seconds{-7200});
-       assert(info.save == minutes{60});
-   }
-   { // Europe/Istanbul - abolished DST Sep 2016
-       auto tz = my_tzdb.locate_zone("Europe/Istanbul");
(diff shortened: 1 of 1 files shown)


stl/src/tzdb.cpp
StephanTLavavej: This is failing to build, indicating that you've never verified the fix, which is really important:

```
C:\a\_work\1\s\stl\src\tzdb.cpp(553,61): error C2065: '_Is_daylight': undeclared identifier
    const auto _Abbrev = _Get_timezone_short_id(_Cal.get(), _Is_daylight, _Abbrev_len, _Info->_Err);
                                                            ^
```

Before pushing changes:

1. Please build the STL.
2. Run the affected existing tests. For example:
   ```
   pushd out\x64
   python tests\utils\stl-lit\stl-lit.py -o testing_x64.log -Dnotags=ASAN --order=random --filter=P0355R7 ..\..\tests\std
   ```
3. Run `set_environment.bat` and then manually verify your test case. (I can add it as an automated test, you don't have to worry about that.)

stl/src/tzdb.cpp
```diff
@@ -665,3 +655,4 @@ void __stdcall __std_free_crt(void* p) noexcept {
 }
 
 } // extern "C"
+
```
StephanTLavavej: This extra newline is causing code format validation to fail (we are very picky and don't permit extra newlines at the end of files).

StephanTLavavej (changes requested):

stl/src/tzdb.cpp
StephanTLavavej: `_Pfn_ucal_inDaylightTime` is now dead. (This can be altered without breaking ABI as it's entirely confined to this separately compiled code.)

daksh-goyal: Sorry about that, didn't have the necessary cl version to get it to build, took a shortcut and it clearly didn't work (as if they ever do). Will build and run the test cases. Maybe I can also try to find a place to add the relevant test case.

ps: it is still installing.. :(

daksh-goyal:
@microsoft-github-policy-service agree company="Microsoft"

StephanTLavavej:
Thanks! I've manually verified that the regression test produced all wrong answers before the fix.

StephanTLavavej:

StephanTLavavej:
:earth_asia: :calendar: :clock10:
===

PR #6098, opened 2026-02-22
Vector algorithms: Add compile time checks for ISA features

matbech:
When the STL is compiled with /arch:SSE4.2 or /arch:AVX2, the runtime checks for the respective ISA features (SSE42, AVX2) are unnecessary. This change adds a compile time check using __check_arch_support to avoid the runtime cost.

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -29,7 +29,7 @@ extern "C" long __isa_enabled;
 namespace {
 #if !defined(_M_ARM64) && !defined(_M_ARM64EC)
     bool _Use_avx2() noexcept {
-        return __check_arch_support(__IA_SUPPORT_VECTOR256, 0) || __isa_enabled & (1 << __ISA_AVAILABLE_AVX2);
+        return __check_arch_support(__IA_SUPPORT_VECTOR256, 0) || (__isa_enabled & (1 << __ISA_AVAILABLE_AVX2));
     }
 
     bool _Use_sse42() noexcept {


AlexGuteniev:
But what is the purpose?

For end users, the STL comes precompiled, so `/arch` flag passed by users, only affects public headers, but not separately compiled code. And when it is compiled for shipping it is always the same `/arch` option, which is SSE2 for x86, and baseline for x64.

matbech:
For custom built binaries (msvcp140_oss.dll) with modern `/arch` compiler flags where performance is important.

StephanTLavavej:
For reference, [`__check_arch_support`](https://learn.microsoft.com/en-us/cpp/intrinsics/check-isa-arch-support?view=msvc-170):

> The `__check_arch_support` intrinsic can always be evaluated at compile time, so using it in optimized code adds no extra instructions to execute. Support for these intrinsics was added in Visual Studio 2022 version 17.10.

stl/src/vector_algorithms.cpp
```diff
@@ -29,11 +29,11 @@ extern "C" long __isa_enabled;
 namespace {
 #if !defined(_M_ARM64) && !defined(_M_ARM64EC)
     bool _Use_avx2() noexcept {
-        return __isa_enabled & (1 << __ISA_AVAILABLE_AVX2);
+        return __check_arch_support(__IA_SUPPORT_VECTOR256, 0) || __isa_enabled & (1 << __ISA_AVAILABLE_AVX2);
```
StephanTLavavej: We conventionally add parens when mixing bitwise ops with other ops. They're being added below but not here.

StephanTLavavej:

StephanTLavavej:
:tada: :rocket: :t-rex:
===

Issue #5998, opened 2026-01-07
ARM64EC vector algorithms


AlexGuteniev:
- Why they need to exist always when x64 algorithms exist?
- How to make them executed?
- Can we have coverage for that?

jovibor:
There shouldn't be separate vector algorithms for the `ARM64EC` arch, at all.

In terms of intrinsics there are only two types of them - `x64/x86` and `ARM64`.
And the `ARM64EC` arch should utilize `ARM64` vector algorithms for the best performance.

The `ARM64EC` is basically `ARM64` code with the `x64` compatible ABI, for seamless interability.

[Here](https://techcommunity.microsoft.com/blog/windowsosplatform/getting-to-know-arm64ec-defines-and-intrinsic-functions/2957235) is a good explanation:

> You also have the option to further optimize the processor-specific code in your project by using ARM64 intrinsic functions in your ARM64EC project.  The _M_ARM64EC preprocessor macro allows you to differentiate ARM64EC from x64 and take ARM-specific code paths rather than x64.  For example, if you have code that already handles choosing the best intrinsic functions for x64 and ARM64, you can key off _M_ARM64EC or _M_ARM64 to use the ARM intrinsic functions, as below:

![Image](https://github.com/user-attachments/assets/0eb179af-ebdf-4661-b2df-baaf45537128)

AlexGuteniev:
> In terms of intrinsics there are only two types of them - `x64/x86` and `ARM64`.
> And the `ARM64EC` arch should utilize `ARM64` vector algorithms for the best performance.

The question is mostly about the cases where we don't have ARM64 intrinsics path yet, but vectorized functions still exist and probably can be executed.

StephanTLavavej:
> Why they need to exist always when x64 algorithms exist?

I think the ARM64EC `vector_algorithms.cpp` is used when x64 user .obj files are linked into an ARM64EC binary. That's why the ARM64EC surface must always match the x64 surface.

> How to make them executed?
> Can we have coverage for that?

We now have ARM64EC runtime coverage.

I believe that this is now superseded by #813. I expect we'll get ARM64 vectorized implementations of all algorithms that are currently done for x64/x86, and we are now simultaneously activating them for ARM64EC.

AlexGuteniev:
> > How to make them executed?
> > Can we have coverage for that?
> 
> We now have ARM64EC runtime coverage.
> 
[...]

PR #6100, opened 2026-02-22
Vector algorithms: runtime coverage for ARM64EC non-vectorized fallbacks

AlexGuteniev:
Use them when `_ENABLE_STL_INTERNAL_CHECK`. is defined.

We have `!_USE_STD_VECTOR_ALGORITHMS` coverage, so no much concerns about the missing non-vectoziaed coverage.

Putting `_VECTORIZED_MINMAX_ELEMENT_64BIT_INT` on display, as in-place expression for this quirk is getting complicated.

### DIFF
--- a/stl/inc/algorithm
+++ b/stl/inc/algorithm
@@ -64,9 +64,9 @@ __declspec(noalias) void __cdecl __std_reverse_copy_trivially_copyable_8(
 _Min_max_element_t __stdcall __std_minmax_element_1(const void* _First, const void* _Last, bool _Signed) noexcept;
 _Min_max_element_t __stdcall __std_minmax_element_2(const void* _First, const void* _Last, bool _Signed) noexcept;
 _Min_max_element_t __stdcall __std_minmax_element_4(const void* _First, const void* _Last, bool _Signed) noexcept;
-#if !defined(_M_ARM64) && !defined(_M_ARM64EC)
+#if _VECTORIZED_MINMAX_ELEMENT_64BIT_INT
 _Min_max_element_t __stdcall __std_minmax_element_8(const void* _First, const void* _Last, bool _Signed) noexcept;
-#endif // ^^^ !defined(_M_ARM64) && !defined(_M_ARM64EC) ^^^
+#endif // ^^^ _VECTORIZED_MINMAX_ELEMENT_64BIT_INT ^^^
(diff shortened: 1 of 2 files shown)


StephanTLavavej:
This is doing two very different things and I think they should be separated:

1. Extracting the int64 complexity. Should be fine but should be done in isolation.
2. (Ab)using stl-internal-check to activate fallback coverage. I really, really don't like this. This means that we're testing something wildly different than what we're shipping. In general, product code shouldn't behave differently for stl-internal-checks; we've made a highly limited number of exceptions for this (e.g. disabling access control to allow tests to mess with invariants) but this is going too far in my opinion.

I am confused as to what this fallback coverage actually is (I am a little more familiar with ARM64EC these days but my understanding is still fragile). Can you explain in simpler terms what this is trying to do, and why we have to exercise modes that aren't usually enabled for users? I could potentially be open to adding an off-by-default mode that the tests could optionally activate (in addition to testing vanilla modes) but I want to understand why.

AlexGuteniev:
> * Extracting the int64 complexity. Should be fine but should be done in isolation.

Then it should go first. Exercising the fallback without extracting int64 complexity becomes way too complicated.

> Can you explain in simpler terms what this is trying to do, and why we have to exercise modes that aren't usually enabled for users? I could potentially be open to adding an off-by-default mode that the tests could optionally activate (in addition to testing vanilla modes) but I want to understand why.

Here you are mentioning https://github.com/microsoft/STL/issues/5998#issuecomment-3936914209:
> I think the ARM64EC `vector_algorithms.cpp` is used when x64 user .obj files are linked into an ARM64EC binary. That's why the ARM64EC surface must always match the x64 surface.

I'm confident that if this explanation is correct, then _used_ in it means _called_. Since there's a scenario where they area called exists, we need some coverage to avoid #5993 regression.

Additionally, this will help catching complete omission of such callbacks in newly vectorized function without having to check exports.

> This means that we're testing something wildly different than what we're shipping.

I understand this disadvantage. But I think the need for coverage of these outweighs the disadvantage.

StephanTLavavej:
Ok, let's do the int64 separation first, then we can think about exercising the ARM64EC scenario.

AlexGuteniev:
Closing, will open the int64 PR instead

PR #2460, opened 2022-01-06
Don't move from contained value of optional<const T>

CaseyCarter:
...unless `T` supports it, of course.

Fixes #2458
Fixes DevCom-1628644 / VSO-1458298 / AB#1458298

PR #6102, opened 2026-02-22
fix: implement LWG-4015 to make `optional` conversions ADL-proof

daksh-goyal:
LWG-4015 identified that LWG-3973's changes introduced ADL vulnerability in `optional`'s converting constructors and assignment operators. The expressions `*_Right` and `_STD move(*_Right)` on `optional<_Ty2>` can be hijacked by ADL-found `operator*` overloads in the namespace of `_Ty2`.

Changed to directly access `_Right._Value` and `_Right._Has_value` (via friend access), avoiding any `operator*` call and thus any ADL.

Added a compile-only test using a deleted non-member `operator*` as an ADL trap to verify the fix.

fixes: #5871

### DIFF
--- a/tests/std/tests/LWG4015_optional_adl_proof_operations/test.compile.pass.cpp
+++ b/tests/std/tests/LWG4015_optional_adl_proof_operations/test.compile.pass.cpp
@@ -11,15 +11,17 @@
 #include <utility>
 
 namespace bad {
-    template <class T>
-    void operator*(const std::optional<T>&) = delete;
-
     struct S {
         int val;
         operator int() const {
             return val;
         }
     };
+
+    void operator*(std::optional<S>&)        = delete;
+    void operator*(std::optional<S>&&)       = delete;
+    void operator*(const std::optional<S>&)  = delete;
+    void operator*(const std::optional<S>&&) = delete;
 } // namespace bad
 
 namespace std {


stl/inc/optional
StephanTLavavej, line 1: This is definitely missing `swap()` which is mentioned in the LWG issue. We should probably also change the comparison operators.

StephanTLavavej:
Thanks for looking into this. I identified that a bunch of changes were missing, and that we really should be saying `.operator*()` as the Standard now depicts to be safe (see #2460 where direct use of `._Value` is not acceptable), so I pushed major changes.

I will revise the test, but I need to drop everything and come back to this later. No action required.

StephanTLavavej:
I pushed additional test coverage for all of the product code changes, and to make the test actually effective. The templated `operator*()` was *not* sufficiently poisonous, as it wasn't causing compiler errors even with the original product code. (I believe this is because templates are less preferred during overload resolution when they are equally as good as non-templates.) We need non-templated poisonous overloads in order to actually exercise this fix. (I realized this after chasing the rabbit down the hole to LWG-3969.)

Finally, I observed that even with the maximally poisonous `operator*()` overloads, the member functions (such as the converting constructors/assigns and `swap()`) aren't actually vulnerable to ADL hijacking. Only the non-member comparisons and `hash` can be hijacked. This is because member lookup is stronger than ADL, see WG21-N5032 \[basic.lookup.argdep\]/1:

> When the *postfix-expression* in a function call is an *unqualified-id*, and unqualified lookup for the name in the *unqualified-id* does not find any declaration of a class member, or \[...\] then lookup for the name also includes the result of *argument-dependent lookup* \[...\]

So we'd actually be fine saying `*opt` in member functions. However, because the Standard depicts `.operator*()` (and Do What The Standard Says is one of our cherished principles) and because it's needed in the non-members, being consistent is just simpler and worth the extra verbosity.

StephanTLavavej:

daksh-goyal:
Didn't realize the union member is `remove_cv_t<_Ty>`, and so it strips const. That would've been a regression.

I would've probably got stuck at [WG21-N5032](https://wg21.link/N5032) [over.match.oper] and it would have taken me forever to realize that the explicit member function call syntax bypasses ADL.

And thanks a lot for pointing to LWG-3969, I don't how I'd have gotten there.

StephanTLavavej:
:tada: :cat2: :cat:
===

PR #6104, opened 2026-02-23
`<random>`: Implement modified ziggurat algorithm for `normal_distribution`

### LINKED ISSUE #1003: <random>: std::normal_distribution four times slower than the corresponding boost version
**Describe the bug**

A benchmark by Alexander Neumann (original issue reporter) showed that `std::normal_distribution` with `std::mt19937_64` was 4 times slower than `boost::normal_distribution` with `std::mt19937_64`.

**Additional context**

Part of the performance deficiency is due to #1000.

Also tracked by DevCom-86909 and Microsoft-internal VSO-486661.

statementreply:
This PR implements modified ziggurat algorithm for `normal_distribution`, based on [McFarland, C. D. (2015). A modified ziggurat algorithm for generating exponentially and normally distributed pseudorandom numbers](https://doi.org/10.1080/00949655.2015.1060234).

The main differences between this PR and the paper:
- The layers are moved upwards to reduce the area of non-rectangular regions, creating a bottommost layer of which the size of the rectangular region isn't equal to 1/N.
- Different criteria are used for the always sampled and always rejected regions of ziggurat overhangs.

Benchmark results for x64 on Intel Core i5-13600KF:

Benchmark | Time | Speed Up
--- | :---: | :---:
`BM_Generator<std::mt19937_64>` | 3.39 ns
`BM_Generator<b_r::mt19937_64>` | 2.52 ns
`BM_Distribution<std::mt19937_64, std_old::normal_distribution<double>>` | 8.91 ns | 1.00
`BM_Distribution<std::mt19937_64, std_new::normal_distribution<double>>` | 4.52 ns | 1.97
`BM_Distribution<std::mt19937_64, boost_r::normal_distribution<double>>` | 5.48 ns | 1.63
`BM_Distribution<b_r::mt19937_64, boost_r::normal_distribution<double>>` | 4.50 ns | 1.98
`BM_Generator<std::mt19937>` | 3.00 ns
`BM_Generator<b_r::mt19937>` | 2.12 ns
`BM_Distribution<std::mt19937, std_old::normal_distribution<float>>` | 8.12 ns | 1.00
`BM_Distribution<std::mt19937, std_new::normal_distribution<float>>` | 4.41 ns | 1.84
`BM_Distribution<std::mt19937, boost_r::normal_distribution<float>>` | 5.28 ns | 1.54
`BM_Distribution<b_r::mt19937, boost_r::normal_distribution<float>>` | 4.33 ns | 1.88

Removed `tests\GH_004618_normal_distribution_avoids_resets` which became moot.

Fixes #1003.


stl/inc/__msvc_random_ziggurat_tables.hpp
```diff
@@ -0,0 +1,341 @@
...
+        0.0804412728339469, 0.07826062733867703, 0.0760917753028642, 0.07393478777870469, 0.07178974080741468,
+        0.0696567157331283, 0.06753579954797788, 0.06542708527228344, 0.06333067237439781, 0.06124666723548977,
+        0.05917518366542869, 0.057116343476991766, 0.05507027712689312, 0.053037124433685696, 0.051017035384481935,
+        0.04901017104476877, 0.04701670458847328, 0.04503682246902435, 0.04307072575665931, 0.04111863167292178,
+        0.0391807753605653, 0.03725741193643623, 0.03534881888708009, 0.033455298882813636, 0.03157718310728027,
+        0.029714835228161415, 0.027868656173851813, 0.02603908993517536, 0.024226630687753957, 0.02243183164061715,
+        0.020655316178002918, 0.018897792103686122, 0.017160070171080462, 0.015443088676963571, 0.013747946876007678,
+        0.012075951653800543, 0.010428684923300423, 0.008808104992035565, 0.007216707019283454, 0.005657794393437552,
+        0.004135980658085006, 0.0026582451648572286, 0.0012366551429421327},
+    {5772624080256872389, 18376497011742254427, 15174068518956773574, 18119521015726707089, 9689093257773371827,
```
StephanTLavavej: The benchmarks are failing to build for Clang (no worries about missing this, it is a non-default config and it trips me up occasionally too):
```
C:\stlBuild\out\inc\__msvc_random_ziggurat_tables.hpp(159,27): error: integer literal is too large to be represented in a signed integer type, interpreting as unsigned [-Werror,-Wimplicitly-unsigned-literal]
  159 |     {5772624080256872389, 18376497011742254427, 15174068518956773574, 18119521015726707089, 9689093257773371827,
      |                           ^
```

StephanTLavavej (changes requested):

stl/inc/__msvc_random_ziggurat_tables.hpp
```diff
@@ -0,0 +1,345 @@
...
+    static constexpr _Ty _Width_scale = (_Signed ? _Ty{1} : _Ty{2}) * _Ty{(numeric_limits<_Uty>::max)() / 2u + 1u};
+
+    _Ty _Layer_widths[_Lx + 1];
+    _Ty _Layer_heights[_Lx + 1];
+    _Uty _Alias_probabilities[256];
+    uint8_t _Alias_indices[256];
+};
+
+template <class _Ty>
+_INLINE_VAR constexpr _Modified_ziggurat_tables<_Ty, void, true, 0> _Normal_distribution_tables;
```
StephanTLavavej: This is making CUDA very unhappy:

```
C:/stlBuild/out/inc\__msvc_random_ziggurat_tables.hpp(24): error: static assertion failed with "invalid table size"
      static_assert(_Lx >= 2, "invalid table size");
      ^
          detected during:
            instantiation of class "std::_Modified_ziggurat_tables<_Ty, _Uty, _Signed, _Lx> [with _Ty=double, _Uty=void, _Signed=true, _Lx=0]" at line 43
            instantiation of "const std::_Modified_ziggurat_tables<_Ty, void, true, 0> std::_Normal_distribution_tables [with _Ty=double]" at line 43
```

Could the primary template just be an `int`, so there's no question of instantiating `_Modified_ziggurat_tables` with bogus arguments? (I'm pretty sure CUDA is wrong here, not that it makes life any easier.)

StephanTLavavej (changes requested):

stl/inc/__msvc_random_ziggurat_tables.hpp
```diff
@@ -0,0 +1,345 @@
...
+
+_STD_BEGIN
+
+template <class _Ty, class _Uty, bool _Signed, int _Lx>
+struct _Modified_ziggurat_tables {
+    static_assert(_Lx >= 2, "invalid table size");
+    static_assert(_Lx <= 254, "invalid table size");
+
+    using _Uint_type = _Uty;
+    using _Xtype     = conditional_t<_Signed, make_signed_t<_Uty>, _Uty>;
```
StephanTLavavej, line 31: This is failing the header units test with:

```
C:\stlBuild\out\inc\__msvc_random_ziggurat_tables.hpp(28): error C7568: argument list missing after assumed function template 'make_signed_t'
```
And indeed this header includes only `<cstdint>` which is insufficient.

stl/inc/__msvc_random_ziggurat_tables.hpp
```diff
@@ -0,0 +1,345 @@
+// __msvc_random_ziggurat_tables.hpp internal header
+
```
StephanTLavavej, line 2: GH_004618_normal_distribution_avoids_resets is failing with:
```
Assertion failed: calls_before == calls_after, file C:\a\_work\1\s\tests\std\tests\GH_004618_normal_distribution_avoids_resets\test.cpp, line 43
```

StephanTLavavej (changes requested):

tools/scripts/random_ziggurat_tables_generate.cpp
```diff
@@ -0,0 +1,337 @@
...
+#include <string_view>
+#include <utility>
+#include <vector>
+using namespace std;
+
+// get rid of the squiggles
+#ifdef __INTELLISENSE__
+using __float128 = long double;
+
+inline constexpr __float128 operator"" q(long double value) noexcept {
```
frederick-vs-ja: Why redundant `inline` and space in the literal operator (see also CWG-2521)?
```suggestion
constexpr __float128 operator""q(long double value) noexcept {
```
(I hope we can get rid of this literal operator function is not necessary, though.)

tools/scripts/random_ziggurat_tables_generate.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#ifndef __MSVC_RANDOM_ZIGGURAT_TABLES_HPP
+#define __MSVC_RANDOM_ZIGGURAT_TABLES_HPP
+#include <yvals_core.h>
+#if _STL_COMPILER_PREPROCESSOR
+
+#include <cstdint>
```
frederick-vs-ja, line 261:
```suggestion
#include <cstdint>
#include <type_traits>
```

tools/scripts/random_ziggurat_tables_generate.cpp
```diff
@@ -0,0 +1,337 @@
...
+#pragma pack(push, _CRT_PACKING)
+#pragma warning(push, _STL_WARNING_LEVEL)
+#pragma warning(disable : _STL_DISABLED_WARNINGS)
+_STL_DISABLE_CLANG_WARNINGS
+#pragma push_macro("new")
+#undef new
+
+_STD_BEGIN
+
+template <class _Ty, class _Uty, bool _Signed, int _Lx>
```
frederick-vs-ja: Can we remove `_Signed` is it's always `true`?
```suggestion
template <class _Ty, class _Uty, int _Lx>
```

tools/scripts/random_ziggurat_tables_generate.cpp
```diff
@@ -0,0 +1,337 @@
...
+struct _Modified_ziggurat_tables {
+    static_assert(_Lx >= 2, "invalid table size");
+    static_assert(_Lx <= 254, "invalid table size");
+
+    using _Uint_type = _Uty;
+    using _Xtype     = conditional_t<_Signed, make_signed_t<_Uty>, _Uty>;
```
frederick-vs-ja, line 281:
```suggestion
    using _Xtype     = make_signed_t<_Uty>;
```
Also, would it be better to rename it to `_Sint_type`?

tools/scripts/random_ziggurat_tables_generate.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+    static constexpr int _Layer_num   = _Lx;
+    static constexpr _Ty _Width_scale = (_Signed ? _Ty{1} : _Ty{2}) * _Ty{(numeric_limits<_Uty>::max)() / 2u + 1u};
```
frederick-vs-ja: Let's avoid `numeric_limits`.
```suggestion
    static constexpr _Ty _Width_scale = _Ty{static_cast<_Uty>(-1) / 2u + 1u};
```

tools/scripts/random_ziggurat_tables_generate.cpp
statementreply: It's reserved for `exponential_distribution`, where `_Signed` would be `false`.

stl/inc/random
```diff
@@ -2645,53 +2646,90 @@ public:
...
+        constexpr const auto& _Tables = _Normal_distribution_tables<_Fty>::_Value;
+        using _Traits                 = remove_reference_t<decltype(_Tables)>;
+        _Rng_from_urng_v2<typename _Traits::_Uint_type, _Engine> _Generator(_Eng);
+
+        const auto _Xbits         = static_cast<typename _Traits::_Xtype>(_Generator._Get_all_bits());
+        const auto _Regular_layer = static_cast<uint8_t>(_Xbits);
+
+        if (_Regular_layer < _Tables._Layer_num - 1) {
+            // rectangular region of a middle layer
+            return static_cast<_Ty>(_Tables._Layer_widths[_Regular_layer + 1] * static_cast<_Fty>(_Xbits));
```
MattStephanson: In common cases (e.g., CVTSI2SD with ties-to-even) this conversion to floating point is slightly biased toward values where the significand ends in 0. It's also correlated with `_Layer` because a tie only happens when the lower eight or eleven bits of `_Xbits` are 1 followed by all 0s. Occurs below.

I'm unsure how much effect this has, but if it doesn't affect the performance much, maybe it would be cleaner to mask or shift out the lower bits as they're used.

[Edit: I shouldn't say that round-to-even bias happens _only_ with those value of the lower bits. I think it's sufficient but not necessary.]

stl/inc/random
```diff
@@ -2645,53 +2646,90 @@ public:
...
+                // tail of a non-bottommost layer
+                const _Fty _Xx0   = _Tables._Layer_widths[_Irregular_layer] * _Tables._Width_scale;
+                const _Fty _Xx1   = _Tables._Layer_widths[_Irregular_layer + 1] * _Tables._Width_scale;
+                const _Fty _Yy0   = _Tables._Layer_heights[_Irregular_layer];
+                const _Fty _Yy1   = _Tables._Layer_heights[_Irregular_layer + 1];
+                const _Fty _Xdiff = _Xx1 - _Xx0;
+                const _Fty _Ydiff = _Yy1 - _Yy0;
+
+                for (;;) {
+                    const _Fty _Dx = _Xdiff * _STD _Nrand_impl<_Fty>(_Eng);
```
MattStephanson: The upper bits of `_Xbits` could be used on the first iteration. Also for the tail, below.

stl/inc/random
```diff
@@ -2645,53 +2646,90 @@ public:
...
+                // tail of the bottommost layer (infinite width)
+                constexpr _Fty _Tail            = _Tables._Layer_widths[_Tables._Layer_num] * _Tables._Width_scale;
+                constexpr _Fty _Tail_reciprocal = _Fty{1.0} / _Tail;
+                for (;;) {
+                    const _Fty _Xval = -_STD log(_Fty{1.0} - _STD _Nrand_impl<_Fty>(_Eng)) * _Tail_reciprocal;
+                    const _Fty _Yval = _STD _Nrand_impl<_Fty>(_Eng);
+                    if (_Yval < _STD exp(_Xval * _Xval * _Fty{-0.5})) {
+                        const _Fty _Xorig = _Tail + _Xval;
+                        return static_cast<_Ty>(_Xbits < 0 ? -_Xorig : _Xorig);
+                    }
```
MattStephanson: Marsaglia's tail method might be faster if we don't have a fast exponential generator, as it only needs `sqrt` when returning, instead of an additional `log` or `exp`. It can be found in [Luc Devroye's book](https://luc.devroye.org/rnbookindex.html), section IX.1.2, or [this DTIC report](https://apps.dtic.mil/sti/tr/pdf/AD0423993.pdf).

mohamed1rashad (approved):

PR #6106, opened 2026-02-24
Azure Pipelines: Use Fasv7 for x64 testing

StephanTLavavej:
I've obtained quota for a small number of [Fasv7](https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/compute-optimized/fasv7-series?tabs=sizebasic) VMs (AMD Zen 5). With only 20 VMs (each 32 cores), using them for everything would be counterproductive. However, if we use them only for the compute-dominated x64 testing stage, which is on the critical path, then we have capacity for up to 2 PRs simultaneously.

### DIFF
--- a/azure-devops/config.yml
+++ b/azure-devops/config.yml
@@ -7,6 +7,9 @@ variables:
 - name: x64SlowPoolName
   value: 'Stl-2026-02-20T0952-x64-Pool'
   readonly: true
+- name: x64FastPoolName
+  value: 'Stl-2026-02-24T1452-x64-Fasv7-Pool'
+  readonly: true
 - name: arm64PoolName
   value: 'Stl-2026-02-20T0952-arm64-Pool'
   readonly: true
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -137,7 +137,7 @@ stages:
       - Configure_Tests
     displayName: 'Test x64'
     pool:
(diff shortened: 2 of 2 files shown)


azure-devops/create-1es-hosted-pool.ps1
```diff
@@ -8,39 +8,60 @@ Creates a 1ES Hosted Pool, set up for the STL's CI.
...
+  $PoolSize = 20
+  $AvailableLocations = @('australiaeast', 'northeurope', 'southeastasia')
+} elseif ($VMSku -ieq 'Dpsv6') {
+  $Arch = 'arm64'
+  $VMSize = 'Standard_D32ps_v6'
+  $PoolSize = 32
+  $AvailableLocations = @('eastus2', 'northeurope') # Locations where CPP_STL_GitHub has quota for 1024 cores (32 VMs).
+}
+
+$AvailableLocationIdx = 5 # Increment for each new set of pools, to cycle through the available locations.
```
davidmrdavid, line 51: should this increase now that we have Fasv7?

StephanTLavavej: For future revisions, it will. This is the first Fasv7 pool I'm generating, so I'm starting with the index I used for Feb Patch Tuesday.

StephanTLavavej:

PR #6107, opened 2026-02-25
Vector algorithms: runtime coverage for ARM64EC non-vectorized fallbacks again

AlexGuteniev:
These fallbacks are executed when linking x64 objects to ARM64EC binary, but they are not executed in a non-mixed configuration. As producing mixed configuration is complex, we execute them in a test by introducing another special mode.

Unlike #6100, instead of using `_ENABLE_STL_INTERNAL_CHECK` macro, a new control macro is introduced, and the primary coverage is given to the normal mode. This separate control macro that is supported only on ARM64EC.

~~In `<xutility>` the change is rather verbose, but I'm afraid a more concise way will reduce clarity and/or robustness.~~

~~We don't need a variety of modes like `/permissive` etc, the goal is to get all these algorithms called, so usual latest will do. We still want to execute different compilers and different run-time library linking.~~

~~ASan would be desired too, but we don't have it available for ARM64EC. I've added test cases anyway for possible future ASan addition.~~

Only `VSO_0000000_vector_algorithms` need this, as `VSO_0000000_vector_algorithms_mismatch_and_lex_compare`, `VSO_0000000_vector_algorithms_floats`, and `VSO_0000000_vector_algorithms_search_n` don't test any algorithms that are vectorized on x64 not vectorized on ARM64.

Edit: stroke out what is no longer valid after review.

### DIFF
--- a/stl/inc/xutility
+++ b/stl/inc/xutility
@@ -58,19 +58,12 @@ _STL_DISABLE_CLANG_WARNINGS
 // In the STL's implementation, we inspect the following detection macros,
 // because various algorithms are vectorized for various architectures.
 
-// _CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC is a macro to test ARM64EC fallbacks for x64 algorithms
-
-#if defined(_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC) && !defined(_M_ARM64EC)
-#error _CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC should only be defined on _M_ARM64EC
-#endif // defined(_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC) && !defined(_M_ARM64EC)
+// _CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC is a macro to test ARM64EC fallbacks for x64 algorithms.
 
 #if !_USE_STD_VECTOR_ALGORITHMS
 #define _VECTORIZED_FOR_X64_X86               0
 #define _VECTORIZED_FOR_X64_X86_ARM64_ARM64EC 0
-#elif defined(_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC)
-#define _VECTORIZED_FOR_X64_X86               1
-#define _VECTORIZED_FOR_X64_X86_ARM64_ARM64EC 1
-#elif defined(_M_ARM64) || defined(_M_ARM64EC)
+#elif defined(_M_ARM64) || (defined(_M_ARM64EC) && !defined(_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC))
 #define _VECTORIZED_FOR_X64_X86               0
 #define _VECTORIZED_FOR_X64_X86_ARM64_ARM64EC 1
 #elif defined(_M_X64) || defined(_M_IX86)


tests/utils/stl/test/tests.py
```diff
@@ -316,6 +316,8 @@ def _parseFlags(self, litConfig):
                 self.requires.append('edg') # available for x64, see features.py
             elif flag[1:] == 'arch:AVX2':
                 self.requires.append('arch_avx2') # available for x86 and x64, see features.py
+            elif flag[1:] == 'D_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC':
```
AlexGuteniev, line 319: `'D' + '_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC'` might be a better spelling for search.

tests/std/tests/arm64ec_vector_fallbacks_suffix.lst
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+PM_CL="/w14640 /Zc:threadSafeInit- /EHsc /std:c++latest /MTd /D_ITERATOR_DEBUG_LEVEL=0 /D_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC"
```
StephanTLavavej: This is testing debug IDL=0 which is weird. We want to test normal here.

tests/std/tests/arm64ec_vector_fallbacks_suffix.lst
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+PM_CL="/w14640 /Zc:threadSafeInit- /EHsc /std:c++latest /MTd /D_ITERATOR_DEBUG_LEVEL=0 /D_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC"
+ASAN	PM_CL="/w14640 /Zc:threadSafeInit- /EHsc /std:c++latest /MTd /D_ITERATOR_DEBUG_LEVEL=0 /D_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC -fsanitize=address /Zi" PM_LINK="/debug"
```
StephanTLavavej: No change requested: The ASan configs don't matter because ARM64EC doesn't support it, but maybe someday in the far future...

tests/std/tests/arm64ec_vector_fallbacks_suffix.lst
```diff
@@ -0,0 +1,13 @@
...
+PM_CL="/w14640 /Zc:threadSafeInit- /EHsc /std:c++latest /MDd /D_ITERATOR_DEBUG_LEVEL=0 /D_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC"
+ASAN	PM_CL="/w14640 /Zc:threadSafeInit- /EHsc /std:c++latest /MDd /D_ITERATOR_DEBUG_LEVEL=0 /D_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC -fsanitize=address /Zi" PM_LINK="/debug"
+PM_CL="/w14640 /Zc:threadSafeInit- /EHsc /std:c++latest /MD /D_ITERATOR_DEBUG_LEVEL=0 /D_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC"
+ASAN	PM_CL="/w14640 /Zc:threadSafeInit- /EHsc /std:c++latest /MD /D_ITERATOR_DEBUG_LEVEL=0 /D_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC -fsanitize=address /Zi" PM_LINK="/debug"
+PM_COMPILER="clang-cl" PM_CL="-fno-ms-compatibility -fno-delayed-template-parsing -Wno-unqualified-std-cast-call /permissive- /EHsc /MD /D_ITERATOR_DEBUG_LEVEL=0 /D_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC --start-no-unused-arguments"
+PM_COMPILER="clang-cl" PM_CL="-fno-ms-compatibility -fno-delayed-template-parsing -Wno-unqualified-std-cast-call /permissive- /EHsc /MT /D_ITERATOR_DEBUG_LEVEL=0 /D_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC --start-no-unused-arguments"
```
StephanTLavavej: It would be much better to test Clang in latest mode, giving us more coverage, instead of just the classic algorithms.

These are also missing `/EHsc`!

After thinking about it, I don't think I like this approach - it is too fragile (we have enough config complexity already) and it's actually creating a fair number of useless configs, since this is crossed with `PM_CL="/D_USE_STD_VECTOR_ALGORITHMS=0"`. I think we should just cross this and accept that the matrix will be larger for ARM64EC.

stl/inc/xutility
```diff
@@ -58,9 +58,18 @@ _STL_DISABLE_CLANG_WARNINGS
 // In the STL's implementation, we inspect the following detection macros,
 // because various algorithms are vectorized for various architectures.
 
+// _CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC is a macro to test ARM64EC fallbacks for x64 algorithms
+
+#if defined(_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC) && !defined(_M_ARM64EC)
+#error _CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC should only be defined on _M_ARM64EC
+#endif // defined(_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC) && !defined(_M_ARM64EC)
```
StephanTLavavej: The internal test harness is going to be too dumb to exclude this. We should just remove the check.

tests/utils/stl/test/tests.py
StephanTLavavej: Thanks for thinking about searchability; in practice with macros it is always necessary to avoid whole-word searching so this is fine.

StephanTLavavej:

StephanTLavavej:
Thanks for making sure we don't damage this exotic scenario! :white_check_mark: :tada: :rocket:

PR #6109, opened 2026-02-25
Update llvm-project

StephanTLavavej:
* Update llvm-project.
* llvm/llvm-project#179780 was merged.
* llvm/llvm-project#181560 was merged.
* llvm/llvm-project#179038 was merged.
* Two `locale.time.get.members` tests are now passing, possibly related to llvm/llvm-project#179986 being merged.
* Mark `ranges_shift_right.pass.cpp` as SKIPPED due to `constexpr` step limits, like `ranges.shift_left.pass.cpp`.
* `any.cons/value.pass.cpp` is failing for Clang after llvm/llvm-project#177082 was merged.
* Attempt to enable the filesystem tests.
  + These skips date back to #4263 on 2023-12-15 where I commented:
    > I had to undo an attempted change and go back to using SKIPPED for the big group of filesystem tests. I'm adding a comment that these tests are SKIPPED due to unknown differences between local test runs (where they fail normally) and Azure Pipelines (where they fail mysteriously in a way that FAIL doesn't handle, and that doesn't result in usable test logs.)
  + Presumably, infrastructure changes in the last 2 years have eliminated whatever differences were responsible.
* Categorize filesystem failures.
* Update the expected results for the MSVC-internal harness.

### DIFF
--- a/tests/libcxx/expected_results.txt
+++ b/tests/libcxx/expected_results.txt
@@ -304,6 +304,7 @@ std/ranges/range.factories/range.iota.view/indices.pass.cpp FAIL
 # Add 'std-at-least-c++26' to tests/utils/stl/test/tests.py when beginning work on C++26.
 std/utilities/optional/optional.iterator/begin.pass.cpp FAIL
 std/utilities/optional/optional.iterator/borrowed_range.compile.pass.cpp FAIL
+std/utilities/optional/optional.iterator/compare.pass.cpp FAIL
 std/utilities/optional/optional.iterator/end.pass.cpp FAIL
 std/utilities/optional/optional.iterator/iterator.pass.cpp FAIL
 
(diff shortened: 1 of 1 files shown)


tests/libcxx/expected_results.txt
```diff
@@ -1224,6 +1111,35 @@ std/input.output/file.streams/fstreams/filebuf.virtuals/xsputn.pass.cpp:1 FAIL
 std/utilities/meta/meta.unary/meta.unary.prop/is_implicit_lifetime.pass.cpp:0 FAIL
 std/utilities/meta/meta.unary/meta.unary.prop/is_implicit_lifetime.pass.cpp:1 FAIL
 
+# Not analyzed. Clang error: no member named 'value' in 'std::is_copy_constructible<ConvertibleFromAndToAny>'
+std/utilities/any/any.class/any.cons/value.pass.cpp:2 FAIL
```
frederick-vs-ja, line 1116: It turned out that our implementation strategy for `conjunction` can cause meta-recursion for other compilers ([Godbolt link](https://godbolt.org/z/9ahxz8aoa)).

Perhaps we should use a different way to achieve shortcut instantiation in either `conjunction` or the related `any` constructor.

StephanTLavavej:

PR #6114, opened 2026-02-26
`<any>`: Short-cut constraints of single-argument `any` constructor, again

frederick-vs-ja:
The issue was originally reported as LLVM-176877 and recently found reproducible with MSVC STL + Clang (see #6109).

It's a bit unfortunately that our implementation strategy for `conjunction` is a bit too eager to achieve shortcut instantiation for the example posted in LLVM-176877. The strategy currently works for MSVC but not other compilers ([Godbolt link](https://godbolt.org/z/9ahxz8aoa)).

This patch chooses to modify the implementation strategy for the involved `any` constructor to avoid using `conjunction`.

Test file changes are mirrored from LLVM-177082.

### DIFF
--- a/tests/libcxx/expected_results.txt
+++ b/tests/libcxx/expected_results.txt
@@ -1112,9 +1112,6 @@ std/input.output/file.streams/fstreams/filebuf.virtuals/xsputn.pass.cpp:1 FAIL
 std/utilities/meta/meta.unary/meta.unary.prop/is_implicit_lifetime.pass.cpp:0 FAIL
 std/utilities/meta/meta.unary/meta.unary.prop/is_implicit_lifetime.pass.cpp:1 FAIL
 
-# Not analyzed. Clang error: no member named 'value' in 'std::is_copy_constructible<ConvertibleFromAndToAny>'
-std/utilities/any/any.class/any.cons/value.pass.cpp:2 FAIL
-
 # Not analyzed. Assertion failed: globalMemCounter.checkOutstandingNewLessThanOrEqual(1)
 std/input.output/filesystems/class.path/path.member/path.assign/move.pass.cpp FAIL
 std/input.output/filesystems/class.path/path.member/path.construct/move.pass.cpp FAIL


tests/std/tests/P0220R1_any/test.cpp
```diff
@@ -1160,15 +1160,34 @@ void test_sfinae_constraints() {
...
     test_copy_value_throws<large_throws_on_copy>();
     test_move_value_throws();
     test_sfinae_constraints();
+    test_no_constraint_recursion();
 
-  return 0;
+    return 0;
 }
 } // namespace ctor::value
 // -- END: test/std/utilities/any/any.class/any.cons/value.pass.cpp
```
AlexGuteniev, line 1194: This code is marked as originated from LLVM test.
I think new code should be added elsewhere.

frederick-vs-ja: However, the new code is also originated from LLVM test. I initially added it to LLVM's `libcxx/test/std/utilities/any/any.class/any.cons/value.pass.cpp` in LLVM-177082.

AlexGuteniev: Ah, that's fine then

StephanTLavavej: I fixed your citation which had a digit transposition.

stl/inc/any
```diff
@@ -130,10 +130,11 @@ public:
...
 
-    template <class _ValueType, enable_if_t<conjunction_v<negation<is_same<decay_t<_ValueType>, any>>,
-                                                negation<_Is_specialization<decay_t<_ValueType>, in_place_type_t>>,
-                                                is_copy_constructible<decay_t<_ValueType>>>,
-                                    int> = 0>
+    template <class _ValueType,
+        enable_if_t<!is_same_v<decay_t<_ValueType>, any> && !_Is_specialization_v<decay_t<_ValueType>, in_place_type_t>,
+            int> = 0, // Separate these conditions into two `enable_if_t`s to avoid meta-recursion.
+                      // See LLVM-176877 and GH-6109.
+        enable_if_t<is_copy_constructible_v<decay_t<_ValueType>>, int> = 0>
```
StephanTLavavej, line 139: I forgot that an ordering guarantee existed and had to look it up, so let's cite it.

StephanTLavavej:
Thanks! I'll remember the stealth merge conflict with #6109.

StephanTLavavej:

StephanTLavavej (approved):

StephanTLavavej:
Thanks for jumping on this test failure! :heart_eyes_cat: :tada: :white_check_mark:

PR #6117, opened 2026-02-27
Cleanup documentation URLs

StephanTLavavej:
Today I learned that our documentation team recommends omitting locale/version information from URLs (unless linking to a specific version is desired). So while I generally prefer to avoid URLs that get rewritten (except when intentionally using a redirector), following this recommendation is desirable and makes the URLs a bit shorter. I've manually verified that all of these updated URLs still work.

* `(https://learn\.microsoft\.com/)en-us/([^ ?#]+)(?:\?view=([^ #]+))?` => `$1$2`
* Reduce the Feedback Hub URL.
  + Like DevCom URLs, everything after the number is ignored.
* Expand shortened URLs, reduce A/W file differences.
  + I had created these shortened URLs purely to stay within 120 columns.
* Drop `?tabs=powershell`.
* Drop redundant DevCom URL.

There's one URL which still incurs a major redirect, where `.github/ISSUE_TEMPLATE/question.md` cites https://help.github.com/en/github/managing-your-work-on-github/creating-a-permanent-link-to-a-code-snippet which redirects to https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-a-permanent-link-to-a-code-snippet, but that exceeds 120 columns so I'm not changing it (and it isn't worth the additional confusion of linking to the parent doc node).

### DIFF
--- a/tests/std/tests/P0896R4_ranges_range_machinery/test.cpp
+++ b/tests/std/tests/P0896R4_ranges_range_machinery/test.cpp
@@ -2090,7 +2090,6 @@ constexpr bool complicated_algorithm_test() {
 }
 
 // Regression test for DevCom-739010 (aka VSO-985597)
-// https://developercommunity.visualstudio.com/content/problem/739010/meow.html
 // which allows overload resolution to prefer a hidden friend `const T&` overload of `begin`
 // for an rvalue `T` over the deleted `begin(T&&)` instantiated from the poison pill.
 template <class T>


README.md
```diff
@@ -616,14 +616,14 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 [enhancement tag]: https://github.com/microsoft/STL/issues?q=is%3Aopen+is%3Aissue+label%3Aenhancement
 [gbenchmark]: https://github.com/google/benchmark
 [gbenchmark:docs]: https://github.com/google/benchmark/blob/main/docs/user_guide.md
-[hub]: https://support.microsoft.com/en-us/help/4021566/windows-10-send-feedback-to-microsoft-with-feedback-hub-app
+[hub]: https://support.microsoft.com/help/4021566
```
michaelbprice, line 619: This one seems different than all the rest. It looks like it just ignores whatever comes after the `4021566` when doing the redirection. Very weird, but it the edit looks good.

StephanTLavavej: Yep, I intentionally dropped that stuff. It behaves like a DevCom URL. Since it's ignored, and the whole thing is rewritten anyways, I figured it was simplest to prune the URL down to what matters.

stl/src/StlCompareStringW.cpp
```diff
@@ -14,10 +14,9 @@
 //
 // Entry:
 //   LPCWSTR  LocaleName  - locale context for the comparison.
-//   DWORD    dwCmpFlags  - see https://aka.ms/stl/comparestringex
+//   DWORD    dwCmpFlags  - https://learn.microsoft.com/windows/win32/api/stringapiset/nf-stringapiset-comparestringex
```
AlexGuteniev, line 17: These two URLs are different from the other two

StephanTLavavej: Yes, there were two pairs of URLs. I replaced each with what they expand to, then dropped the `/en-us` component.

StephanTLavavej:

Issue #6118, opened 2026-02-27
`<regex>`: `regex_search()` sometimes incorrectly matches capturing groups


muellerj2:
# Describe the bug
`regex_search()` fails to reset the capturing group state correctly between match attempts. Because of this, it might claim that some capturing groups are matched even though they shouldn't be.

# Test case
```
#include <iostream>
#include <regex>
#include <string>

using namespace std;

int main() {
    smatch captures;
    regex re("a|(b)c");
    string input("ba");

    auto result = regex_search(input, captures, re);

    cout << "search succeeded: " << result << '\n';

    if (result) {
        cout << "matched character sequence: " << captures[0].str() << '\n';
        cout << "capturing group 1 matched: " << captures[1].matched << '\n';
        cout << "contents of capturing group 1: " << captures[1].str();
    }

    return 0;
}
```
Godbolt link: https://godbolt.org/z/Wsb6aEjTf

This program produces the following output:

```
search succeeded: 1
matched character sequence: a
capturing group 1 matched: 1
contents of capturing group 1: b
```

# Expected behavior
The program should produce the following output:

```
search succeeded: 1
matched character sequence: a
capturing group 1 matched: 0
contents of capturing group 1: 
```

# STL version
This bug appears to have been introduced in MSVC Build Tools 19.50.

# Additional context
The setup code for the matcher in `_Matcher(x)::_Match()` has never contained explicit code to reset the capturing group state. But until recently, the matcher tried to reset captures when it encountered an `_N_capture` node: https://github.com/microsoft/STL/blob/713dd95901f79aad7d51bb43435fa40727390d46/stl/inc/regex#L3605-L3608

But this loop was an inadequate attempt to implement ECMAScript's capturing group reset rule and a major source of bugs in this area. #5456 applied changes to reset capturing groups according to the ECMAScript standard, and one change was to remove this loop. But this had the subtle consequence that capturing groups were no longer reset when the `_N_capture` node for group 0 was encountered, so the matched capturing groups from a prior failed match attempt could now spill over into the following matches.

muellerj2:
I had assumed this still reproduces, but it seems that this is fixed as a side effect of #5865 and #5918, because capture groups get reset by the manual stack unwinding mechanism now when matching fails completely.

But this just shows how incredibly subtle this problem is, so I will still open a PR to add some test coverage to avoid regressing this in the future.

PR #5918, opened 2025-11-29
`<regex>`: Remove capture validity vectors from stack frames

muellerj2:
This removes the capturing group validity vectors in stack frames that were used to snapshot the validity/matched status of capturing groups whenever a stack frame was pushed. The restoration of this status is now achieved by processing new or modified opcodes during unwinding plus some changed logic when the pattern of a lookahead assertion matched:
* The opcode `_Capture_restore_end` is split into `_Capture_restore_matched_end` and `_Capture_restore_unmatched_end`, which are pushed depending on whether a capturing group is already matched while the `_N_end_capture` node is processed. `_Capture_restore_matched_end` keeps the capturing group matched (so doesn't do anything about the matched status), `_Capture_restore_unmatched_end` resets it to unmatched.
* The reset of the status to unmatched at the start of a loop in ECMAScript is now performed by `_Matcher3::_Reset_capture_groups()`, replacing the prior calls to `std::fill()`. This function pushes a new stack frame with opcode `_Capture_restore_matched` for every capture group whose status is changed to unmatched.
  - The new test cases fail when the status is reset to unmatched, but the reset is not undone when backtracking.
* For successful positive lookahead assertions, we keep the stack frames with opcodes `_Capture_restore_unmatched_end` and any with opcode `_Capture_restore_begin` before them on the stack. We don't have to keep those with opcode `_Capture_restore_matched_end` because ECMAScript rules guarantee that the capturing groups inside the lookahead assertion are unmatched when processing the lookahead assertion starts, so the stack frame pushed for the first modification of a capturing group's end pointer in this lookahead assertion must have opcode `_Capture_restore_unmatched_end`.
* For a failed negative lookahead assertion (i.e., one whose asserted pattern matched), we have to reset all the capturing groups to unmatched status. When going through the stack frames pushed by such an assertion, the code is changed to track the first and last matched capturing group inside this assertion, and then calls `std::fill()` to reset their status to unmatched. (We don't have to worry about restoring the begin and end pointers of the capturing groups because the capturing groups are always unmatched when leaving a negative lookahead assertion, so the pointers are meaningless.)

With this PR, the worst-case number of allocations is logarithmic in the size of the input (pattern + searched string) and no longer linear. But even for patterns like "a*", where the capture extent and validity vectors in the stack frames did not actually allocate, we still see some major performance improvement, probably because the overhead of managing these vectors is gone.

[...]

PR #6122, opened 2026-02-28
`<regex>`: Add test coverage for capture group reset between match attempts in `regex_search()`

muellerj2:
We failed to notice that we regressed #6118 and that we later fixed it again. The regression and the fix were subtle side effects of changes that were intended to fix other issues in the capture group handling in the regex matcher (#5456, #5865 and #5918). But we didn't realize the impact on the capture group reset between match attempts in `regex_search()` for any of them.

This PR adds some minimal test coverage for the capture group reset in `regex_search()` so that we will not accidentally regress this again. Each test case covers one of the two `_Matcher3::_Match()` calls in `regex_search` after the initial match attempt.

I confirmed that MSVC Build Tools 19.50 fails both test cases.


StephanTLavavej:

StephanTLavavej:
:beetle: :no_entry: :tada:
===

PR #6125, opened 2026-03-01
Try to unskip another test; it works on my machine

AlexGuteniev:
I even disabled long path support and tested x86 to be sure

### DIFF
--- a/tests/libcxx/expected_results.txt
+++ b/tests/libcxx/expected_results.txt
@@ -911,10 +911,6 @@ std/ranges/range.adaptors/range.lazy.split/range.lazy.split.inner/iter_swap.pass
 # Clang error: non-type template argument is not a constant expression
 std/utilities/tuple/tuple.tuple/tuple.cnstr/convert_const_move.pass.cpp:2 FAIL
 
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
This still fails on my machine even though I enabled long path support. What's your repo root?

```
D:\GitHub\STL\out\x64\tests\libcxx\std\thread\thread.mutex\thread.mutex.requirements\thread.sharedtimedmutex.requirements\thread.sharedtimedmutex.class\Output\try_lock_until_deadlock_bug.pass.cpp.dir\0\lnk{AE6FDA3F-20B6-489C-A3A5-4266E78AD28B}.tmp(2) : error RC2136 : missing '=' in EXSTYLE=<flags>
LINK : fatal error LNK1327: failure during running rc.exe
```

AlexGuteniev:
`C:\Project\STL`

StephanTLavavej:
Surprising since it's one character longer than mine.

In any event, because it's flaky on my machine, I don't think we can merge this. I still wish upstream would change this ludicrously long path (this is in the category of little favors that I wish my charconv contribution would get us).

PR #6132, opened 2026-03-02
`<filesystem>`: Set the error code for `fs::absolute("", ec)`.

### LINKED ISSUE #6120: `<filesystem>`: workaround dubious `absolute("", ec)` test case in libcxx suite
# Describe the bug
I was investigating a libcxx [failure](https://github.com/microsoft/STL/blob/117ca965fe2f7157540edd9df5a62b0e42d7ece2/tests/libcxx/expected_results.txt#L1125-L1126) on test `std\input.output\filesystems\fs.op.funcs\fs.op.absolute\absolute.pass.cpp`.
It fails because of a dubious [test case](https://github.com/llvm/llvm-project/blob/58f4da463e5b3cd3531cace17cc3f2d8d860964e/libcxx/test/std/input.output/filesystems/fs.op.funcs/fs.op.absolute/absolute.pass.cpp#L42) that involves calling `fs::absolute("", ec)`.
Interestingly, libstdc++, libc++, and MSVC STL all handle this differently.

## Code
https://godbolt.org/z/MK97zP3jT

```c++
#include <filesystem>
#include <iostream>

int main()
{
    const fs::path cwd = fs::current_path();
    std::error_code ec{};

    std::cout << "CWD is " << cwd << std::endl;

    const fs::path ret = fs::absolute("", ec);

    std::cout << ret << '\t' << ec << std::endl;
}
```

## Output
|           | `ret`   | `ec`       |
|-----------|---------|------------|
| libstdc++ | `""`    |`generic:22`|
| libc++    | `cwd /` |`system:0`  |
| MSVC STL  | `""`    |`system:0`  | 

# Command-line test case
See Godbolt link.

# Expected behavior
- I think that returning an empty `path()` is fine; it's probably the desired behavior, unlike libc++.
  - However, since it's an 'erroneous' case, we may consider changing the implementation such that `ec` is set with an appropriate error code.
- If no implementation change is made, it should suffice to just document the reason of the test failure and move on.

# STL version
```
Microsoft (R) C/C++ Optimizing Compiler Version 19.50.35725 for x64
```
# Additional context

leejy12:
Closes #6120 

`fs::absolute("", ec)` was returning an empty `path`, indicating an error has occurred, but didn't set `ec` to the appropriate `error_code`.

libstdc++ treats this call as an error, whereas libc++ treats it as a success case.

[[fs.op.absolute#1]](https://eel.is/c++draft/fs.op.absolute#1) says the function "Composes an absolute path referencing the same file system location as p according to the operating system". On Windows, `GetFullPathNameW(L"", ...)` fails, so following libstdc++’s behavior is more consistent with the standard wording.

This change updates the implementation to set `ec.value()` to 123 (`ERROR_INVALID_NAME`).


YexuanXiao:
The error code for calling `CreateFile` with an empty string is `ERROR_PATH_NOT_FOUND` (`3`). Is it more appropriate to use it? This means the result obtained when not converting it to an absolute path.

leejy12:
@YexuanXiao I intended to let `__std_fs_get_full_path_name` populate it automatically.

https://github.com/microsoft/STL/blob/117ca965fe2f7157540edd9df5a62b0e42d7ece2/stl/src/filesystem.cpp#L156-L160

But libstdc++ sets the error to `EINVAL` (22), so a better error may be `ERROR_INVALID_PARAMETER` (87). I think if we just return this when `_Input._Text.empty()` is true, we can avoid calling the Win32 API.

StephanTLavavej:
I like the behavior of allowing empty strings through to the Windows API and returning whatever it returns, instead of attempting to synthesize our own error code.

StephanTLavavej:

StephanTLavavej:
Thanks for investigating this libcxx failure and fixing our implementation! :floppy_disk: :hammer_and_wrench: :tada:

Issue #6130, opened 2026-03-02
`<xloctime>`: `time_get::do_get` doesn't parse a literal `%`


cpplearner:
# Describe the bug
MSVC STL's [`time_get::do_get`](https://github.com/microsoft/STL/blob/117ca965fe2f7157540edd9df5a62b0e42d7ece2/stl/inc/xloctime#L436) doesn't know the `%%` specifier. It should recognize this specifier and consume a literal `%`.

# Command-line test case
```
D:\test>type test-time-get-do-get.cpp
```
```c++
#include <cassert>
#include <ctime>
#include <iomanip>
#include <sstream>

int main() {
    std::tm t{};
    std::stringstream ss{"% "};
    ss >> std::get_time(&t, "%%");
    assert(!ss.fail());
}
```
```
D:\test>cl /EHsc test-time-get-do-get.cpp
```
```
Microsoft (R) C/C++ Optimizing Compiler Version 19.50.35725 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

test-time-get-do-get.cpp
Microsoft (R) Incremental Linker Version 14.50.35725.0
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:test-time-get-do-get.exe
test-time-get-do-get.obj
```
```
D:\test>test-time-get-do-get.exe
```
```
Assertion failed: !ss.fail(), file test-time-get-do-get.cpp, line 10
```

# Expected behavior
Assertion passed.

# STL version
https://github.com/microsoft/STL/commit/117ca965fe2f7157540edd9df5a62b0e42d7ece2

# Additional context
Both #6129 and this issue concern `time_get::do_get` (and affect the same `libcxx` test file), but they can be fixed independently.

In fact, IMO this issue is much easier to fix and could be a "good first issue".

StephanTLavavej:
Locale bugs are too scary for me to mark as good first issue.

PR #6133, opened 2026-03-03
`<xloctime>`: Fix `time_get::do_get` format strings for `%c` and `%x`

### LINKED ISSUE #6129: `<xloctime>`: `time_get::do_get` uses the wrong format for `%c` and `%x`
# Describe the bug
MSVC STL's `time_get::do_get` hardcodes `%c` as equivalent to `%b %d %H : %M : %S %Y`, and `%x` as equivalent to `%d / %m / %y`.

https://github.com/microsoft/STL/blob/117ca965fe2f7157540edd9df5a62b0e42d7ece2/stl/inc/xloctime#L455-L457
https://github.com/microsoft/STL/blob/117ca965fe2f7157540edd9df5a62b0e42d7ece2/stl/inc/xloctime#L549-L551

This doesn't match what `strftime` produces:
- In the "C" locale, `%c` is equivalent to `%a %b %e %T %Y`, and `%x` is equivalent to `%m/%d/%y`. MSVC notably doesn't parse `%a` for the former, and has the wrong order for the latter.
- Other locales might have very different formats.

Fixing this would be a behavior change (even if ABI-preserving). I don't know if it's worth the trouble.

This affects `libcxx` test `std/localization/locale.categories/category.time/locale.time.get/locale.time.get.members/get_one.pass.cpp`.

# Command-line test case
```
D:\test>type test-time-get-do-get.cpp
```
```c++
#include <cassert>
#include <ctime>
#include <iostream>
#include <locale>

using I = const char*;

class my_facet : public std::time_get<char, I>
{
public:
    explicit my_facet(std::size_t refs = 0)
        : time_get(refs) {}
};

int main() {
    const my_facet f(1);
    std::ios str(nullptr);
    std::ios_base::iostate err;
    std::tm t;

    {
        const char in[] = "Thu Jun  6 09:49:10 2009";
        err = std::ios_base::goodbit;
        t = std::tm();
        I i = f.get(I(in), I(in+sizeof(in)-1), str, err, &t, 'c');
        assert(i == in+sizeof(in)-1);
    }
}
```
```
D:\test>cl /EHsc test-time-get-do-get.cpp
```
```
Microsoft (R) C/C++ Optimizing Compiler Version 19.50.35725 for x64

test-time-get-do-get.cpp
Microsoft (R) Incremental Linker Version 14.50.35725.0

/out:test-time-get-do-get.exe
test-time-get-do-get.obj
```
```
D:\test>test-time-get-do-get.exe
```
```
Assertion failed: i == in+sizeof(in)-1, file test-time-get-do-get.cpp, line 26
```

# Expected behavior
Assertion passed.

# STL version
https://github.com/microsoft/STL/commit/117ca965fe2f7157540edd9df5a62b0e42d7ece2

# Additional context

BrandonPacewic:
Fixes #6129.

`time_get::do_get` hardcoded the wrong format strings for `%c` and `%x` in the C locale:

  - `%c` was `"%b %d %H : %M : %S %Y"`, missing the leading weekday field entirely. It should be `"%a %b %e %T %Y"` to match what `strftime("%c")` produces.
  - `%x` was `"%d / %m / %y"` (day/month order), but should be `"%m/%d/%y"` (month/day order) to match `strftime("%x")`.

As the original open issue describes, this is an ABI preserving behavior change. The existing test `Dev11_0836436_get_time` is updated accordingly and its `%x` test inputs were written for the old day first format and its comment cited `"%d / %m / %y"` directly.

The libcxx test `std/localization/locale.categories/category.time/locale.time.get/locale.time.get.members/get_one.pass.cpp` still fails due to a separate pre-existing bug where `do_get` does not handle the `%%` specifier (falls to `default: failbit`) (#6130). Its entry in `expected_results.txt` is left in place.

Also updates `P0355R7_calendars_and_time_zones_io` test inputs for `%c` and `%x`, since `std::chrono::parse` routes those format specifiers through `time_get::do_get` and the existing inputs were written against the old incorrect format strings.

<details>
<summary>Force push note 🐱 </summary>
<br>
I force pushed during early preparation of this PR to fix a commit message and add a missing `test.lst` entry before any review had begun. I understand that force pushing during active review is disruptive and I will not do so going forward. 🙃 

</details>

### DIFF
--- a/tests/std/tests/Dev11_0836436_get_time/test.cpp
+++ b/tests/std/tests/Dev11_0836436_get_time/test.cpp
@@ -144,9 +144,9 @@ int main() {
     assert(read_hour("11 PM") == 23);
     assert(read_hour("11 pm") == 23);
 
-    assert(read_date("04/22/77") == make_tuple(22, /*NOTE DIFFERENCE:*/ 3, 77));
+    assert(read_date("04 / 22 / 77") == make_tuple(22, /*NOTE DIFFERENCE:*/ 3, 77));
 
-    assert(read_date("04/22/11") == make_tuple(22, /*NOTE DIFFERENCE:*/ 3, /*NOTE DIFFERENCE:*/ 111));
+    assert(read_date("04 / 22 / 11") == make_tuple(22, /*NOTE DIFFERENCE:*/ 3, /*NOTE DIFFERENCE:*/ 111));
 
     assert(read_time("15 : 47 : 58") == make_tuple(15, 47, 58));
 
@@ -930,7 +930,13 @@ void test_gh_6129() {
 
     // %x in the C locale uses "%m / %d / %y"
     {
-        const auto t = helper("03/15/09", "%x");
+        const auto t = helper("03 / 15 / 09", "%x"); // test with whitespace
+        assert(t.tm_mon == 2);
+        assert(t.tm_mday == 15);
+        assert(t.tm_year == 109);
+    }
+    {
+        const auto t = helper("03/15/09", "%x"); // test without whitespace
         assert(t.tm_mon == 2);
         assert(t.tm_mday == 15);
         assert(t.tm_year == 109);


tests/std/tests/GH_006129_time_get_do_get_format/test.cpp
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Microsoft Corporation.
```
cpplearner: Instead of adding a new directory, can you add the test code to `tests/std/tests/Dev11_0836436_get_time/test.cpp`?

In this repo, extending an existing test (if reasonable) is preferred over adding a new test, because a new test has a higher runtime cost. This is not a strict requirement though.

BrandonPacewic: Makes sense 👍

stl/inc/xloctime
```diff
@@ -547,7 +547,7 @@ protected:
             break;
 
         case 'x':
-            _First = _Getfmt(_First, _Last, _Iosbase, _State, _Pt, "%d / %m / %y");
+            _First = _Getfmt(_First, _Last, _Iosbase, _State, _Pt, "%m/%d/%y");
```
MattStephanson: Similar specifiers, like `%D` and `%T`, continue to allow whitespace. This is consistent with the [POSIX spec](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/functions/strptime.html) up to 2018, although I note that the [2024 version](https://pubs.opengroup.org/onlinepubs/9799919799/functions/strptime.html) removes the whitespace. For consistency, I think it's better to keep it here.

BrandonPacewic: Just to confirm, current implementation correct?

MattStephanson: Sorry, I'm not sure what you're asking.

BrandonPacewic: I just misread your suggestion.

tests/std/tests/Dev11_0836436_get_time/test.cpp
```diff
@@ -142,9 +144,9 @@ int main() {
     assert(read_hour("11 PM") == 23);
     assert(read_hour("11 pm") == 23);
 
-    assert(read_date("22 / 4 / 77") == make_tuple(22, /*NOTE DIFFERENCE:*/ 3, 77));
+    assert(read_date("04/22/77") == make_tuple(22, /*NOTE DIFFERENCE:*/ 3, 77));
 
-    assert(read_date("22 / 4 / 11") == make_tuple(22, /*NOTE DIFFERENCE:*/ 3, /*NOTE DIFFERENCE:*/ 111));
+    assert(read_date("04/22/11") == make_tuple(22, /*NOTE DIFFERENCE:*/ 3, /*NOTE DIFFERENCE:*/ 111));
 
     assert(read_time("15 : 47 : 58") == make_tuple(15, 47, 58));
```
MattStephanson, line 149: I'd also update the test to put the whitespace back in, to make sure that the parser accepts it. Sorry I neglected to mention it earlier. But otherwise LGTM, thanks!

MattStephanson (approved):

StephanTLavavej:
Thanks! :heart_eyes_cat: I pushed a test change as recommended by @MattStephanson.

StephanTLavavej:

StephanTLavavej:
Thanks for fixing this bug! :stopwatch: :world_map: :earth_asia:

PR #935, opened 2020-06-29
<complex>: Improve numerical accuracy of sqrt and log

statementreply:
Goals:
- [x] Fix `sqrt` overflow when input is huge.
- [x] Fix `sqrt` inaccuracy when input is tiny, due to internal underflow.
- [x] Fix `log` inaccuracy when input is tiny, due to internal underflow.
- [x] Improve accuracy of `log` when |z| ≈ 1, and fix `log(complex<float>{1, 0}) != 0` under certain combinations of compile time and runtime settings. (DevCom-1093507 is a symptom of the inaccuracy. It is itself non-bug.)
  - [x] Use hardware FMA on x86/x64 when available.
  - [x] Use hardware FMA on ~arm~/arm64 when available.
- [x] Add test coverage.

The changes contain algorithm described in W. Kahan (1987), _Branch Cuts for Complex Elementary Functions, or Much Ado About Nothing's Sign Bit_.

PR #6137, opened 2026-03-03
`<complex>`: Include full `<intrin.h>` for Clang 22 compatibility

StephanTLavavej:
Reported to me internally by Chris Pulido. The STL is currently tested with what VS currently ships, Clang 20. However, the just-released Clang 22 is incompatible:

```
C:\Temp>type meow.cpp
#include <complex>

C:\Temp>clang-cl -v
clang version 22.1.0 (https://github.com/llvm/llvm-project 4434dabb69916856b824f68a64b029c67175e532)
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Program Files\LLVM\bin

C:\Temp>clang-cl /EHsc /nologo /W4 /std:c++latest /MTd /Od /c meow.cpp
In file included from meow.cpp:1:
C:\Program Files\Microsoft Visual Studio\18\Insiders\VC\Tools\MSVC\14.51.36014\include\complex(28,28): error:
      non-constexpr declaration of '_mm_fmsub_sd' follows constexpr declaration
   28 | extern "C" __m128d __cdecl _mm_fmsub_sd(__m128d, __m128d, __m128d);
      |                            ^
C:\Program Files\LLVM\lib\clang\22\include\fmaintrin.h(228,1): note: previous declaration is here
  228 | _mm_fmsub_sd(__m128d __A, __m128d __B, __m128d __C) {
      | ^
1 error generated.
```

This happened because `<complex>` tried to cleverly hand-declare an intrinsic to avoid pulling in a heavyweight intrinsics header (in #935 merged 2020-11-09). However, such hand-declarations have been enormously problematic in the past, so we've avoided this elsewhere. While this particular `constexpr` problem is new, it just highlights that we should have been following our conventions more strictly.

`<complex>` is a leaf header (included by only the C++17-deprecated `<ccomplex>` and `<ctgmath>` headers) so its throughput isn't as critical as other headers. We should include full `<intrin.h>` for now, for all compilers. This also eliminates the only inclusion of a `<MEOWmmintrin.h>` header.

### DIFF
--- a/stl/inc/complex
+++ b/stl/inc/complex
@@ -24,8 +24,7 @@
 #elif defined(_M_IX86) || defined(_M_X64)
 #define _FMP_USING_X86_X64_INTRINSICS
 #include <__msvc_bit_utils.hpp>
-#include <emmintrin.h>
-extern "C" __m128d __cdecl _mm_fmsub_sd(__m128d, __m128d, __m128d);
+#include <intrin.h>
 #endif // ^^^ defined(_M_IX86) || defined(_M_X64) ^^^
 
 #pragma pack(push, _CRT_PACKING)


AlexGuteniev:
> (they contain an ARM64EC `#error this header should only be included through <intrin.h>` that I'm surprised wasn't causing trouble of its own).

A possible reason is that Windows SDK pulls `<intrin.h>` in elsewhere; see #5596

AlexGuteniev:
> (they contain an ARM64EC `#error this header should only be included through <intrin.h>` that I'm surprised wasn't causing trouble of its own).

Actually, isn't ARM64EC handled just above?

StephanTLavavej:
You're right, thanks. Edited.

AlexGuteniev:
You can avoid including `<intrin.h>` and compile in Clang by using `__builtin_fma` like this:
```c++
[[gnu::target("fma")]]
double fma(double a, double b, double c)
{
    return __builtin_fma(a, b, -c);
}
```

https://godbolt.org/z/oczaPoK6W

AlexGuteniev:
(`__builtin_fma` also works on ARM64/ARM64EC, though for these targets `[[gnu::target("fma")]]` should not be present)

StephanTLavavej:
I would be open to a followup although I'd prefer to see documentation (I don't see clear documentation for it in Clang). I don't want to radically change the FMA logic in this PR.

AlexGuteniev:
It is exactly the same machine code, but ok.

AlexGuteniev:
That's maybe indeed nuanced actually. See llvm/llvm-project#126025.

There are conditions where `__builtin_fma` may instead of emitting one instruction potentially set `errno`.

I _think_ it does not happen with `[[gnu::target("fma")]]`, but apparently there's no documentation where this is written.

It is also possible to use a different intrinsic llvm/llvm-project#126288, which is also undocumented.

Are you still open to that?

StephanTLavavej:
No, I don't want to use undocumented intrinsics (too much risk). Thanks for looking into this.

StephanTLavavej:

PR #6141, opened 2026-03-05
Add Neon implementations of `find_last_of`

hazzlim:
This PR adds Neon implementations (Shuffle, Bitmap, and Find-Haystack-in-Needle) of `find_last_of`, `find_last_not_of`.

(Hopefully slighty better formatted) Benchmark results 🕐:
  | MSVC | Clang
-- | -- | --
`bm<AlgType::str_member_last, char>/2/3` | 0.926 | 0.977
`bm<AlgType::str_member_last, char>/6/81` | 0.95 | 0.722
`bm<AlgType::str_member_last, char>/7/4` | 0.774 | 0.747
`bm<AlgType::str_member_last, char>/9/3` | 0.986 | 0.944
`bm<AlgType::str_member_last, char>/22/5` | 1.169 | 1.069
`bm<AlgType::str_member_last, char>/58/2` | 2.288 | 2.394
`bm<AlgType::str_member_last, char>/75/85` | 1.304 | 1.071
`bm<AlgType::str_member_last, char>/102/4` | 2.178 | 2.229
`bm<AlgType::str_member_last, char>/200/46` | 2.133 | 1.966
`bm<AlgType::str_member_last, char>/325/1` | 5.349 | 5.629
`bm<AlgType::str_member_last, char>/400/50` | 2.666 | 2.588
`bm<AlgType::str_member_last, char>/1011/11` | 3.707 | 3.833
`bm<AlgType::str_member_last, char>/1280/46` | 3.889 | 3.798
`bm<AlgType::str_member_last, char>/1502/23` | 4.051 | 4.128
`bm<AlgType::str_member_last, char>/2203/54` | 4.167 | 3.895
`bm<AlgType::str_member_last, char>/3056/7` | 4.255 | 4.356
`bm<AlgType::str_member_last, wchar_t>/2/3` | 0.975 | 0.957
`bm<AlgType::str_member_last, wchar_t>/6/81` | 0.96 | 0.787
`bm<AlgType::str_member_last, wchar_t>/7/4` | 0.852 | 0.821
`bm<AlgType::str_member_last, wchar_t>/9/3` | 0.933 | 0.816
`bm<AlgType::str_member_last, wchar_t>/22/5` | 1.175 | 1.035
`bm<AlgType::str_member_last, wchar_t>/58/2` | 2.399 | 1.923
`bm<AlgType::str_member_last, wchar_t>/75/85` | 1.442 | 1.116
[...]

PR #6139, opened 2026-03-03
Add Neon implementations of `search`, `find_end`

hazzlim:
This PR adds Neon implementations of `search`, `find_end`.

Note: the code in this PR is affected by DevCom-11053447 "ARM64 suboptimal codegen: reload of stack spills on every loop iteration".

Benchmark results 🕐 (measured on a Neoverse N2 machine):

Benchmark | MSVC SU | Clang SU
---|---|---
`c_strstr/0` | 1.032 | 1
`c_strstr/1` | 1.006 | 0.983
`c_strstr/2` | 0.977 | 1
`c_strstr/3` | 0.987 | 1
`c_strstr/4` | 1.045 | 1
`c_strstr/5` | 1 | 1
`classic_search<std::uint8_t>/0` | 5.769 | 5.476
`classic_search<std::uint8_t>/1` | 6.071 | 6.273
`classic_search<std::uint8_t>/2` | 5.622 | 6.061
`classic_search<std::uint8_t>/3` | 4.067 | 4.457
`classic_search<std::uint8_t>/4` | 1.433 | 1.463
`classic_search<std::uint8_t>/5` | 2.333 | 2.397
`classic_search<std::uint16_t>/0` | 3.837 | 4.737
`classic_search<std::uint16_t>/1` | 4.694 | 5.051
`classic_search<std::uint16_t>/2` | 3.516 | 4.718
`classic_search<std::uint16_t>/3` | 2.986 | 3.333
`classic_search<std::uint16_t>/4` | 1.462 | 1.75
`classic_search<std::uint16_t>/5` | 1.33 | 1.486
`classic_search<std::uint32_t>/0` | 2.647 | 2.867
`classic_search<std::uint32_t>/1` | 2.77 | 3.077
`classic_search<std::uint32_t>/2` | 2.8 | 2.75
`classic_search<std::uint32_t>/3` | 2.381 | 2.33
`classic_search<std::uint32_t>/4` | 0.486 | 0.537
`classic_search<std::uint32_t>/5` | 2.059 | 1.605
`classic_search<std::uint64_t>/0` | 1.099 | 1.194
`classic_search<std::uint64_t>/1` | 1.133 | 1.223
`classic_search<std::uint64_t>/2` | 1.256 | 0.959
`classic_search<std::uint64_t>/3` | 1.604 | 1.462
`classic_search<std::uint64_t>/4` | 0.913 | 0.52
`classic_search<std::uint64_t>/5` | 1.231 | 0.882
[...]

PR #6142, opened 2026-03-05
Vector algorithms: trim runtime coverage for ARM64EC non-vectorized fallbacks

AlexGuteniev:
Follow up on #6107.

We can't have complexity in test matrices, especially for internal test runs, so let's trim the excess coverage otherwise.

I know it still adds some complexity in another place, so it is still controversial.

Trim random coverage, which is time-consuming, for the algorithms that are vectorized everywhere and don't have fallbacks.
After #6141, #6139, and #6143 this excludes all string algorithms. The following are left: `bitset`, `replace`, and `minmax_element` for 64-bit integers and pointers.

Deliberately avoided checking for `_VECTORIZED_*` macros. We still need `_USE_STD_VECTOR_ALGORITHMS=0` coverage intact, and we need the machinery to be tested independently.

### DIFF
--- a/tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
+++ b/tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
@@ -1441,18 +1441,6 @@ void test_vector_algorithms(mt19937_64& gen) {
     test_swap_arrays<uint64_t, 512>(gen);
 }
 
-void test_vector_algorithms_fallbacks(mt19937_64& gen) {
-    test_min_max_element<long long>(gen);
-    test_min_max_element<unsigned long long>(gen);
-
-    test_min_max_element_pointers(gen);
-
-    test_replace<int>(gen);
-    test_replace<unsigned int>(gen);
-    test_replace<long long>(gen);
-    test_replace<unsigned long long>(gen);
-}
-
 template <typename Container1, typename Container2>
 void test_two_containers() {
     Container1 one                  = {10, 20, 30, 40, 50};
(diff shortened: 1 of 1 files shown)


tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -1439,6 +1446,7 @@ void test_vector_algorithms(mt19937_64& gen) {
     test_swap_arrays<uint16_t, 512>(gen);
     test_swap_arrays<uint32_t, 512>(gen);
     test_swap_arrays<uint64_t, 512>(gen);
+#endif // ^^^ !defined(_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC) ^^^
```
AlexGuteniev: Another approach could be having all affected in a separate function.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -1441,6 +1441,18 @@ void test_vector_algorithms(mt19937_64& gen) {
     test_swap_arrays<uint64_t, 512>(gen);
 }
 
+void test_vector_algorithms_fallbacks(mt19937_64& gen) {
```
StephanTLavavej: This is used only for `_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC`. But instead of guarding it, it's small enough that we ought to just inline it.

tests/std/tests/VSO_0000000_vector_algorithms/test.cpp
```diff
@@ -1955,9 +1967,14 @@ int main() {
...
     run_randomized_tests_with_different_isa_levels([](mt19937_64& gen) {
+#ifndef _CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC
         test_vector_algorithms(gen);
         test_various_containers();
         test_bitset(gen);
         test_string(gen);
+#else // ^ !defined(_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC) / defined(_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC) v
+        test_vector_algorithms_fallbacks(gen);
+        test_bitset(gen);
+#endif // ^^^ !defined(_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC) ^^^
```
StephanTLavavej: This `#endif` comment is incorrect. With the verbosity of the `#else` comment, we should probably make an exception to our usual new rule of "just repeat the condition" and be descriptive instead.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for keeping our test harness running efficiently! :stopwatch: :white_check_mark: :heart_eyes_cat:

PR #6145, opened 2026-03-06
`<xutility>`: Improve the all-bits-zero optimization in the `fill()` family

StephanTLavavej:
We've implemented a couple of library optimizations in the `fill()` family. For byte-sized objects, we can just call `memset()`, which is expected to be implemented in ultrafast assembly and faster than anything we can write in the Core Language. For other scalars, we can detect all-bits-zero values, and call `memset()` with zero bytes. (Zero is sufficiently common that this is a valuable check, whereas detecting all-bits-one or `0xABABABAB` would be counterproductive.)

Due to the existence of floating-point negative zero (a value which compares equal to zero, but is very much *not* an all-bits-zero value), we implemented this optimization with a `memcmp()` to preserve correctness. However, this had negative performance effects. The optimizer can't see through the `memcmp()`, so it was emitting substantial codegen even when the argument is a compile-time value that's been constant-propagated through.

We can improve this optimization in 3 ways:

* Detect integers and pointers, which don't have sneaky values, and directly compare them against `_Ty{}`. (Not `0` because even though that would compile, we despise `0` as a null pointer constant.)
* Now that we have downlevel support for C++20 `bit_cast` (available as `_Bit_cast`), we can send 1/2/4/8-byte types through it, to detect their bit patterns in an optimizer-friendly way. Note that this absorbs enums and floating-point types.
  + We could, of course, send integers and pointers down this path, but that would result in function calls in non-optimized debug mode. Better to spend 2 lines.
* Retain the `memcmp` codepath just in case (in practice it will only be activated for our unofficial quasi-support of 80-bit `long double`). Use `static constexpr` to avoid pushing bytes on the stack (usually we don't use `static` for non-arrays, but this `_Zero` is address-taken.)

This has 3 very nice effects for performance:

* When the argument is a compile-time constant, constant-propagated down so the optimizer can see it,
  + If it's all-bits-zero, the check is completely avoided and we directly call `memset()` with `0`.
  + If it isn't all-bits-zero, the check is completely avoided and we implement `fill()` normally.
* When the argument isn't a compile-time constant, we just compare values instead of setting up a `memcmp` function call.

Because this is a classic, central algorithm, I've added extensive test coverage even though we already had fairly good coverage.

### DIFF
--- a/tests/std/tests/VSO_0180469_fill_family/test.cpp
+++ b/tests/std/tests/VSO_0180469_fill_family/test.cpp
@@ -12,6 +12,7 @@
 #include <cassert>
 #include <cmath>
 #include <cstddef>
+#include <cstdint>
 #include <cstdio>
 #include <cstdlib>
 #include <memory>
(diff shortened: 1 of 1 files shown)


stl/inc/xutility
```diff
@@ -5540,8 +5540,18 @@ _NODISCARD bool _Is_all_bits_zero(const _Ty& _Val) {
...
+        return _STD _Bit_cast<uint8_t>(_Val) == 0;
+    } else if constexpr (sizeof(_Ty) == 2) {
+        return _STD _Bit_cast<uint16_t>(_Val) == 0;
+    } else if constexpr (sizeof(_Ty) == 4) {
+        return _STD _Bit_cast<uint32_t>(_Val) == 0;
+    } else if constexpr (sizeof(_Ty) == 8) {
+        return _STD _Bit_cast<uint64_t>(_Val) == 0;
     } else {
-        constexpr _Ty _Zero{};
+        static constexpr _Ty _Zero{};
```
frederick-vs-ja, line 5554: Just some questions: What's `_Ty` in this case, maybe Intel 80-bit `long double`? Does `static` make the execution faster for such `_Ty` which is still a scalar type?

AlexGuteniev: Yes, 3rd bullet in the PR description.

StephanTLavavej:

PR #6153, opened 2026-03-11
Add Neon implementation of `bitset_to_string`

hazzlim:
This PR adds a Neon implementation of `bitset_to_string`.

There's a slightly ugly workaround for DevCom-11056873 which hopefully isn't deemed to egregious.

## Benchmark results ⏲️ 
  | MSVC | Clang
-- | -- | --
`BM_bitset_to_string<15, char>` |	1 | 0.996
`BM_bitset_to_string<64, char>` |	2.516 | 1.294
`BM_bitset_to_string_large_single<512, char>` |	7.873 | 7.064
`BM_bitset_to_string_large_single<2048, char>` |	11.442 | 9.977
`BM_bitset_to_string<7, wchar_t>` |	1 | 1.008
`BM_bitset_to_string<64, wchar_t>` |	2.304 | 1.242
`BM_bitset_to_string_large_single<512, wchar_t>` |	5.8 | 4.293
`BM_bitset_to_string_large_single<2048, wchar_t>` |	7.778 | 5.714

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -11085,7 +11085,7 @@ namespace {
                 return vdupq_n_u8(_Val);
             }
 
-            static _Vec_t _Load_constant() {
+            static _Vec_t _Load_constant() noexcept {
                 // We do not omit static here, despite DevCom-11055227, because codegen is worse - see DevCom-11056805.
                 static constexpr uint32_t _Idx_arr[4] = {0x01010101, 0x01010101, 0x00000000, 0x00000000};
                 const auto _Idx                       = vld1q_u8(reinterpret_cast<const uint8_t*>(_Idx_arr));
@@ -11116,7 +11116,7 @@ namespace {
                 return vdupq_n_u16(_Val);
             }
 
-            static _Vec_t _Load_constant() {
+            static _Vec_t _Load_constant() noexcept {
                 // We do not omit static here, despite DevCom-11055227, because codegen is worse - see DevCom-11056805.
                 static constexpr uint64_t _Wx_arr[2] = {0x0010002000400080, 0x0001000200040008};
                 const auto _Wx                       = vld1q_u64(_Wx_arr);
@@ -11142,7 +11142,7 @@ namespace {
                 _mm256_zeroupper();
             }
 
-            static int _Load_constant() {
+            static int _Load_constant() noexcept {
                 return 0;
             }
         };
@@ -11154,7 +11154,7 @@ namespace {
 
             static void _Exit_vectorized() noexcept {}
 
-            static int _Load_constant() {
+            static int _Load_constant() noexcept {
                 return 0;
             }
         };


stl/src/vector_algorithms.cpp
```diff
@@ -11069,15 +11069,70 @@ __declspec(noalias) bool __stdcall __std_includes_less_8u(
...
+            static void _Out(void* const _Dest, const _Vec_t _Elems) noexcept {
+                vst1q_u8(static_cast<uint8_t*>(_Dest), _Elems);
+            }
+
+            static _Vec_t _Set(const uint8_t _Val) noexcept {
+                return vdupq_n_u8(_Val);
+            }
+
+            static _Vec_t _Load_constant() {
+                // We do not omit static here, despite DevCom-11055227, because codegen is worse - see DevCom-11056805.
```
hazzlim, line 11089: I actually wonder if this doesn't belong in a comment now (and below), but I thought I would explain why I didn't avoid static here (and below).

AlexGuteniev: I'm 💯 sure these things are good to comment, and it is not too verbose to comment every occurrence like you did.

stl/src/vector_algorithms.cpp
```diff
@@ -11069,15 +11069,70 @@ __declspec(noalias) bool __stdcall __std_includes_less_8u(
...
+                static constexpr uint32_t _Idx_arr[4] = {0x01010101, 0x01010101, 0x00000000, 0x00000000};
+                const auto _Idx                       = vld1q_u8(reinterpret_cast<const uint8_t*>(_Idx_arr));
+                return _Idx;
+            }
+
+            static _Vec_t __forceinline _Step(
+                const uint16_t _Val, const _Vec_t _Px0, const _Vec_t _Px1, const _Vec_t _Idx) noexcept {
+                const auto _Vx0 = vdupq_n_u16(_Val);
+                const auto _Vx1 = vqtbl1q_u8(vreinterpretq_u8_u16(_Vx0), _Idx);
```
AlexGuteniev, line 11098: But is this the most efficient approach?
if `vqtbl` is not very efficient, there may be other approach.
Making vector of halves which are broadcasted `_Val & 0xFF` and `_Val >> 8` bytes may work.

stl/src/vector_algorithms.cpp
```diff
@@ -11069,15 +11069,70 @@ __declspec(noalias) bool __stdcall __std_includes_less_8u(
...
+            static void _Out(void* const _Dest, const _Vec_t _Elems) noexcept {
+                vst1q_u16(static_cast<uint16_t*>(_Dest), _Elems);
+            }
+
+            static _Vec_t _Set(const uint16_t _Val) noexcept {
+                return vdupq_n_u16(_Val);
+            }
+
+            static _Vec_t _Load_constant() {
+                // We do not omit static here, despite DevCom-11055227, because codegen is worse - see DevCom-11056805.
```
AlexGuteniev, line 11120: I especially dislike that constant has different meaning in `_Traits_1_neon` and `_Traits_2_neon`.
But different argument naming should be sufficient to draw attention to this (worked for me).

stl/src/vector_algorithms.cpp
```diff
@@ -11069,15 +11069,70 @@ __declspec(noalias) bool __stdcall __std_includes_less_8u(
...
+                static constexpr uint32_t _Idx_arr[4] = {0x01010101, 0x01010101, 0x00000000, 0x00000000};
+                const auto _Idx                       = vld1q_u8(reinterpret_cast<const uint8_t*>(_Idx_arr));
+                return _Idx;
+            }
+
+            static _Vec_t __forceinline _Step(
+                const uint16_t _Val, const _Vec_t _Px0, const _Vec_t _Px1, const _Vec_t _Idx) noexcept {
+                const auto _Vx0 = vdupq_n_u16(_Val);
+                const auto _Vx1 = vqtbl1q_u8(vreinterpretq_u8_u16(_Vx0), _Idx);
+                const auto _Msk = vandq_u8(_Vx1, vreinterpretq_u8_u64(vdupq_n_u64(0x0102040810204080)));
```
AlexGuteniev, line 11099: Is compiler capable of moving out of the loop this constant, but not the other constant? 👀

AlexGuteniev:
> `BM_bitset_to_string<64, wchar_t>`	2.304

This is small. 

Is this threshold large enough then?
https://github.com/microsoft/STL/blob/26b7ca58ee3151c81736088e554e5797cafca641/stl/inc/bitset#L554

stl/src/vector_algorithms.cpp
hazzlim: TBL is performant on Neoverse cores, e.g. on Neoverse N2 it has 2c latency and throughput of 2 (it executes on all vector pipelines). The TBL approach (once we've loaded the indices) uses one DUP, and one TBL instruction. Whereas non-TBL would require two DUPs and one additional instruction to combine the two vectors.

It can be wise to avoid TBL when optimizing for AArch64 little core CPUs, but I don't think we care much about these here.

stl/src/vector_algorithms.cpp
hazzlim: Yes it's not ideal... but hopefully the argument naming softens the blow

stl/src/vector_algorithms.cpp
hazzlim: Yes (for some reason) - you can see in [DevCom-11056873](https://developercommunity.visualstudio.com/t/MSVC-fails-to-hoist-constants-out-of-loo/11056873) that v20 is not loaded inside the loop:
```
|$LL7@std_bitset|
        ldrh        w8,[x1],#2
        ldr         q17,[x10]
        dup         v16.8h,w8
        tbl         v16.16b,{v16.16b},v17.16b
        and         v16.16b,v16.16b,v20.16b
        cmeq        v16.16b,v16.16b,v21.16b
        bsl         v16.16b,v19.16b,v18.16b
        str         q16,[x9,#-0x10]!
        cmp         x9,x11
        bne         |$LL7@std_bitset|
```
It's almost as though the compiler is willing to hoist only one Neon constant per loop (and no more!)

hazzlim:
> > `BM_bitset_to_string<64, wchar_t>`	2.304
> 
> This is small.
> 
> Is this threshold large enough then?
> 
> https://github.com/microsoft/STL/blob/26b7ca58ee3151c81736088e554e5797cafca641/stl/inc/bitset#L554

I added a couple more sizes and it seems reasonable to keep threshold at 32:

  | MSVC | Clang
-- | -- | --
`BM_bitset_to_string<32, char>`	| 1.83	| 0.98
`BM_bitset_to_string<48, char>`	| 2.14	| 1.22
`BM_bitset_to_string<32, wchar_t>`	| 1.64	| 0.93
`BM_bitset_to_string<48, wchar_t>`	| 2.01	| 1.17

stl/src/vector_algorithms.cpp
```diff
@@ -11069,15 +11069,70 @@ __declspec(noalias) bool __stdcall __std_includes_less_8u(
...
+
+            static void _Out(void* const _Dest, const _Vec_t _Elems) noexcept {
+                vst1q_u8(static_cast<uint8_t*>(_Dest), _Elems);
+            }
+
+            static _Vec_t _Set(const uint8_t _Val) noexcept {
+                return vdupq_n_u8(_Val);
+            }
+
+            static _Vec_t _Load_constant() {
```
StephanTLavavej: Missing `noexcept`.

StephanTLavavej:

StephanTLavavej:
:zero: :one: :mechanical_arm: :six: :four:
===

Issue #6155, opened 2026-03-12
`P1004R2_constexpr_vector`: Report MSVC `constexpr` error affecting `std::construct_at` array support


StephanTLavavej:
Followup to #5110 and #5920.

After fixing DevCom-10798069 "Placement `operator new[]` in `<vcruntime.h>` should be marked `_MSVC_CONSTEXPR`", I'm removing our workarounds, including but not limited to:

https://github.com/microsoft/STL/blob/26b7ca58ee3151c81736088e554e5797cafca641/stl/inc/xutility#L610-L636

However, a couple of the tests in `P1004R2_constexpr_vector` are still failing. Somewhat reduced:

```
C:\Temp>type meow.cpp
```
```cpp
#include <cassert>
#include <vector>
using namespace std;

struct Udt {
    int x{1729};
};

constexpr bool test_vector_of_array_impl() {
    vector<Udt[3]> vec(5);

    for (const auto& arr : vec) {
        for (const auto& elem : arr) {
            assert(elem.x == 1729);
        }
    }

    return true;
}

int main() {
    static_assert(test_vector_of_array_impl());
}
```
```
C:\Temp>cl /EHsc /nologo /W4 /std:c++latest /MT /c meow.cpp
meow.cpp
meow.cpp(22): error C2131: expression did not evaluate to a constant
meow.cpp(14): note: failure was caused by accessing a data member of pointer past the end of object
meow.cpp(22): note: the call stack of the evaluation (the oldest call first) is
meow.cpp(22): note: while evaluating function 'bool test_vector_of_array_impl(void)'
```

I'm guessing this is an MSVC compiler bug that needs to be reported, possibly after being further reduced. (These are typically easier to report to the compiler devs after merging any STL product code changes, such as the workaround removals here.) I'm filing this as a tracking issue.

frederick-vs-ja:
I think I've roughly located the bug. Currently, MSVC doesn't correctly compute some properties of the pointer obtained by `constexpr` allocation. There's a surprising workaround (`p = p + 0` or `p = &p[0]`).

<details><summary>Reduced example (<a href="https://godbolt.org/z/48YcEc5hn">Godbolt link</a>)</summary>
<p>

```C++
using size_t = decltype(sizeof(0));

[[nodiscard]] [[msvc::constexpr]] inline void* operator new(size_t, void* p) noexcept {
    return p;
}

[...]

PR #6157, opened 2026-03-13
Toolset update: MSVC Compiler 19.51.36122, CUDA 13.2

StephanTLavavej:
Changelog
===
- Improved test coverage:
  * Added ARM64 ASan test coverage.
- Code cleanups:
  * Updated dependencies.
    + Updated MSVC Compiler to 19.51.36122 (now required).
    + Updated CMake to 4.2.3 (now required).
    + Updated CUDA to 13.2 (now required).

Commits
===
* Cycle locations.
* CUDA 13.2 (now required).
* CMake 4.2.3 (now required).
* MSVC Compiler 19.51.36122 (now required).
  + Internal toolset is still 19.50.
* New pools.
* cxx23-feature.md: Delete.
  + Users have occasionally misused this, it's outdated for C++26, and it isn't a huge time-saver anyways.
* README.md, CMakeLists.txt: Improve guidance for the Preview build tools.
  + If the compiler or WinSDK version checks fail, we should point to the README's centralized explanation of the VS Installer and `vcvarsall.bat` options.
* bug-report.md: Recommend the Latest or Preview version.
* pull_request_template.md: Add AI guidance.
* Azure Pipelines: Consolidate Setup TMP Directory into Clean After Checkout.
  + Deleting and recreating tmpDir is essentially cleaning, so we don't need to adjust the displayName. The checkout doesn't use tmpDir, so we can do this in any order. This will avoid displaying an extremely boring step in Azure Pipelines.
* All: Avoid warnings about comparing unsigned integers with zero.
  + MSVC warning C4296: '<': expression is always false
  + CUDA warning 186-D: pointless comparison of unsigned integer with zero
* All: Remove workaround for CWG-1129, implemented aeons ago.
* CUDA: Perma-workaround for designated aggregate initialization.
  + This was introduced by #5662 on 2025-08-08. Just initialize the return value with `{}` to avoid garbage. Use the name `_Tm` (instead of the previous `_Time`) for consistency with other returns. Avoids host compiler error C2760: syntax error: '}' was unexpected here; expected 'expression'
* CUDA: Remove workaround for Down With Typename.
  + I'm adding test coverage for this.
* CUDA: Remove workaround for CWG-2518 static_assert(false).
  + No test coverage necessary; the whole issue was that static_assert(false) was previously an instant error.
* CUDA: Remove workaround for intrinsics.
  + I'm adding test coverage for this.
* CUDA: Remove workaround for nocheck type traits.
  + I'm adding test coverage for this.
* CUDA: Remove workaround for static call operators.
  + See followup mechanical replacement. I'm adding test coverage for this.
* CUDA: Remove workaround for parenthesized aggregate initialization.
  + I'm adding test coverage for this.
* MSVC: Remove workaround: VSO-1285783 was fixed by MSVC-PR-694629 on 2025-12-12.
  + Including the 4th and final workaround, which appears to have been fixed by later changes.
* MSVC: Remove workaround: VSO-2714784 was fixed by MSVC-PR-707352 on 2026-02-10.
* MSVC: Remove workaround: VSO-2714798 was fixed by MSVC-PR-708374 on 2026-02-17.
* MSVC: Remove workaround: DevCom-10798069 was fixed by MSVC-PR-704666 on 2026-02-02.
  + P1004R2_constexpr_vector still needs 2 workarounds, filed #6155 for followup.
* MSVC: Remove workaround: DevCom-10416247 was fixed before 2024-10-15.
* MSVC: Remove workaround: VSO-570323 was a `/Za` bug, which we no longer test or care about.
* EDG: Remove workaround: VSO-892694 was fixed by VS-PR-527543 on 2024-02-09.
* /analyze: CppCoreCheck.dll was removed by MSVC-PR-674029 on 2025-10-29.
* Tests: GH_000639_nvcc_include_all: Add coverage for removed CUDA workarounds.
* Tests: P1682R3_to_underlying: Remove SFINAE workaround, use concepts.
* Tests: VSO_0000000_instantiate_type_traits: Use is_permissive.hpp instead of handwritten logic.
  + Drop the "TRANSITION, Evil Extension" comment because all usage of is_permissive is inherently TRANSITION. De Morgan the condition to clarify its behavior.
* Tests: Dev11_1074023_constexpr: Test the constexpr mutex constructor.
* Tests: Enable ASan annotation tests for ARM64, except Clang.
  + Clang 23 fixed llvm/llvm-project#184902.
* Comments: Clarify the CUDA restrict workaround.
  + Cite #5061 and #5097 so we don't attempt to remove this without proper care. Use "workaround / no workaround" for clarity.
* Comments: CUDA still needs the `basic_format_arg()` workaround.
* Comments: Update citation from WG21-N4928 + WG21-P2655R3 to WG21-N5032.
  + The paragraph number changed, but the content was otherwise identical.
* Comments: Consistently say "TRANSITION, ABI".
* Comments: Drop TRANSITION comment in experimental code.
  + We're never going to mess with this again.
* Comments: Update citation from DevCom-10265237 to VSO-2826400.
  + And VSO-2826400 was fixed by MSVC-PR-717544 on 2026-03-13.
* Mechanical: `_STATIC_CALL_OPERATOR` => `static`, `_CONST_CALL_OPERATOR` => nothing
* Improve the `_Signed_integer_like` fallback.
  + The concept short-circuits, but the fallback doesn't. Because the fallback `_Integer_like` is just `_Is_nonbool_integral`, we can use `is_signed_v`, which is preconditionless.
* Simplify `_HAS_POPCNT_INTRINSICS` condition.

STL-ASan-CI [passed](https://dev.azure.com/vclibs/STL/_build/results?buildId=20475&view=results).

### DIFF
--- a/stl/inc/__msvc_bit_utils.hpp
+++ b/stl/inc/__msvc_bit_utils.hpp
@@ -267,7 +267,7 @@ _NODISCARD int _Checked_x86_x64_countr_zero(const _Ty _Val) noexcept {
 
 #endif // _HAS_TZCNT_BSF_INTRINSICS
 
-#if (defined(_M_IX86) || defined(_M_X64) || defined(_M_ARM64)) && !defined(_M_CEE_PURE)
+#if !defined(_M_CEE_PURE)
 #define _HAS_POPCNT_INTRINSICS 1
 #if defined(__AVX__) || defined(_M_ARM64) || defined(_M_ARM64EC)
 #define _POPCNT_INTRINSICS_ALWAYS_AVAILABLE 1


stl/inc/__msvc_bit_utils.hpp
```diff
@@ -267,7 +267,7 @@ _NODISCARD int _Checked_x86_x64_countr_zero(const _Ty _Val) noexcept {
 
 #endif // _HAS_TZCNT_BSF_INTRINSICS
 
-#if (defined(_M_IX86) || defined(_M_X64) || defined(_M_ARM64)) && !defined(_M_CEE_PURE) && !defined(__CUDACC__)
+#if (defined(_M_IX86) || defined(_M_X64) || defined(_M_ARM64)) && !defined(_M_CEE_PURE)
```
AlexGuteniev: This can be further simplified to `#ifndef _M_CEE_PURE` I assume.
Though this would be less future proof if RISC-V, web assembly, or whatever else emerges.

CMakeLists.txt
```diff
@@ -1,13 +1,13 @@
...
-cmake_minimum_required(VERSION 4.2.1)
+cmake_minimum_required(VERSION 4.2.3)
 
 set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
 project(msvc_standard_libraries LANGUAGES CXX)
 
-if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "19.51.36014")
-    message(FATAL_ERROR "The STL must be built with MSVC Compiler 19.51.36014 or later.")
+if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "19.51.36122")
+    message(FATAL_ERROR "The STL must be built with MSVC Compiler 19.51.36122 or later. Follow the README instructions.")
```
AlexGuteniev, line 10: This is way too close to the edge, will not endure version having one more digit, though I admit this would be a very unlikely event.

StephanTLavavej: CMakeLists.txt files are exempt from the 120 column limitation.

AlexGuteniev, line 10: Since there's no limitation on line length, it could be README.md pedantically.

StephanTLavavej: It could, but the extension is not necessary for clarity. The tab on https://github.com/microsoft/STL just describes it as README. There's no potential for confusion here.

stl/inc/__msvc_bit_utils.hpp
StephanTLavavej: Agreed, thanks. This became possible after ARM32 removal.

I'm going to keep this condition around `hardware_constructive_interference_size` and `hardware_destructive_interference_size` in `<new>`, since those `64` constants won't immediately break if we port the codebase to isolinear chips in the 24th century.

PR #2526, opened 2022-02-07
P0627R6: Function to mark unreachable code

AlexGuteniev:
Resolves #2531

PR #6159, opened 2026-03-14
`<utility>`: make `unreachable()` call the doom function in debug mode

### LINKED ISSUE #6151: `<utility>`: `unreachable`  does not abort in debug builds
# Describe the bug
Calling `std::unreachable` is not diagnosed in debug builds.

# Expected behavior
In debug builds, I would expect calling `std::unreachable` to abort execution.

# STL version
e2115badf22f04439d8dd359a1a0202bee292146

# Additional context
This was discovered after `std::terminate` failed to capture undefined behavior due to a logic error. Aborting execution is compatible with undefined behavior, and it helps catch bugs. In that regard, the old ` assert(false)`-based implementations are preferable to `std::unreachable()`.

The original implementation (#2526) attempted to call `abort`  in debug builds:

https://github.com/microsoft/STL/blob/b6b977885809f185189b4e3eb01a30967eba7a30/stl/inc/utility#L757

This was later removed in (#5560), with no obvious explanation for the rationale.

https://github.com/microsoft/STL/commit/e2115badf22f04439d8dd359a1a0202bee292146#diff-b40b803e74da64226183307c996e495cce0bf23a9c83d9822fea849bb99b7144L962

AlexGuteniev:
Fix #6151.

In debug mode, `unreachable()` will deterministically call the doom function. In release mode, users get the UB they asked for.

Extract doom function to a new core header to achieve that. As @frederick-vs-ja suggested, core headers do call some functions. In this case, a function will be called by default for `/clr` configurations only.

Extended test coverage to make sure the machinery does not break compilation.
No runtime coverage -- it is still UB to call `unreachable()` according to the Standard.

### DIFF
--- a/tests/std/tests/GH_001411_core_headers/test.cpp
+++ b/tests/std/tests/GH_001411_core_headers/test.cpp
@@ -8,6 +8,7 @@
 
 // Also test GH-3103 "<xatomic.h>: Investigate making this a core header" and other internal core headers
 #include <__msvc_int128.hpp>
+#include <__msvc_minmax.hpp>
 #include <__msvc_system_error_abi.hpp>
 #include <__msvc_threads_core.hpp>
 #include <__msvc_xlocinfo_types.hpp>
@@ -25,6 +26,7 @@
 #endif // ^^^ _HAS_CXX20 ^^^
 
 // <__msvc_bit_utils.hpp> is included by <bit> and <limits>
+// <__msvc_doom_core.hpp> is included by <utility>
 // <__msvc_iter_core.hpp> is included by <tuple>
 // <xkeycheck.h> should not be included outside of <yvals_core.h>
 // <xtr1common> is included by <cstddef>


stl/inc/header-units.json
```diff
@@ -8,6 +8,7 @@
         "__msvc_bit_utils.hpp",
         "__msvc_chrono.hpp",
         "__msvc_cxx_stdatomic.hpp",
+        // "__msvc_doom_core.hpp", // internal header, provides macro that handles fatal conditions
```
StephanTLavavej: This doesn't need to be commented out; it isn't providing a control macro that affects how other headers are dragged in.

stl/inc/utility
```diff
@@ -1014,6 +1015,9 @@ _NODISCARD constexpr underlying_type_t<_Ty> to_underlying(_Ty _Value) noexcept {
 
 _EXPORT_STD [[noreturn]] __forceinline void unreachable() noexcept /* strengthened */ {
     _STL_UNREACHABLE;
+#ifdef _DEBUG
+    _MSVC_STL_DOOM_FUNCTION("unreachable called");
+#endif // defined(_DEBUG)
```
StephanTLavavej: The unreachable macro should occur after calling the doom function.

stl/inc/utility
```diff
@@ -1014,6 +1015,9 @@ _NODISCARD constexpr underlying_type_t<_Ty> to_underlying(_Ty _Value) noexcept {
 
 _EXPORT_STD [[noreturn]] __forceinline void unreachable() noexcept /* strengthened */ {
     _STL_UNREACHABLE;
+#ifdef _DEBUG
+    _MSVC_STL_DOOM_FUNCTION("unreachable called");
```
StephanTLavavej: Because it's an ordinary word, the message is potentially ambiguous. I recommend saying "std::unreachable() called" here.

tests/std/tests/P0627R6_unreachable/test.cpp
```diff
@@ -18,13 +17,14 @@ constexpr char test_impl(const int arg) {
...
-    assert(test_impl(1) == 'a');
-    assert(test_impl(2) == 'z');
-    return true;
+    return test_impl(1) == 'a' && test_impl(2) == 'z';
 }
 
 int main() {
-    test();
+    if (!test()) {
+        return 1729;
```
StephanTLavavej: 1 is a safer value than 1729 IMO (avoids modulo-256 behavior which I vaguely recall may be a thing), and more conventional.

tests/std/tests/P0627R6_unreachable/test.cpp
```diff
@@ -18,13 +17,14 @@ constexpr char test_impl(const int arg) {
...
+    }
     static_assert(test());
     static_assert(noexcept(std::unreachable())); // strengthened
+    return 0;
```
StephanTLavavej: This is unnecessary.

stl/inc/utility
AlexGuteniev: I like this idea, but remember that it contradicts to what has been decided previously and captured in the comments: https://github.com/microsoft/STL/pull/2526#discussion_r801980346

stl/inc/header-units.json
AlexGuteniev: I observe it was in `yvals.h` which was and is commented out,

stl/inc/utility
StephanTLavavej: Let's please capture it in a comment, then. Not even my memory is that perfect.

stl/inc/header-units.json
StephanTLavavej: Yes, specifically for the reason "// internal header, provides macros that control header inclusion". It's the latter part that's critical. Tons of internal headers are otherwise listed.

tests/std/tests/P0627R6_unreachable/test.cpp
AlexGuteniev: Yeah, let's support previous century shells

stl/inc/utility
AlexGuteniev: First please clarify how do you want it -- as it was agreed upon previously, and as you suggested it and I want it to be (I pushed a change already before seeing your comment).

AlexGuteniev: If we decide on doom before UB, we can also have death runtime coverage. Though it will be complicated to test `_DEBUG` specific behavior.

fortsnek9348:
Yes, this just tripped me up. I used `std::unreachable` *assuming* the debug build will check it. But finding out that it doesn't, I may just have to never ever use the function. No debug checks at all is too dangerous for me.

stl/inc/utility
StephanTLavavej: After thinking about this again, I'm happy with doom before unreachable (i.e. reversing the previous decision in #2526). My understanding is that people call `std::unreachable()` because they want efficient codegen in release mode, and it's clear there's demand for deterministic doom in debug mode. Nobody actually specifically wants undefined behavior as an end in itself.

I don't think we need runtime coverage.

tests/std/tests/GH_001411_core_headers/test.cpp
```diff
@@ -25,6 +26,7 @@
 #endif // ^^^ _HAS_CXX20 ^^^
 
 // <__msvc_bit_utils.hpp> is included by <bit> and <limits>
+// <__msvc_doom_core.hpp> is included by <utility>
```
StephanTLavavej, line 29: I updated this test to record the new inclusion.

tests/std/tests/GH_001411_core_headers/test.cpp
```diff
@@ -8,6 +8,7 @@
 
 // Also test GH-3103 "<xatomic.h>: Investigate making this a core header" and other internal core headers
 #include <__msvc_int128.hpp>
+#include <__msvc_minmax.hpp>
```
StephanTLavavej, line 11: Drive-by: Fixed pre-existing omission.

StephanTLavavej:

StephanTLavavej:
:supervillain_man: :skull_and_crossbones: :beetle:
===

K-ballo:
@AlexGuteniev thanks again for your work! 🎉

PR #6160, opened 2026-03-14
Consistently comment `_WIN64` macro usage

AlexGuteniev:


tests/std/tests/VSO_0000000_fancy_pointers/test.compile.pass.cpp
```diff
@@ -104,7 +104,7 @@ class fancy_pointer {
         rep += rhs;
         return *this;
     }
-#endif // _WIN64
+#endif // ^^^ 64 bit ^^^
```
StephanTLavavej: Should hyphenate.

stl/inc/atomic
StephanTLavavej, line 1: There are a few more inconsistent occurrences in `<atomic>`.

AlexGuteniev: I intentionally did not change them, keeping this change within existing `_WIN64`.

I think we also have remaining _M_IX86 that are good to be `_WIN64` here:
https://github.com/microsoft/STL/blob/2626cf1ee9d26be701b9bdbd2fb30db240456456/stl/inc/atomic#L1045
https://github.com/microsoft/STL/blob/2626cf1ee9d26be701b9bdbd2fb30db240456456/stl/inc/atomic#L1484

And here:

https://github.com/microsoft/STL/blob/2626cf1ee9d26be701b9bdbd2fb30db240456456/stl/inc/__msvc_bit_utils.hpp#L71
https://github.com/microsoft/STL/blob/2626cf1ee9d26be701b9bdbd2fb30db240456456/stl/inc/__msvc_bit_utils.hpp#L95
https://github.com/microsoft/STL/blob/2626cf1ee9d26be701b9bdbd2fb30db240456456/stl/inc/__msvc_bit_utils.hpp#L205
https://github.com/microsoft/STL/blob/2626cf1ee9d26be701b9bdbd2fb30db240456456/stl/inc/__msvc_bit_utils.hpp#L236
https://github.com/microsoft/STL/blob/2626cf1ee9d26be701b9bdbd2fb30db240456456/stl/inc/__msvc_bit_utils.hpp#L291

StephanTLavavej: I'd be open to a followup PR.

StephanTLavavej:

StephanTLavavej:
:six: :four: :scroll: :three: :two:
===

PR #6161, opened 2026-03-15
Use `memcpy` in `uninitialized_meow`

### LINKED ISSUE #6029: `<algorithm>`: std::uninitialized_copy used memmove instead of memcpy
# Describe the bug
std::uninitialized_copy uses memmove instead of memcpy. This is a codegen issue. memcpy is faster as it is an intrinsic.

# Command-line test case
```
C:\Temp>type repro.cpp
#include <iostream>
struct thirtytwo {
    char bytes[32];
};

void f(thirtytwo& target, const thirtytwo& source) {
    std::uninitialized_copy(std::begin(source.bytes), std::end(source.bytes), std::begin(target.bytes));
}
```

# Expected behavior
memcpy

# Additional context
Discussed in MSVC STL discord server in #stl

AlexGuteniev:
Resolves #6029

`memcpy` is the same as `memmove` when it is usual library call.

But intrinsic optimizations are possible for  known size for `memcpy` but not for `memmove`.
These optimizations happen when size is known at compile time (compile time constant or propagated constant), and this size is small enough (the boundary is at about kilobyte order of magnitude).

Uninitialized algorithms require non-overlapped range: See [uninitialized.copy]/1, and the others. It would have been strange to have overlap between initialized memory and uninitialized memory.

There hasn't been `_Copy_memcpy` / `_Copy_memcpy_n`, so I've added them. They might be useful in other places, besides uninitialized algorithms, as some other algorithms have useful non-overlap guarantee, but for now lets focus on uninitialized algorithms.

Benchmark results show overall improvement, although for some larger amount of data that is 800 bytes it flips. Probably the compiler should have its internal threshold to always use the library `memcpy` set to a lower value. For very large amounts where the compiler does emit `memcpy` the results before and after the change do match.

Benchmark                                                 |   Before  |  After   |  Speedup
----------------------------------------------------------|-----------|----------|------------
bm_uninitialized_copy<1, uint8_t, highly_aligned>         |  1.18 ns  | 1.16 ns  |   1.02
bm_uninitialized_copy<5, uint8_t, highly_aligned>         |  1.41 ns  | 1.17 ns  |   1.21
bm_uninitialized_copy<15, uint8_t, highly_aligned>        |  1.63 ns  | 1.42 ns  |   1.15
bm_uninitialized_copy<26, uint8_t, highly_aligned>        |  1.41 ns  | 1.19 ns  |   1.18
bm_uninitialized_copy<32, uint8_t, highly_aligned>        |  1.41 ns  | 1.18 ns  |   1.19
bm_uninitialized_copy<38, uint8_t, highly_aligned>        |  2.32 ns  | 1.40 ns  |   1.66
bm_uninitialized_copy<60, uint8_t, highly_aligned>        |  2.39 ns  | 1.43 ns  |   1.67
bm_uninitialized_copy<64, uint8_t, highly_aligned>        |  2.33 ns  | 1.41 ns  |   1.65
bm_uninitialized_copy<125, uint8_t, highly_aligned>       |  2.33 ns  | 1.91 ns  |   1.22
bm_uninitialized_copy<800, uint8_t, highly_aligned>       |  5.68 ns  | 7.00 ns  |   0.81
bm_uninitialized_copy<3000, uint8_t, highly_aligned>      |  15.0 ns  | 15.1 ns  |   0.99
bm_uninitialized_copy<9000, uint8_t, highly_aligned>      |  41.7 ns  | 41.9 ns  |   1.00
bm_uninitialized_copy<1, uint8_t, not_highly_aligned>     |  1.30 ns  | 1.18 ns  |   1.10
bm_uninitialized_copy<5, uint8_t, not_highly_aligned>     |  1.41 ns  | 1.18 ns  |   1.19
bm_uninitialized_copy<15, uint8_t, not_highly_aligned>    |  1.66 ns  | 1.40 ns  |   1.19
bm_uninitialized_copy<26, uint8_t, not_highly_aligned>    |  1.40 ns  | 1.16 ns  |   1.21
bm_uninitialized_copy<32, uint8_t, not_highly_aligned>    |  1.39 ns  | 1.17 ns  |   1.19
bm_uninitialized_copy<38, uint8_t, not_highly_aligned>    |  2.31 ns  | 1.42 ns  |   1.63
bm_uninitialized_copy<60, uint8_t, not_highly_aligned>    |  2.35 ns  | 1.88 ns  |   1.25
bm_uninitialized_copy<64, uint8_t, not_highly_aligned>    |  2.32 ns  | 1.87 ns  |   1.24
bm_uninitialized_copy<125, uint8_t, not_highly_aligned>   |  2.82 ns  | 2.85 ns  |   0.99
bm_uninitialized_copy<800, uint8_t, not_highly_aligned>   |  5.57 ns  | 12.4 ns  |   0.45
bm_uninitialized_copy<3000, uint8_t, not_highly_aligned>  |  15.0 ns  | 15.3 ns  |   0.98
bm_uninitialized_copy<9000, uint8_t, not_highly_aligned>  |  42.4 ns  | 42.6 ns  |   1.00

### DIFF
--- a/benchmarks/src/uninitialized_copy.cpp
+++ b/benchmarks/src/uninitialized_copy.cpp
@@ -1,11 +1,11 @@
 // Copyright (c) Microsoft Corporation.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-#include <algorithm>
 #include <benchmark/benchmark.h>
 #include <cstddef>
 #include <cstdint>
-#include <vector>
+#include <cstring>
+#include <memory>
 
 #include "skewed_allocator.hpp"
 


benchmarks/CMakeLists.txt
```diff
@@ -141,6 +141,7 @@ add_benchmark(std_copy src/std_copy.cpp)
 add_benchmark(sv_equal src/sv_equal.cpp)
 add_benchmark(swap_ranges src/swap_ranges.cpp)
 add_benchmark(unique src/unique.cpp)
+add_benchmark(uninitialized_copy src/uninitialized_copy.cpp)
```
StephanTLavavej: Should be sorted.

stl/inc/xutility
```diff
@@ -4981,6 +4981,47 @@ _CONSTEXPR20 void _Verify_ranges_do_not_overlap(const _Iter1& _First1, const _Se
...
 }
 
+template <class _OutCtgIt>
+_OutCtgIt _Copy_memcpy_tail(
+    const char* const _First_ch, const _OutCtgIt _Dest, const size_t _Byte_count, const size_t _Object_count) {
+    _STL_INTERNAL_CHECK(_Byte_count == _Object_count * sizeof(*_Dest));
+    // (pre-verified contiguous iterator)
+    const auto _Dest_ptr = _STD _To_address(_Dest);
+    const auto _Dest_ch  = const_cast<char*>(reinterpret_cast<const volatile char*>(_Dest_ptr));
+    _CSTD memcpy(_Dest_ch, _First_ch, _Byte_count);
```
StephanTLavavej: Suggestion: This is pretty verbose and the code is *exactly* identical except for the `memcpy` vs. `memmove` call, so unifying their implementations would be worth it.

benchmarks/src/uninitialized_copy.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include "skewed_allocator.hpp"
+
+using namespace std;
+
+template <size_t N, class T, template <class> class Padder>
+void bm_uninitialized_copy(benchmark::State& state) {
+    Padder<T[N]> padded_a;
+    auto a = &padded_a.value[0];
+    memset(a, 'a', sizeof(T) * N);
```
StephanTLavavej, line 18: Need `<cstring>` for `memset`.

benchmarks/src/uninitialized_copy.cpp
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <algorithm>
+#include <benchmark/benchmark.h>
+#include <cstddef>
+#include <cstdint>
+#include <vector>
```
StephanTLavavej: `<vector>` is unused.

benchmarks/src/uninitialized_copy.cpp
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <algorithm>
```
StephanTLavavej: Need `<memory>`, not `<algorithm>`, for `uninitialized_copy`.

StephanTLavavej:

StephanTLavavej:
:smile_cat: :cat2: :black_cat:
===

PR #6165, opened 2026-03-17
Skip a libcxx test for upcoming compiler changes

StephanTLavavej:
This is a reverse mirror of @Codiferous's MSVC-PR-720700 implementing WG21-P2564R3 "`consteval` Needs To Propagate Up" (aka Immediate Escalation). This increases compiler memory consumption, which is a problem for the accursed x86-hosted toolset.


StephanTLavavej:
And just as I created this, I realized we should skip it for :0 and :1 only (i.e. MSVC, not Clang). Not worth resetting the expensive MSVC-internal checks, I can adjust that in a followup.

AlexGuteniev:
We currently do native AArm64 compilation and cross compilation. Can do the same for x86?
The memory skips tend to accumulate.

StephanTLavavej:
We could easily configure GitHub to use the x64-hosted x86-targeting cross-compiler, but the MSVC-internal test harness is not so easily malleable.

StephanTLavavej:
Resolved adjacent-add merge conflict the same way that Cody did (with 3 newlines).

PR #6168, opened 2026-03-19
`<xlocnum>`: Fix `num_put::do_put()` to avoid `frexp()`'s unspecified behavior for inf/nan

StephanTLavavej:
# Background
This bug was reported to me by @Codiferous, reversing the natural order of things. (I'm supposed to report compiler bugs to him; he's not supposed to report library bugs to me! :joy_cat:) He found this while doing brain surgery to replace the UCRT's math functions with LLVM libc's math functions.

# CRT `frexp` behavior
[C23 WG14-N3220][C23] 7.12.6.7 "The `frexp` functions"/3:
> If `value` is not a floating-point number or if the integral power is outside the range of `int`, the results are unspecified.

## Microsoft UCRT
When `frexp()` is called with inf/nan, our behavior is that we always set the integer exponent to `-1`. (This is internal `minkernel/crts/ucrt/src/appcrt/tran/frexp.c` assigning `INT_NAN` which is defined in `minkernel/crts/ucrt/inc/trans.h`; we ship most UCRT sources but I don't believe we ship these ones, sorry.)

## LLVM libc
Their behavior is that they don't assign anything to the integer exponent by default, although they do support a mode to control this behavior: [`libc/src/__support/FPUtil/ManipulationFunctions.h`](https://github.com/llvm/llvm-project/blob/25f0d08bafa447b4d02a24bb7472ba187bc3b823/libc/src/__support/FPUtil/ManipulationFunctions.h#L31-L43)

# STL
## Bug
The problem was that `num_put::do_put()` was calling `frexp()` while calculating how much to increase its internal buffer size for fixed-precision formatting (which might need lots of characters for large positive or negative exponents). Unfortunately, this wasn't guarded for finite values only, so we were allowing inf/nan to go through this codepath when they didn't need it. This went unnoticed for decades because the Microsoft UCRT (and DevDiv's CRT before that, presumably) set the integer exponent to `-1`, which we then took the `abs()` of, with a final result of adding exactly `0` to `_Bufsize`.

However, because we were garbage-initing `int _Ptwo;`, if we replace `frexp()`'s implementation with LLVM libc's, we end up using that garbage integer while calculating how much to adjust `_Bufsize`, which is catastrophic.

## Fix
From previous fixes (#4212), we're already determining whether the value is finite (i.e. neither inf nor nan). Move that up, so we'll call `frexp()` to perform the buffer size adjustment for finite values only. inf/nan don't need this because immediately afterwards, we add 50 as a "fudge factor"; this isn't pretty but it is absolutely enough for our possible inf/nan outputs:

https://github.com/microsoft/STL/blob/2626cf1ee9d26be701b9bdbd2fb30db240456456/stl/inc/xlocnum#L1395

Additionally, I have a commit to avoid garbage-init for `frexp()`'s integer exponent. While this isn't needed for correctness (the previous commit is the correctness fix), zero is a strict subset of garbage, and this may help avoid analysis warnings in the future. Note that this zero-init isn't sufficient for correctness either, since there's nothing stopping `frexp()`'s unspecified behavior from being "assign a billion".

Finally, #4212 introduced comprehensive test coverage for #4210, which is where @Codiferous actually found this bug. It was almost perfect to leave as-is, but I noticed that we weren't expecting specific strings. They're implementation-defined ([C23 WG14-N3220][C23] 7.23.6.1 "The `fprintf` function"/8), so expecting these exact outputs will improve the strictness of the test.

[C23]: https://open-std.org/JTC1/SC22/WG14/www/docs/n3220.pdf

### DIFF
--- a/tests/std/tests/GH_003867_output_nan/test.cpp
+++ b/tests/std/tests/GH_003867_output_nan/test.cpp
@@ -58,7 +58,8 @@ void test_gh_3867() {
 // Also test GH-4210: With setprecision(0) showpoint fixed, a bogus '.' is emitted for infinity and NaN
 
 template <class FloatingPoint>
-void test_output_nonfinite_value(const FloatingPoint x) {
+void test_output_nonfinite_value(
+    const FloatingPoint x, const string& expected_noshowpos, const string& expected_showpos) {
     const auto s1 = [x] {
         ostringstream os;
         os << setprecision(0) << showpoint << fixed;
@@ -72,6 +73,7 @@ void test_output_nonfinite_value(const FloatingPoint x) {
         return os.str();
     }();
     assert(s1 == s2);
+    assert(s1 == expected_noshowpos);
 
     const auto s3 = [x] {
         ostringstream os;
@@ -86,6 +88,7 @@ void test_output_nonfinite_value(const FloatingPoint x) {
         return os.str();
     }();
     assert(s3 == s4);
+    assert(s3 == expected_showpos);
 }
 
 template <class FloatingPoint>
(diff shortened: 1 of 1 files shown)


StephanTLavavej:

PR #6177, opened 2026-03-23
Workaround for MSVC's `constexpr` array allocation bug for `vector`

### LINKED ISSUE #6155: `P1004R2_constexpr_vector`: Report MSVC `constexpr` error affecting `std::construct_at` array support

frederick-vs-ja:
This PR adds workaround for DevCom-11060638 needed for `vector`. Closes #6155.

Due the implementation strategy of MSVC, if a user directly uses `allocator<T[N]>::allocate`, the behavior in constant evaluation is still broken. There doesn't seem any possible workaround in the library implementation for this.


StephanTLavavej:

StephanTLavavej:
Thanks for reporting and working around this bug! :heart_eyes_cat: :detective: :beetle:

PR #6178, opened 2026-03-23
__msvc_cxx_stdatomic.hpp: Ensure C++ linkage

lb90:
Users can include the header in `extern "C"` blocks. This fixes error C2894:

```
__msvc_cxx_stdatomic.hpp(36): error C2894: templates cannot be declared to have 'C' linkage
```

### DIFF
--- a/stl/inc/__msvc_cxx_stdatomic.hpp
+++ b/stl/inc/__msvc_cxx_stdatomic.hpp
@@ -24,6 +24,8 @@
 _EMIT_STL_WARNING(STL4038, "The contents of <stdatomic.h> are available only with C++23 or later.");
 #else // ^^^ !_HAS_CXX23 / _HAS_CXX23 vvv
 
+_EXTERN_CXX_WORKAROUND
+
 #include <atomic>
 
 #pragma pack(push, _CRT_PACKING)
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
> Users can include the header in `extern "C"` blocks.

That's forbidden. WG21-N5032 \[using.headers\]/3:

> A translation unit shall include a header only outside of any declaration or definition

\[dcl.pre\]/1:

> Declarations generally specify how names are to be interpreted. Declarations have the form \[...\]
> * *declaration:*
>   + *name-declaration*
>   + \[...\]
> * *name-declaration:*
>   + \[...\]
>   + *linkage-specification*

\[dcl.link\]/2:

> * *linkage-specification:*
>   + `extern` *unevaluated-string* `{` *declaration-seq<sub>opt</sub>* `}`
>   + `extern` *unevaluated-string name-declaration*

Therefore, it is forbidden to write:

```cpp
extern "C" {
    #include <stdatomic.h>
}
```

lb90:
Hi @StephanTLavavej, thanks for the info! :)

Actualy I'm not a language expert, but I'd ask about this use-case: Suppose one has to consume a C library (`aclib.h`) from C++ code. Suppose that `aclib` is a 3rd party library and [can't be modified](https://isocpp.org/wiki/faq/mixing-c-and-cpp#include-c-hdrs-nonsystem). The C++ source can only do:

```c++
extern "C" {
#include <aclib.h>
}
```

Since `mylib.h` is a C header it can include `stdatomic.h` (a C header). That leads to error C2894 (`stdatomic.h` includes `__msvc_cxx_stdatomic.hpp` [in C++ mode](https://devblogs.microsoft.com/cppblog/c11-atomics-in-visual-studio-2022-version-17-5-preview-2/#:~:text=The%20Microsoft%20C%2B%2B%20standard%20library%20has%20implemented))

StephanTLavavej:
That's a bug in aclib.h. It's been known throughout the industry for 30+ years that C headers must be minimally aware of C++ and wrap their own contents, but not those of included headers, in `extern "C"`.

PR #6179, opened 2026-03-23
Fix CodeQL warnings after `_wsetlocale()` changes

StephanTLavavej:
Fixes internal VSO-2845667 and VSO-2845668.

CodeQL is emitting SM02986: cpp/incorrect-string-type-conversion, saying "Conversion from `char *` to `__wchar_t *`. Use of invalid string can lead to undefined behavior." for `<xlocinfo>` and "Conversion from `const char *` to `const __wchar_t *`. Use of invalid string can lead to undefined behavior." for `locale0.cpp`.

These warnings are desirable, we're just doing something that's intentionally sketchy-looking but correct. Accordingly, I'm adding suppression comments, citing #5781 where these changes were introduced.

CodeQL appears to warn only when `reinterpret_cast`ing from `char*` to `wchar_t*`, not the reverse (possibly because casts to `char*` are very common). Just in case, I am additionally adding the suppression to the `reinterpret_cast<char*>` where we actually store the wide string.

### DIFF
--- a/stl/inc/xlocinfo
+++ b/stl/inc/xlocinfo
@@ -188,6 +188,7 @@ public:
             _Tidy();
 
             if (_Right) { // new is not empty, copy it
+                // CodeQL [SM02986] We are intentionally storing `wchar_t*` as `char*` due to ABI, see GH-5781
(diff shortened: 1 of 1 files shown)


MahmoudGSaleh (approved):

StephanTLavavej:

PR #6182, opened 2026-03-25
Don't test ARM64EC vectorized bitset fallbacks

AlexGuteniev:
There's no need, once these are vectorized everywhere.

This does not save much of run time, unlike the previous trimming, but good to do anyway on principle.


StephanTLavavej:

StephanTLavavej:
Thanks for remembering to update this! :zero: :one: :brain:

PR #6183, opened 2026-03-25
Use `_WIN64` test where applicable

AlexGuteniev:
Follow up to #6160

Use `_WIN64` test instead of `_M_IX86` test where it looks consistent with previous usages:
 * In bit functions, where it tests for native 64-bit bit intrinsics availability
 * In atomics, where it tests for native 64-bit interlocked intrinsics availability

Keep `_M_IX86` when it means something else. 
There are places where it means calling conventions or name mangling, for example.


StephanTLavavej:

StephanTLavavej:
Thanks for making the codebase more consistent! :chart_with_upwards_trend: :heart_eyes_cat: :tada:

PR #6185, opened 2026-03-25
`<regex>`: Correctly set up `match_results` for match attempts on default-constructed regexes

### LINKED ISSUE #6181: `<regex>`: `match_results` are not filled correctly when `regex_match()` and `regex_search()` are called on a default-constructed `basic_regex`
# Describe the bug

N5032 [\[re.alg.match\]/2-3](https://eel.is/c++draft/re.alg.match#3) specify for `regex_match()`:

> *Effects*: Determines whether there is some sub-sequence within *[first, last)* that matches the regular expression `e`. [...] Returns `true` if such a sequence exists, `false` otherwise.  
> *Postconditions*: `m.ready() == true` in all cases. If the function returns `false`, then the effect on [`match_results`] parameter `m` is unspecified except that `m.size()` returns `0` and `m.empty()` returns `true`.

Furthermore, the default constructor of `basic_regex` satisfies the following postcondition [\[re.regex.construct\]/1](https://eel.is/c++draft/re.regex.construct#1):

> *Postconditions*: `*this` does not match any character sequence.

This means that `m.ready()` should be `true` and `m.size()` should be `0` for argument `m` when `regex_match()` is called on a default-constructed `basic_regex`, but the current implementation fails to adjust the `match_results` internals accordingly.

Similar behavior is specified for `regex_search()`.

# Test case
```cpp
#include <iostream>
#include <regex>
#include <string>

using namespace std;

int main() {

    const string input = "abc";
    {
        smatch sm;
        regex re;
        regex_match(input, sm, re);

        cout << "match_results ready: " << sm.ready() << '\n';
        cout << "match_results size: " << sm.size() << '\n';
    }

    {
        smatch sm;
        regex re;
        regex re2("abc");
        regex_match(input, sm, re2);

        regex_match(input, sm, re);

        cout << "match_results ready: " << sm.ready() << '\n';
        cout << "match_results size: " << sm.size() << '\n';
    }
}
```

This program currently prints:
```
match_results ready: 0
match_results size: 0
match_results ready: 1
match_results size: 1
```
https://godbolt.org/z/5j543r517 (which also shows that the same bug is present in libstdc++)

# Expected behavior
The program should print (as it does using libc++):
```
match_results ready: 1
match_results size: 0
match_results ready: 1
match_results size: 0
```

muellerj2:
Fixes #6181.

This PR moves the code to manipulate the `match_results` out of the `_Matcher3::_Match` function.

The code resetting the `match_results` becomes the first step in `_Regex_match1` and `_Regex_search2`, thus ensuring that they are set up correctly for the early return when these functions are called on default-constructed regexes.

The code filling the `match_results` is moved into a new function `_Matcher3::_Copy_captures`. This function is called by `_Regex_match1` and `_Regex_search2` now after finding a successful match.

The changes in this PR and #6164 mean that `_Matcher3` no longer depends on the (potentially wrapped) iterator type `_BidIt` except for the code in `_Matcher3::_Copy_captures`, so we can remove the type as a template parameter from the class and only template this specific member function on it. This also exposed one place in `_Matcher3::_Skip` where I accidentally failed to clean up some wrong `_BidIt` usage in #6164.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -2052,7 +2052,7 @@ private:
     _Compressed_pair<_Alloc, _Vector_state> _Mypair;
 };
 
-template <class _BidIt, class _Elem, class _RxTraits, class _It, class _Alloc>
+template <class _Elem, class _RxTraits, class _It, class _Alloc>
 class _Matcher3 { // provides ways to match a regular expression to a text sequence
 private:
     using _Loop_vals_type   = _Loop_vals_v3_t<_Iter_diff_t<_It>>;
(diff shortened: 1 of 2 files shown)


stl/inc/regex
```diff
@@ -4849,8 +4828,41 @@ bool _Matcher3<_BidIt, _Elem, _RxTraits, _It, _Alloc>::_Match_pat(_Node_base* _N
...
+            _Matches._At(_Idx).second  = _Result._Grps[_Idx - 1U]._End;
+        } else { // copy failed match
+            _Matches._At(_Idx).matched = false;
+            _Matches._At(_Idx).first   = _End;
+            _Matches._At(_Idx).second  = _End;
+        }
+    }
+    _Matches._Pfx().second = _Matches._At(0).first;
+
+    _Matches._Sfx().first   = _Matches._At(0).second;
```
StephanTLavavej, line 4856: Pre-existing, no change requested: I believe that `_Matches._At(0)` could be replaced with `_Submatch0`.

StephanTLavavej:

StephanTLavavej:
Thanks for noticing and fixing this bug, Regex Lord! :crown: :detective: :heart_eyes_cat:

PR #6187, opened 2026-03-28
`<filesystem>`: Fix `filesystem::equivalent()` (and `filesystem::copy()` with `overwrite_existing`) on FAT32 etc.

### LINKED ISSUE #6180: `<filesystem>`: `copy()` with `overwrite_existing` fails with `ERROR_INVALID_PARAMETER` on filesystems without `FILE_ID_128` support (FAT32, exFAT, SMB/non-NTFS) — regression from PR #5434
# Describe the bug
`std::filesystem::copy()` with `copy_options::overwrite_existing` fails with error 87 (`ERROR_INVALID_PARAMETER`) when the source file resides on a filesystem that does not support 128-bit file IDs, such as FAT32, exFAT, or SMB shares backed by a non-NTFS filesystem.

The failure only occurs when the destination file **already exists**. In that case, `fs::copy()` internally calls `equivalent()` to confirm that source and destination are not the same file. `equivalent()` queries file identity via `GetFileInformationByHandleEx(FileIdInfo)`, which returns `ERROR_INVALID_PARAMETER` on any filesystem that does not support `FILE_ID_128`. The error is propagated back through `copy()` to the caller.

This is a regression introduced by PR #5434 ("Win8 baseline: Simplify `<filesystem>` API calls"), merged April 2025, first shipped in v14.50.

# Command-line test case
```cpp
#include <filesystem>
#include <fstream>
#include <iostream>
#include <system_error>

int main(int argc, char *argv[]) {
  if (argc < 3) {
    std::cerr << "Usage: repro <source-on-mapped-drive> <local-dest>\n"
              << "Example: repro R:\\testfile %TEMP%\\testfile\n";
    return 1;
  }

  fs::path src = argv[1];
  fs::path dst = argv[2];

  std::cout << "_MSVC_STL_VERSION = " << _MSVC_STL_VERSION
            << ", _MSVC_STL_UPDATE = " << _MSVC_STL_UPDATE << "\n";

  { std::ofstream(dst).flush(); } // pre-create destination
  std::cout << "Source: " << src << "\n";
  std::cout << "Destination: " << dst << "\n";

  std::error_code ec;
  fs::copy(src, dst, fs::copy_options::overwrite_existing, ec);
  if (ec)
    std::cout << "[FAIL] fs::copy() error " << ec.value() << " (" << ec.message() << ")\n";
  else
    std::cout << "[OK]   fs::copy() succeeded\n";
}
```

Build with both toolsets from a VS 2026 installation:

```bat
cmd /c "call "C:\Program Files\Microsoft Visual Studio\18\Professional\VC\Auxiliary\Build\vcvars64.bat" -vcvars_ver=14.50 && cl /std:c++17 /EHsc repro.cpp /Fe:repro_v145.exe"
cmd /c "call "C:\Program Files\Microsoft Visual Studio\18\Professional\VC\Auxiliary\Build\vcvars64.bat" -vcvars_ver=14.44 && cl /std:c++17 /EHsc repro.cpp /Fe:repro_v143.exe"
```

Run both against a file on a FAT32 volume (see setup instructions below):

```
repro_v145.exe R:\testfile %TEMP%\testfile
repro_v143.exe R:\testfile %TEMP%\testfile
```

Output:

```
# v145 (regression):
_MSVC_STL_VERSION = 145, _MSVC_STL_UPDATE = 202508
Source: "R:\testfile"
Destination: "C:\Users\...\AppData\Local\Temp\testfile"
[FAIL] fs::copy() error 87 (The parameter is incorrect.)

# v143 (baseline, works correctly):
_MSVC_STL_VERSION = 143, _MSVC_STL_UPDATE = 202503
Source: "R:\testfile"
Destination: "C:\Users\...\AppData\Local\Temp\testfile"
[OK]   fs::copy() succeeded
```

# Expected behavior
`fs::copy(src, dst, fs::copy_options::overwrite_existing, ec)` should succeed (and `ec` should be clear) regardless of whether the source filesystem supports `FILE_ID_128`.

# STL version
Regression introduced in 14.50. Version 14.44 is not affected.

# Additional context
PR #5130 ("Fix false positives by `filesystem::equivalent` on file systems with transient file IDs") introduced `GetFileInformationByHandleEx(FileIdInfo)` to query `FILE_ID_128` in `_Get_file_id_by_handle`, but crucially kept a fallback: when `FileIdInfo` failed with `ERROR_INVALID_PARAMETER` or `ERROR_NOT_SUPPORTED`, the code would fall back to `GetFileInformationByHandle` (`BY_HANDLE_FILE_INFORMATION`), synthesizing a 128-bit ID from the 32-bit file index. This fallback made v143 work correctly on FAT32.

PR #5434 then removed this fallback as part of collapsing the `#ifdef _CRT_APP` / non-`_CRT_APP` bifurcation to a single Win8-API code path. The assumption was that Windows 8+ always supports `FileIdInfo`. However, `FILE_ID_128` support is a **filesystem** capability, not a Windows version guarantee — FAT32 and exFAT do not support it on any Windows version.

The removed code was in `stl/src/filesystem.cpp`, `_Get_file_id_by_handle`:

```cpp
// Before PR #5434 (v143, works):
switch (_Last_error) {
case __std_win_error::_Not_supported:
case __std_win_error::_Invalid_parameter:
    break; // try more things
default:
    return _Last_error;
}
// try GetFileInformationByHandle as a fallback
BY_HANDLE_FILE_INFORMATION _Info;
if (GetFileInformationByHandle(_Handle, &_Info)) {
    _Id->VolumeSerialNumber = _Info.dwVolumeSerialNumber;
    memcpy(&_Id->FileId.Identifier[0], &_Info.nFileIndexHigh, 4);
    memcpy(&_Id->FileId.Identifier[4], &_Info.nFileIndexLow, 4);
    memset(&_Id->FileId.Identifier[8], 0, 8);
    return __std_win_error::_Success;
}

// After PR #5434 (v145, broken):
return __std_win_error{GetLastError()};
```

## Test setup: create a FAT32 VHD mounted as R: (requires admin)

```powershell
# Run as Administrator
$vhdPath = "$PSScriptRoot\repro.vhd"
$disk = New-VHD -Path $vhdPath -SizeBytes 64MB -Fixed |
        Mount-VHD -Path $vhdPath -PassThru |
        Initialize-Disk -PartitionStyle MBR -PassThru
New-Partition -DiskNumber $disk.DiskNumber -UseMaximumSize -DriveLetter R
Format-Volume -DriveLetter R -FileSystem FAT32 -Confirm:$false
echo test > R:\testfile
```

StephanTLavavej:
Fixes #6180, a regression that was introduced by #5434 in the MSVC Build Tools 14.50.

# Commits
## (1) Partial revert

Partially revert #5434. This reverts only the alternatives to `GetFileInformationByHandleEx()` with `FileIdInfo`. All other changes, including the unconditional use of `GetFileInformationByHandleEx()` with `FileStandardInfo` to get `NumberOfLinks` in legacy `filesys.cpp` `_Hard_links()`, are being allowed to stand as correct. Note that that case matches what modern `filesystem.cpp` `__std_fs_get_stats()` has always done:

https://github.com/microsoft/STL/blob/bed5dd5b9e354c1fa3d287d291afd1d0356f9e1b/stl/src/filesys.cpp#L294-L296
https://github.com/microsoft/STL/blob/bed5dd5b9e354c1fa3d287d291afd1d0356f9e1b/stl/src/filesystem.cpp#L960-L969

## (2) Fix modern `filesystem.cpp`

`filesystem.cpp`: Always build the fallback, improve comment.

We don't need `_CRT_APP` guards anymore. @amyw-msft explained:

> All the banned API restrictions were lifted back in 2020, and all the devices/OS versions that would have had a runtime issue with using a desktop-only API have been out of support for a long time now. Shouldn't be any issue with this. \[...\] there are still APIs that are banned in the sense that you can't use the user32 UI stuff in a UWP sandbox. It's the deployment time blocks that historically prevented common Win32 API use in Store apps (and forced us to create the uwp variants) that are gone.

## (3) Fix legacy `filesys.cpp`

`filesys.cpp`: Add fallback behavior, improve comments.

Previously, the legacy implementation had **no** fallback behavior.

For `_CRT_APP`, 128-bit file IDs were the only codepath. That might have been okay, depending on what filesystems were available, but it might have been a mistake.

For non-App, 64-bit file indexes were the only codepath, so we weren't handling modern filesystems in an ideal way.

I'm bringing the logic closer to the modern implementation. It isn't an exact match (we don't look at the error codes, so if neither file exists, we try the fallback only to discover again that neither file exists). However, it should be good enough for the legacy implementation, and a strict improvement over the status quo. This is already a fairly major change, and I don't want it to become even more invasive.

# Manual testing
For obvious reasons, adding automated coverage for this would be far more difficult than it's worth. I've manually verified both the legacy and modern implementations. With the new comments, the risk of regression in the future should be low.

<details><summary>Click to expand before-and-after testing on NTFS and FAT32:</summary>

```
C:\Temp>type meow.cpp
```
```cpp
#include <exception>
#include <filesystem>
#include <fstream>
#include <print>
#include <string>
#include <system_error>

#include <experimental_filesystem.hpp>
using namespace std;

int main() {
    try {
        println("_MSVC_STL_VERSION: {}", _MSVC_STL_VERSION);
        println("_MSVC_STL_UPDATE:  {}", _MSVC_STL_UPDATE);
        println();

        const filesystem::path src{"DELME_src.txt"};
        const filesystem::path dst{"DELME_dst.txt"};

        for (int api = 0; api < 2; ++api) {
            filesystem::remove(src);
            filesystem::remove(dst);

            {
                ofstream f{src.string()};
                f << "Cats!\n";
            }
            {
                ofstream f{dst.string()};
                f << "Dogs!\n";
            }

            error_code ec{};

            if (api == 0) {
                println("Testing Standard <filesystem>.");
                filesystem::copy(src, dst, filesystem::copy_options::overwrite_existing, ec);
            } else {
                println("Testing legacy <experimental/filesystem>.");
                namespace exfs = experimental::filesystem;
                exfs::copy(src.string(), dst.string(), exfs::copy_options::overwrite_existing, ec);
            }

            if (ec) {
                println(R"(Copy failed! Error code: {}, "{}")", ec.value(), ec.message());
            } else {
                println("Copy succeeded!");
            }

            {
                print("After copy, {} contains: ", dst.string());
                ifstream f{dst.string()};
                for (string s; getline(f, s);) {
                    println("{}", s);
                }
                println();
            }

            filesystem::remove(src);
            filesystem::remove(dst);
        }
        println(R"(Exception: "{}")", e.what());
    }
}
```
```
C:\Temp>cl /EHsc /nologo /W4 /std:c++latest /MTd /Od /I D:\GitHub\STL\tests\std\include /D_SILENCE_EXPERIMENTAL_FILESYSTEM_DEPRECATION_WARNING meow.cpp
meow.cpp
```

NTFS is unaffected:

```
C:\Temp>pwsh -Command "(Get-Volume -DriveLetter C).FileSystemType"
NTFS

C:\Temp>meow_broken.exe
_MSVC_STL_VERSION: 145
_MSVC_STL_UPDATE:  202601

Testing Standard <filesystem>.
Copy succeeded!
After copy, DELME_dst.txt contains: Cats!

Testing legacy <experimental/filesystem>.
Copy succeeded!
After copy, DELME_dst.txt contains: Cats!


C:\Temp>meow_fixed.exe
_MSVC_STL_VERSION: 145
_MSVC_STL_UPDATE:  202603

Testing Standard <filesystem>.
Copy succeeded!
After copy, DELME_dst.txt contains: Cats!

Testing legacy <experimental/filesystem>.
Copy succeeded!
After copy, DELME_dst.txt contains: Cats!
```

FAT32 is fixed:

```
E:\>pwsh -Command "(Get-Volume -DriveLetter E).FileSystemType"
FAT32

E:\>C:\Temp\meow_broken.exe
_MSVC_STL_VERSION: 145
_MSVC_STL_UPDATE:  202601

Testing Standard <filesystem>.
Copy failed! Error code: 87, "The parameter is incorrect."
After copy, DELME_dst.txt contains: Dogs!

Testing legacy <experimental/filesystem>.
Copy failed! Error code: 1, "operation not permitted"
After copy, DELME_dst.txt contains: Dogs!


E:\>C:\Temp\meow_fixed.exe
_MSVC_STL_VERSION: 145
_MSVC_STL_UPDATE:  202603

Testing Standard <filesystem>.
Copy succeeded!
After copy, DELME_dst.txt contains: Cats!

Testing legacy <experimental/filesystem>.
Copy succeeded!
After copy, DELME_dst.txt contains: Cats!
```
</details>

### DIFF
--- a/stl/src/filesys.cpp
+++ b/stl/src/filesys.cpp
@@ -374,66 +374,65 @@ extern "C" _CRTIMP2_PURE space_info __CLRCALL_PURE_OR_CDECL _Statvfs(const wchar
 extern "C" _CRTIMP2_PURE int __CLRCALL_PURE_OR_CDECL _Equivalent(
     const wchar_t* _Fname1, const wchar_t* _Fname2) noexcept { // test for equivalent file names
     // See GH-3571: File IDs are only guaranteed to be unique and stable while handles remain open
-#ifdef _CRT_APP
-    _FILE_ID_INFO _Info1 = {0};
-    _FILE_ID_INFO _Info2 = {0};
-    bool _Ok1            = false;
-    bool _Ok2            = false;
-
     HANDLE _Handle1 = _FilesysOpenFile(_Fname1, FILE_READ_ATTRIBUTES, FILE_FLAG_BACKUP_SEMANTICS);
-    if (_Handle1 != INVALID_HANDLE_VALUE) { // get file1 info
-        _Ok1 = GetFileInformationByHandleEx(_Handle1, FileIdInfo, &_Info1, sizeof(_Info1)) != 0;
+    HANDLE _Handle2 = _FilesysOpenFile(_Fname2, FILE_READ_ATTRIBUTES, FILE_FLAG_BACKUP_SEMANTICS);
+
+    bool _Ok1   = false;
+    bool _Ok2   = false;
+    int _Result = -1; // negative indicates error
+
+    {
+        // If we can get FILE_ID_INFO, use that as the source of truth.
+        _FILE_ID_INFO _Info1 = {0};
+        _FILE_ID_INFO _Info2 = {0};
+
+        if (_Handle1 != INVALID_HANDLE_VALUE) {
+            _Ok1 = GetFileInformationByHandleEx(_Handle1, FileIdInfo, &_Info1, sizeof(_Info1)) != 0;
+        }
+
+        if (_Handle2 != INVALID_HANDLE_VALUE) {
+            _Ok2 = GetFileInformationByHandleEx(_Handle2, FileIdInfo, &_Info2, sizeof(_Info2)) != 0;
+        }
+
+        if (_Ok1 && _Ok2) { // test existing files for equivalence
+            _Result = memcmp(&_Info1, &_Info2, sizeof(_FILE_ID_INFO)) == 0 ? 1 : 0;
+        } else if (_Ok1 || _Ok2) { // one file exists, the other doesn't
+            _Result = 0;
+        }
     }
 
-    HANDLE _Handle2 = _FilesysOpenFile(_Fname2, FILE_READ_ATTRIBUTES, FILE_FLAG_BACKUP_SEMANTICS);
-    if (_Handle2 != INVALID_HANDLE_VALUE) { // get file2 info
-        _Ok2 = GetFileInformationByHandleEx(_Handle2, FileIdInfo, &_Info2, sizeof(_Info2)) != 0;
-        CloseHandle(_Handle2);
+    if (_Result < 0) {
+        // Some filesystems don't support FILE_ID_INFO's 128-bit file identifiers.
+        // Try GetFileInformationByHandle() as a fallback.
+        BY_HANDLE_FILE_INFORMATION _Info1 = {0};
+        BY_HANDLE_FILE_INFORMATION _Info2 = {0};
+
+        if (_Handle1 != INVALID_HANDLE_VALUE) {
+            _Ok1 = GetFileInformationByHandle(_Handle1, &_Info1) != 0;
+        }
+
+        if (_Handle2 != INVALID_HANDLE_VALUE) {
+            _Ok2 = GetFileInformationByHandle(_Handle2, &_Info2) != 0;
+        }
+
+        if (_Ok1 && _Ok2) { // test existing files for equivalence
+            _Result = static_cast<int>(_Info1.dwVolumeSerialNumber == _Info2.dwVolumeSerialNumber
+                                       && _Info1.nFileIndexHigh == _Info2.nFileIndexHigh
+                                       && _Info1.nFileIndexLow == _Info2.nFileIndexLow);
+        } else if (_Ok1 || _Ok2) { // one file exists, the other doesn't
+            _Result = 0;
+        }
     }
 
     if (_Handle1 != INVALID_HANDLE_VALUE) {
         CloseHandle(_Handle1);
     }
 
-    if (!_Ok1 && !_Ok2) {
-        return -1;
-    } else if (!_Ok1 || !_Ok2) {
-        return 0;
-    } else { // test existing files for equivalence
-        return memcmp(&_Info1, &_Info2, sizeof(_FILE_ID_INFO)) == 0 ? 1 : 0;
-    }
-#else // ^^^ defined(_CRT_APP) / !defined(_CRT_APP) vvv
-    BY_HANDLE_FILE_INFORMATION _Info1 = {0};
-    BY_HANDLE_FILE_INFORMATION _Info2 = {0};
-    bool _Ok1                         = false;
-    bool _Ok2                         = false;
-
-    HANDLE _Handle1 = _FilesysOpenFile(_Fname1, FILE_READ_ATTRIBUTES, FILE_FLAG_BACKUP_SEMANTICS);
-    if (_Handle1 != INVALID_HANDLE_VALUE) { // get file1 info
-        _Ok1 = GetFileInformationByHandle(_Handle1, &_Info1) != 0;
-    }
-
-    HANDLE _Handle2 = _FilesysOpenFile(_Fname2, FILE_READ_ATTRIBUTES, FILE_FLAG_BACKUP_SEMANTICS);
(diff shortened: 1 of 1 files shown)


StephanTLavavej:

PR #6188, opened 2026-03-28
`<__msvc_threads_core.hpp>`: Revert `alignas` changes that broke `mutex` and `condition_variable` for `/clr`

StephanTLavavej:
We merged #5811 which is about to ship in MSVC Build Tools 14.51, but it's causing major user heartburn: DevCom-11059932 VSO-2841023 "C++ toolset 14.51.36014 Preview alignment changes breaks C++/CLI code".

This cleanup would have been nice to have, but we need to revert it until the `/clr` `alignas` bugs VSO-1659383 and VSO-2616030 are fixed.

* Revert product code changes from #5811.
* Comment workarounds.


StephanTLavavej:

PR #6190, opened 2026-03-29
Remove many alternate codepaths for `_CRT_APP` and `_ONECORE`

StephanTLavavej:
As I mentioned in #6187, we don't need most of our `_CRT_APP` guards anymore. @amyw-msft explained:

> All the banned API restrictions were lifted back in 2020, and all the devices/OS versions that would have had a runtime issue with using a desktop-only API have been out of support for a long time now. Shouldn't be any issue with this. \[...\] there are still APIs that are banned in the sense that you can't use the user32 UI stuff in a UWP sandbox. It's the deployment time blocks that historically prevented common Win32 API use in Store apps (and forced us to create the uwp variants) that are gone.

I understand `_ONECORE` less, and I believe our codebase has been equally confused about the difference, but I believe that we can also remove most of those guards too. Since deployment time blocks are a thing of the past, I think we can rely on "if it builds, it's good". (My internal MSVC-PR-632910 merged on 2025-05-05 got our build using the WinSDK's `mincore.lib` for OneCore builds, instead of weird "mystery meat" libs, increasing my confidence that our builds are cromulent.)

Getting rid of these alternate codepaths increases maintainability (since we almost exclusively test desktop configurations). In a few cases they may even improve functionality (e.g. by preferentially using `GetTempPath2W`). I want to merge this at the beginning of the 14.52 cycle, so if we discover problems, we can easily revert part or all of these changes.

# Commits
* filesystem.cpp: Use `CreateSymbolicLinkW`.
* filesystem.cpp: Use `CreateHardLinkW`.
* filesystem.cpp: Use `CreateDirectoryExW`.
* filesys.cpp: Use `CreateSymbolicLinkW`.
* filesys.cpp: Use `CreateHardLinkW`.
* filesys.cpp: Use `_wgetcwd`.
* filesys.cpp: Use `_wchdir`.
* winapisupp.cpp: Use `CreateSymbolicLinkW`.
* taskscheduler.cpp: Use `GetModuleHandleExW`.
* \[OneCore\] filesystem.cpp: Use `AreFileApisANSI`.
* \[OneCore\] filesys.cpp: Use `AreFileApisANSI`.
* \[OneCore\] filesystem.cpp: Use `GetModuleHandleW` to load `GetTempPath2W`.
* \[OneCore\] winapisupp.cpp: Use `GetModuleHandleW` to load `GetTempPath2W`.

This has passed an internal build and `stlimport` test pass.

### DIFF
--- a/stl/src/winapisupp.cpp
+++ b/stl/src/winapisupp.cpp
@@ -8,7 +8,6 @@
 
 #include <Windows.h>
 
-#if !defined(_ONECORE)
 namespace {
 
 // Use this macro for defining the following function pointers
@@ -24,7 +23,6 @@ namespace {
 #define IFDYNAMICGETCACHEDFUNCTION(fn_name) if (const auto pf##fn_name = __KERNEL32Function_##fn_name)
 
 } // unnamed namespace
-#endif // ^^^ !defined(_ONECORE) ^^^
 
 #if !defined(_CRT_WINDOWS) && !defined(UNDOCKED_WINDOWS_UCRT)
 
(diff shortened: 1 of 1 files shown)


StephanTLavavej:

PR #6189, opened 2026-03-29
`<regex>`: Optimize skip heuristic for searches of patterns with initial dot wildcards

muellerj2:
Towards #5468.

This closes one of the remaining gaps in the skip heuristic. It doesn't really make sense to use the dot wildcard itself as the basis for skipping ahead, as it essentially matches everything (except for newlines in ECMAScript or NUL in POSIX grammars). So instead, this uses the following NFA node to compute the skip.

## Benchmark

Relevant changes highlighted.

benchmark | before [ns] | after [ns] | speedup
-- | -- | -- | --
bm_lorem_search/"^bibe"/2 | 50.2232 | 51.5625 | 0.97
bm_lorem_search/"^bibe"/3 | 50.2232 | 50.2232 | 1.00
bm_lorem_search/"^bibe"/4 | 50 | 50.8161 | 0.98
bm_lorem_search/"bibe"/2 | 2887.83 | 2845.99 | 1.01
bm_lorem_search/"bibe"/3 | 5625 | 5580.36 | 1.01
bm_lorem_search/"bibe"/4 | 11718.8 | 11474.6 | 1.02
bm_lorem_search/"bibe".collate/2 | 3013.39 | 2887.83 | 1.04
bm_lorem_search/"bibe".collate/3 | 5580.36 | 5580.36 | 1.00
bm_lorem_search/"bibe".collate/4 | 10742.2 | 11230.5 | 0.96
bm_lorem_search/"(bibe)"/2 | 3529.57 | 3529.57 | 1.00
bm_lorem_search/"(bibe)"/3 | 7114.96 | 6975.45 | 1.02
bm_lorem_search/"(bibe)"/4 | 13811.3 | 13671.9 | 1.01
bm_lorem_search/"(bibe)+"/2 | 4603.8 | 4652.62 | 0.99
bm_lorem_search/"(bibe)+"/3 | 8998.29 | 8998.29 | 1.00
bm_lorem_search/"(bibe)+"/4 | 17578.3 | 18415.3 | 0.95
bm_lorem_search/"(?:bibe)+"/2 | 4185.27 | 4010.88 | 1.04
bm_lorem_search/"(?:bibe)+"/3 | 7847.38 | 7672.99 | 1.02
bm_lorem_search/"(?:bibe)+"/4 | 15694.7 | 15380.8 | 1.02
bm_lorem_search/R"(\bbibe)"/2 | 64174.1 | 64174.1 | 1.00
[...]

PR #6191, opened 2026-03-29
`<regex>`: Optimize searches for patterns with initial branching

muellerj2:
Towards #5468.

This PR extends the skip heuristic to regexes with initial branching, i.e., disjunctions or loops that allow no repetition.

The skip heuristic included an optimization attempt for disjunctions in the past. But this attempt was based on a depth-first search of the NFA. #5452 showed that this could result in quadratic worst-case complexity for some regexes and inputs, which is unacceptable if the input would have been processed in linear time without this optimization attempt. It was therefore removed in #5457.

This quadratic slowdown can be avoided by searching the NFA breadth-first instead. But this search strategy is typically slower and hostile to vectorization.

This PR tries to strike a balance between depth-first and breadth-first search: The input string is split into search windows of a maximum length. We search for a potential skip position within a window and only move to the next window if we couldn't find a potential match. This means that the heuristic performs depth-first search within each search window (essentially resurrecting the code removed in #5457), but behaves like breadth-first search between search windows. This ensures that it only reads a constant number of characters in the input beyond the actually determined skip position, thus achieving a linear time complexity in the worst case, but also allows to take advantage of vectorization and the faster processing of the depth-first search within each search window.

The current choice of the window size is a bit arbitrary: It shouldn't be too low so that vectorization can still give some benefit, but it also shouldn't be too large to limit the amount of unnecessary work. I tried to strike a bit of a balance when I chose the value, but it isn't based on any measurements on real-world data.

The boundaries of the search windows require some special attention: The skip heuristic for some NFA nodes can search for whole character sequences, not just single characters. If a matching string were to sit exactly on a boundary between search windows, a naive implementation would not find it because the match is not contained in the first search window and the search in the following window would only start in the middle of the match. For this reason, the heuristic for `_N_str` nodes had to be adapted such that it can read beyond the search window by as much as necessary. (This doesn't break the runtime analysis because the amount of characters read beyond the window is determined by the regex and the analysis assumes that the regex is fixed.) No adjustment had to be made to the heuristic for `_N_class` nodes because it has never considered the value of the `_Last` iterator in the `_Skip` function.

I considered implementing all this using `std::distance()` and `std::advance()`, but instead opted for a new `_Advance_at_most()` function because the former implementation would have resulted in cubic worst-case complexity for bidirectional iterators.

## Benchmark
The adjusted benchmark code is included in #6189. Relevant changes are highlighted.

-- | -- | -- | --
bm_lorem_search/"^bibe"/2 | 50.2232 | 51.5625 | 0.97
bm_lorem_search/"^bibe"/3 | 50.2232 | 49.6689 | 1.01
bm_lorem_search/"^bibe"/4 | 50 | 50.2232 | 1.00
bm_lorem_search/"bibe"/2 | 2887.83 | 2964.57 | 0.97
bm_lorem_search/"bibe"/3 | 5625 | 5580.36 | 1.01
bm_lorem_search/"bibe"/4 | 11718.8 | 11997.8 | 0.98
bm_lorem_search/"bibe".collate/2 | 3013.39 | 2915.74 | 1.03
bm_lorem_search/"bibe".collate/3 | 5580.36 | 5719.87 | 0.98
bm_lorem_search/"bibe".collate/4 | 10742.2 | 10986.3 | 0.98
bm_lorem_search/"(bibe)"/2 | 3529.57 | 3930.66 | 0.90
bm_lorem_search/"(bibe)"/3 | 7114.96 | 7149.83 | 1.00
bm_lorem_search/"(bibe)"/4 | 13811.3 | 13811.5 | 1.00
bm_lorem_search/"(bibe)+"/2 | 4603.8 | 5156.25 | 0.89
bm_lorem_search/"(bibe)+"/3 | 8998.29 | 9835.34 | 0.91
bm_lorem_search/"(bibe)+"/4 | 17578.3 | 17996.8 | 0.98
bm_lorem_search/"(?:bibe)+"/2 | 4185.27 | 4289.91 | 0.98
bm_lorem_search/"(?:bibe)+"/3 | 7847.38 | 8370.5 | 0.94
bm_lorem_search/"(?:bibe)+"/4 | 15694.7 | 15485.6 | 1.01
bm_lorem_search/R"(\bbibe)"/2 | 64174.1 | 69754.5 | 0.92
bm_lorem_search/R"(\bbibe)"/3 | 131138 | 136021 | 0.96
bm_lorem_search/R"(\bbibe)"/4 | 256696 | 256696 | 1.00
bm_lorem_search/R"(\Bibe)"/2 | 144385 | 142299 | 1.01
bm_lorem_search/R"(\Bibe)"/3 | 288771 | 341797 | 0.84
bm_lorem_search/R"(\Bibe)"/4 | 610352 | 625000 | 0.98
bm_lorem_search/R"((?=....)bibe)"/2 | 3989.95 | 5022.33 | 0.79
bm_lorem_search/R"((?=....)bibe)"/3 | 8021.76 | 8196.15 | 0.98
bm_lorem_search/R"((?=....)bibe)"/4 | 15346 | 16043.5 | 0.96
bm_lorem_search/R"((?=bibe)....)"/2 | 3759.77 | 3934.14 | 0.96
bm_lorem_search/R"((?=bibe)....)"/3 | 7324.22 | 7324.22 | 1.00
bm_lorem_search/R"((?=bibe)....)"/4 | 13950.9 | 14125.2 | 0.99
bm_lorem_search/R"((?!lorem)bibe)"/2 | 3449.35 | 3683.04 | 0.94
bm_lorem_search/R"((?!lorem)bibe)"/3 | 6835.94 | 7149.83 | 0.96
bm_lorem_search/R"((?!lorem)bibe)"/4 | 13253.3 | 14439.1 | 0.92
**bm_lorem_search/"bibe\|soda"/2** | **429688** | **8893.69** | **48.31**
**bm_lorem_search/"bibe\|soda"/3** | **836680** | **18833.7** | **44.42**
**bm_lorem_search/"bibe\|soda"/4** | **1727580** | **34527.8** | **50.03**
**bm_lorem_search/"(id )?bibe"/2** | **486592** | **8370.54** | **58.13**
**bm_lorem_search/"(id )?bibe"/3** | **1004020** | **16741.2** | **59.97**
**bm_lorem_search/"(id )?bibe"/4** | **1968830** | **32889.7** | **59.86**
bm_lorem_search/".bibe"/2 | 190438 | 199507 | 0.95
bm_lorem_search/".bibe"/3 | 374930 | 359869 | 1.04
bm_lorem_search/".bibe"/4 | 784738 | 784738 | 1.00

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -5093,13 +5093,22 @@ _It _Matcher3<_Elem, _RxTraits, _It, _Alloc>::_Skip(
                 return _First;
             }
 
+        case _N_dot:
+            {
+                if (_Recursion_depth >= _Max_recursion_depth) {
+                    return _First;
+                }
+
+                _First = _Skip(++_First, _Last, _Nx->_Next, _Recursion_depth + 1U);
+                return --_First;
+            }
+
         case _N_begin:
         case _N_endif:
             break;
 
         case _N_end:
         case _N_none:
-        case _N_dot:
         case _N_back:
         case _N_end_rep:
         default:
--- a/tests/std/tests/VSO_0000000_regex_use/test.cpp
+++ b/tests/std/tests/VSO_0000000_regex_use/test.cpp
@@ -2509,6 +2509,21 @@ void test_gh_6181() {
     }
 }
 
+void test_gh_6189() {
+    // GH-6189: Optimize skip heuristic for searches of patterns with initial dot wildcards
+    test_regex re(&g_regexTester, ".abc");
+    re.should_search_match("dabc", "dabc");
+    re.should_search_match("dabcdddd", "dabc");
+    re.should_search_match("ddabc", "dabc");
+    re.should_search_match("ddabcdddd", "dabc");
+    re.should_search_match("ddddddddddddddddabcdddddddddddd", "dabc");
+    re.should_search_match("ddddddddddddddddddddddddabc", "dabc");
+    re.should_search_match("ddabcddd", "dabc");
+    re.should_search_fail("abcddddd");
+    re.should_search_fail("ddddddddddd\nabcdddddddddd");
+    re.should_search_fail("d");
+}
+
 void test_gh_6191() {
     // GH-6191: Optimize searches for patterns with initial branching
     // We must check that we handle matches near search window boundaries correctly.
@@ -2629,6 +2644,7 @@ int main() {
     test_gh_6118();
     test_gh_6147();
     test_gh_6181();
+    test_gh_6189();
     test_gh_6191();
 
     return g_regexTester.result();
--- a/benchmarks/src/regex_search.cpp
+++ b/benchmarks/src/regex_search.cpp
@@ -47,5 +47,8 @@ BENCHMARK_CAPTURE(bm_lorem_search, R"(\Bibe)", R"(\Bibe)")->Apply(common_args);
 BENCHMARK_CAPTURE(bm_lorem_search, R"((?=....)bibe)", R"((?=....)bibe)")->Apply(common_args);
 BENCHMARK_CAPTURE(bm_lorem_search, R"((?=bibe)....)", R"((?=bibe)....)")->Apply(common_args);
 BENCHMARK_CAPTURE(bm_lorem_search, R"((?!lorem)bibe)", R"((?!lorem)bibe)")->Apply(common_args);
+BENCHMARK_CAPTURE(bm_lorem_search, "(bibe|soda)", "(bibe|soda)")->Apply(common_args);
+BENCHMARK_CAPTURE(bm_lorem_search, "(id )?bibe", "(id )?bibe")->Apply(common_args);
+BENCHMARK_CAPTURE(bm_lorem_search, ".bibe", ".bibe")->Apply(common_args);
 
 BENCHMARK_MAIN();


stl/inc/regex
```diff
@@ -4952,27 +4982,83 @@ _It _Matcher3<_Elem, _RxTraits, _It, _Alloc>::_Skip(
...
+
+                // GH-5452: If this node has two or more branches,
+                // examining all alternatives in a disjunction
+                // until the end of the input string
+                // could result in quadratic worst-case complexity.
+                // For this reason, we split the input string
+                // into search windows with a constant maximum length.
+                // This ensures that this heuristic has linear time complexity,
+                // because this bounds the number of characters that were read
+                // beyond the finally determined skip position by a constant.
```
StephanTLavavej: Style nitpick: This is aggressively narrowly wrapped, consuming a lot of vertical space, which interrupts reading the code. We're flexible about comment wrapping but we usually use most of the 120-column limit.

stl/inc/regex
```diff
@@ -4869,7 +4886,8 @@ _It _Matcher3<_Elem, _RxTraits, _It, _Alloc>::_Skip(
     static constexpr char _Line_terminators_char[]       = {static_cast<char>(_Meta_cr), static_cast<char>(_Meta_nl)};
     static constexpr wchar_t _Line_terminators_wchar_t[] = {static_cast<wchar_t>(_Meta_cr),
         static_cast<wchar_t>(_Meta_nl), static_cast<wchar_t>(_Meta_ls), static_cast<wchar_t>(_Meta_ps)};
-    constexpr unsigned int _Max_recursion_depth          = 50U;
+    static constexpr unsigned int _Max_recursion_depth   = 50U;
+    static constexpr short _Max_lookahead                = 512;
```
StephanTLavavej: Scalars should be plain `constexpr`, not `static constexpr` (unless they're address-taken, which is highly unusual).

stl/inc/regex
```diff
@@ -4952,27 +4982,83 @@ _It _Matcher3<_Elem, _RxTraits, _It, _Alloc>::_Skip(
...
+                // into search windows with a constant maximum length.
+                // This ensures that this heuristic has linear time complexity,
+                // because this bounds the number of characters that were read
+                // beyond the finally determined skip position by a constant.
+
+                for (;;) {
+                    _It _Lookahead_last = _First;
+                    _STD _Advance_at_most(_Lookahead_last, _Last, _Max_lookahead);
+                    _First = _Skip(_First, _Skip(_First, _Lookahead_last, _Node->_Next, _Recursion_depth + 1U),
+                        _Node->_End_rep->_Next, _Recursion_depth + 1U);
```
StephanTLavavej: Style: These nested `_Skip` calls are doing way more work in a single expression than we generally like to do these days. It would be better to extract an intermediate statement.

tests/std/tests/GH_005204_regex_collating_ranges/test.cpp
```diff
@@ -656,11 +657,29 @@ void test_gh_5437() {
 #endif // !defined(SKIP_COLLATE_TESTS)
 }
 
+void test_gh_6191() {
+    // GH-6191: Optimize searches for patterns with initial branching
+    // Check that collating elements are handled correctly at search window boundaries.
+    const gh_994_regex re("[[.dzs.]]|abc");
+    smatch sm;
+    for (size_t count = 510; count < 516; ++count) {
```
StephanTLavavej: Ultra-nitpick: Should include `<cstddef>` for `size_t`.

StephanTLavavej: This includes `<algorithm>` so it's shadowing `count` rather directly.

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -2507,6 +2508,60 @@ void test_gh_6181() {
...
+    // test bidirectional iterators
+    {
+        const regex alt_re("abcdef|uvwxyz");
+        const string suffix = "abcdefhhhhhhhuvwxyz";
+        for (size_t count = 510; count < 516; ++count) {
+            list<char> input(count, 'h');
+            input.insert(input.end(), suffix.begin(), suffix.end());
+
+            match_results<list<char>::iterator> results;
+            assert(regex_search(input.begin(), input.end(), results, alt_re));
```
StephanTLavavej: We could use `const_iterator` here, for a little bit of safety/clarity.

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -2507,6 +2508,60 @@ void test_gh_6181() {
...
+        }
+    }
+
+    // test bidirectional iterators
+    {
+        const regex alt_re("abcdef|uvwxyz");
+        const string suffix = "abcdefhhhhhhhuvwxyz";
+        for (size_t count = 510; count < 516; ++count) {
+            list<char> input(count, 'h');
+            input.insert(input.end(), suffix.begin(), suffix.end());
```
StephanTLavavej: This is surprisingly expensive (~260K allocated nodes). But we're increasing the prefix length, which is exactly what `list` is good at! We can just reuse the `input`.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Resolved trivial adjacent-add conflicts with #6189 in `VSO_0000000_regex_use`.

StephanTLavavej:
:deciduous_tree: :detective: :rocket:
===

PR #3011, opened 2022-08-08
Avoid `<atomic>` in `filesystem.cpp`

StephanTLavavej:
# Background
@zacklj89 found this while updating the STL in Windows. (Thanks Zack, for investigating this for over a week - it wasn't easy! :smile_cat:)

#2302 (merged for VS 2022 17.4 Preview 2) made `filesystem.cpp` (which is injected into the STL's import lib) include `<atomic>` to cache a function pointer. Unfortunately, `<atomic>` is a non-core header, and it's the first non-core header included by `filesystem.cpp`:

https://github.com/microsoft/STL/blob/2263d931f48a4e196036b2523af7a0048bb96b46/stl/inc/atomic#L9

This drags in `_Init_locks`, which is separately compiled and has an implicitly defined assignment operator:

https://github.com/microsoft/STL/blob/2263d931f48a4e196036b2523af7a0048bb96b46/stl/inc/yvals.h#L456-L475

In a certain project, this was leading to linker errors, as the STL's DLL exports `_Init_locks::operator=` and the linker was seeing a duplicate definition in the injected `filesystem.obj`.

# GitHub vs. MSVC difference
Curiously, I'm unable to observe this assignment operator in the import lib produced by microsoft/STL `main`, but I can see it in the MSVC-internal build. I'm not sure this is due to a very recent compiler change, or some difference in the CMake vs. MSBuild build systems. In any event, dragging in this non-core header is risky, and we should avoid it due to the potential for similar linker errors when 17.4 ships.

## microsoft/STL `main`
```
D:\GitHub\STL\out\build\x64\out\lib\amd64>dumpbin /symbols msvcprt.lib | grep -P __std_fs_remove
172 00000000 SECT85 notype ()    External     | __std_fs_remove
258 00000000 SECTC3 notype       Static       | $unwind$__std_fs_remove
25B 00000000 SECTC4 notype       Static       | $pdata$__std_fs_remove

D:\GitHub\STL\out\build\x64\out\lib\amd64>dumpbin /symbols msvcprt.lib | grep -P _Init_locks

D:\GitHub\STL\out\build\x64\out\lib\amd64>
```

## MSVC-internal `prod/fe`
```
C:\Temp>dumpbin /symbols \\vcfs\builds\msvc\fe\20220808.01\binaries.amd64ret\lib\amd64\msvcprt.lib | grep -P __std_fs_remove
179 00000000 SECT88 notype ()    External     | __std_fs_remove
[...]

PR #6192, opened 2026-03-30
`vector_algorithms.cpp`: Add `[[nodiscard]]` to `_Use_avx2()` and `_Use_sse42()`

jovibor:
Adds `[[nodiscard]]` to `_Use_avx2()` and `_Use_sse42()` functions.

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -25,15 +25,11 @@ extern "C" long __isa_enabled;
 namespace {
 #if !defined(_M_ARM64) && !defined(_M_ARM64EC)
     [[nodiscard]] bool _Use_avx2() noexcept {
-        const static bool _HasAVX2 =
-            __check_arch_support(__IA_SUPPORT_VECTOR256, 0) || (__isa_enabled & (1 << __ISA_AVAILABLE_AVX2));
-        return _HasAVX2;
+        return __check_arch_support(__IA_SUPPORT_VECTOR256, 0) || (__isa_enabled & (1 << __ISA_AVAILABLE_AVX2));
     }
 
     [[nodiscard]] bool _Use_sse42() noexcept {
-        const static bool _HasSSE42 =
-            __check_arch_support(__IA_SUPPORT_SSE42, 0) || (__isa_enabled & (1 << __ISA_AVAILABLE_SSE42));
-        return _HasSSE42;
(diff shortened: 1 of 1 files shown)


stl/src/vector_algorithms.cpp
```diff
@@ -24,12 +24,16 @@ extern "C" long __isa_enabled;
 
 namespace {
 #if !defined(_M_ARM64) && !defined(_M_ARM64EC)
-    bool _Use_avx2() noexcept {
-        return __check_arch_support(__IA_SUPPORT_VECTOR256, 0) || (__isa_enabled & (1 << __ISA_AVAILABLE_AVX2));
+    [[nodiscard]] bool _Use_avx2() noexcept {
+        const static bool _HasAVX2 =
```
StephanTLavavej: This introduces magic statics, which we avoid (they have unusual interactions with certain code).

I would accept just the `[[nodiscard]]`s, though.

YexuanXiao:
The query results can be cached as `std::atomic<bool>`s with namespace scope and internal linkage. This pattern helps generate cleaner and more stable code than magic static.

YexuanXiao:
MSVC can optimize the function-scoped static atomic to be the same as one with internal linkage at namespace scope, without using the magic static mechanism, and magic static is worse, see: https://godbolt.org/z/r4zr9oE7s

StephanTLavavej:
No - `<atomic>` is a non-core header, which we can't drag into the import lib. See #3011.

Aside from that, the complexity of introducing atomics isn't worth it here. We're reading an integer, not doing anything outrageously expensive, and branch prediction is going to be really happy with a result that doesn't change.

jovibor:
Dragging-in atomics for such a tiny function would definitely be an overkill, imo.
I reverted changes back, leaving only `[[nodiscard]]`.

> @StephanTLavavej 
> This introduces magic statics, which we avoid (they have unusual interactions with certain code).

Can you give any hints, what exactly these unusual interactions is, or where to read about it?
As far as I know the "magic statics" are guaranteed to be thread-safe and one-time initialized since **C++11**.

StephanTLavavej:
As I recall, they create a TLS section in the resulting binary, which had weird interactions with an optimization that Windows wanted to do.

YexuanXiao:
I have re-showd my comments, which compare magic static and atomic. The conclusion is that if you want the best performance, then use lock-free atomics, or do not cache the result. Although I have only shown the results for MSVC, this actually applies to all compilers and operating systems.

StephanTLavavej:

StephanTLavavej:
Thanks for adding our favorite attribute! :heart_eyes_cat:

Issue #4433, opened 2024-03-02
`std::replace` vectorization and redundant writes


AlexGuteniev:
## Vectorization non-options

Vectorized `replace` algorithm would write to its source.

To preserve the existing elements with only replacing some in a single vector register, one of these approaches could be used:
1. Masked stores
2. Element-wise stores
3. Storing the old elements along with replaced elements

Before AVX512, masked stores are only for 32 and 64 sized elements and only for AVX2: `mm_maskstore_epi32`, `_mm256_maskstore_epi32`, `_mm_maskstore_epi64`, `_mm256_maskstore_epi64`. The SSE2 `_mm_maskmoveu_si128` is not suitable due to a non-temporal memory hint. So, only AVX2 and only large element size is not a full set.

Element-wise stores will defeat the purpose of vectorization, at least the most of it, since the scalar replacement is fast, and need to do several elements at once to outperform it.

## Redundant writes caveat

So having to resort to redundant writes. That is computing a vector with preserved and replaced elements, and writing it at once.

Redundant writes are not transparent, in a sense that one can pass read-only array with no elements to be replaced, and expect no crashes. With sophisticated `VirtualAlloc`/`VirtualProtect` use, only a part of array can be read only. This makes arbitrary redundant writes non-transparent. 

There's a sophisticated approach possible to make the redundant element writes transparent. We can align the pointer at the beginning, and then store only when we have something to replace. So we still have redundant writes, but within the same page we would normally write, so these are transparent.

Sure the sophisticated approach adds complexity to maintain.

## ✨  Compiler magic! ✨

If instead of this:
https://github.com/microsoft/STL/blob/8b081e26ba016970ce3338cb483ff10f2ade30a5/stl/inc/algorithm#L3707-L3711

We would use `*_UFirst =  *_UFirst == _Oldval ?_Newval : *_UFirst`, to enable these redundant writes, the compiler would auto-vectorize the algorithm almost perfectly. I created DevCom-10606365 to improve codegen, but the improvement effect is small. 

So we can just specialize the algorithm with `if constexpr` for proper types, if we accept these redundant stores.

## Are redundant stores indeed harmful?

I dunno.

From the standard perspective, I'm not a language lawyer.

From the beyond standard perspective, assuming standard permits redundant stores, I think these are fine: there's no realistic use case that would rely on read-only array modifying attempt being successful if no modification happens.

From performance perspective, they might be potentially slightly harmful, but the benefit from vectorization clearly outweighs this potential harm

[...]

Issue #814, opened 2020-05-08
Provide AVX-512 implementations of vector algorithms


StephanTLavavej:
Currently, `stl/src/vector_algorithms.cpp` provides optimized implementations for SSE2, SSE4.2, and AVX2; for example:

https://github.com/microsoft/STL/blob/11150ca6679507461ae9f64e90c678c498cfe908/stl/src/vector_algorithms.cpp#L39

Presumably, AVX-512 would further increase performance. 
`C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\VC\Tools\MSVC\14.26.28801\include\isa_availability.h` defines `__ISA_AVAILABLE_AVX512` which can be used to detect AVX-512 support.

Inspired by #813.

StephanTLavavej:
Closing, as in #2104 we were unable to find positive performance effects. We're still open to using AVX-512 if benchmarks can be found indicating significant performance improvements.

AlexGuteniev:
<details>
<summary>It is time!</summary>
<img width="1000" height="812" alt="Image" src="https://github.com/user-attachments/assets/dd1ea1e3-7fa6-4f27-810a-391670e183c7" />
</details>

It is time to reconsider.

#2104 may still be unsuccessful. Back then we only had swizzling algorithms that are very likely to be memory operations bound, and so double vector width would not help.

Now we have some other vector algorithms that would clearly benefit from AVX-512:
 * **replace**
   - AVX-512 can add 8-bit and 16-bit flavor. This will avoid oddness of vectorizing only for some integer sizes and make the vectorization coverage on par with possible future Arm64 SVE implementation.
   - AVX-512 can dramatically improve AMD performance. See #5068. The CPU mentioned by @muellerj2 is Zen 4, so it has AVX-512 that doesn't have masked store perf problem, but AVX2 masked stores are problematic, and resulted in almost no vectorization benefit.
   - This will make these algorithms on par with ARM64 SVE implementation.
      * We may simplify fallback testing then
      * And simplify away header macros complexity
 * **remove**, **unique**, **remove_copy**, **unique_copy**
   - `vpcompress` variants will allow not using additional memory for lookup tables, reducing memory operations number
   - Not having a lookup table will lift the elements number from 8 to full vector
   - `_copy` variant will not need to use intermediate buffer
   - This all again will make these algorithms on par with possible future ARM64 SVE implementation.
 * **find_meow_of**
[...]

PR #6195, opened 2026-03-31
Add SVE implementation of `replace`

hazzlim:
This PR adds an SVE implementation of `replace`. This algorithm was previously not vectorized using Neon, due to the absence of masked stores in the instruction set. See https://github.com/microsoft/STL/issues/4433 for why this is an issue.

### Benchmark results ⏲️ 
Results are speedup values relative to the existing C code as a baseline - higher is better. Benchmark results were obtained running on a Neoverse N2 machine.
  | MSVC Speedup | Clang Speedup
-- | -- | --
`r<std::uint8_t>`  | 17.03 | 7.024
`r<std::uint16_t>` | 10.17 | 3.767
`r<std::uint32_t>` | 4.592 | 2.109
`r<std::uint64_t>` | 2.475 | 1.23


stl/inc/algorithm
```diff
@@ -117,6 +117,13 @@ __declspec(noalias) bool __stdcall __std_includes_less_8u(
 #endif // ^^^ _VECTORIZED_INCLUDES ^^^
 
 #if _VECTORIZED_REPLACE
+#if defined(_M_ARM64) || defined(_M_ARM64EC)
```
AlexGuteniev: Let's repeat the current minmax elemen 64 bit integers scheme here instead of directly testing for arch

AlexGuteniev: A possible option could be also providing all versions for all archs, as #814  will have it. But it seems too early as we don't know when we start doing AVX-512

stl/src/vector_algorithms.cpp
```diff
@@ -9745,10 +9856,29 @@ namespace {
 
 extern "C" {
 
-#ifndef _M_ARM64
+#if defined(_M_ARM64) || defined(_M_ARM64EC)
+__declspec(noalias) void __stdcall __std_replace_1(
+    void* const _First, const void* const _Last, const uint8_t _Old_val, const uint8_t _New_val) noexcept {
```
StephanTLavavej: `const void* const _Last` doesn't match the declaration.

stl/src/vector_algorithms.cpp
```diff
@@ -9607,6 +9608,116 @@ __declspec(noalias) size_t __stdcall __std_mismatch_8(
 namespace {
     namespace _Replacing {
 #if defined(_M_ARM64) || defined(_M_ARM64EC)
+        struct _Traits_1_sve {
+            static svuint8_t _Load(svbool_t _Pred, const void* const _Ptr) noexcept {
```
StephanTLavavej: I believe `_Pred` can be const.

stl/src/vector_algorithms.cpp
```diff
@@ -9607,6 +9608,116 @@ __declspec(noalias) size_t __stdcall __std_mismatch_8(
...
+                const auto _True = svptrue_b8();
+                while (_First != _Stop_at) {
+                    const auto _Data = _Traits::_Load(_True, _First);
+                    const auto _Mask = _Traits::_Cmp(_True, _Data, _Comparand);
+                    _Traits::_Store(_Mask, _First, _Replacement);
+                    _Advance_bytes(_First, _Sve_vl);
+                }
+
+                if (const size_t _Tail_length = _Size_bytes & size_t{_Sve_vl - 1}; _Tail_length != 0) {
+                    auto _Tail_mask  = svwhilelt_b8(size_t{0}, _Tail_length);
```
StephanTLavavej: I believe `_Tail_mask` could be const.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Turns out I can't merge this until the MSVC-internal checked-in compiler is updated. The current 14.50 compiler can't understand the new `<arm_sve.h>` and ICEs quite horribly, and I don't see any possible workaround. (`vector_algorithms.cpp` has to be built by the checked-in compiler.) The good news is that we should only have to wait a month.

stl/src/vector_algorithms.cpp
```diff
@@ -9607,6 +9608,116 @@ __declspec(noalias) size_t __stdcall __std_mismatch_8(
...
+
+        template <class _Traits, class _Ty>
+        __declspec(noalias) void __stdcall _Replace_impl(
+            void* _First, void* const _Last, const _Ty _Old_val, const _Ty _New_val) noexcept {
+
+            if (_Use_FEAT_SVE()) {
+                const size_t _Sve_vl        = svcntb();
+                const size_t _Size_bytes    = _Byte_length(_First, _Last);
+                const size_t _Full_vl_bytes = _Size_bytes & ~size_t{_Sve_vl - 1};
+
```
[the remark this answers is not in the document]
hazzlim: SVE vector length is (now) architecturally defined to be a power of 2 in range [128, 2048] 

> The architecturally defined SVL set is all powers of two from 128 to 2048 bits inclusive
([ARM ARM B1.4.2](https://developer.arm.com/documentation/ddi0487/maa/-Part-B-The-AArch64-Application-Level-Architecture/-Chapter-B1-The-AArch64-Application-Level-Programmers--Model/-B1-4-The-Scalable-Vector-and-Scalable-Matrix-Extensions--SVE---SME-/-B1-4-2-Configurable-SVE-vector-lengths?lang=en))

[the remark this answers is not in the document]
StephanTLavavej: Thanks, I'll add a comment.

stl/inc/algorithm
```diff
@@ -491,10 +509,18 @@ _Ty* _Unique_copy_vectorized(const _Ty* const _First, const _Ty* const _Last, _T
 #endif // ^^^ _VECTORIZED_UNIQUE_COPY ^^^
 
 #if _VECTORIZED_REPLACE
+#if _VECTORIZED_REPLACE_1_2
+template <class _Iter>
+constexpr bool _Have_masked_op_for_iter = true;
+#else // ^^^ _VECTORIZED_REPLACE_1_2 / !_VECTORIZED_REPLACE_1_2 vvv
+template <class _Iter>
+constexpr bool _Have_masked_op_for_iter = sizeof(_Iter_value_t<_Iter>) >= 4; // avx masked op compatible size
+#endif // ^^^ !_VECTORIZED_REPLACE_1_2 ^^^
```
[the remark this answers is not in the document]
StephanTLavavej: Excellent catch, I'll fix this.

StephanTLavavej:
Internal MSVC-PR-747214, merged 1 hour ago, updated the checked-in toolset from 19.50 to 19.51, unblocking this PR.

I've pushed commits to address Copilot's review, citing the Arm ARM for the SVE vector length, and (importantly!) extending test coverage for 1/2 bytes on ARM64/ARM64EC.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Something is wrong with this PR in the MSVC-internal repo again and I need to pull it from the merge batch.

This is breaking the ARM64-native (arm64chk) compiler such that it ICEs. On a D32ps_v6 VM, the error is uninformative:

```
### CLI output:
Microsoft (R) C/C++ Optimizing Compiler Version 19.52.36529.98 for ARM64 (PREVIEW)

test.cpp
c1xx: fatal error C1001: Internal compiler error.
(compiler file 'msc1.cpp', line 1673)
 To work around this problem, try simplifying or changing the program near the locations listed above.
If possible please provide a repro here: https://developercommunity.visualstudio.com 
Please choose the Technical Support command on the Visual C++ 
 Help menu, or open the Technical Support help file for more information
INTERNAL COMPILER ERROR in 'C:\repos\msvc\binaries\arm64chk\bin\arm64\cl.exe'
    Please choose the Technical Support command on the Visual C++
    Help menu, or open the Technical Support help file for more information
```

@cdacamar helped out tremendously by getting more repro details, and our partners at Arm are now investigating.

StephanTLavavej:
Pushed a clean merge with `main` to get an updated run with the latest toolset (I expect it will pass).

StephanTLavavej:
I just read (in MSVC-PR-757601) that "The ARM64EC ABI does not include the SVE register state (z/p/FFR are not preserved across EC/x64 transitions - armasm64 rejects them with A4266)", so should we be activating this for original recipe ARM64 only and not extra crispy ARM64EC?

hazzlim:
> I just read (in [MSVC-PR-757601](https://devdiv.visualstudio.com/DevDiv/_git/msvc/pullrequest/757601)) that "The ARM64EC ABI does not include the SVE register state (z/p/FFR are not preserved across EC/x64 transitions - armasm64 rejects them with A4266)", so should we be activating this for original recipe ARM64 only and not extra crispy ARM64EC?

I am just consulting with people more in the know regarding ARM64EC than I, but as far as I understand https://learn.microsoft.com/en-us/windows/arm/arm64ec-abi does indeed imply that SVE Z and P registers are forbidden entirely. 

So I guess the fix is to remove `_M_ARM64EC` from our guards

AlexGuteniev:
> So I guess the fix is to remove `_M_ARM64EC` from our guards

But can't we keep Neon for `_M_ARM64EC` and guard away only SVE+_M_ARM64EC ?

StephanTLavavej:
Yes, my understanding is that ARM64EC supports NEON, so we shouldn't mess with that.

I believe that we should (thanks @AlexGuteniev for remembering the tests):

1. Restore the definition of `_VECTORIZED_FOR_X64_X86_ARM64` which had been removed by #6084.
1. Change the "dashboard" for `_VECTORIZED_REPLACE` to be `_VECTORIZED_FOR_X64_X86_ARM64`.
1. Change `_Use_FEAT_SVE()` to *not exist* for ARM64EC, ensuring that all codepaths that attempt to call it are also guarded for native-ARM64 only.
   * I am not sure exactly which features should be unavailable for ARM64EC, but everything higher than SVE is presumably a slam dunk. It would be great if Arm could confirm the precise set.
1. In the tests, `replace_is_vectorized` should be always true only for ARM64
1. Removal of `test_replace` in `_CALL_ALL_X64_VECTOR_ALGORITHMS_ON_ARM64EC` should be reverted

AlexGuteniev:
> I am just consulting with people more in the know regarding ARM64EC than I, but as far as I understand https://learn.microsoft.com/en-us/windows/arm/arm64ec-abi does indeed imply that SVE Z and P registers are forbidden entirely.

Here is another page with explicit table: https://learn.microsoft.com/en-us/cpp/build/arm64ec-windows-abi-conventions?view=msvc-170#register-mapping-for-integer-registers

I'm wondering if the decision not to support SVE at all on ARM64EC is permanent in long term, or it may be lifted in a near future.

StephanTLavavej:
I've pushed a merge with `main` followed by the changes that I believe are needed to deal with ARM64EC, and to use the updated `<isa_availability.h>`. Unfortunately this still doesn't work in the MSVC-internal repo.

StephanTLavavej:
I think I've figured this out.

MSVC-internal Dpsv6, Win11 24H2:
```
        __arm64_xstate_features: 0x0000000000000000
      __processor_features_0_63: 0x00003804EB8C1004
    __processor_features_64_127: 0x000000000000001F
     _Sve_in_processor_features: false
_Sve_in_enabled_xstate_features: false
```

GitHub Dpdsv6, Server 2025 25H2:
```
        __arm64_xstate_features: 0x0000000000000004
      __processor_features_0_63: 0x039AF804EB8C1004
    __processor_features_64_127: 0x000000000000001F
     _Sve_in_processor_features: true
_Sve_in_enabled_xstate_features: true
```

It appears that (contrary to the documentation I saw) Win11 24H2 doesn't support SVE. (It doesn't claim support, and attempting to execute SVE instructions will crash.) I'm not absolutely certain of this without local confirmation, though.

Additionally, we've verified through inspecting codegen (special thanks to Fulgen on the STL Discord who provided https://godbolt.org/z/nv5vs8rsW showing the SVE instruction `addvl` outside of the `_Use_FEAT_SVE` branch) that any use of SVE intrinsics will generate an SVE prologue outside of branches like `if (_Use_FEAT_SVE())`. This explains the previously-baffling observation that the code would crash despite being guarded by `if (false)`. I am verifying a fix for this now.

It's especially wacky that we discovered this very real bug due to the unexpected behavior of the MSVC CI system not supporting SVE despite its use of Cobalt 100.

stl/src/vector_algorithms.cpp
```diff
@@ -9608,6 +9577,134 @@ __declspec(noalias) size_t __stdcall __std_mismatch_8(
...
+            const auto _True = svptrue_b8();
+            while (_First != _Stop_at) {
+                const auto _Data = _Traits::_Load(_True, _First);
+                const auto _Mask = _Traits::_Cmp(_True, _Data, _Comparand);
+                _Traits::_Store(_Mask, _First, _Replacement);
+                _Advance_bytes(_First, _Sve_vl);
+            }
+
+            if (const size_t _Tail_length = _Size_bytes & size_t{_Sve_vl - 1}; _Tail_length != 0) {
+                const auto _Tail_mask = svwhilelt_b8(size_t{0}, _Tail_length);
```
StephanTLavavej, line 9690: Opus 5 insistently remarks that the use of `svwhilelt_b8` with `_Tail_length` in bytes is confusing, because other things are being done in terms of elements. It wants to move `_While_lt` into the `_Traits`, switching between the 8/16/32/64 forms, and pass an element count with `_Tail_length / sizeof(_Ty)`.

While that makes some sense to me, I don't want to mess with the vectorized code without asking first. Should we change this?

(In contrast, it was fine with the unconditional usage of `svptrue_b8`, which makes sense to me.)

StephanTLavavej (approved):

stl/src/vector_algorithms.cpp
hazzlim: I would be inclined to disagree slightly with Opus 5, as we have the SVE VL in byte units and we use this for the loop control via `_Advance_bytes` 

However I'm not against this refactoring if you would prefer it?

StephanTLavavej: Thanks. Actually I was confused, it was Copilot code review that was insistent; Opus 5 thinks the code is correct as-is and doesn't need to be changed. Let's leave it alone.

StephanTLavavej:

StephanTLavavej:
Thanks for this major optimization, we've finally figured out how to unlock a whole new set of intrinsics! :heart_eyes_cat: :rocket: :tada:

PR #6211, opened 2026-04-03
Allow ASan container annotations to activate piecemeal

### LINKED ISSUE #6186: Partially disabling STL ASan annotations leads to harmful ODR violation
# Describe the bug
The STL ASan annotations are designed to allow for "annotations inserted, but disabled" in order to help with static libraries that need to work in ASan and non-ASan contexts. The idea is that the ODR violation that may occur when some code is ASan annotated and some isn't is benign if the annotations are disabled. However, currently the "annotations enabled" switch is coarse and activates all annotations regardless of what was requested. 

This affects annotations for `vector`, `string`, and `optional`.

As an example:
If a static library `a.lib` is built without ASan and with `/D_ANNOTATE_STL` (it is a static library and wants to work with ASan and non-ASan projects), then if the code to check whether to annotate each container is added to the function definitions in this .lib.

If `a.lib` is linked to an ASan project (`/fsanitize=address`), each annotation will be activated, there are no ODR violations, and the containers will be checked for memory safety errors.

If `a.lib` is linked to a non-ASan project, there will be an ODR violation where some source files will get a version of these containers with annotation code inserted, and some will not. However, when the inserted annotation code checks for whether to call ASan to mark the shadow memory (ex: `_Asan_vector_should_annotate` for `std::vector`, etc) that value will be false, so even though there is an ODR violation, the behavior between the two competing definitions will be the same.

If `a.lib` is linked to an ASan project that does not want vector annotations (`/fsanitize=address /D_DISABLE_VECTOR_ANNOTATIONS`), `_INSERT_VECTOR_ANNOTATIONS` will be unset, so just `std::vector` will no longer have the inserted annotation code, causing an ODR violation similar to the case when linked to a non-ASan project. In this case though, `stl_asan.lib` has been included, setting `_Asan_vector_should_annotate` to `true`, even though we only want the annotations activated for `string` and `optional`, meaning that the behavior between the two competing definitions is no longer the same. In the resulting binary, some `std::vector` functions will call ASan to poison/unpoison the container, and some will not, likely eventually causing an error.

# Command-line test case (for `std::vector`, this issue exists for `std::string` and `std::optional` as well)
```
C:\Users\amyw\source\vector_anno>type test.cpp

// cl /c /EHsc /MD /O2 /D_ANNOTATE_STL /DANNOTATED annotated_lib.cpp
// lib /OUT:annotated_lib.lib annotated_lib.obj
// cl /EHsc /MD /fsanitize=address /Zi /D_DISABLE_VECTOR_ANNOTATION asan_main.cpp annotated_lib.lib
#include <vector>
#include <cstring>
#include <cstdio>
#ifdef ANNOTATED
__declspec(noinline)
void fill_vector(std::vector<char>* v) {
    v->reserve(100);
    for (int i = 0; i < 10; i++)
        v->push_back('A' + static_cast<char>(i));
}

__declspec(noinline)
void copy_from_vector(const std::vector<char>* v, char* dst, size_t n) {
    std::memcpy(dst, v->data(), n);
}
#else
// Defined in annotated_lib.lib
extern "C" const bool _Asan_vector_should_annotate;
void fill_vector(std::vector<char>* v);
void copy_from_vector(const std::vector<char>* v, char* dst, size_t n);
int main() {
    std::printf("Annotate string: %s\n", _Asan_string_should_annotate ? "true" : "false");
    std::printf("Annotate vector: %s\n", _Asan_vector_should_annotate ? "true" : "false");

    std::vector<char> v;
    fill_vector(&v);
    char buf[50];
    copy_from_vector(&v, buf, 50);
    std::printf("PASS: copied %d bytes\n", 50);
    return 0;
}
#endif

C:\Users\amyw\source\vector_anno>cl /c /EHsc /MD /O2 test.cpp /D_ANNOTATE_STL /DANNOTATED /Foannotated_lib.obj
Microsoft (R) C/C++ Optimizing Compiler Version 19.51.26504.99 for x64 (PREVIEW)

test.cpp

C:\Users\amyw\source\vector_anno>lib /OUT:annotated_lib.lib annotated_lib.obj
Microsoft (R) Library Manager Version 14.51.36127.96 (PREVIEW)


C:\Users\amyw\source\vector_anno>cl /EHsc /MD /fsanitize=address /Zi /D_DISABLE_VECTOR_ANNOTATION test.cpp annotated_lib.lib
Microsoft (R) C/C++ Optimizing Compiler Version 19.51.26504.99 for x64 (PREVIEW)

test.cpp
Microsoft (R) Incremental Linker Version 14.51.36127.96 (PREVIEW)

/out:test.exe
/InferAsanLibs
/debug
test.obj
annotated_lib.lib

C:\Users\amyw\source\vector_anno>test.exe
Annotate string: true
Annotate vector: true
=================================================================
==45176==ERROR: AddressSanitizer: container-overflow on address 0x12358a7a00fa at pc 0x7ffac533bea5 bp 0x00a2cf97fcf0 sp 0x00a2cf97f480
READ of size 50 at 0x12358a7a00fa thread T0
    #0 0x7ffac533bea4 in memcpy X:\msvc\src\vctools\asan\llvm\compiler-rt\lib\sanitizer_common\sanitizer_common_interceptors_memintrinsics.inc:147
    #1 0x7ff7959d120d in main C:\Users\amyw\source\vector_anno\test.cpp:32
    #2 0x7ff7959d4283 in invoke_main C:\repos\msvc\src\vctools\crt\vcstartup\src\startup\exe_common.inl:78
    #3 0x7ff7959d4283 in __scrt_common_main_seh C:\repos\msvc\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288
    #4 0x7ffbc3e2e8d6  (C:\WINDOWS\System32\KERNEL32.DLL+0x18002e8d6)
    #5 0x7ffbc522c53b  (C:\WINDOWS\SYSTEM32\ntdll.dll+0x18008c53b)

0x12358a7a00fa is located 10 bytes inside of 100-byte region [0x12358a7a00f0,0x12358a7a0154)
allocated by thread T0 here:
    #0 0x7ff7959d3ad5 in operator new(unsigned __int64) X:\msvc\src\vctools\asan\llvm\compiler-rt\lib\asan\asan_win_new_scalar_thunk.cpp:40
    #1 0x7ff7959d36d4 in fill_vector(class std::vector<char, class std::allocator<char>> *) (C:\Users\amyw\source\vector_anno\test.exe+0x1400036d4)
    #2 0x7ff7959d11f5 in main C:\Users\amyw\source\vector_anno\test.cpp:30
    #3 0x7ff7959d4283 in invoke_main C:\repos\msvc\src\vctools\crt\vcstartup\src\startup\exe_common.inl:78
    #4 0x7ff7959d4283 in __scrt_common_main_seh C:\repos\msvc\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288
    #5 0x7ffbc3e2e8d6  (C:\WINDOWS\System32\KERNEL32.DLL+0x18002e8d6)
    #6 0x7ffbc522c53b  (C:\WINDOWS\SYSTEM32\ntdll.dll+0x18008c53b)

HINT: if you don't care about these errors you may set ASAN_OPTIONS=detect_container_overflow=0.
Or if supported by the container library, pass -D__SANITIZER_DISABLE_CONTAINER_OVERFLOW__ to the compiler to disable  instrumentation.
If you suspect a false positive see also: https://github.com/google/sanitizers/wiki/AddressSanitizerContainerOverflow.
SUMMARY: AddressSanitizer: container-overflow C:\Users\amyw\source\vector_anno\test.cpp:32 in main
Shadow bytes around the buggy address:
  0x12358a79fe00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x12358a79fe80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x12358a79ff00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x12358a79ff80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x12358a7a0000: fa fa fa fa fa fa fa fa fd fd fd fd fd fd fd fd
=>0x12358a7a0080: fd fd fd fd fd fd fa fa fa fa fa fa fa fa 00[02]
  0x12358a7a0100: fc fc fc fc fc fc fc fc fc fc fc fa fa fa fa fa
  0x12358a7a0180: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x12358a7a0200: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x12358a7a0280: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x12358a7a0300: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==45176==ABORTING
```

# Expected behavior
Annotations should be disabled for vector, but not string if `/D_DISABLE_VECTOR_ANNOTATION` is specified.

Above repro should produce:
```
C:\Users\amyw\source\vector_anno>test.exe
Annotate string: true
Annotate vector: false
PASS: copied 50 bytes
```

amyw-msft:
Fixes #6186.

Previously, ASan annotations for `string`, `vector`, and `optional` were activated by linking with `stl_asan.lib` which set each activation variable to true, leading to ODR violations if some TUs were built with annotations inserted but partially activated.

This change removes the `stl_asan.lib` dependency and allows each option to set the activation variable individually. 

I've retained `stl_asan.lib` and modified the definitions with `__declspec(selectany)` to match those in the headers, and also to ensure compatibility if linking with a static library from a previous version of the toolset.

The test I've included just ensures the annotations can be targeted individually instead of constructing an ASan error. I did this for simplicity (esp as new annotations are added) while still covering every scenario. I limited it to one disabled out of the set instead of exhaustive validation to not devote too much test time to this. 

I've verified the scenario from #6186 locally, as well as ensuring it is backwards compatible with old versions of the headers (meaning, I built the annotated_lib.lib with an old toolset and test.obj with a toolset with this change and manually tried out every combination). Building a static library with the new toolset and linking with an old toolset produces a link error (the old `stl_asan.lib` does not contain `__declspec(selectany)`, so clashes with the definition added in `__msvc_sanitizer_annotate_container.hpp`), but my understanding is that this isn't a supported scenario.

### DIFF
--- a/stl/inc/__msvc_sanitizer_annotate_container.hpp
+++ b/stl/inc/__msvc_sanitizer_annotate_container.hpp
@@ -129,15 +129,15 @@ _STL_DISABLE_CLANG_WARNINGS
 #endif // ^^^ !defined(_INSERT_OPTIONAL_ANNOTATION) ^^^
 
 #ifdef _ACTIVATE_STRING_ANNOTATION
-#pragma comment(lib, "stl_asan")
+extern "C" __declspec(selectany) const bool _Asan_string_should_annotate = true;
 #pragma detect_mismatch("annotate_string", "1")
 #endif // ^^^ defined(_ACTIVATE_STRING_ANNOTATION) ^^^
 #ifdef _ACTIVATE_VECTOR_ANNOTATION
-#pragma comment(lib, "stl_asan")
+extern "C" __declspec(selectany) const bool _Asan_vector_should_annotate = true;
 #pragma detect_mismatch("annotate_vector", "1")
 #endif // ^^^ defined(_ACTIVATE_VECTOR_ANNOTATION) ^^^
 #ifdef _ACTIVATE_OPTIONAL_ANNOTATION
-#pragma comment(lib, "stl_asan")
+extern "C" __declspec(selectany) const bool _Asan_optional_should_annotate = true;
 #pragma detect_mismatch("annotate_optional", "1")
 #endif // ^^^ defined(_ACTIVATE_OPTIONAL_ANNOTATION) ^^^
 
--- a/stl/inc/__msvc_sanitizer_annotate_container.hpp
+++ b/stl/inc/__msvc_sanitizer_annotate_container.hpp
@@ -128,16 +128,17 @@ _STL_DISABLE_CLANG_WARNINGS
 #pragma detect_mismatch("annotate_optional", "0")
 #endif // ^^^ !defined(_INSERT_OPTIONAL_ANNOTATION) ^^^
 
+extern "C" {
 #ifdef _ACTIVATE_STRING_ANNOTATION
-extern "C" __declspec(selectany) const bool _Asan_string_should_annotate = true;
+__declspec(selectany) const bool _Asan_string_should_annotate = true;
 #pragma detect_mismatch("annotate_string", "1")
 #endif // ^^^ defined(_ACTIVATE_STRING_ANNOTATION) ^^^
 #ifdef _ACTIVATE_VECTOR_ANNOTATION
-extern "C" __declspec(selectany) const bool _Asan_vector_should_annotate = true;
+__declspec(selectany) const bool _Asan_vector_should_annotate = true;
 #pragma detect_mismatch("annotate_vector", "1")
 #endif // ^^^ defined(_ACTIVATE_VECTOR_ANNOTATION) ^^^
 #ifdef _ACTIVATE_OPTIONAL_ANNOTATION
-extern "C" __declspec(selectany) const bool _Asan_optional_should_annotate = true;
+__declspec(selectany) const bool _Asan_optional_should_annotate = true;
 #pragma detect_mismatch("annotate_optional", "1")
 #endif // ^^^ defined(_ACTIVATE_OPTIONAL_ANNOTATION) ^^^
 
@@ -145,7 +146,6 @@ extern "C" __declspec(selectany) const bool _Asan_optional_should_annotate = tru
 #undef _ACTIVATE_VECTOR_ANNOTATION
 #undef _ACTIVATE_OPTIONAL_ANNOTATION
 
-extern "C" {
 #ifdef _INSERT_VECTOR_ANNOTATION
 extern const bool _Asan_vector_should_annotate;
 #endif
--- a/stl/inc/__msvc_sanitizer_annotate_container.hpp
+++ b/stl/inc/__msvc_sanitizer_annotate_container.hpp
@@ -130,15 +130,15 @@ _STL_DISABLE_CLANG_WARNINGS
 
 extern "C" {
 #ifdef _ACTIVATE_STRING_ANNOTATION
-__declspec(selectany) const bool _Asan_string_should_annotate = true;
+__declspec(selectany) extern const bool _Asan_string_should_annotate = true;
 #pragma detect_mismatch("annotate_string", "1")
 #endif // ^^^ defined(_ACTIVATE_STRING_ANNOTATION) ^^^
 #ifdef _ACTIVATE_VECTOR_ANNOTATION
-__declspec(selectany) const bool _Asan_vector_should_annotate = true;
+__declspec(selectany) extern const bool _Asan_vector_should_annotate = true;
 #pragma detect_mismatch("annotate_vector", "1")
 #endif // ^^^ defined(_ACTIVATE_VECTOR_ANNOTATION) ^^^
 #ifdef _ACTIVATE_OPTIONAL_ANNOTATION
-__declspec(selectany) const bool _Asan_optional_should_annotate = true;
+__declspec(selectany) extern const bool _Asan_optional_should_annotate = true;
 #pragma detect_mismatch("annotate_optional", "1")
 #endif // ^^^ defined(_ACTIVATE_OPTIONAL_ANNOTATION) ^^^
 
--- a/stl/src/asan.cpp
+++ b/stl/src/asan.cpp
@@ -1,16 +1,14 @@
 // Copyright (c) Microsoft Corporation.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-namespace std {
-    extern "C" {
-    // TRANSITION, ABI: preserved for compatibility with old headers, which
-    // added stl_asan.lib to the link line. We use __declspec(selectany) to be
-    // compatible with the new headers that define these variables as
-    // __declspec(selectany) in __msvc_sanitizer_annotate_container.hpp.
-    // The new method is preferred because previously enabling just string
-    // would also enable vector and optional. See GH-6186 for details.
-    __declspec(selectany) extern const bool _Asan_string_should_annotate   = true;
-    __declspec(selectany) extern const bool _Asan_vector_should_annotate   = true;
-    __declspec(selectany) extern const bool _Asan_optional_should_annotate = true;
-    } // extern "C"
-} // namespace std
+extern "C" {
+// TRANSITION, ABI: preserved for compatibility with old headers, which
+// added stl_asan.lib to the link line. We use __declspec(selectany) to be
+// compatible with the new headers that define these variables as
+// __declspec(selectany) in __msvc_sanitizer_annotate_container.hpp.
+// The new method is preferred because previously enabling just string
+// would also enable vector and optional. See GH-6186 for details.
+__declspec(selectany) extern const bool _Asan_string_should_annotate   = true;
+__declspec(selectany) extern const bool _Asan_vector_should_annotate   = true;
+__declspec(selectany) extern const bool _Asan_optional_should_annotate = true;
+} // extern "C"
--- a/stl/src/asan.cpp
+++ b/stl/src/asan.cpp
@@ -9,8 +9,8 @@ namespace std {
     // __declspec(selectany) in __msvc_sanitizer_annotate_container.hpp.
     // The new method is preferred because previously enabling just string
     // would also enable vector and optional. See GH-6186 for details.
-    extern __declspec(selectany) const bool _Asan_string_should_annotate   = true;
-    extern __declspec(selectany) const bool _Asan_vector_should_annotate   = true;
-    extern __declspec(selectany) const bool _Asan_optional_should_annotate = true;
+    __declspec(selectany) extern const bool _Asan_string_should_annotate   = true;
+    __declspec(selectany) extern const bool _Asan_vector_should_annotate   = true;
+    __declspec(selectany) extern const bool _Asan_optional_should_annotate = true;
     } // extern "C"
 } // namespace std
--- a/stl/src/asan.cpp
+++ b/stl/src/asan.cpp
@@ -3,11 +3,12 @@
 
 namespace std {
     extern "C" {
-    // Retained for compatibility with old headers that add stl_asan.lib to the link.
-    // Use __declspec(selectany) to be compatible with new version that define these
-    // variables as __declspec(selectany) in __msvc_sanitizer_annotate_container.hpp.
-    // The new method is preferred because previously enabling just string would also
-    // enable vector and optional.
+    // TRANSITION, ABI: preserved for compatibility with old headers, which
+    // added stl_asan.lib to the link line. We use __declspec(selectany) to be
+    // compatible with the new headers that defines these variables as
+    // __declspec(selectany) in __msvc_sanitizer_annotate_container.hpp.
+    // The new method is preferred because previously enabling just string
+    // would also enable vector and optional. See GH-6186 for details.
     extern __declspec(selectany) const bool _Asan_string_should_annotate   = true;
     extern __declspec(selectany) const bool _Asan_vector_should_annotate   = true;
     extern __declspec(selectany) const bool _Asan_optional_should_annotate = true;
--- a/stl/src/asan.cpp
+++ b/stl/src/asan.cpp
@@ -3,6 +3,11 @@
 
 namespace std {
     extern "C" {
+    // Retained for compatibility with old headers that add stl_asan.lib to the link.
+    // Use __declspec(selectany) to be compatible with new version that define these
+    // variables as __declspec(selectany) in __msvc_sanitizer_annotate_container.hpp.
+    // The new method is preferred because previously enabling just string would also
+    // enable vector and optional.
     extern __declspec(selectany) const bool _Asan_string_should_annotate   = true;
     extern __declspec(selectany) const bool _Asan_vector_should_annotate   = true;
     extern __declspec(selectany) const bool _Asan_optional_should_annotate = true;
--- a/stl/src/asan.cpp
+++ b/stl/src/asan.cpp
@@ -5,7 +5,7 @@ namespace std {
     extern "C" {
     // TRANSITION, ABI: preserved for compatibility with old headers, which
     // added stl_asan.lib to the link line. We use __declspec(selectany) to be
-    // compatible with the new headers that defines these variables as
+    // compatible with the new headers that define these variables as
     // __declspec(selectany) in __msvc_sanitizer_annotate_container.hpp.
     // The new method is preferred because previously enabling just string
     // would also enable vector and optional. See GH-6186 for details.
--- a/stl/src/asan.cpp
+++ b/stl/src/asan.cpp
@@ -3,8 +3,8 @@
 
 namespace std {
     extern "C" {
-    extern const bool _Asan_string_should_annotate   = true;
-    extern const bool _Asan_vector_should_annotate   = true;
-    extern const bool _Asan_optional_should_annotate = true;
+    extern __declspec(selectany) const bool _Asan_string_should_annotate   = true;
+    extern __declspec(selectany) const bool _Asan_vector_should_annotate   = true;
+    extern __declspec(selectany) const bool _Asan_optional_should_annotate = true;
     } // extern "C"
 } // namespace std
--- a/tests/std/tests/GH_006186_asan_annotate_partial/env.lst
+++ b/tests/std/tests/GH_006186_asan_annotate_partial/env.lst
@@ -1,0 +1,41 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# This test matrix is the usual test matrix, with all currently unsupported options removed, crossed with the ASan flags.
+
+# TRANSITION, google/sanitizers#328: clang-cl does not support /MDd or /MTd with ASan
+RUNALL_INCLUDE ..\prefix.lst
+RUNALL_CROSSLIST
+PM_CL="/fsanitize=address /DTEST_ENSURE_VECTOR_ENABLED /DTEST_ENSURE_STRING_ENABLED /DTEST_ENSURE_OPTIONAL_ENABLED"
+PM_CL="/fsanitize=address /D_DISABLE_VECTOR_ANNOTATION /DTEST_ENSURE_STRING_ENABLED /DTEST_ENSURE_OPTIONAL_ENABLED"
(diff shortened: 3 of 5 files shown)


tests/std/tests/GH_006186_asan_annotate_partial/env.lst
```diff
@@ -0,0 +1,41 @@
...
+PM_CL="/EHsc /MDd /std:c++20 /permissive-"
+PM_CL="/EHsc /MDd /std:c++latest /permissive- /Zc:wchar_t-"
+PM_CL="/EHsc /MDd /std:c++latest /permissive-"
+PM_CL="/EHsc /MT /std:c++latest /permissive- /analyze:only /analyze:autolog-"
+PM_CL="/EHsc /MT /std:c++latest /permissive-"
+PM_CL="/EHsc /MTd /std:c++latest /permissive"
+PM_CL="/EHsc /MTd /std:c++latest /permissive- /analyze:only /analyze:autolog-"
+PM_CL="/EHsc /MTd /std:c++latest /permissive- /fp:strict"
+PM_CL="/EHsc /MTd /std:c++latest /permissive-"
+# TRANSITION, we don't use /ALTERNATENAME for Clang (see GH-5224) so we cannot test /D_ANNOTATE_VECTOR without -fsanitize=address
```
StephanTLavavej: This comment was copy-pasted from `GH_002030_asan_annotate_vector/env.lst`, but as you're crossing with `PM_CL="/D_ANNOTATE_STL"`, I think this comment is inapplicable and should be removed here.

stl/src/asan.cpp
```diff
@@ -3,8 +3,13 @@
 
 namespace std {
     extern "C" {
-    extern const bool _Asan_string_should_annotate   = true;
-    extern const bool _Asan_vector_should_annotate   = true;
-    extern const bool _Asan_optional_should_annotate = true;
+    // Retained for compatibility with old headers that add stl_asan.lib to the link.
```
StephanTLavavej: This should use the magic words "TRANSITION, ABI".

stl/src/asan.cpp
```diff
@@ -3,8 +3,13 @@
...
-    extern const bool _Asan_vector_should_annotate   = true;
-    extern const bool _Asan_optional_should_annotate = true;
+    // Retained for compatibility with old headers that add stl_asan.lib to the link.
+    // Use __declspec(selectany) to be compatible with new version that define these
+    // variables as __declspec(selectany) in __msvc_sanitizer_annotate_container.hpp.
+    // The new method is preferred because previously enabling just string would also
+    // enable vector and optional.
+    extern __declspec(selectany) const bool _Asan_string_should_annotate   = true;
+    extern __declspec(selectany) const bool _Asan_vector_should_annotate   = true;
+    extern __declspec(selectany) const bool _Asan_optional_should_annotate = true;
```
StephanTLavavej: I'm concerned that bundling these 3 definitions into a single object can still lead to scenarios where annotations are partially activated, the linker drags in `asan.obj` here, and (through the magic of selectany) we end up activating annotations that the user didn't request.

Conversely, I am even more concerned that `asan_noop.cpp` isn't being changed, and is still a monolithic source file producing a single object file.

Should we break up `asan.cpp` and `asan_noop.cpp` into 3 separate files each, prominently commented that they need to remain separate and never fused? `stl_asan.lib` would still be a single LIB for back-compat containing the 3 OBJs. The linker's behavior operates at an OBJ level of granularity.

stl/src/asan.cpp
```diff
@@ -3,8 +3,13 @@
...
 namespace std {
     extern "C" {
-    extern const bool _Asan_string_should_annotate   = true;
-    extern const bool _Asan_vector_should_annotate   = true;
-    extern const bool _Asan_optional_should_annotate = true;
+    // Retained for compatibility with old headers that add stl_asan.lib to the link.
+    // Use __declspec(selectany) to be compatible with new version that define these
+    // variables as __declspec(selectany) in __msvc_sanitizer_annotate_container.hpp.
+    // The new method is preferred because previously enabling just string would also
+    // enable vector and optional.
```
StephanTLavavej: Adding ", see GH-6186." would be helpful (we try to avoid the need to `git blame` for programmer-archaeology).

stl/src/asan.cpp
```diff
@@ -3,8 +3,13 @@
 
 namespace std {
     extern "C" {
-    extern const bool _Asan_string_should_annotate   = true;
-    extern const bool _Asan_vector_should_annotate   = true;
-    extern const bool _Asan_optional_should_annotate = true;
+    // Retained for compatibility with old headers that add stl_asan.lib to the link.
+    // Use __declspec(selectany) to be compatible with new version that define these
```
StephanTLavavej: "with new version that define these" isn't grammatical. Should probably be "with the new version that defines these".

stl/inc/__msvc_sanitizer_annotate_container.hpp
```diff
@@ -129,15 +129,15 @@ _STL_DISABLE_CLANG_WARNINGS
 #endif // ^^^ !defined(_INSERT_OPTIONAL_ANNOTATION) ^^^
 
 #ifdef _ACTIVATE_STRING_ANNOTATION
-#pragma comment(lib, "stl_asan")
+extern "C" __declspec(selectany) const bool _Asan_string_should_annotate = true;
```
StephanTLavavej: Instead of individually marking these as `extern "C"`, I believe it would be better to move up the `extern "C" {` scope beginning at right line 148, so that these definitions look more similar to the declarations below. The only stuff in between are pragmas and undefs.

tests/std/tests/GH_006186_asan_annotate_partial/test.cpp
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// REQUIRES: x64 || x86 || arm64
```
StephanTLavavej, line 4: Nitpick: We conventionally put an empty line after the license banner.

StephanTLavavej (changes requested):

stl/src/asan.cpp
amyw-msft: The normal selectany concerns shouldn't be present here because we are using `/alternatename` to select the intended value. The definitions in asan_noop.cpp are appended with 'default' (like `_Asan_string_should_annotate_default`) and always set to `false`, and all non-`default` values (like `_Asan_string_should_annotate`) are always set to `true`. 

Using `/alternatename:A=B` will cause the linker to search for A like it's a normal symbol, and only once it cannot be found anywhere, result to resolving it to B. So, if `_Asan_string_should_annotate` is defined anywhere in the link, it will be `true`. If it is not, it will be resolved to `_Asan_string_should_annotate_default` and be `false`. 

In order to partially activate the annotations with a seperate .lib, we would need a different .lib for each setting, which is why I went with this strategy of just replacing stl_asan.lib with a selectany definition in the header. Likewise, because of the `/alternatename` mechanism, asan_noop.cpp doesn't need to be broken up: if `_Asan_string_should_annotate` is present in the link, `_Asan_string_should_annotate_default` simply exists as a separate symbol and does not interact.

stl/inc/__msvc_sanitizer_annotate_container.hpp
```diff
@@ -128,24 +128,24 @@ _STL_DISABLE_CLANG_WARNINGS
 #pragma detect_mismatch("annotate_optional", "0")
 #endif // ^^^ !defined(_INSERT_OPTIONAL_ANNOTATION) ^^^
 
+extern "C" {
 #ifdef _ACTIVATE_STRING_ANNOTATION
-#pragma comment(lib, "stl_asan")
+__declspec(selectany) const bool _Asan_string_should_annotate = true;
```
StephanTLavavej: Ah, this is now causing the tests to fail with:

```
C:\stlBuild\out\inc\__msvc_sanitizer_annotate_container.hpp(133): error C2496: '__declspec(selectany)' can only be applied to data items with external linkage
```

because now that the `extern "C"` is a scope, the variables aren't getting `extern` linkage. Gotta love this language.

StephanTLavavej (changes requested):

stl/src/asan.cpp
StephanTLavavej: Thanks. I wonder if that means we could avoid `asan_noop.cpp` by having its definitions be extern and selectany, but that would just add extra complexity since we'd still need to retain it.

stl/src/asan.cpp
```diff
@@ -3,8 +3,14 @@
 
 namespace std {
     extern "C" {
-    extern const bool _Asan_string_should_annotate   = true;
-    extern const bool _Asan_vector_should_annotate   = true;
-    extern const bool _Asan_optional_should_annotate = true;
+    // TRANSITION, ABI: preserved for compatibility with old headers, which
+    // added stl_asan.lib to the link line. We use __declspec(selectany) to be
+    // compatible with the new headers that defines these variables as
```
StephanTLavavej: Now that you're saying "the new headers" we need to go back to "define".

stl/src/asan.cpp
```diff
@@ -3,8 +3,14 @@
...
-    extern const bool _Asan_optional_should_annotate = true;
+    // TRANSITION, ABI: preserved for compatibility with old headers, which
+    // added stl_asan.lib to the link line. We use __declspec(selectany) to be
+    // compatible with the new headers that defines these variables as
+    // __declspec(selectany) in __msvc_sanitizer_annotate_container.hpp.
+    // The new method is preferred because previously enabling just string
+    // would also enable vector and optional. See GH-6186 for details.
+    extern __declspec(selectany) const bool _Asan_string_should_annotate   = true;
+    extern __declspec(selectany) const bool _Asan_vector_should_annotate   = true;
+    extern __declspec(selectany) const bool _Asan_optional_should_annotate = true;
```
StephanTLavavej: According to the [`__declspec(selectany)`](https://learn.microsoft.com/cpp/cpp/selectany) documentation, we can place it before or after the `extern`, but I think it makes more sense to not interleave it with the Standard keywords.

tests/std/tests/GH_006186_asan_annotate_partial/test.cpp
```diff
@@ -0,0 +1,38 @@
...
+#pragma comment(linker, "/INFERASANLIBS")
+int main() {}
+#else // ^^^ workaround / no workaround vvv
+
+#include <cassert>
+#include <vector> // include __msvc_sanitizer_annotate_container.hpp
+
+extern "C" const bool _Asan_vector_should_annotate;
+extern "C" const bool _Asan_string_should_annotate;
+extern "C" const bool _Asan_optional_should_annotate;
```
StephanTLavavej: Style: We can also update these to match the headers.

stl/src/asan.cpp
```diff
@@ -3,8 +3,14 @@
 
 namespace std {
     extern "C" {
```
StephanTLavavej: Pre-existing: This is bogus, although fortunately innocuous. `<__msvc_sanitizer_annotate_container.hpp>` doesn't have `namespace std`. The namespace is simply ignored for `extern "C"`, WG21-N5032 \[dcl.link\]/7:

> Two declarations declare the same entity if they (re)introduce the same name, one declares a function or variable with C language linkage, and the other declares such an entity or declares a variable that belongs to the global scope.

StephanTLavavej: Before removing `namespace std`:

```
D:\GitHub\STL\out\x64\out\lib\amd64>git branch --show-current
main

D:\GitHub\STL\out\x64\out\lib\amd64>dumpbin /symbols stl_asan.lib | rg _Asan
008 00000000 SECT3  notype       External     | _Asan_string_should_annotate
009 00000001 SECT3  notype       External     | _Asan_vector_should_annotate
00A 00000002 SECT3  notype       External     | _Asan_optional_should_annotate
```

After removing `namespace std`:

```
D:\GitHub\STL\out\x64\out\lib\amd64>git branch --show-current
partial-asan-annotation

D:\GitHub\STL\out\x64\out\lib\amd64>dumpbin /symbols stl_asan.lib | rg _Asan
008 00000000 SECT3  notype       External     | _Asan_string_should_annotate
00B 00000000 SECT4  notype       External     | _Asan_vector_should_annotate
00E 00000000 SECT5  notype       External     | _Asan_optional_should_annotate
```

StephanTLavavej (approved):

tests/std/tests/GH_006186_asan_annotate_partial/env.lst
```diff
@@ -0,0 +1,40 @@
...
+
+# This test matrix is the usual test matrix, with all currently unsupported options removed, crossed with the ASan flags.
+
+# TRANSITION, google/sanitizers#328: clang-cl does not support /MDd or /MTd with ASan
+RUNALL_INCLUDE ..\prefix.lst
+RUNALL_CROSSLIST
+PM_CL="/fsanitize=address /DTEST_ENSURE_VECTOR_ENABLED /DTEST_ENSURE_STRING_ENABLED /DTEST_ENSURE_OPTIONAL_ENABLED"
+PM_CL="/fsanitize=address /D_DISABLE_VECTOR_ANNOTATION /DTEST_ENSURE_STRING_ENABLED /DTEST_ENSURE_OPTIONAL_ENABLED"
+PM_CL="/fsanitize=address /DTEST_ENSURE_VECTOR_ENABLED /D_DISABLE_STRING_ANNOTATION /DTEST_ENSURE_OPTIONAL_ENABLED"
+PM_CL="/fsanitize=address /DTEST_ENSURE_VECTOR_ENABLED /DTEST_ENSURE_STRING_ENABLED /D_DISABLE_OPTIONAL_ANNOTATION"
```
zacklj89, line 12: Nice 👍 

Should we also add a test for multiple disabled in case for sanity checks or is that overkill?

```
PM_CL="/fsanitize=address /D_DISABLE_VECTOR_ANNOTATION /D_DISABLE_STRING_ANNOTATION /DTEST_ENSURE_OPTIONAL_ENABLED"
PM_CL="/fsanitize=address /D_DISABLE_VECTOR_ANNOTATION /DTEST_ENSURE_STRING_ENABLED /D_DISABLE_OPTIONAL_ANNOTATION"
PM_CL="/fsanitize=address /DTEST_ENSURE_VECTOR_ENABLED /D_DISABLE_STRING_ANNOTATION /D_DISABLE_OPTIONAL_ANNOTATION"
``` 

edit: missed the description outlining this, I'll resolve it then

zacklj89: And also `PM_CL="/fsanitize=address /D_DISABLE_STL_ANNOTATION"` variants?

zacklj89 (approved):
Thanks for fixing this :)

tests/std/tests/GH_006186_asan_annotate_partial/env.lst
amyw-msft: Yeah - I tested all the configurations manually, I just didn't want to explode the number of tests we run regularly.

StephanTLavavej:

StephanTLavavej:
Internal build failure that I don't understand, pulling from the merge batch.

PR #6224, opened 2026-04-05
Implement LWG-4510 Ambiguity of `std::ranges::advance` and `std::ranges::next` when the difference type is also a sentinel type

### LINKED ISSUE #6218: LWG-4510 Ambiguity of `std::ranges::advance` and `std::ranges::next` when the difference type is also a sentinel type
LWG-4510 Ambiguity of `std::ranges::advance` and `std::ranges::next` when the difference type is also a sentinel type

mirion-dev:
Fixes #6218.

### DIFF
--- a/tests/std/tests/P0896R4_ranges_iterator_machinery/test.cpp
+++ b/tests/std/tests/P0896R4_ranges_iterator_machinery/test.cpp
@@ -3686,7 +3686,7 @@ namespace lwg4510 {
         using difference_type = int;
 
         IterType& operator++();
-        IterType& operator++(int);
+        IterType operator++(int);
         IterType& operator*() const;
     };
 
--- a/tests/std/tests/P0896R4_ranges_iterator_machinery/test.cpp
+++ b/tests/std/tests/P0896R4_ranges_iterator_machinery/test.cpp
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+#include <__msvc_int128.hpp>
 #include <cassert>
 #include <compare>
 #include <concepts>
@@ -1544,6 +1545,27 @@ namespace iterator_concept_sentinel_test {
         return (unpack_iterator<Is>(std::make_index_sequence<iterator_archetype_max + 1>{}) && ...);
     }
     static_assert(unpack_sentinel(std::make_index_sequence<sentinel_archetype_max + 1>{}));
+
+    // Test LWG-4510
+    // "Ambiguity of std::ranges::advance and std::ranges::next when the difference type is also a sentinel type"
+    template <class T>
+    struct iterator {
+        using difference_type = int;
+
+        iterator& operator++();
+        iterator& operator++(int);
+        iterator& operator*() const;
+    };
+
+    struct any {
+        any(const auto&);
+
+        friend bool operator==(const any&, const any&);
+    };
+
+    static_assert(std::input_or_output_iterator<iterator<any>>);
+    static_assert(!std::sentinel_for<int, iterator<any>>);
+    static_assert(!std::sentinel_for<std::_Unsigned128, iterator<any>>);
 } // namespace iterator_concept_sentinel_test
 
 namespace iterator_concept_sizedsentinel_test {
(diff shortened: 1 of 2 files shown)


tests/std/tests/P0896R4_ranges_iterator_machinery/test.cpp
```diff
@@ -1544,6 +1545,27 @@ namespace iterator_concept_sentinel_test {
         return (unpack_iterator<Is>(std::make_index_sequence<iterator_archetype_max + 1>{}) && ...);
     }
     static_assert(unpack_sentinel(std::make_index_sequence<sentinel_archetype_max + 1>{}));
+
+    // Test LWG-4510
+    // "Ambiguity of std::ranges::advance and std::ranges::next when the difference type is also a sentinel type"
+    template <class T>
+    struct iterator {
```
StephanTLavavej: These are quasi-shadowing `std::iterator` and `std::any` which is unnecessarily confusing.

tests/std/tests/P0896R4_ranges_iterator_machinery/test.cpp
```diff
@@ -1544,6 +1545,27 @@ namespace iterator_concept_sentinel_test {
...
+        using difference_type = int;
+
+        iterator& operator++();
+        iterator& operator++(int);
+        iterator& operator*() const;
+    };
+
+    struct any {
+        any(const auto&);
```
StephanTLavavej: We generally avoid abbreviated function templates (they are a bit too subtle).

tests/std/tests/P0896R4_ranges_iterator_machinery/test.cpp
```diff
@@ -1544,6 +1545,27 @@ namespace iterator_concept_sentinel_test {
         return (unpack_iterator<Is>(std::make_index_sequence<iterator_archetype_max + 1>{}) && ...);
     }
     static_assert(unpack_sentinel(std::make_index_sequence<sentinel_archetype_max + 1>{}));
+
+    // Test LWG-4510
```
StephanTLavavej: This is a good file to add test coverage to, but we don't specifically need anything from `namespace iterator_concept_sentinel_test`. It would probably be better to move this next to `namespace lwg3420`.

tests/std/tests/P0896R4_ranges_iterator_machinery/test.cpp
```diff
@@ -1544,6 +1545,27 @@ namespace iterator_concept_sentinel_test {
...
     static_assert(unpack_sentinel(std::make_index_sequence<sentinel_archetype_max + 1>{}));
+
+    // Test LWG-4510
+    // "Ambiguity of std::ranges::advance and std::ranges::next when the difference type is also a sentinel type"
+    template <class T>
+    struct iterator {
+        using difference_type = int;
+
+        iterator& operator++();
+        iterator& operator++(int);
```
StephanTLavavej: Postincrement should return by value.

tests/std/tests/P0896R4_ranges_iterator_machinery/test.cpp
```diff
@@ -1544,6 +1545,27 @@ namespace iterator_concept_sentinel_test {
...
+
+    struct any {
+        any(const auto&);
+
+        friend bool operator==(const any&, const any&);
+    };
+
+    static_assert(std::input_or_output_iterator<iterator<any>>);
+    static_assert(!std::sentinel_for<int, iterator<any>>);
+    static_assert(!std::sentinel_for<std::_Unsigned128, iterator<any>>);
```
StephanTLavavej: I think we should have a positive test for `sentinel_for`. This helps ensure that something isn't causing the concept to be false across the board.

StephanTLavavej (approved):

tests/std/tests/P0896R4_ranges_iterator_machinery/test.cpp
mirion-dev: My intention is to use the weakest iterator that satisfies the requirements of sentinel_for, so I returned It& instead of It.

StephanTLavavej: Thanks. That seems unusually squirrely though, and not necessary to exercise this LWG resolution, so I'd prefer the more realistic iterator here.

StephanTLavavej:

StephanTLavavej:
:heart_eyes_cat: :smile_cat: :cat2:
===

PR #6228, opened 2026-04-06
Implement LWG-4497 `std::nullopt_t` should be comparable

### LINKED ISSUE #6215: LWG-4497 `std::nullopt_t` should be comparable
LWG-4497 `std::nullopt_t` should be comparable

alexprabhat99:
Fixes: #6215


stl/inc/optional
```diff
@@ -33,6 +33,9 @@ _STD_BEGIN
 _EXPORT_STD struct nullopt_t { // no-value state indicator
     struct _Tag {};
     constexpr explicit nullopt_t(_Tag) {}
+#if _HAS_CXX20
+    _NODISCARD friend constexpr strong_ordering operator<=>(nullopt_t, nullopt_t) noexcept = default;
+#endif // _HAS_CXX20
```
frederick-vs-ja: I'd like to backport this to C++17 mode.
```suggestion
#if _HAS_CXX20
    _NODISCARD friend constexpr auto operator<=>(nullopt_t, nullopt_t) noexcept /* strengthened */ = default;
#else // ^^^ _HAS_CXX20 / !_HAS_CXX20 vvv
    _NODISCARD friend constexpr bool operator==(nullopt_t, nullopt_t) noexcept /* strengthened */ {
        return true;
    }

    _NODISCARD friend constexpr bool operator!=(nullopt_t, nullopt_t) noexcept /* strengthened */ {
        return false;
    }

    _NODISCARD friend constexpr bool operator<(nullopt_t, nullopt_t) noexcept /* strengthened */ {
        return false;
    }

    _NODISCARD friend constexpr bool operator>(nullopt_t, nullopt_t) noexcept /* strengthened */ {
        return false;
    }

    _NODISCARD friend constexpr bool operator<=(nullopt_t, nullopt_t) noexcept /* strengthened */ {
        return true;
    }

    _NODISCARD friend constexpr bool operator>=(nullopt_t, nullopt_t) noexcept /* strengthened */ {
        return true;
    }
#endif // ^^^ !_HAS_CXX20 ^^^
```

tests/std/tests/P2405R1_nullopt_t_comparable/env.lst
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+RUNALL_INCLUDE ..\strict_20_matrix.lst
```
frederick-vs-ja: Ditto, I'd like to test these in C++17 mode (with some code guarded with `_HAS_CXX20`).
```suggestion
RUNALL_INCLUDE ..\usual_17_matrix.lst
```

tests/std/test.lst
```diff
@@ -697,6 +697,7 @@ tests\P2387R3_bind_back
 tests\P2387R3_pipe_support_for_user_defined_range_adaptors
 tests\P2401R0_conditional_noexcept_for_exchange
 tests\P2404R3_move_only_types_for_comparison_concepts
+tests\P2405R1_nullopt_t_comparable
```
frederick-vs-ja: P2405R1 wasn't the accepted paper. Could we put the test cases into `P0220R1_optional`?

StephanTLavavej: Agreed. In fact, P2405R1 is nonexistent.

stl/inc/optional
StephanTLavavej: Agreed. In general, LWG issues are always treated as defect reports, implemented unconditionally (i.e. down to the oldest applicable Standard level, which would be C++17 for anything involving `optional`).

tests/std/tests/P2405R1_nullopt_t_comparable/env.lst
StephanTLavavej: Agreed. And yes, replacing "strict" with "usual" is desirable here. We use the usual matrices by default, and "strict" only when code is fundamentally incompatible with `/permissive` mode.

tests/std/tests/P2405R1_nullopt_t_comparable/test.cpp
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <cassert>
+#include <compare>
+#include <concepts>
+#include <optional>
+#include <ranges>
```
StephanTLavavej: `<ranges>` isn't necessary for `ranges::find`, but `<algorithm>` is.

tests/std/tests/P0220R1_optional/test.cpp
```diff
@@ -8363,10 +8363,16 @@ int run_test()
...
+#include <algorithm>
 #include <cassert>
 #include <optional>
 #include <type_traits>
 #include <utility>
+#if _HAS_CXX20
+#include <compare>
+#include <concepts>
+#endif // _HAS_CXX20
+#include <vector>
```
StephanTLavavej: Nitpick: Interleaving conditional includes like this interferes with clang-format auto-sorting. There are some pre-existing occurrences but we're trying to avoid this now.

tests/std/tests/P0220R1_optional/test.cpp
```diff
@@ -8576,6 +8582,33 @@ namespace msvc {
...
+        static_assert(std::three_way_comparable<std::nullopt_t, std::strong_ordering>);
+        static_assert(std::is_eq(std::nullopt <=> std::nullopt));
+#endif // _HAS_CXX20
+
+        void run_test() {
+            std::vector<std::optional<int>> v = {1, std::nullopt, 3};
+
+#if _HAS_CXX20
+            auto it = std::ranges::find(v, std::nullopt);
+#else
```
StephanTLavavej: IMO this is unnecessary - classic vs. ranges algorithms are orthogonal to the element/value comparisons, so this is just introducing unnecessary complexity into the test.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:zero: :cat2: :heart_eyes_cat:
===

PR #6230, opened 2026-04-06
Fix CodeQL warnings in Python

StephanTLavavej:
* Fix CodeQL "Unnecessary lambda".
* Fix CodeQL "Empty except".
  + Instead of calling index() and ignoring ValueError, we can guard the logic with `if VALUE in LIST`.
* Suppress CodeQL "Empty except".
  + The warnings say "Except doesn't do anything and has no comment", so we should be able to add comments to suppress the warnings. Unlike C++ CodeQL suppressions, we don't have an opaque ID with a specific suppression syntax to follow, but prefixing the comment with "CodeQL:" will be a good reminder that these comments serve a purpose.
* Attempt to fix CodeQL "Mismatch in multiple assignment".
  + For the line `cmd, out, err, rc = self.runStep(step, litConfig)` this is complaining: "Assigning multiple variables without ensuring that you define a value for each variable causes an exception at runtime." I speculate that this is happening because CodeQL doesn't understand what we're doing with the 3-tuple returned by executeCommand(). I'm hoping that explicitly unpacking it into 3 variables, before composing runStep()'s 4-tuple result, will make it happy.

### DIFF
--- a/tests/utils/stl/util.py
+++ b/tests/utils/stl/util.py
@@ -132,7 +132,9 @@ def killProcessAndChildren(pid):
             try:
                 child.kill()
             except psutil.NoSuchProcess:
+                # CodeQL: Intentionally ignore this exception; if the process has already terminated, that's fine.
                 pass
         psutilProc.kill()
     except psutil.NoSuchProcess:
+        # CodeQL: Intentionally ignore this exception; if the process has already terminated, that's fine.
         pass
--- a/tests/utils/stl/test/format.py
+++ b/tests/utils/stl/test/format.py
@@ -130,7 +130,9 @@ class STLTestFormat:
 
         env = _mergeEnvironments(os.environ, testStep.env)
 
-        return testStep.cmd, *stl.util.executeCommand(testStep.cmd, cwd=testStep.workDir, env=env)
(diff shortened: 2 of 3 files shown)


tests/utils/stl/util.py
```diff
@@ -132,7 +132,9 @@ def killProcessAndChildren(pid):
             try:
                 child.kill()
             except psutil.NoSuchProcess:
+                # CodeQL: Intentionally ignore this exception; if the process has already terminated, that's fine.
                 pass
```
cpplearner, line 136: I recently learned that Python has [`contextlib.suppress`](https://docs.python.org/3/library/contextlib.html#contextlib.suppress) for ignoring specific exceptions.

```suggestion
            with contextlib.suppress(psutil.NoSuchProcess):
                child.kill()
```

(Need to import `contextlib` first, of course.)

StephanTLavavej: Thanks. Nested use of this looks weird, so I'd prefer to keep the current structure.

BillyONeal (approved):

davidmrdavid (approved):

StephanTLavavej:

PR #6235, opened 2026-04-08
Implement LWG-3662 string::append/assign(NTBS, pos, n) should not construct a temporary string

### LINKED ISSUE #6203: LWG-3662 `basic_string::append/assign(NTBS, pos, n)` suboptimal
LWG-3662 `basic_string::append/assign(NTBS, pos, n)` suboptimal

YexuanXiao:
Fixes #6203.


tests/std/tests/VSO_0174871_string_replace/test.cpp
```diff
@@ -142,4 +186,5 @@ int main() {
 
     test_index_boundary_cases();
     test_shrink_to_fit();
+    test_LWG3662();
```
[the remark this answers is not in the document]
frederick-vs-ja: Nope. These overloads should be added to C++14 mode (the earliest mode we support),

stl/inc/xstring
```diff
@@ -1556,6 +1556,16 @@ public:
...
 
+    _CONSTEXPR20 basic_string& append(
+        _In_reads_(_Count) const _Elem* const _Ptr, const size_type _Off, _CRT_GUARDOVERFLOW const size_type _Count) {
+        // append(string_view(_Ptr).substr(_Off, _Count))
+        const size_type _Length = _Convert_size<size_type>(_Traits::length(_Ptr));
+        if (_Off > _Length) {
+            _Scary_val::_Xran();
+        }
+        return _Append(_Ptr + _Off, (_STD min) (_Length - _Off, _Count));
+    }
```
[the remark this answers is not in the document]
StephanTLavavej: Nope, LWG issue resolutions are defect reports to be implemented downlevel/unconditionally.

stl/inc/xstring
```diff
@@ -1556,6 +1556,16 @@ public:
     }
 #endif // _HAS_CXX17
 
+    _CONSTEXPR20 basic_string& append(
+        _In_reads_(_Count) const _Elem* const _Ptr, const size_type _Off, _CRT_GUARDOVERFLOW const size_type _Count) {
```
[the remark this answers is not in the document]
StephanTLavavej: Correct, excellent catch. I'll fix this.

StephanTLavavej: `substr()` itself isn't marked with `_CRT_GUARDOVERFLOW`, so I don't think it's appropriate here.

`_CRT_GUARDOVERFLOW` is for when we're directly allocating memory, where it's possible for an integer overflow to produce a small number, so only a small number of elements are allocated and then indexed with a larger number, producing an overrun.

Here, we're selecting a substring of a given string. In this case, if `_Count` has overflowed before it's passed to this function, we'll end up with a differently-sized substring than the user may have expected (the count is clamped so if it's too big that's no problem either). Using `_CRT_GUARDOVERFLOW` to force integer overflows to the maximum `size_type` (which would be `npos` here) provides no particular benefit.

This was presumably copy-pasted from the overload below, where we append `[_Ptr, _Ptr + _Count)`, which is very different.

tests/std/tests/VSO_0174871_string_replace/test.cpp
```diff
@@ -103,6 +103,50 @@ void test_shrink_to_fit() {
     assert(example == longerStr);
 }
 
+template <typename T = char>
+struct CharAllocator {
```
StephanTLavavej: This allocator isn't doing anything; it's not instrumented to count allocations. And if it were, the string data involved is too short.

I think we don't specifically need to test the "doesn't construct a temporary string" intent of the LWG issue resolution here.

tests/std/tests/VSO_0174871_string_replace/test.cpp
```diff
@@ -103,6 +103,50 @@ void test_shrink_to_fit() {
...
+    T* allocate(std::size_t n) {
+        return new T[n];
+    }
+    void deallocate(T* p, std::size_t) noexcept {
+        delete[] p;
+    }
+};
+
+void test_LWG3662() {
+    // append/assign(NTBS, pos, n) should not construct a temporary string
```
StephanTLavavej: We should say LWG-3662 in the comment for searchability.

tests/std/tests/VSO_0174871_string_replace/test.cpp
```diff
@@ -103,6 +103,50 @@ void test_shrink_to_fit() {
...
+    basic_string<char, char_traits<char>, CharAllocator<char>> s(CharAllocator<char>(0));
+
+    s.append("hello", 1, 3);
+    assert(s == "ell");
```
StephanTLavavej: Because the string starts empty, this doesn't detect the difference between append and assign.

tests/std/tests/VSO_0174871_string_replace/test.cpp
```diff
@@ -103,6 +103,50 @@ void test_shrink_to_fit() {
...
+    s.assign("world", 1, 3);
+    assert(s == "orl");
+
+    s.clear();
```
StephanTLavavej: Only EH tests follow this, so there's no reason to clear.

tests/std/tests/VSO_0174871_string_replace/test.cpp
```diff
@@ -103,6 +103,50 @@ void test_shrink_to_fit() {
...
+
+void test_LWG3662() {
+    // append/assign(NTBS, pos, n) should not construct a temporary string
+    basic_string<char, char_traits<char>, CharAllocator<char>> s(CharAllocator<char>(0));
+
+    s.append("hello", 1, 3);
+    assert(s == "ell");
+    s.assign("world", 1, 3);
+    assert(s == "orl");
+
```
StephanTLavavej, line 115: More things to test:

* `out_of_range` is thrown when the offset is strictly greater than the length. For an equal offset, we should append/assign a zero-length string.
* We should clamp the count to the remaining size of the string.
* We should return the string.

tests/std/tests/VSO_0174871_string_replace/test.cpp
```diff
@@ -103,6 +103,50 @@ void test_shrink_to_fit() {
...
+    T* allocate(std::size_t n) {
+        return new T[n];
+    }
+    void deallocate(T* p, std::size_t) noexcept {
+        delete[] p;
+    }
+};
+
+void test_LWG3662() {
```
StephanTLavavej, line 106: We should also test the `constexpr` part.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:smile_cat: :tada: :rocket:
===

PR #6236, opened 2026-04-08
Implement LWG-4259 P1148R0 changed the return values of searching functions of `std::basic_string` on some platforms

### LINKED ISSUE #6208: LWG-4259 P1148R0 changed the return values of searching functions of `std::basic_string` on some platforms
LWG-4259 WG21-P1148R0 changed the return values of searching functions of `std::basic_string` on some platforms

YexuanXiao:
Fixes #6208.


stl/inc/xstring
```diff
@@ -2512,42 +2512,51 @@ public:
...
     }
 
+    _NODISCARD static _CONSTEXPR20 size_type _Find_result_cast(_STD size_t _Off) {
+        // On 32-bit platforms, basic_string may use 64-bit size_type (depending on the allocator), while
+        // basic_string_view uses 32-bit size_t. This makes basic_string_view::npos smaller than basic_string::npos.
+        if (_Off == _STD size_t(-1)) {
+            return npos;
+        }
+        return static_cast<size_type>(_Off);
+    }
```
frederick-vs-ja: There're some conventions in MSVC STL, IIUC.
- Types shouldn't be namespace-qualified unless necessary. This seems to be conventional in other standard library implementations.
- Parameters should be `const` whenever suitable.
- Casts via brace-init-list or keywords are preferable to C-style casts or casts in equivalent functional notation.

I personally think that this function should be `noexcept`, and it's better to wrap the branch in `if constexpr`. I'm not sure whether `static_cast<size_t>(-1)` is the preferred spelling, though.

```suggestion
    _NODISCARD static _CONSTEXPR20 size_type _Find_result_cast(const size_t _Off) noexcept {
        if constexpr (sizeof(size_type) > sizeof(size_t)) { // only happens on 32-bit platforms with 64-bit size_type
            if (_Off == static_cast<size_t>(-1)) {
                return npos;
            }
        }

        return static_cast<size_type>(_Off);
    }
```

stl/inc/xstring
```diff
@@ -2512,42 +2512,53 @@ public:
...
     }
 
+private:
+    _NODISCARD static _CONSTEXPR20 size_type _Find_result_cast(const size_t _Off) noexcept {
+        if constexpr (sizeof(size_type) > sizeof(size_t)) { // only happens on 32-bit platforms with 64-bit size_type
+            if (_Off == static_cast<size_t>(-1)) {
+                return npos;
+            }
+        }
+        return static_cast<size_type>(_Off);
```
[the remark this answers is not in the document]
StephanTLavavej: In isolation, I would be inclined to disagree with Copilot here. The found location is an offset, so `_Off` is a perfectly normal name for such a parameter. Copilot's stated rationale isn't really persuasive. However, what tips the scales for me is that the callers have a different parameter named `_Off`. It's confusing when a callstack gives parameters with different meanings the same names. I'll apply this suggestion. I'll also cite the latest WP.

tests/std/tests/GH_005546_containers_size_type_cast/test.cpp
```diff
@@ -517,10 +517,87 @@ CONSTEXPR20 bool test() {
...
+
+    WideSizeAllocator() = default;
+
+    template <typename U>
+    constexpr WideSizeAllocator(const WideSizeAllocator<U>&) noexcept {}
+
+    constexpr T* allocate(size_type n) {
+        return allocator<T>{}.allocate(static_cast<size_t>(n));
+    }
+    constexpr void deallocate(T* p, size_type n) noexcept {
```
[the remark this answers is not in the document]
StephanTLavavej: Strong yes for `CONSTEXPR20`. Weak yes for `check_alloc_in_range_of_size_t` (I'm not sure why the test bothered with this paranoia since it doesn't seem to be exercised, but it would be correct for production, so okay). Applying to `allocate()` only, matching `redifference_allocator`. `deallocate()` is `noexcept`, and already has to trust that the given `size_type` matches the allocation.

tests/std/tests/GH_005546_containers_size_type_cast/test.cpp
```diff
@@ -517,10 +517,87 @@ CONSTEXPR20 bool test() {
     return true;
 }
 
-#if _HAS_CXX20
-static_assert(test());
-#endif // _HAS_CXX20
+template <typename T = char>
```
StephanTLavavej: We conventionally use `class`. (Some test code is inconsistent but this file already uses `class`.)

StephanTLavavej: This default template arg isn't used.

tests/std/tests/GH_005546_containers_size_type_cast/test.cpp
```diff
@@ -517,10 +517,87 @@ CONSTEXPR20 bool test() {
...
+};
+
+CONSTEXPR20 bool test_LWG4259() {
+    // On 32-bit platforms, basic_string may use 64-bit size_type (depending on the allocator), while basic_string_view
+    // uses 32-bit size_t. This makes basic_string_view::npos smaller than basic_string::npos.
+    using WideSizeString = basic_string<char, char_traits<char>, WideSizeAllocator<char>>;
+
+    WideSizeString str           = "Hello World";
+    WideSizeString not_found_str = "XYZ";
+    const char* not_found_cstr   = "XYZ";
```
StephanTLavavej: It's just test code, but in a fairly large function, it's nice for things to be as const as possible.

tests/std/tests/GH_005546_containers_size_type_cast/test.cpp
```diff
@@ -517,10 +517,87 @@ CONSTEXPR20 bool test() {
...
+    assert(str.find(not_found_str) == WideSizeString::npos);
+    assert(str.find(not_found_cstr, 0, 3) == WideSizeString::npos);
+    assert(str.find(not_found_cstr) == WideSizeString::npos);
+    assert(str.find('Z') == WideSizeString::npos);
+#if _HAS_CXX17
+    string_view sv = "XYZ";
```
StephanTLavavej: `not_found_sv` would be a better name.

tests/std/tests/GH_005546_containers_size_type_cast/test.cpp
```diff
@@ -517,10 +517,87 @@ CONSTEXPR20 bool test() {
...
+#endif
+    assert(str.find_first_of(not_found_str) == WideSizeString::npos);
+    assert(str.find_first_of(not_found_cstr, 0, 3) == WideSizeString::npos);
+    assert(str.find_first_of(not_found_cstr) == WideSizeString::npos);
+    assert(str.find_first_of('Z') == WideSizeString::npos);
+#if _HAS_CXX17
+    assert(str.find_first_of(sv) == WideSizeString::npos);
+#endif
+    WideSizeString all_in_str = "Hello World!";
+    WideSizeString search_all = "Hello World";
```
StephanTLavavej: `search_all` is identical to `str`.

tests/std/tests/GH_005546_containers_size_type_cast/test.cpp
```diff
@@ -517,10 +517,87 @@ CONSTEXPR20 bool test() {
...
+    assert(search_all.find_first_not_of(all_in_str) == WideSizeString::npos);
+    assert(search_all.find_first_not_of("Hello World!", 0, 12) == WideSizeString::npos);
```
StephanTLavavej: We should extract `"Hello World!"` to a named variable, following how `not_found_cstr` extracts `"XYZ"`.

tests/std/tests/GH_005546_containers_size_type_cast/test.cpp
```diff
@@ -517,10 +517,87 @@ CONSTEXPR20 bool test() {
...
+    assert(search_all.find_first_not_of("Hello World!") == WideSizeString::npos);
+    WideSizeString repeated = "AAAAA";
+    assert(repeated.find_first_not_of('A') == WideSizeString::npos);
+#if _HAS_CXX17
+    string_view sv_all = "Hello World!";
```
StephanTLavavej: `all_in_sv` would be a more consistent name.

tests/std/tests/GH_005546_containers_size_type_cast/test.cpp
StephanTLavavej: It would be better to deduplicate the string content.

tests/std/tests/GH_005546_containers_size_type_cast/test.cpp
```diff
@@ -517,10 +517,87 @@ CONSTEXPR20 bool test() {
...
+CONSTEXPR20 bool test_LWG4259() {
+    // On 32-bit platforms, basic_string may use 64-bit size_type (depending on the allocator), while basic_string_view
+    // uses 32-bit size_t. This makes basic_string_view::npos smaller than basic_string::npos.
+    using WideSizeString = basic_string<char, char_traits<char>, WideSizeAllocator<char>>;
+
+    WideSizeString str           = "Hello World";
+    WideSizeString not_found_str = "XYZ";
+    const char* not_found_cstr   = "XYZ";
+    assert(str.find(not_found_str) == WideSizeString::npos);
+    assert(str.find(not_found_cstr, 0, 3) == WideSizeString::npos);
```
StephanTLavavej: Finally, the numbers 3 and 12 are excessively magical, we should extract them.

tests/std/tests/GH_005546_containers_size_type_cast/test.cpp
```diff
@@ -517,10 +517,87 @@ CONSTEXPR20 bool test() {
...
+    assert(str.find_first_of(not_found_str) == WideSizeString::npos);
+    assert(str.find_first_of(not_found_cstr, 0, 3) == WideSizeString::npos);
+    assert(str.find_first_of(not_found_cstr) == WideSizeString::npos);
+    assert(str.find_first_of('Z') == WideSizeString::npos);
+#if _HAS_CXX17
+    assert(str.find_first_of(sv) == WideSizeString::npos);
+#endif
+    WideSizeString all_in_str = "Hello World!";
+    WideSizeString search_all = "Hello World";
+    assert(search_all.find_first_not_of(all_in_str) == WideSizeString::npos);
```
StephanTLavavej: Following the order find, rfind, find_first_of, find_last_of, find_first_not_of, and find_last_not_of as depicted in the Standard would also help keep similar code together.

StephanTLavavej:
Thanks! :heart_eyes_cat: I pushed a bunch of nitpicky changes.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for implementing this LWG issue resolution, and apologies for how long it took to review this! :joy_cat: :tada: :rocket:

PR #2732, opened 2022-05-19
remove `const` from `<random>`:distribution::`operator()`

fsb4000:
Fixes #1912 

`normal_distribution` didn't have the `const`

https://github.com/microsoft/STL/blob/1a20fe1133d711a647bbb135d98743f91b7be323/stl/inc/random#L2938-L2947

```
$ type main.cpp
#include <random>

int main() {
    std::default_random_engine random;
    const /*<-*/ std::uniform_int_distribution<int> dist;
    return dist(random);
}
```
```
cl /EHsc /W4 /WX /nologo main.cpp
main.cpp
main.cpp(6): error C3848: expression having type 'const std::uniform_int_distribution<int>' would lose some const-volatile qualifiers in order to call 'int std::uniform_int<_Ty>::operator ()<std::default_random_engine>(_Engine &)'
        with
        [
            _Ty=int,
            _Engine=std::default_random_engine
        ]
```
```
cl /EHsc /W4 /WX /nologo /D_ALLOW_RANDOM_DISTRIBUTION_CONST_OPERATOR main.cpp
main.cpp
```

PR #6246, opened 2026-04-10
`<random>`: Remove escape hatch for `const` distribution operators

StephanTLavavej:
This escape hatch was added by #2732 on 2022-05-23 shipping in VS 2022 17.3. We've never needed to give it to any customers, and it hasn't been necessary in our Real World Code test suite. It's time to remove this escape hatch and make the fix unconditional.


BillyONeal (approved):
🤯 about this existing in the first place

StephanTLavavej:

PR #6251, opened 2026-04-13
`<mdspan>`: Require `mdspan`'s element type to be an object type

### LINKED ISSUE #6250: `<mdspan>`: `std::mdspan` should reject reference `ElementType`
# Describe the bug
`std::mdspan` accepts a reference type as its `ElementType`.

# Reproduction
```cpp
#include <mdspan>

struct BadAccessor {
	using offset_policy = BadAccessor;
	using element_type = int&;
	using reference = int&;
	using data_handle_type = int*;

	reference access(data_handle_type p, std::size_t i) const {
		return p[i];
	}
};

int main() {
	std::mdspan<int&, std::dextents<int, 2>, std::layout_right, BadAccessor> m;
}
```

# Expected behavior
According to [[mdspan.mdspan.overview](https://eel.is/c++draft/mdspan.mdspan.overview#2.1)]

>ElementType is a complete object type that is neither an abstract class type nor an array type,

The program should be rejected as ill-formed.

# STL version
202508

# Additional context
libstdc++ and libc++ also accept this.
https://godbolt.org/z/PeP6beWWb

frederick-vs-ja:
Fixes #6250.

There's a similar requirement in [[mdspan.accessor.default.overview]/2](https://eel.is/c++draft/mdspan.accessor.default.overview#2), but we don't need to add another `statis_assert`, because for `default_accessor` we're already requiring both `_ElementType&` and `_ElementType*` to be well-formed.

The reference working draft numbers are somehow inconsistent, but I'd like to harmonize all these numbers after the publishment of the C++26 final working draft.

### DIFF
--- a/stl/inc/mdspan
+++ b/stl/inc/mdspan
@@ -1257,6 +1257,8 @@ private:
     static_assert(
         sizeof(element_type) > 0, "ElementType must be a complete type (N4950 [mdspan.mdspan.overview]/2.1).");
     static_assert(
+        is_object_v<element_type>, "ElementType must be an object type (N5032 [mdspan.mdspan.overview]/2.1).");
+    static_assert(
         !is_abstract_v<element_type>, "ElementType cannot be an abstract type (N4950 [mdspan.mdspan.overview]/2.1).");
     static_assert(
         !is_array_v<element_type>, "ElementType cannot be an array type (N4950 [mdspan.mdspan.overview]/2.1).");


StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:tada: :cat2: :smile_cat:
===

PR #6233, opened 2026-04-07
Implement LWG-3891 LWG-3870 breaks `std::expected<cv T, E>`

frederick-vs-ja:
Fixes #6220.

Remarks:
- The final resolution of LWG-3891 introduced some potential inconsistency - some member functions are detected but not actually called when `value_type` is cv-qualified. The `assert(false);` occurrences correspond to such inconsistency.
- MSVC may fail to activate union member of an empty class type during constant evaluation, and DevCom-11070772 is reported for this. For MSVC, `FakeCvAssignable` is made non-empty for now.

PR #6197, opened 2026-04-01
Toolset update: MSVC Compiler 19.51.36223.2

StephanTLavavej:
Changelog
===
- Code cleanups:
  * Removed compiler bug workarounds.
- Infrastructure improvements:
  * Updated dependencies.
    + Updated MSVC Compiler to 19.51.36223.2 (now required).

Commits
===
* Cycle locations.
* Simplify `VisualStudioUrl`.
* PowerShell 7.6.0.
* MSVC Compiler 19.51.36223.2.
* New pools.
* 2026-02-24: MSVC-PR-711310 fixed VSO-2744645 "Standard Library Header Units: Declaring and defining `constexpr` member functions in different headers fails to compile"
* 2026-02-27: MSVC-PR-689579 fixed VSO-2742607 "C1XX `__reference_constructs_from_temporary` mishandles a couple of cases"
  + Followup: #6196
* 2026-03-13: MSVC-PR-717544 fixed VSO-2826400 "C1XX still mishandles parameter pack size mismatch in `requires`-expressions"
* 2026-03-17: MSVC-PR-718489 fixed VSO-2225134 DevCom-10729775 "Standard Library Modules: Random number machinery emits bogus error C2512: '`std::_Unsigned128`': no appropriate default constructor available"
* 2026-03-18: MSVC-PR-718945 fixed VSO-2226569 "Standard Library Modules: `generate_canonical` causes C1XX ICE"
* Preserve `tuple` workaround for the old MSVC-internal toolset only.

STL-ASan-CI [passed](https://dev.azure.com/vclibs/STL/_build/results?buildId=20566&view=results).

Issue #3567, opened 2023-03-13
Ranges tests consuming excessive compiler memory


StephanTLavavej:
Here's the C1XX compiler memory consumption of all tests named `*ranges*`, as reported by VS 2022 17.6 Preview 1 x64 `/d1reportMemorySummary`. There are 4 tests that exceed 2 GB of RAM, which we should investigate reducing (either by pruning their tested type combinations, or by splitting the test up into multiple parts as we've done elsewhere).

This was brought to my attention by Russell Johnston in MS; while working with EDG to investigate and fix issues for the STL (:heart_eyes_cat:), the top 3 tests here were failing due to out-of-memory errors (and presumably the others in the ~1.8 to 2.2 GB range for C1XX were perilously close to OOM for EDG). This doesn't indicate anything pathological or undesirable about what EDG is doing; we should simply avoid writing such enormously expensive tests. This is problematic even for C1XX, since it means that x86-hosted test runs (internally) run close to OOM, and multithreaded test runs can come close to swapping (e.g. I have "only" 64 GB for 32 cores).

Peak Working<br>Set Size (> 2 GB) | Test
-------------:|:-------------------------
3,177,598,976 | P0896R4_ranges_alg_mismatch
2,631,102,464 | P1659R3_ranges_alg_ends_with
2,321,670,144 | P0896R4_ranges_alg_equal
2,204,807,168 | P1659R3_ranges_alg_starts_with

Peak Working<br>Set Size (< 2 GB) | Test
-------------:|:-------------------------
1,970,331,648 | P0896R4_ranges_alg_replace_copy_if
1,950,568,448 | P0896R4_ranges_alg_replace_copy
1,811,857,408 | P0896R4_ranges_alg_move
1,650,929,664 | P0896R4_ranges_alg_transform_unary
1,642,221,568 | P0896R4_ranges_alg_copy_if
1,512,505,344 | P0896R4_ranges_alg_copy
1,423,818,752 | P0896R4_ranges_alg_find_end
1,387,089,920 | P0896R4_ranges_alg_rotate_copy
1,235,529,728 | P0896R4_ranges_alg_copy_backward
1,184,952,320 | P0896R4_ranges_alg_move_backward
1,157,095,424 | P0896R4_ranges_alg_reverse_copy

<details>
<summary>Click to expand smaller (< 1 GB) tests:</summary>

Peak Working<br>Set Size (< 1 GB) | Test
-------------:|:-------------------------
[...]

PR #6260, opened 2026-04-17
Toolset update: MSVC Compiler 19.52.36307.1

StephanTLavavej:
* Got 1024 cores of Dpsv6 in uksouth, add comments.
* VS-PR-713550 on 2026-03-03 added `-Preview -NoLogo` to `Launch-VsDevShell.ps1`.
* Cycle locations.
* Remove TRANSITION comment about DefaultOutboundAccess.
  + To be safe, I'm going to keep passing `-DefaultOutboundAccess $false`.
* Python 3.14.4 (now required).
* CUDA 13.2.1.
  + Intentionally not increasing the requirement for patch versions.
* MSVC Compiler 19.52.36307.1 (now required).
  + Still can't update yvals_core.h due to the internal toolset.
* Windows 11 SDK 10.0.28000 (now required for STL development, but not usage).
  + Verified that warnings still repro.
* VS-PR-713693 on 2026-03-03 fixed DevCom-11012299 "EDG crashes when handling placement-new".
* VS-PR-713693 on 2026-03-03 fixed VSO-2581623 "EDG should implement `__builtin_is_implicit_lifetime`".
* libcxx/expected_results.txt: Cleanup an unnecessary MSVC-internal skip.
  + Followup to #6197. Any mention of a test as FAIL or SKIPPED for any configuration causes the MSVC-internal harness to skip it for all configurations.
* Some constexpr FAILs are now passing.
* Extend Clang workaround to EDG.
  + Reported VSO-2941370 "EDG `__builtin_is_implicit_lifetime` isn't checking for trivial constructors".
* New pools.
* MSVC-PR-720628 on 2026-03-26 fixed DevCom-11070772 "MSVC fails to activate a union member of certain empty class type during constant evaluation".
  + This workaround was very recently added by #6233.
* Work around VSO-2946529 "`/analyze` emits warning C28285 for `_BitScanForward` declared by winnt.h".
  + This has already been fixed by [OS-PR-14602802](https://microsoft.visualstudio.com/DefaultCollection/OS/_git/os.2020/pullrequest/14602802) on 2026-01-22, but I'm not sure when it'll ship (i.e. in a patch release of 10.0.28000.x, or in a whole new version which would take much longer).
* Work around VSO-1639191 "Excessive memory consumption with `/analyze` targeting x86".
  + Fine-grained workarounds don't appear to be viable, so I'm disabling the entire tests. Because this is specific to x86 `/analyze`, we still have reasonable coverage of x86 vanilla and non-x86 `/analyze`.
* Work around #3567.
  + The tests for `ranges::replace_copy` and `ranges::replace_copy_if` are now OOMing for x86 `/permissive` configurations. Add a targeted workaround for this. Note that we still have reasonable coverage of x86 strict and non-x86 `/permissive`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ Just try to follow these rules, so we can spend more time fixing bugs and implem
     - "MSVC Build Tools for x64/x86 (Preview)"
     - "C++ CMake tools for Windows"
     - "MSVC AddressSanitizer"
-    - "Windows 11 SDK (10.0.26100)" or later
+    - "Windows 11 SDK (10.0.28000)" or later
     - "C++ Clang tools for Windows (20.1.8 - x64/x86)"
     - *Optional, see Note 2 below:* "MSVC Build Tools for ARM64/ARM64EC (Preview)"
 * Install [Python][] 3.14.4 or later.
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ Just try to follow these rules, so we can spend more time fixing bugs and implem
     - "Windows 11 SDK (10.0.26100)" or later
     - "C++ Clang tools for Windows (20.1.8 - x64/x86)"
     - *Optional, see Note 2 below:* "MSVC Build Tools for ARM64/ARM64EC (Preview)"
-* Install [Python][] 3.14.3 or later.
+* Install [Python][] 3.14.4 or later.
   + Select "Add python.exe to PATH" if you want to follow the instructions below that invoke `python`.
     Otherwise, you should be familiar with alternative methods.
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,12 +13,12 @@ endif()
 include(CheckCXXSourceCompiles)
 check_cxx_source_compiles([=[
 #include <sdkddkver.h>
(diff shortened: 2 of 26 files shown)


README.md
```diff
@@ -151,7 +151,7 @@ Just try to follow these rules, so we can spend more time fixing bugs and implem
     - "Windows 11 SDK (10.0.26100)" or later
     - "C++ Clang tools for Windows (20.1.8 - x64/x86)"
     - *Optional, see Note 2 below:* "MSVC Build Tools for ARM64/ARM64EC (Preview)"
-* Install [Python][] 3.14.3 or later.
+* Install [Python][] 3.14.4 or later.
```
davidmrdavid, line 154: just an observation: **whoa, I just l learned about this 'deferred link' (`[something][]`) syntax. Very cool!**

davidmrdavid (approved):
👍 :shipit:

README.md
StephanTLavavej: Yeah, I use it all the time in my mirror commit messages too (which you can see in MSVC-internal history). Just plain `[something]` also gets picked up, but I don't like that form, it's too implicit.

PR #6261, opened 2026-04-18
`<regex>`: Clean up fullmatch flag handling

muellerj2:
This PR performs some minor cleanups:

* `_Regex_match1()` took a flag `_Full`, but all call sites pass `true`, so we can just remove this argument. (I renamed the function even though it's not necessary, strictly speaking; there will be more changes to its signature.)
* `_Matcher3::_Match` took an argument `_Full_match`, but the value is the same for all calls on the same instance. Let's move this argument to the constructor and remove it from `_Match2()` (rename also not strictly necessary).
* Drive-by change: Add more `const` to updated function definitions.
* Drive-by change: The call sites affected by the other changes are updated to make `_STD`-qualified calls.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -2063,9 +2063,9 @@ public:
     template <size_t _Stack_storage_size>
     _Matcher3(_It _Pfirst, _It _Plast, const _RxTraits& _Tr, _Root_node* const _Re, const unsigned int _Nx,
         const regex_constants::syntax_option_type _Sf, const regex_constants::match_flag_type _Mf,
-        unsigned char (&_Stack_storage)[_Stack_storage_size], const bool _Full)
+        unsigned char (&_Stack_storage)[_Stack_storage_size], const bool _Full_)
         : _Traits(_Tr), _Begin(_Pfirst), _End(_Plast), _Sflags(_Sf), _Mflags(_Mf), _Ncap(_Nx),
-          _Longest((_Re->_Flags & _Fl_longest) && !(_Mf & regex_constants::match_any)), _Full(_Full) {
+          _Longest((_Re->_Flags & _Fl_longest) && !(_Mf & regex_constants::match_any)), _Full(_Full_) {
         _Adl_verify_range(_Pfirst, _Plast);
         if (_Re->_Flags & _Fl_begin_needs_w) {
             _Char_class_w = _Lookup_char_class(static_cast<_Elem>('W'));
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -2061,11 +2061,11 @@ private:
 
 public:
     template <size_t _Stack_storage_size>
-    _Matcher3(_It _Pfirst, _It _Plast, const _RxTraits& _Tr, _Root_node* _Re, unsigned int _Nx,
-        regex_constants::syntax_option_type _Sf, regex_constants::match_flag_type _Mf,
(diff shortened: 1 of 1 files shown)


stl/inc/regex
```diff
@@ -2061,11 +2061,11 @@ private:
...
     template <size_t _Stack_storage_size>
-    _Matcher3(_It _Pfirst, _It _Plast, const _RxTraits& _Tr, _Root_node* _Re, unsigned int _Nx,
-        regex_constants::syntax_option_type _Sf, regex_constants::match_flag_type _Mf,
-        unsigned char (&_Stack_storage)[_Stack_storage_size])
+    _Matcher3(_It _Pfirst, _It _Plast, const _RxTraits& _Tr, _Root_node* const _Re, const unsigned int _Nx,
+        const regex_constants::syntax_option_type _Sf, const regex_constants::match_flag_type _Mf,
+        unsigned char (&_Stack_storage)[_Stack_storage_size], const bool _Full)
         : _Traits(_Tr), _Begin(_Pfirst), _End(_Plast), _Sflags(_Sf), _Mflags(_Mf), _Ncap(_Nx),
-          _Longest((_Re->_Flags & _Fl_longest) && !(_Mf & regex_constants::match_any)) {
+          _Longest((_Re->_Flags & _Fl_longest) && !(_Mf & regex_constants::match_any)), _Full(_Full) {
```
StephanTLavavej: Style nitpick: We conventionally avoid shadowing between parameters and data members.

stl/inc/regex
```diff
@@ -2061,11 +2061,11 @@ private:
 
 public:
     template <size_t _Stack_storage_size>
-    _Matcher3(_It _Pfirst, _It _Plast, const _RxTraits& _Tr, _Root_node* _Re, unsigned int _Nx,
-        regex_constants::syntax_option_type _Sf, regex_constants::match_flag_type _Mf,
-        unsigned char (&_Stack_storage)[_Stack_storage_size])
+    _Matcher3(_It _Pfirst, _It _Plast, const _RxTraits& _Tr, _Root_node* const _Re, const unsigned int _Nx,
+        const regex_constants::syntax_option_type _Sf, const regex_constants::match_flag_type _Mf,
+        unsigned char (&_Stack_storage)[_Stack_storage_size], const bool _Full)
```
StephanTLavavej: No change requested: This is a coordinated change between member functions which is a hidden ABI risk, because you need the new constructor (which sets `_Full`) to be paired with calls to `_Match2()` (which assumes that `_Full` has been set). Here, we're ABI-safe because the signature change in the constructor produces a different mangled name, which is sufficient.

As you mentioned, the `_Match2()` rename wasn't strictly necessary because it also underwent a signature change, but I absolutely agree that the rename is desirable. For the constructor, of course we can't rename it. I don't believe there's any risk of this `bool` parameter disappearing during further refactoring (how would the data member get set?) so there's no need to rename the whole matcher class again.

stl/inc/regex
StephanTLavavej: No change requested: Because `_Full` doesn't change for the lifetime of the matcher, we could actually make it a template parameter and remove the data member. That would incur a class rename, and the gains would appear to be minimal.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:pirate_flag: :broom: :white_flag:
===

PR #6262, opened 2026-04-18
`<regex>`: Remove match mode `_Skip_zero_length`

muellerj2:
This removes a strange internal match mode `_Skip_zero_length`. Its only purpose was to shift the start of the pattern search one position later, but to set `prefix().first` in the `match_results` object as if there were no such shift. This match mode was only used to implement the increment operator of `regex_iterator`.

Moreover, `_Regex_search2` has an additional parameter `_Org` which is only used to set the member `_Org` in the `match_results` object.

This PR  removes this internal match mode `_Skip_zero_length` and replaces it in `regex_iterator` by using another internal match mode, `_Match_not_null`. This sometimes reduces the number of constructed matcher objects in the zero-length match case, which can improve performance by avoiding some unnecessary allocations. But the fusion of two `regex_search` calls means that the implementation deviates a bit further from the specification in [\[re.regiter.incr\]](https://eel.is/c++draft/re.regiter.incr): This necessitates some additional logic to emulate the `match_prev_avail` update to the match flags, because the original spec allows the caller to differentiate in some cases whether the first or the second call to `regex_search()` succeeded when incrementing from an initial zero-length match. The new tests make sure that this alternate logic behaves like the spec. (We do not have to emulate this logic strictly if the regex search fails, because the value of the match flags member is irrelevant for the end-of-sequence iterator.)

Moreover, this PR moves the update of the `match_results` field `_Org` outside `_Regex_search2/3`. Instead, the callers have to set this member correctly now. This allows the removal of the `_Org` parameter from `_Regex_search2/3`. Note that the callers were already responsible for choosing the value of `_Org`; they just didn't assign it themselves, but instead passed it to `_Regex_search` which performed the assignment (and did nothing else with this argument).

Drive-by change: `_STD`-qualify a few more calls.

Drive-by bugfix: I think `regex_iterator::operator++()` failed to call `_Adopt(nullptr)` in debug mode in one case when no more match was found.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -143,7 +143,7 @@ namespace regex_constants {
         format_no_copy    = 0x0800,
         format_first_only = 0x1000,
         _Match_not_null   = 0x2000,
-        _Skip_zero_length = 0x4000,
+        // TRANSITION, ABI: Flag 0x4000 was used by the matcher in the past
     };
 
     _BITMASK_OPS(_EXPORT_STD, match_flag_type)
@@ -2745,8 +2745,8 @@ _NODISCARD bool regex_match(const basic_string<_Elem, _StTraits, _StAlloc>& _Str
 }
 
 template <class _BidIt, class _Alloc, class _Elem, class _RxTraits, class _It>
-bool _Regex_search2(_It _First, _It _Last, match_results<_BidIt, _Alloc>* _Matches,
-    const basic_regex<_Elem, _RxTraits>& _Re, regex_constants::match_flag_type _Flgs, _It _Org) {
+bool _Regex_search3(_It _First, _It _Last, match_results<_BidIt, _Alloc>* _Matches,
+    const basic_regex<_Elem, _RxTraits>& _Re, regex_constants::match_flag_type _Flgs) {
     // search for regular expression match in target text
     static_assert(_Is_ranges_bidi_iter_v<_It>,
         "regex_search requires bidirectional iterators or stronger. See N5014 [re.alg.search]/1.");
@@ -2760,11 +2760,7 @@ bool _Regex_search2(_It _First, _It _Last, match_results<_BidIt, _Alloc>* _Match
         return false;
     }
 
-    bool _Found      = false;
-    const _It _Begin = _First;
-    if ((_Flgs & regex_constants::_Skip_zero_length) && _First != _Last) {
-        ++_First;
-    }
+    bool _Found = false;
 
     alignas(_Loop_vals_v3_t<_Iter_diff_t<_It>>) alignas(_Rx_capture_range_t<_It>) //
         alignas(_Rx_state_frame_t<_It>) unsigned char _Stackbuf[4096];
@@ -2790,8 +2786,6 @@ bool _Regex_search2(_It _First, _It _Last, match_results<_BidIt, _Alloc>* _Match
 
     if (_Found && _Matches) { // update _Matches
         _Mx._Copy_captures(*_Matches);
-        _Matches->_Org           = _Org;
-        _Matches->_Pfx().first   = _Begin;
         _Matches->_Pfx().matched = _Matches->_Pfx().first != _Matches->_Pfx().second;
     }
     return _Found;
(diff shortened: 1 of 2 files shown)


stl/inc/regex
```diff
@@ -2868,8 +2866,14 @@ _OutIt _Regex_replace1(_OutIt _Result, _BidIt _First, _BidIt _Last, const basic_
...
-    while (
-        _Regex_search2(_Pos, _Last, _STD addressof(_Matches), _Re, _Flags | _Not_null, _Pos)) { // replace at each match
+    for (;;) {
+        _Matches._Org         = _Pos;
+        _Matches._Pfx().first = _Pos;
+        if (!_STD _Regex_search3(_Pos, _Last, _STD addressof(_Matches), _Re, _Flags | _Not_null)) {
+            break;
+        }
+
+        // replace at each match
```
muellerj2: Maybe we should just call `_STD regex_search()` instead. Then we can retain the previous `while` loop structure at the cost of an unnecessary `_Adl_verify_range()` call (and unnecessary `_Get_unwrapped()` calls in the future).

stl/inc/regex
```diff
@@ -3026,43 +3032,43 @@ public:
         _STL_VERIFY(_MyRe, "regex_iterator not incrementable");
 #endif // _ITERATOR_DEBUG_LEVEL != 0
 
-        _BidIt _Start = _MyVal._At(0).second;
+        _BidIt _Start       = _MyVal._At(0).second;
+        _MyVal._Pfx().first = _Start;
```
muellerj2: Note that we don't have to set `_MyVal._Org` here, because the constructor already did this. This is even the case when an old version of the constructor gets picked by the linker.

stl/inc/regex
StephanTLavavej: I'd be fine with that if it simplifies the code, particularly if we've already unwrapped before calling `_Regex_replace1()`.

muellerj2: I looked at it more closely and calling `regex_search()` instead of `_Regex_search3()` does have one major disadvantage: The internal match mode `_Match_not_null` (not to be confused with the Standard-mandated match mode `match_not_null`) gets baked into the ABI.

I guess the other alternative is to define an overload of `_Regex_search3()` that essentially looks like `regex_search()`.

muellerj2: Ah, but the internal match mode `_Match_not_null` might be a way to achieve the same result more elegantly in the first place. Give me some time to try that.

muellerj2: That worked reasonably well. Bonus: We avoid constructing two matcher objects in `regex_iterator::operator++()`.

But as a result, the implementation of the algorithm deviates even a bit more from the specification than the original implementation. But I think this is worth it, as this can save some allocations during matching.

muellerj2: PR description adjusted.

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -2600,6 +2600,33 @@ void test_gh_6249() {
...
 
+void test_gh_6262() {
+    // GH-6262: Remove match mode _Skip_zero_length
+    // Check that replacement with internal match mode _Match_not_null in regex_iterator behaves equivalently
+
+    const regex re{"^|b"};
+    const char* input = "aab";
+
+    {
+        regex_iterator<const char*> it1{input + 2, input + 3, re}, it2{input + 2, input + 3, re, match_prev_avail};
```
StephanTLavavej: We conventionally avoid defining multiple variables on a single line.

StephanTLavavej: Minor style nitpick: We could use the `cregex_iterator` typedef, as this test does above.

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -2600,6 +2600,33 @@ void test_gh_6249() {
...
+
+    {
+        regex_iterator<const char*> it1{input + 1, input + 3, re}, it2{input + 1, input + 3, re, match_prev_avail};
+        ++it1;
+        assert(it1 == it2);
+    }
+
+    const regex re2{"^a|b"};
+    {
+        regex_iterator<const char*> it1{input + 1, input + 3, re}, it2{input + 1, input + 3, re, match_prev_avail};
```
StephanTLavavej: Bug: `re2` is unused. (This is a good example of why tightly controlling variable scope is useful in tests.)

tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -2600,6 +2600,33 @@ void test_gh_6249() {
     }
 }
 
+void test_gh_6262() {
+    // GH-6262: Remove match mode _Skip_zero_length
+    // Check that replacement with internal match mode _Match_not_null in regex_iterator behaves equivalently
+
+    const regex re{"^|b"};
+    const char* input = "aab";
```
StephanTLavavej: `input` could be top-level const.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:fire: :zero: :fire_extinguisher:
===

PR #6264, opened 2026-04-28
Fix excessive comparisons in minmax algorithms

mirion-dev:
According to [[alg.min.max#32]](https://eel.is/c++draft/alg.min.max#32), the algorithms should perform at most `max(floor(3/2 (N−1)),0)` comparisons, but the current implementations of `std::{ranges::}minmax_element` violate this requirement when `N` is even. This can be reproduced with the added test cases in this PR.

Although `std::ranges::minmax` does not violate it because its upper bound is `(3/2) N` [[alg.min.max#23]](https://eel.is/c++draft/alg.min.max#23), I have optimized it as well.

### DIFF
--- a/tests/std/tests/Dev11_0532622_minmax_element/test.cpp
+++ b/tests/std/tests/Dev11_0532622_minmax_element/test.cpp
@@ -107,5 +107,5 @@ int main() {
     test_cmp_count({1, 2, 3, 4});
     test_cmp_count({3, 2, 1});
     test_cmp_count({4, 3, 2, 1});
-#endif
+#endif // _ITERATOR_DEBUG_LEVEL < 2
 }
--- a/tests/std/tests/Dev11_0532622_minmax_element/test.cpp
+++ b/tests/std/tests/Dev11_0532622_minmax_element/test.cpp
@@ -37,6 +37,9 @@ void test_all_permutations(vector<int>& v) {
     } while (next_permutation(v.begin(), v.end()));
 }
 
+// N5032 [alg.min.max]/32:
+// Complexity: Let N be last - first. At most max(floor(3/2 (N − 1)), 0) comparisons
+// and twice as many applications of the projection, if any.
 void test_cmp_count(initializer_list<int> v) {
     size_t cmp_count = 0;
     (void) minmax_element(v.begin(), v.end(), [&cmp_count](int left, int right) {
--- a/tests/std/tests/Dev11_0532622_minmax_element/test.cpp
+++ b/tests/std/tests/Dev11_0532622_minmax_element/test.cpp
@@ -37,7 +37,7 @@ void test_all_permutations(vector<int>& v) {
     } while (next_permutation(v.begin(), v.end()));
 }
 
-void test_cmp_count(std::initializer_list<int> v) {
+void test_cmp_count(initializer_list<int> v) {
     size_t count = 0;
     auto _       = minmax_element(v.begin(), v.end(), [&count](int left, int right) {
         ++count;
--- a/tests/std/tests/Dev11_0532622_minmax_element/test.cpp
+++ b/tests/std/tests/Dev11_0532622_minmax_element/test.cpp
@@ -38,12 +38,12 @@ void test_all_permutations(vector<int>& v) {
 }
 
 void test_cmp_count(initializer_list<int> v) {
-    size_t count = 0;
-    auto _       = minmax_element(v.begin(), v.end(), [&count](int left, int right) {
-        ++count;
+    size_t cmp_count = 0;
+    (void) minmax_element(v.begin(), v.end(), [&cmp_count](int left, int right) {
+        ++cmp_count;
         return left < right;
     });
-    assert(count <= 3 * (v.size() - 1) / 2);
+    assert(cmp_count <= 3 * (v.size() - 1) / 2);
 }
 
 int main() {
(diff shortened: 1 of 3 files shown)


mirion-dev:
It's weird that the failed test `P0811R3_midpoint_lerp` works on my machine and contains no `minmax`.

StephanTLavavej:
@zacklj89 had also encountered `lerp(1, 2, 2) == 3; expected 3` on ARM64 internally. I think this test is now flaky and I'll need to separately add workarounds to it.

tests/std/tests/Dev11_0532622_minmax_element/test.cpp
```diff
@@ -37,6 +37,15 @@ void test_all_permutations(vector<int>& v) {
     } while (next_permutation(v.begin(), v.end()));
 }
 
+void test_cmp_count(std::initializer_list<int> v) {
```
StephanTLavavej: This test is already `using namespace std;`.

tests/std/tests/Dev11_0532622_minmax_element/test.cpp
```diff
@@ -37,6 +37,15 @@ void test_all_permutations(vector<int>& v) {
     } while (next_permutation(v.begin(), v.end()));
 }
 
+void test_cmp_count(std::initializer_list<int> v) {
+    size_t count = 0;
```
StephanTLavavej: Shadows `std::count()`, should use another name.

tests/std/tests/Dev11_0532622_minmax_element/test.cpp
```diff
@@ -37,6 +37,15 @@ void test_all_permutations(vector<int>& v) {
     } while (next_permutation(v.begin(), v.end()));
 }
 
+void test_cmp_count(std::initializer_list<int> v) {
+    size_t count = 0;
+    auto _       = minmax_element(v.begin(), v.end(), [&count](int left, int right) {
```
StephanTLavavej: We conventionally `(void)` to ignore return values, instead of introducing an underscore variable.

tests/std/tests/Dev11_0532622_minmax_element/test.cpp
```diff
@@ -92,4 +101,11 @@ int main() {
         assert(max_element(begin(data), end(data)) == begin(data) + 3);
         assert(minmax_element(begin(data), end(data)) == make_pair(begin(data) + 9, begin(data) + 16));
     }
+
+#if _ITERATOR_DEBUG_LEVEL < 2 // _DEBUG_LT_PRED affects comparison count
+    test_cmp_count({1, 2, 3});
+    test_cmp_count({1, 2, 3, 4});
+    test_cmp_count({3, 2, 1});
+    test_cmp_count({4, 3, 2, 1});
+#endif
```
StephanTLavavej: We conventionally comment `#endif` directives when they're more than 1-2 lines away.

tests/std/tests/P0896R4_ranges_alg_minmax/test.cpp
```diff
@@ -339,6 +339,81 @@ constexpr void mm_constexpr_tests() {
...
+    constexpr iterator begin() const {
+        return {ptr};
+    }
+
+    constexpr iterator end() const {
+        return {ptr + size};
+    }
+};
+
+constexpr void test_cmp_count(std::initializer_list<int> v) {
```
StephanTLavavej: Same comments for this file: `using namespace std;`, `(void)` cast, shadowing, etc.

tests/std/tests/Dev11_0532622_minmax_element/test.cpp
```diff
@@ -37,6 +37,15 @@ void test_all_permutations(vector<int>& v) {
     } while (next_permutation(v.begin(), v.end()));
 }
 
+void test_cmp_count(std::initializer_list<int> v) {
+    size_t count = 0;
+    auto _       = minmax_element(v.begin(), v.end(), [&count](int left, int right) {
+        ++count;
+        return left < right;
+    });
+    assert(count <= 3 * (v.size() - 1) / 2);
```
StephanTLavavej: Can you please cite the Standardese for these assertions, particularly when they vary between N and N - 1? See other citations for the expected form (including the WP number).

StephanTLavavej (changes requested):

tests/std/tests/Dev11_0532622_minmax_element/test.cpp
```diff
@@ -92,4 +104,11 @@ int main() {
         assert(max_element(begin(data), end(data)) == begin(data) + 3);
         assert(minmax_element(begin(data), end(data)) == make_pair(begin(data) + 9, begin(data) + 16));
     }
+
+#if _ITERATOR_DEBUG_LEVEL < 2 // _DEBUG_LT_PRED affects comparison count
```
frederick-vs-ja: Do we need this guard? IIUC `assert(cmp_count <= 3 * (v.size() - 1) / 2);` should always pass.

If we can't remove it at this moment, let's mention #1006 in the comment (perhaps in form `// TRANSITION, GH-1006, [...]`).

mirion-dev: Yes. I have verified locally that this test fails without the guard.

mirion-dev:
Failed on `P0811R3_midpoint_lerp` again.

StephanTLavavej:
Thanks (I will deal with the midpoint lerp failures, somehow).

PR #6265, opened 2026-04-28
Update llvm-project

StephanTLavavej:
* Update llvm-project.
  + llvm/llvm-project#188613
  + llvm/llvm-project#194707
* `left_folds.pass.cpp` was renamed to `ranges.fold_left.pass.cpp`.
  + llvm/llvm-project#180214
* `numeric.ops.sat` tests were renamed.
  + llvm/llvm-project#189574
* Categorize `range.stride.view` failures.

### DIFF
--- a/tests/libcxx/expected_results.txt
+++ b/tests/libcxx/expected_results.txt
@@ -1161,7 +1161,7 @@ std/algorithms/alg.nonmodifying/alg.equal/equal.pass.cpp:9 SKIPPED
 std/algorithms/alg.nonmodifying/alg.find.last/ranges.find_last.pass.cpp:9 SKIPPED
 std/algorithms/alg.nonmodifying/alg.find/find.pass.cpp:9 SKIPPED
 std/algorithms/alg.nonmodifying/alg.find/ranges.find.pass.cpp:9 SKIPPED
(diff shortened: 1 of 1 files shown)


davidmrdavid (approved):

StephanTLavavej:

tests/libcxx/expected_results.txt
```diff
@@ -1151,6 +1155,10 @@ std/utilities/meta/meta.unary/meta.unary.prop/reference_constructs_from_temporar
 std/strings/basic.string/string.modifiers/string_replace/size_size_string_size_size.pass.cpp:0 SKIPPED
 std/strings/basic.string/string.modifiers/string_replace/size_size_string_size_size.pass.cpp:1 SKIPPED
 
+# Not analyzed.
+# MSVC error C2036: '_Sent': unknown size
+# Clang error: arithmetic on a pointer to void
+std/ranges/range.adaptors/range.stride.view/end.pass.cpp FAIL
```
frederick-vs-ja, line 1184: This is probably a bug in MSVC STL.

https://github.com/microsoft/STL/blob/fe3b0ac933ceb088a428cc77829cde8a95f72860/stl/inc/xutility#L2913-L2914

That test case uses `void*` as the sentinel type, so we can't perform `_Se + 0` for decay. IIUC `+_Se` works. Perhaps we can fix this in a later PR.

PR #6266, opened 2026-04-29
`P0811R3_midpoint_lerp`: Avoid flaky ARM64/ARM64EC failures

StephanTLavavej:
This extends the workaround for #5685, originally introduced by #5687, to cover all occurrences of `check_feexcept()`.

We've been seeing flaky ARM64/ARM64EC failures, both internally and externally, always of the form `lerp(1, 2, 2) == 3; expected 3`. This indicates that the value bits are correct, but something's wrong with the floating-point exception state. This only started happening recently, without having messed with `lerp()`, so I speculate that something in the compiler changed.


StephanTLavavej:

zacklj89 (approved):

BillyONeal (approved):
This approval is contingent on a bug being filed with the compiler folks.

StephanTLavavej:
We have a tracking issue in the STL repo so we won't forget, but we don't yet have a reduced repro to send to the compiler team. I agree that should be the next step in this saga, but it cannot block resolving these flaky failures both internally and externally.

PR #6268, opened 2026-05-04
Implement LWG-3599 The `const` overload of `lazy_split_view::begin` should be constrained by `const Pattern`

### LINKED ISSUE #6202: LWG-3599 The `const` overload of `lazy_split_view::begin` should be constrained by `const Pattern`
LWG-3599 The `const` overload of `lazy_split_view::begin` should be constrained by `const Pattern`

frederick-vs-ja:
Fixes #6202.

Blocked libcxx test(s):
- `std/ranges/range.adaptors/range.lazy.split/begin.pass.cpp`

### DIFF
--- a/tests/std/tests/P0896R4_views_lazy_split/test.cpp
+++ b/tests/std/tests/P0896R4_views_lazy_split/test.cpp
@@ -351,6 +351,22 @@ constexpr bool instantiation_test() {
     return true;
 }
 
+constexpr bool test_lwg_3599() {
+    auto p = views::iota(0) | views::take(1) | views::reverse;
+    auto r = views::single(42) | views::lazy_split(p);
+    auto f = r.front();
+
+    static_assert(ranges::forward_range<remove_cvref_t<decltype(r)>>);
+    static_assert(!ranges::forward_range<const remove_cvref_t<decltype(r)>>);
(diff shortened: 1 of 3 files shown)


tests/std/tests/P0896R4_views_lazy_split/test.cpp
```diff
@@ -351,6 +351,22 @@ constexpr bool instantiation_test() {
...
 }
 
+constexpr bool test_lwg_3599() {
+    auto p = views::iota(0) | views::take(1) | views::reverse;
+    auto r = views::single(42) | views::lazy_split(p);
+    auto f = r.front();
+
+    static_assert(ranges::forward_range<remove_cvref_t<decltype(r)>>);
+    static_assert(!ranges::forward_range<const remove_cvref_t<decltype(r)>>);
+    static_assert(ranges::forward_range<remove_cvref_t<decltype(f)>>);
```
StephanTLavavej, line 361: No change requested: The `remove_cvref_t`s aren't doing anything because the variables are plain `auto`, but I suppose this makes it very clear that we don't want any cvref qualifiers.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for improving this lazy kitty! :cat: :zzz: :cat2:

PR #6289, opened 2026-05-17
`<regex>`: Emit only complete `_N_str` nodes during NFA construction

muellerj3:
This is a side quest to #5962: If the parser no longer generates `_N_group` nodes, the current node in the NFA is the node just before the group when parsing of the group contents begins. This might be an `_N_str` node and if the groups starts with some character sequence, the current logic will attach these characters to the `_N_str` node just before the group. This would effectively move these characters outside of the non-capturing group without the `_N_group` node, thus miscompiling the regex.

For this reason, this PR changes this character sequence pooling logic a bit: Characters now get added to a string `_Chars` internal to the `_Builder3` class, until the following subexpression in the regex results in the attachment of some other kind of node to the NFA or the parser must remember the current position in the NFA. At that point, the builder will emit the string node and then append the other kind of node to the string node or return a pointer to the string node. This means that string nodes are no longer modified after they have been generated.

Another nice bonus for the future: The parser does not really need the geometric growth logic inside the `_Buf` class anymore for `_N_str` nodes (although it is still used for now). I plan to rework the parsing logic for character classes as well and intend to remove all of that resize logic inside the `_Buf` class at that time.

The `_Builder2` and `_Parser2` classes are renamed because the layout of the builder class changed, so these classes must be renamed to avoid an ABI break. The version numbers of `_Parser3::_Alternative()` and `_Builder3::_Else_if()` are removed because the class renaming renders them unnecessary.


muellerj3:

stl/inc/regex
```diff
@@ -3588,20 +3590,22 @@ void _Builder2<_FwdIt, _Elem, _RxTraits>::_Add_equiv(const _Elem* const _First,
...
     // add collation element to bracket expression
     _Node_class<_Elem, _RxTraits>* _Node = static_cast<_Node_class<_Elem, _RxTraits>*>(_Current);
     _Sequence<_Elem>** _Cur              = _STD addressof(_Node->_Coll);
     _Char_to_elts(_First, _Last, _Cur);
 }
 
 template <class _FwdIt, class _Elem, class _RxTraits>
-_Node_base* _Builder2<_FwdIt, _Elem, _RxTraits>::_Begin_group() { // add group node
+_Node_base* _Builder3<_FwdIt, _Elem, _RxTraits>::_Begin_group() { // add group node
+    _Emit_str_node();
```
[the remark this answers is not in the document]
muellerj3: Good catch.

stl/inc/regex
```diff
@@ -1637,10 +1633,11 @@ private:
     _Node_base* _Current;
     regex_constants::syntax_option_type _Flags;
     const _RxTraits& _Traits;
+    vector<_Elem> _Chars;
```
muellerj2: In hindsight, a string is the more appropriate container here.

StephanTLavavej:
Updated the title to avoid ambiguity about the meaning of "only".

stl/inc/regex
StephanTLavavej: `_Begin_assert_group2` and `_End_assert_group2` should also receive a suffix-ectomy.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for the ongoing regex cleanups, and sorry for the delay in merging your PR - it's been a wacky couple of months! :joy_cat: :robot: :tada:

PR #6271, opened 2026-05-04
`<yvals_core.h>`: Update `_MSVC_STL_UPDATE`

clin1234:
Bump per https://github.com/microsoft/STL/wiki/Macro-_MSVC_STL_UPDATE

PR #6193, opened 2026-03-30
Bump `_MSVC_STL_UPDATE`

clin1234:
Bump per https://github.com/microsoft/STL/wiki/Macro-_MSVC_STL_UPDATE

PR #6301, opened 2026-06-01
Bump _MSVC_STL_UPDATE

clin1234:


StephanTLavavej (approved):

StephanTLavavej:
Thanks! I'll add this to our next merge batch (it may be in a week or two, I am extremely overloaded with other work at the moment).

We noticed that this is your third macro update in a row (after #6193 and #6271). I'm happy to merge this, but the reason we don't just do this ourselves is to give truly first-time contributors the chance to go through the easiest possible experience of contributing. Now that you've done this a few times, we think you're ready for bigger and better things, whether contributing to our repo or others. I've got plenty of skipped libcxx tests that need to be investigated / reported / fixed, if you're looking for something useful to do!

StephanTLavavej:

StephanTLavavej:
Thanks again, got it in with a day to spare! :joy_cat:

PR #6306, opened 2026-06-09
Declare `__isa_available` and `__isa_enabled` properly

StephanTLavavej:
In conjunction with my upcoming changes to VCRuntime/VCStartup for low-overhead `IsProcessorFeaturePresent()` results, I'm cleaning up `<isa_availability.h>` and noticed several issues in the STL.

First, these variables are of type `int`. Second, they live in the global namespace. Third, they don't exist for actual ARM64.

These STL changes will work with both the old and new VCRuntime/VCStartup.

### DIFF
--- a/stl/inc/__msvc_bit_utils.hpp
+++ b/stl/inc/__msvc_bit_utils.hpp
@@ -20,10 +20,13 @@ _STL_DISABLE_CLANG_WARNINGS
 #pragma push_macro("new")
 #undef new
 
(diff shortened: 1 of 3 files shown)


zacklj89 (approved):

jaykrell:
This change is ok, but IsProcessorFeaturePresent you have to be careful with, since it is not present in Server2022 VBS Enclaves, you cannot import it, load will fail.

PR #6350, opened 2026-07-01
`<random>`: Fix catastrophic cancellation in `piecewise_linear_distribution` evaluation

### LINKED ISSUE #6323: `<random>`: `piecewise_linear_distribution` generates very non-uniform values for PDF inputs VERY close to 1.0 on the 0.0 to 1.0 interval
```cpp
// Piecewise Linear Distribution Bug.cpp 
//
// Problem:
// STL <random> piecewise_linear_distribution
// generates VERY non-uniform values
// for PDF inputs very close to 1.0
// on the 0.0 to 1.0 X interval.

// MS Visual Studio Community 2022 (64 bit)
// Release 17.14.35+37411.7.-june.2026-
// 00482-90000-00000-AA902

// Nelson Bridwell

#include <iostream>
#include <vector>
#include <random>

using namespace std;

int main() {

    const unsigned int fixed_seed = 42u;     // any value you like
    mt19937 uniform;
    uniform.seed(fixed_seed);
    const int NUM_SAMPLES = 1000000;

    // EXAMPLE 1: Linear PDF from 0.0 to 1.0
    vector<double> x = { 0.0, 1.0 };
    double pdf0 = 0.99999999999999990;
    double pdf1 = 1.00000000000000010;
    vector<double> pdf = {pdf0, pdf1};
    printf("Simple Linear PDF from 0.0 to 1.0:\n");
    printf("pdf(0.0): %.20f   pdf(1.0): %.20f \n", pdf0, pdf1);
    piecewise_linear_distribution<double> dist =
        piecewise_linear_distribution<double>(x.begin(), x.end(), pdf.begin());

    int binCountA[11] = { 0,0,0,0,0,0,0,0,0,0,0 };
    printf("First 25 samples:\n Only exactly 1 or exactly 0!\n");
    for (long n = 1; n <= NUM_SAMPLES; ++n) {
        double observation = dist(uniform);
        if (n <= 25)
            printf(" %d  %.20f\n", n, observation);
        int idx = trunc(10 * observation);
        binCountA[idx]++;
    }
    printf("After 1,000,000 samples, the counts for 11 equal bins from 0.0 up to just less than 1.1.\n");
    for (int idx = 0; idx<=10; idx++)
        printf(" %d  ", binCountA[idx]);
    printf("\n");
    printf("Counts for first 10 bins should be similar, with none in bin 11!\n");
    printf("\n\n");


    // EXAMPLE 2: Another, very slightly different linear PDF from 0.0 to 1.0
    pdf0 = 0.9999999999999990;
    pdf1 = 1.0000000000000010;
    pdf = { pdf0, pdf1 };
    printf("Another Linear PDF curve from 0.0 to 1.0:\n");
    printf("pdf(0.0): %.20f   pdf(1.0): %.20f \n", pdf0, pdf1);
    dist = piecewise_linear_distribution<double>(x.begin(), x.end(), pdf.begin());
    int binCountB[11] = {0,0,0,0,0,0,0,0,0,0,0};
    printf("First 25 samples.\n");
    printf("Samples 4, 7, 9, 14, 18, 23, and 24 the exact same number!\n");
    printf("3, 8, and 25 the same. 10, 16, and 21 the same...\n");

    for (long n = 1; n <= NUM_SAMPLES; ++n) {
        double observation = dist(uniform);
        if (n <= 25)
            printf(" %d  %.20f\n", n, observation);
        int idx = trunc(10 * observation);
        binCountB[idx]++;
    }
    printf("After 1,000,000 samples, the counts for 11 equal bins from 0.0 up to just less than 1.1.\n");
    for (int idx = 0; idx <= 10; idx++)
        printf(" %d  ", binCountB[idx]);
    printf("\n");

}
```

Console Output:
```
Simple Linear PDF from 0.0 to 1.0:
pdf(0.0): 0.99999999999999988898   pdf(1.0): 1.00000000000000000000
First 25 samples:
 Only exactly 1 or exactly 0!
 1  0.00000000000000000000
 2  0.00000000000000000000
 3  0.00000000000000000000
 4  0.00000000000000000000
 5  1.00000000000000000000
 6  1.00000000000000000000
 7  0.00000000000000000000
 8  0.00000000000000000000
 9  0.00000000000000000000
 10  0.00000000000000000000
 11  0.00000000000000000000
 12  0.00000000000000000000
 13  0.00000000000000000000
 14  1.00000000000000000000
 15  1.00000000000000000000
 16  0.00000000000000000000
 17  1.00000000000000000000
 18  0.00000000000000000000
 19  0.00000000000000000000
 20  1.00000000000000000000
 21  1.00000000000000000000
 22  0.00000000000000000000
 23  1.00000000000000000000
 24  0.00000000000000000000
 25  0.00000000000000000000
After 1,000,000 samples, the counts for 11 equal bins from 0.0 up to just less than 1.1.
 624881   0   0   0   0   0   0   0   0   0   375119
Counts for first 10 bins should be similar, with none in bin 11!


Another Linear PDF curve from 0.0 to 1.0:
pdf(0.0): 0.99999999999999900080   pdf(1.0): 1.00000000000000111022
First 25 samples.
Samples 4, 7, 9, 14, 18, 23, and 24 the exact same number!
3, 8, and 25 the same. 10, 16, and 21 the same...
 1  0.26315789473684209065
 2  0.21052631578947367252
 3  0.47368421052631576318
 4  0.78947368421052632748
 5  0.00000000000000000000
 6  0.05263157894736841813
 7  0.78947368421052632748
 8  0.47368421052631576318
 9  0.78947368421052632748
 10  0.68421052631578949121
 11  0.57894736842105265495
 12  0.89473684210526316374
 13  0.31578947368421050879
 14  0.78947368421052632748
 15  0.26315789473684209065
 16  0.68421052631578949121
 17  0.15789473684210525439
 18  0.78947368421052632748
 19  0.57894736842105265495
 20  0.89473684210526316374
 21  0.68421052631578949121
 22  0.10526315789473683626
 23  0.78947368421052632748
 24  0.78947368421052632748
 25  0.47368421052631576318
After 1,000,000 samples, the counts for 11 equal bins from 0.0 up to just less than 1.1.
 83229   110962   111230   106522   129218   110975   118173   104798   100025   0   24868
```

MattStephanson:
Fixes #6323.

The formula used to invert the CDF is unstable if two consecutive weights are nearly equal. To be concrete, suppose that the pdf is $f(x) = 1+\delta(2x-1)$ over the domain $[0,1]$, so that the range is $[1-\delta,1+\delta]$. Then the CDF is $F(x) = \delta x^2 + (1-\delta)x$. Then, given a random U(0,1) variate $u$ (`_Xx0` in the code), this is inverted with the formula
```math
x = \frac{\sqrt{(1-\delta)^2(1-u)+(1+\delta)^2u}-(1-\delta)}{2\delta} \approx \frac{[1+\delta(2u-1)]-(1-\delta)}{2\delta} = u,
```
which is correct to zeroth order in $\delta$, since the PDF is nearly constant. But there is catastrophic cancellation in the numerator, which causes the extreme quantization in some of OP's examples.

An alternative form of the quadratic formula,
```math
\frac{-b + \sqrt{b^2-4ac}}{2a} = \frac{-2c}{b + \sqrt{b^2-4ac}},
```
is more stable because the two terms being added in the denominator have the same sign.

There is also some risk of overflow in the final line, https://github.com/microsoft/STL/blob/c430582bc1147f4c01973f6a051415110c4eb286/stl/inc/random#L4682

I've extracted the internal `std::lerp` machinery to all language modes to make use of it here, but we could drop that commit if it's undesired.

<details>
<summary>Derivation of new formula</summary>
The weight function on the chosen interval is $(p_0(x-b_1)+p_1(b_0-x))/(b_0-b_1)$, but this can't be inverted directly because it's not normalized. Its integral over the interval is $(b_1-b_0)(p_1+p_0)/2$. Normalizing and reparameterizing from $x\in [b_0,b_1]$ to $x\in [0,1]$ gives the CDF $((p_1-p_0)x^2+2p_0x)/(p_1+p_0)$. Inverting this requires solving $\frac{1}{2}(p_1-p_0)x^2+p_0x-\frac{1}{2}u(p_1+p_0)=0$. Substituting $a=\frac{1}{2}(p_1-p_0)$, $b=p_0$, and $c=-\frac{1}{2}u(p_1+p_0)$ into the two formulas above gives the old and new versions.
</details>

### DIFF
--- a/stl/inc/cmath
+++ b/stl/inc/cmath
@@ -9,7 +9,7 @@
 #if _STL_COMPILER_PREPROCESSOR
 
 #include <cstdlib>
-#include <xtr1common>
+#include <type_traits>
 
 #if !defined(__clang__) && !defined(_M_CEE) // TRANSITION, VSO-1663104
 #define _HAS_CMATH_INTRINSICS 1
@@ -20,8 +20,6 @@
 #if _HAS_CMATH_INTRINSICS
 #include _STL_INTRIN_HEADER
 #endif // _HAS_CMATH_INTRINSICS
-
-#include <type_traits>
 
 #pragma pack(push, _CRT_PACKING)
 #pragma warning(push, _STL_WARNING_LEVEL)
--- a/stl/inc/cmath
+++ b/stl/inc/cmath
@@ -21,9 +21,7 @@
 #include _STL_INTRIN_HEADER
 #endif // _HAS_CMATH_INTRINSICS
 
-#if _HAS_CXX20
 #include <type_traits>
-#endif // _HAS_CXX20
 
 #pragma pack(push, _CRT_PACKING)
 #pragma warning(push, _STL_WARNING_LEVEL)
@@ -1120,6 +1118,112 @@ _EXPORT_STD using _CSTD isless;
 _EXPORT_STD using _CSTD islessequal;
 _EXPORT_STD using _CSTD islessgreater;
 _EXPORT_STD using _CSTD isunordered;
+
+template <class _Ty>
+_NODISCARD constexpr _Ty _Linear_for_lerp(const _Ty _ArgA, const _Ty _ArgB, const _Ty _ArgT) noexcept {
+    if (_STD _Is_constant_evaluated()) {
+        auto _Smaller     = _ArgT;
+        auto _Larger      = _ArgB - _ArgA;
+        auto _Abs_smaller = _Float_abs(_Smaller);
+        auto _Abs_larger  = _Float_abs(_Larger);
+        if (_Abs_larger < _Abs_smaller) {
+            _Ty _Tmp = _Smaller;
+            _Smaller = _Larger;
+            _Larger  = _Tmp;
+
+            _Tmp         = _Abs_smaller;
+            _Abs_smaller = _Abs_larger;
+            _Abs_larger  = _Tmp;
+        }
+
+        if (_Abs_smaller > 1) {
+            // _Larger is too large to be subnormal, so scaling by 0.5 is exact, and the product _Smaller * _Larger is
+            // large enough that if _ArgA is subnormal, it will be too small to contribute anyway and this way can
+            // sometimes avoid overflow problems.
+            return 2 * (_Ty{0.5} * _ArgA + _Smaller * (_Ty{0.5} * _Larger));
+        } else {
+            return _ArgA + _Smaller * _Larger;
+        }
+    }
+
+    return _STD fma(_ArgT, _ArgB - _ArgA, _ArgA);
+}
+
+template <class _Ty>
+_NODISCARD constexpr _Ty _Common_lerp(const _Ty _ArgA, const _Ty _ArgB, const _Ty _ArgT) noexcept {
+    // on a line intersecting {(0.0, _ArgA), (1.0, _ArgB)}, return the Y value for X == _ArgT
+
+    const bool _T_is_finite = _Is_finite(_ArgT);
+    if (_T_is_finite && _Is_finite(_ArgA) && _Is_finite(_ArgB)) {
+        // 99% case, put it first; this block comes from P0811R3
+        if ((_ArgA <= 0 && _ArgB >= 0) || (_ArgA >= 0 && _ArgB <= 0)) {
+            // exact, monotonic, bounded, determinate, and (for _ArgA == _ArgB == 0) consistent:
+            return _ArgT * _ArgB + (1 - _ArgT) * _ArgA;
+        }
+
+        if (_ArgT == 1) {
+            // exact
+            return _ArgB;
+        }
+
+        // exact at _ArgT == 0, monotonic except near _ArgT == 1, bounded, determinate, and consistent:
+        const auto _Candidate = _Linear_for_lerp(_ArgA, _ArgB, _ArgT);
+        // monotonic near _ArgT == 1:
+        if ((_ArgT > 1) == (_ArgB > _ArgA)) {
+            if (_ArgB > _Candidate) {
+                return _ArgB;
+            }
+        } else {
+            if (_Candidate > _ArgB) {
+                return _ArgB;
+            }
+        }
+
+        return _Candidate;
+    }
+
+    if (_STD _Is_constant_evaluated()) {
+        if (_Is_nan(_ArgA)) {
+            return _ArgA;
+        }
+
+        if (_Is_nan(_ArgB)) {
+            return _ArgB;
+        }
+
+        if (_Is_nan(_ArgT)) {
+            return _ArgT;
+        }
+    } else {
+        // raise FE_INVALID if at least one of _ArgA, _ArgB, and _ArgT is signaling NaN
+        if (_Is_nan(_ArgA) || _Is_nan(_ArgB)) {
+            return (_ArgA + _ArgB) + _ArgT;
+        }
+
+        if (_Is_nan(_ArgT)) {
+            return _ArgT + _ArgT;
+        }
+    }
+
+    if (_T_is_finite) {
+        // _ArgT is finite, _ArgA and/or _ArgB is infinity
+        if (_ArgT < 0) {
+            // if _ArgT < 0:     return infinity in the "direction" of _ArgA if that exists, NaN otherwise
+            return _ArgA - _ArgB;
+        } else if (_ArgT <= 1) {
+            // if _ArgT == 0:    return _ArgA (infinity) if _ArgB is finite, NaN otherwise
+            // if 0 < _ArgT < 1: return infinity "between" _ArgA and _ArgB if that exists, NaN otherwise
+            // if _ArgT == 1:    return _ArgB (infinity) if _ArgA is finite, NaN otherwise
+            return _ArgT * _ArgB + (1 - _ArgT) * _ArgA;
+        } else {
+            // if _ArgT > 1:     return infinity in the "direction" of _ArgB if that exists, NaN otherwise
+            return _ArgB - _ArgA;
+        }
+    } else {
+        // _ArgT is an infinity; return infinity in the "direction" of _ArgA and _ArgB if that exists, NaN otherwise
+        return _ArgT * (_ArgB - _ArgA);
+    }
+}
 _STD_END
 
 #if _HAS_CXX17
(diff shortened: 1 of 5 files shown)


stl/inc/random
```diff
@@ -4676,10 +4676,10 @@ public:
 
         if (_Px0 != _Px1) {
             _Xx0 = static_cast<result_type>(
-                (_STD sqrt(_Px0 * _Px0 * (1.0 - _Xx0) + _Px1 * _Px1 * _Xx0) - _Px0) / (_Px1 - _Px0));
+                (_Xx0 * (_Px0 + _Px1)) / (_STD sqrt(_Px0 * _Px0 * (1.0 - _Xx0) + _Px1 * _Px1 * _Xx0) + _Px0));
         }
```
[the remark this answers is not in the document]
MattStephanson: If `_Px0 == 0` the formula simplifies significantly, so treat that as a special case regardless of `_Xx0`.

tests/std/tests/GH_006323_pwl_dist_rounding/test.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+    constexpr size_t N = 100'000;
+    vector<double> samples(N);
+    for (size_t i = 0; i < N; ++i) {
+        samples[i] = dist(urbg);
+    }
+
+    sort(samples.begin(), samples.end());
+
+    // Given the empirical CDF for $n$ samples, $F_n(x)$, and a hypothesized CDF $F(x)$, the  Kolmogorov-Smirnov
```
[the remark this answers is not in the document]
MattStephanson: D'oh. @StephanTLavavej, I'll let you decide if this is worth resetting test results for (assuming there's nothing else to fix).

AlexGuteniev:
I see there's multipkication and subtraction, would this not benefit from fma?

StephanTLavavej:
@AlexGuteniev Because FMA is "spicy" (has special interactions with compilers), I would prefer to see any usage of it for performance to be done separately from this correctness fix. Let's avoid risky grab-bags!

stl/inc/cmath
```diff
@@ -21,9 +21,7 @@
 #include _STL_INTRIN_HEADER
 #endif // _HAS_CMATH_INTRINSICS
 
-#if _HAS_CXX20
 #include <type_traits>
```
StephanTLavavej: This should now be fused with the other includes, where it will supersede `<xtr1common>`.

tests/std/tests/GH_006323_pwl_dist_rounding/test.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <cstddef>
+#include <random>
+#include <vector>
+
+using namespace std;
+
+int main() {
+    constexpr double delta = 1.0e-15;
+    double b[2]            = {0.0, 1.0};
+    double p[2]            = {1.0 - delta, 1.0 + delta};
```
StephanTLavavej: `piecewise_linear_distribution` takes InputIterators, so these can be const (I'll make them `static constexpr`).

StephanTLavavej (approved):

AlexGuteniev:
> any usage of it for performance

I meant it here for better precision.
I considered fma for performance as "off label", although apparently that's not the case.

stl/inc/random
```diff
@@ -4674,12 +4674,14 @@ public:
         uniform_real_distribution<_Ty> _Dist;
         result_type _Xx0 = _Dist(_Eng);
 
-        if (_Px0 != _Px1) {
+        if (_Px0 == _Ty{0}) {
```
statementreply:
```suggestion
        if (_Px0 == 0.0) {
```
`_Px0` is `double`

stl/inc/random
```diff
@@ -4674,12 +4674,14 @@ public:
         uniform_real_distribution<_Ty> _Dist;
         result_type _Xx0 = _Dist(_Eng);
 
-        if (_Px0 != _Px1) {
+        if (_Px0 == _Ty{0}) {
+            _Xx0 = _STD sqrt(_Xx0);
+        } else if (_Px0 != _Px1) {
             _Xx0 = static_cast<result_type>(
-                (_STD sqrt(_Px0 * _Px0 * (1.0 - _Xx0) + _Px1 * _Px1 * _Xx0) - _Px0) / (_Px1 - _Px0));
+                (_Xx0 * (_Px0 + _Px1)) / (_STD sqrt(_Px0 * _Px0 * (1.0 - _Xx0) + _Px1 * _Px1 * _Xx0) + _Px0));
```
statementreply, line 4681: Pre-existing: the expression inside `sqrt` may overflow or underflow

tests/std/tests/GH_006323_pwl_dist_rounding/test.cpp
```diff
@@ -0,0 +1,57 @@
...
+        const double ecdf_hi = i / static_cast<double>(N);
+        const double cdf     = x + x * (x - 1.0) * delta;
+
+        const double max_diff = max(abs(cdf - ecdf_lo), abs(cdf - ecdf_hi));
+        if (max_diff > ks_stat) {
+            ks_stat = max_diff;
+        }
+    }
+
+    constexpr double crit_val = 1.224; // critical value for p = 0.1
```
statementreply, line 53: I think p=0.1 is too high for a CI test

stl/inc/random
MattStephanson: Agreed, and I see generally how it could happen (very wide or narrow intervals). But I'm not sure that I'll have time in the foreseeable future to address it and add testing. I'll report back in a week or so.

tests/std/tests/GH_006323_pwl_dist_rounding/test.cpp
MattStephanson: To be clear, you're suggesting a lower p value and therefore a higher critical value? So that the test is less sensitive to the seed? FWIW, in the current code the K-S statistic is 0.0016 and the critical value is 0.0039, so it's not exactly on the threshold of failing.

stl/inc/random
StephanTLavavej: I can push a change here, this is within my cat-sized brain's capability.

stl/inc/random
StephanTLavavej: As this is pre-existing, a followup PR would be fine.

tests/std/tests/GH_006323_pwl_dist_rounding/test.cpp
StephanTLavavej: If I understand that correctly, it sounds like no change is required here.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:x: :heart_eyes_cat: :white_check_mark:
===

PR #6364, opened 2026-07-16
Remove workarounds for DevCom-11055227

### LINKED ISSUE #6363: `vector_algorithms.cpp`: Revisit ARM64 constant arrays
We have a few DevCom-11055227 workarounds and one possible place for it is suppressed because of DevCom-11056805.

https://github.com/microsoft/STL/blob/020513e211529e7be30cb3e0ca310869701286da/stl/src/vector_algorithms.cpp#L11269

Need to check the status on the latest Preview and possibly remove workarounds.

Should be done in #6362 or separately after that PR is merged.

hazzlim:
I have verified that these workarounds are no longer needed.

Closes #6363

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -6089,8 +6089,7 @@ namespace {
             template <class _Ty>
             __forceinline bool _Make_bitmap_large_neon(
                 const void* const _Needle, const size_t _Needle_length, uint8x16x2_t& _Bitmap) noexcept {
-                // TRANSITION, DevCom-11055227
-                constexpr uint8_t _Mask_arr[16] = {
+                static constexpr uint8_t _Mask_arr[16] = {
                     0x01, 0x02, 0x04, 0x08, 0x10, 0x20, 0x40, 0x80, 0x01, 0x02, 0x04, 0x08, 0x10, 0x20, 0x40, 0x80};
                 const auto _Mask = vld1q_u8(_Mask_arr);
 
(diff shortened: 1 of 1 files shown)


StephanTLavavej (approved):

YexuanXiao:
What is the best practice for using static constexpr variables? Unless constrained by syntax, I generally don't proactively declare constexpr variables as static.

AlexGuteniev:
> What is the best practice for using static constexpr variables?

Take address? Make it static!

And arrays have their address implicitly taken.

StephanTLavavej:
The best practice is for arrays to be `static constexpr`, but scalars can be plain `constexpr`. The reason why is that distinct objects need distinct addresses. Even when no code actually cares about the distinctiveness of an object's address, the optimizer typically can't prove that nothing cares about an array being distinct, because arrays are inherently address-taken when used (as soon as a pointer to something is flying around, the optimizer has to worry it might be saved somewhere, and then compared for equality). So if you have a plain `constexpr` array, it will end up emitting code to push every array element onto the stack during the function call. (Try it with `/FAs`, or in Compiler Explorer.) A `static constexpr` array is way better for codegen, as it gets burned into the read-only section in the image (rdata), and the function can just push its address onto the stack.

For ordinary scalar `constexpr` variables, they are typically not address-taken, so you don't need to bother with `static constexpr`. (In rare cases they can be, e.g. if passed to `std::max()`, but sometimes inlining will rescue that anyways.)

There's an annoying exception that in actual `constexpr` evaluation, `static constexpr` arrays weren't usable, so you needed to use plain `constexpr`. I think that finally got fixed recently (C++26?) but I am not keeping as close track of language evolution as I once did.

frederick-vs-ja:
> There's an annoying exception that in actual `constexpr` evaluation, `static constexpr` arrays weren't usable, so you needed to use plain `constexpr`. I think that finally got fixed recently (C++26?) but I am not keeping as close track of language evolution as I once did.

That was fixed by WG21-P2647R1 in C++23. Maybe we can start relying on the change.

StephanTLavavej:

StephanTLavavej:
:mechanical_arm: :broom: :heart_eyes_cat:
===

PR #6381, opened 2026-07-26
Fix vectorized `replace_copy` for `bool` arrays, also test `replace`

AlexGuteniev:
Copilot has [discovered](https://github.com/microsoft/STL/pull/6379#discussion_r3651191850) a bug in `replace` vectorization for 8-bit elements that is being implemented. @StephanTLavavej observed that `replace_copy` has already this bug in. This PR fixes the bug.

The problem happens when vectorized `replace` is called for bool array and the replacement is an integer that is a non-value representation when converted to bool.

Due to `_Could_compare_equal_to_value_type` we don't have the comparand equivalent of this problem.

Since there are only two algorithms that need such handling (`replace` and `replace_copy`), and only one exact type, I think we should just dispatch it separately.


tests/std/tests/GH_006381_replace_non_value_bool/test.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include <iterator>
+
+using namespace std;
+
+int main() {
+    // Test replace_copy against an attempt to replace bool value with a non-value representation
+    // Compare with direct memcmp instead of algorithms or loops: make the possible failure as deterministic as possible
+
+    static constexpr bool source[]            = {true, true, false, true, false, false, false};
+    static constexpr bool expected_all_true[] = {true, true, true, true, true, true, true};
```
StephanTLavavej, line 20: I guess you're not actually trying to engage the vectorization, since the potential bug happens in the header layer, and only the separately compiled layer decides whether it's worth vectorizing or can be handled as a scalar tail. So there's no need to use something like 63 bytes. However, I think this at least deserves a comment, since usually vectorization tests need to worry about a size threshold.

tests/std/tests/GH_006381_replace_non_value_bool/test.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+    constexpr size_t source_len = size(source);
+    static_assert(size(expected_all_true) == source_len, "sizes should match");
+
+    bool dest_false_replaced[source_len];
+    replace_copy(begin(source), end(source), begin(dest_false_replaced), 0, 2);
```
StephanTLavavej: Can you also add coverage for `std::replace` so I can just prevent #6195 from adding this bug, and then your PR will provide test coverage for both algorithms?

tests/std/tests/GH_006381_replace_non_value_bool/test.cpp
```diff
@@ -0,0 +1,39 @@
...
+    assert(memcmp(dest_false_replaced, expected_all_true, source_len) == 0);
+
+    bool dest_true_replaced[source_len];
```
StephanTLavavej: Adding extra scopes around each test array will help to make it clear that information is not being transferred between each test, and to prevent mistakes where an assertion uses the wrong local variable (I have caught these multiple times in the past).

StephanTLavavej (changes requested):

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for fixing and testing this! :hammer_and_wrench: :magic_wand: :heart_eyes_cat:

PR #6382, opened 2026-07-26
P3059R2: Making user-defined constructors of view iterators/sentinels private. 

### LINKED ISSUE #6241: P3059R2 Making User-Defined Constructors Of View Iterators/Sentinels `private`
WG21-P3059R2 Making User-Defined Constructors Of View Iterators/Sentinels `private`

At the 2026-03 meeting in Croydon, this was formally accepted as a Defect Report, which should be implemented retroactively back to C++20.

iskandarem:
- User defined constructors for following view-iterators/sentinels made private:
    - [x] filter_view::iterator/sentinel
    - [x] basic_istream_view::iterator
    - [x] transform_view::iterator/sentinel
    - [x] take_view::sentinel
    - [x] take_while_view::sentinel
    - [x] join_view::sentinel
    - [x] lazy_split_view::outer_iterator
    - [x] lazy_split_view::inner_iterator
    - [x] split_view::iterator/sentinel
    - [x] element_view::iterator/sentinel
    - [x] iota_view::_Ioterator

-  Removed the `_Iotinel` and replaced it with a nested `iota_view::_Sentinel`. This avoids changing the ABI of the existing `_Ioinel` type while making the implementation closer to the Standard wording. 
    - [x] iota_view::_Sentinel

closes #6241


StephanTLavavej:
Please don't just click "Commit suggestion" on Copilot code review comments. That will immediately push a commit, rerunning checks. (I don't personally pay for them, but they're a shared resource, and if we waste it enough, someone will eventually hiss at me.) In this case, there were several code review comments from Copilot, *and* your checks were already failing in Code Format Validation. The proper thing to do is to locally address all feedback, from both the failed checks and Copilot, then validate and push a single time. (You can push one or multiple commits at once.)

iskandarem:
> Please don't just click "Commit suggestion" on Copilot code review comments. That will immediately push a commit, rerunning checks. (I don't personally pay for them, but they're a shared resource, and if we waste it enough, someone will eventually hiss at me.) In this case, there were several code review comments from Copilot, _and_ your checks were already failing in Code Format Validation. The proper thing to do is to locally address all feedback, from both the failed checks and Copilot, then validate and push a single time. (You can push one or multiple commits at once.)

Sorry, I didn't know about that. I'll pay attention in the future.

StephanTLavavej:
No worries, I still need to write up Contribution Guidelines that will explain all of this stuff.

While your PR is in draft state, pushes won't trigger checks. When you want checks to run again, move your PR back to ready to review, *then* push commits. If you do the reverse (push then move to ready), the state change of moving out of draft mode won't trigger checks.

stl/inc/ranges
```diff
@@ -1614,15 +1633,7 @@ namespace ranges {
...
-        public:
-            using iterator_concept = conditional_t<bidirectional_range<_Vw>, bidirectional_iterator_tag,
-                conditional_t<forward_range<_Vw>, forward_iterator_tag, input_iterator_tag>>;
-            using value_type       = range_value_t<_Vw>;
-            using difference_type  = range_difference_t<_Vw>;
-
-            _Iterator()
-                requires default_initializable<iterator_t<_Vw>>
-            = default;
+            friend filter_view; // to so that filter_view had access to the private constructor
```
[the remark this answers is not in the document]
StephanTLavavej: I agree with Copilot here (except that it wants to mess up the indentation, lol). In general, inspect similar code and see if we think something is subtle enough to deserve a comment.

iskandarem:
Hi @StephanTLavavej! I have a question regarding the `lazy_split_view` implementation.

`_Inner_iter` currently declares(i added it):

```
friend lazy_split_view;
```

Copilot suggested replacing it with:

```
friend typename _Outer_iter<_Const>::value_type;
```

arguing that `_Inner_iter` is actually constructed from `value_type::begin()`, not directly from `lazy_split_view`.

However, I tried both versions, and they both compile successfully. I also wrote a test that exercises the normal public path:

```
auto outer = view.begin();
auto inner = (*outer).begin();
```

and it passes with both friendship declarations.

From my understanding, nested classes have the same access rights as members of their enclosing class, so `friend lazy_split_view` already can access the iterator. However friend lazy_split_view also grants broader access than necessary, whereas befriending typename _Outer_iter<_Const>::value_type would restrict access to the actual caller.


And my question would be should i use `friend lazy_split_view` or `friend typename _Outer_iter<_Const>::value_type`?

StephanTLavavej:
`friend lazy_split_view;` is what we've used elsewhere. I'm not really worried about granting more friendship than necessary; the STL trusts itself (and in fact we used to use public `_Ugly` data members by convention). I slightly prefer not having to think too hard about why the friendship is being granted to something unusual. It's simpler to reason about "oh, of course the view can access all of its iterators' internals".

stl/inc/ranges
```diff
@@ -902,8 +909,17 @@ namespace ranges {
...
+        template <weakly_incrementable _Wj, semiregular _Bj>
+            requires _Weakly_equality_comparable_with<_Wj, _Bj> && copyable<_Wj>
+        friend class iota_view;
+
         using _It = _Ioterator<_Wi>;
 
+        /* [[no_unique_address]] */ _Bo _Last{};
+
+        constexpr explicit _Iotinel(_Bo _Last_) noexcept(is_nothrow_move_constructible_v<_Bo>)
+            : _Last(_STD move(_Last_)) {}
```
frederick-vs-ja: Currently this PR changes `_Iotinel` from aggregate to non-aggregate, which may cause ABI change for certain underlying bound types. Is this acceptable?

iskandarem: Yes, you're right, @frederick-vs-ja. However, in the paper I found the following example:
```cpp
namespace std::ranges {
    template<weakly_incrementable W, semiregular Bound>
        requires weakly-equality-comparable-with<W, Bound> && copyable<W>
    struct iota_view<W, Bound>::iterator {
    private:
        W value_ = W();                       // exposition only

        constexpr explicit iterator(W value); // exposition only

    public:
        // ...
        iterator() requires default_initializable<W> = default;
        // ...
    };
}
```

Based on that wording, I thought the closest implementation would be to make the stored bound and the constructor private, matching the paper's exposition as closely as possible.

I'd also be interested to know what the MSVC STL policy is regarding ABI changes in cases like this

iskandarem:
Some tests check if the views internal constructors of iterator/sentinel are accessible to users which this paper makes private.

Example:
`tests\std\tests\P0896R4_views_filter_death\test.cpp:51`
```cpp
void test_constructor_wrong_range() {
    vector<int> vec0{0, 1, 2, 3};
    vector<int> vec1{4, 5, 6, 7};
    auto r0            = views::filter(vec0, lambda);
    using R            = decltype(r0);
    same_as<R> auto r1 = views::filter(vec1, lambda);
    ranges::iterator_t<R> i{r0, r1.begin().base()}; // vector iterators in range are from different containers
}
```
Should we remove these tests or should we just adjust them to right behavior?

stl/inc/ranges
StephanTLavavej: The C++23 ABI is not yet finalized so we can change it arbitrarily.

The C++20 ABI is finalized so we need to avoid breaking it. That said, if this is a Standard-mandated change, and restricted to iota_view sentinels, I have difficulty imagining them being passed across TUs, so we could probably get away with a change. I would like a complete accounting of which types are changing from aggregate to non-aggregate. It may be safer to rename any such types with a `2` suffix, e.g. `_Iotinel2`, which could limit the blast radius.

StephanTLavavej:
If such tests can't be exercised from user code (without granting friendship) then they should simply be removed, as the avenues for misuse have been closed off. (We should possibly consider removing same-parent checks from now-private constructors, or make them STL-internal checks.)

stl/inc/ranges
frederick-vs-ja: > I would like a complete accounting of which types are changing from aggregate to non-aggregate.

I checked this before and only found `_Iotinel`. Other iterator and sentinel types were non-aggregate initially.

I considered modifying `_Iotinel` as the following

```C++
struct _Iotinel {
    struct _Layer {
        /* [[no_unique_address]] */ _Bo _Last{};
    };
    // omitting other members
};
```

, which can keep `_Iotinel` an aggregate and avoid changing size and alignment (IIUC), while making `is_constructible_v<_Iotinel, _Bo>` `false`. Although this method still allows users to construct `_Iotinel` from `_Bo` using list-initialization.

iskandarem:
The libcxx pipeline is failing because the current llvm-project tests still directly construct iterators/sentinels using constructors that P3059R2 makes private. This has already been fixed in llvm/llvm-project, but that version isn't in Microsoft/STL yet (**Is this the issue?**).
```log
Failed Tests (7):
  libc++ :: std/ranges/range.adaptors/range.elements/sentinel/ctor.convert.pass.cpp:2
  libc++ :: std/ranges/range.adaptors/range.lazy.split/range.lazy.split.outer/ctor.parent.pass.cpp:2
  libc++ :: std/ranges/range.adaptors/range.split/iterator/deref.pass.cpp:2
  libc++ :: std/ranges/range.adaptors/range.take.while/sentinel/ctor.convert.pass.cpp:0
  libc++ :: std/ranges/range.factories/range.iota.view/iterator/ctor.value.pass.cpp:2
  libc++ :: std/ranges/range.factories/range.istream.view/ctor.pass.cpp:2
  libc++ :: std/ranges/range.factories/range.istream.view/iterator/ctor.pass.cpp:0
```
What should I do in this case?
Would be happy to follow any guidance.

StephanTLavavej:
I merged #6385 yesterday, updating our LLVM commit to https://github.com/llvm/llvm-project/tree/85305ba2b67ea806e1e04b61dcc8743b19f51fc8 which is 2 days old.

It looks like you want https://github.com/llvm/llvm-project/pull/193891 which merged 8 hours ago?

You can mark the libcxx tests as failed, citing that PR number (look at the top section of expected_results.txt listing issues that are known to libcxx; add the tests in sorted order, and if any :0 configs need to be FAILed, those are plain MSVC and you also need to FAIL the :1 configs which are MSVC ASan.

tests/libcxx/expected_results.txt
```diff
@@ -1275,3 +1275,95 @@ std/iterators/iterator.container/ssize.LWG3207.compile.pass.cpp:9 SKIPPED
...
+# istream_view
+std/ranges/range.factories/range.istream.view/ctor.pass.cpp:2 FAIL
+std/ranges/range.factories/range.istream.view/iterator/ctor.pass.cpp:0 FAIL
+std/ranges/range.factories/range.istream.view/iterator/ctor.pass.cpp:1 FAIL
+std/ranges/range.factories/range.istream.view/ctor.pass.cpp:0 FAIL
+std/ranges/range.factories/range.istream.view/ctor.pass.cpp:1 FAIL
+std/ranges/range.factories/range.istream.view/iterator/ctor.pass.cpp:2 FAIL
+std/ranges/range.factories/range.istream.view/iterator/deref.pass.cpp:0 FAIL
+std/ranges/range.factories/range.istream.view/iterator/deref.pass.cpp:1 FAIL
+std/ranges/range.factories/range.istream.view/iterator/deref.pass.cpp:2 FAIL
```
cpplearner: Since the test is expected to fail in all three configs (plain MSVC, MSVC ASan, Clang-cl), you can fuse these lines and omit the number.

```suggestion
std/ranges/range.factories/range.istream.view/iterator/deref.pass.cpp FAIL
```

Same applies for other lines in this file. Also, remember to sort the added lines.

stl/inc/ranges
iskandarem: what should we do about that?

tests/std/tests/P3059R2_basic_istream_view_iterator_constructor/test.compile.pass.cpp
```diff
@@ -0,0 +1,38 @@
...
+using View     = ranges::basic_istream_view<int, char>;
+using Iterator = ranges::iterator_t<View>;
+
+// P3059R2: users must not be able to construct
+// basic_istream_view::iterator directly from the parent view.
+static_assert(!constructible_from<Iterator, View&>);
+
+// The iterator must remain non-copyable but movable through its public interface.
+static_assert(!copy_constructible<Iterator>);
+static_assert(move_constructible<Iterator>);
```
frederick-vs-ja: Can we merge these into the existing main test files? I think it will be a bit more efficient to avoid more test files.

tests/std/tests/P3059R2_basic_istream_view_iterator_constructor/test.compile.pass.cpp
```diff
@@ -0,0 +1,38 @@
...
+    istringstream input{"10 20 30 40"};
+    auto view = ranges::istream_view<int>(input);
+
+    vector<int> result;
+
+    for (const int value : view) {
+        result.push_back(value);
+    }
+    return result == vector{10, 20, 30, 40};
+}
```
frederick-vs-ja: Is such test coverage missing in the "main" test files? If not, I don't think we should add these cases.

stl/inc/ranges
frederick-vs-ja: If the renaming approach is acceptable, I'd like to do so.

@StephanTLavavej How about changing the sentinel type to a nested class of `iota_view`? This also renames the class, and additionally makes our implementation strategy closer to the standard wording.

StephanTLavavej: I'd be fine with that, Do What The Standard Says is a good principle to follow and as you say it's equivalent to a rename.

iskandarem: Should we remove the old  `_Iotinel` after introducing the new nested sentinel type? It will no longer have any references in the implementation, except in `stl\debugger\STL.natvis`, which would also need to be updated to refer to the new type.

I’m considering naming the new nested type either `_Sentinel` or `_InnerIotinel`. I think _Sentinel is closer to the Standard wording.

StephanTLavavej: Yes, the old unused type should be removed from the product code. In the visualizers, we typically keep the visualizers for old types so debugging old programs is unaffected. If the new type can reuse the same visualizer it can be added as a synonym.

`_Sentinel` seems fine to me, as we've used that in `take_view` and probably elsewhere. As long as there's no confusion about _which_ sentinel is meant (shadowing or near-shadowing is extremely confusing).

StephanTLavavej:
Thanks for continuing to iterate on this PR (I am trying to reply to comments as I see pings, but if I miss any email notifications, please get my attention in the STL Discord).

When you believe you've addressed a code review comment, either human or Copilot, resolving the conversation helps us keep track of which comments remain to be addressed. This is more a concern as PR history gets longer and this one is certainly getting there :joy_cat:

iskandarem:
I rewrote the commits to correct the commit-author because I had accidentally been using the wrong git email.😨

StephanTLavavej:
I'll update LLVM again so this PR doesn't have to work around failures (also need it for `constexpr <cmath>`).

StephanTLavavej:
I've merged the LLVM update.

stl/debugger/STL.natvis
```diff
@@ -1851,6 +1851,13 @@ SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
     </Expand>
   </Type>
 
+  <Type Name="std::ranges::iota_view&lt;*,*&gt;::_Sentinel&lt;*&gt;">
```
frederick-vs-ja, line 1854: There's already an entry for `_Iotinel`. I guess it's better to just add an `AlternativeType` entry for `std::ranges::iota_view&lt;*,*&gt;::_Sentinel&lt;*&gt;` - or the other way around, to make `std::ranges::iota_view&lt;*,*&gt;::_Sentinel&lt;*&gt;` the main `Name` and make `std::ranges::_Iotinel&lt;*&gt;` an `AlternativeType`.

tests/libcxx/expected_results.txt
```diff
@@ -837,6 +818,10 @@ std/iterators/iterator.requirements/iterator.assoc.types/readable.traits/indirec
 # Not analyzed, likely bogus test. constexpr fails with "vector iterators incompatible".
 std/ranges/range.adaptors/range.join.with/range.join.with.iterator/ctor.default.pass.cpp FAIL
 
+# The test dereferences a basic_istream_view iterator after begin() fails on an empty stream.
```
frederick-vs-ja: Now we can change this line as the following.

```suggestion
# LLVM-215589: [libc++][ranges] Fix the LWG 3568 test for basic_istream_view
```

And then we should move the entry to the `*** ISSUES REPORTED/KNOWN UPSTREAM ***` part. The skipping will be removed in the next LLVM submodule update, IIUC.

iskandarem:
Hi, just checking in on this PR since it's been in Initial Review for a few weeks. I've addressed the previous review comments and there isn't anything currently pending from my side.

Please let me know if there's anything else I should update or investigate. Thanks!

PR #6380, opened 2026-07-26
docs: document TESTS_BUILD_ONLY default on some targets

Lngmate:
Resolves #6149

PR #6383, opened 2026-07-26
Document how to build ARM64 natively with `TESTS_BUILD_ONLY`

### LINKED ISSUE #6149: `README.md`: Document `TESTS_BUILD_ONLY` default on some targets
Discord discussion: https://discord.com/channels/737189251069771789/737189251497721887/1432744992627228733

At lease we should document this if no better ideas.

StephanTLavavej:
Fixes #6149.

Thanks to @Lngmate for looking into this with #6380.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,19 @@ To build the ARM64EC target:
 1. `cmake --preset ARM64EC`
 1. `cmake --build --preset ARM64EC`
 
+## Building ARM64 Natively
+
+By default, the x64 and x86 presets are configured to both build and run the tests.
+The ARM64 and ARM64EC presets assume that you're cross-compiling, so they enable an option `TESTS_BUILD_ONLY`
+to build test executables without running them. If you have an ARM64 machine, you'll want to use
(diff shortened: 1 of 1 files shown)


StephanTLavavej:

README.md
```diff
@@ -216,6 +216,19 @@ To build the ARM64EC target:
 1. `cmake --preset ARM64EC`
 1. `cmake --build --preset ARM64EC`
 
+## Building ARM64 Natively
```
BillyONeal, line 219: This seems reasonable but it also seems to be something that the build should be fixing itself rather than trying to document around :/

I submitted https://github.com/microsoft/STL/pull/6384

BillyONeal (approved):

BillyONeal:
Approved in case you would prefer this solution rather than that one as I don't think this is *wrong*.

README.md
StephanTLavavej: At a minimum we should explicitly depict the `arm64` parameter to `vcvarsall.bat`.

BillyONeal: That's true but also applies to the other platforms; they need to become (for example) arm64_x86

PR #6384, opened 2026-07-27
Determine `TESTS_BUILD_ONLY`'s default by probing.

BillyONeal:
... rather than hard coding ARM64==off

This should make the docs change in https://github.com/microsoft/STL/pull/6383/ unnecessary by making the build system do the correct thing by default.

GPT 5.6-Sol was used in preparing these changes.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -65,9 +65,6 @@
           "arm64",
           ""
         ]
-      },
-      "cacheVariables": {
-        "TESTS_BUILD_ONLY": true
       }
     },
     {
@@ -87,7 +84,6 @@
         ]
       },
       "cacheVariables": {
-        "TESTS_BUILD_ONLY": true,
         "VCLIBS_TARGET_ARCHITECTURE": "ARM64EC"
       }
     }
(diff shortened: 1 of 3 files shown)


BillyONeal:
I opened this as "draft" because I didn't bring my ARM64 machine to the office today and I want to verify that we're doing the right thing before being sure this does the right thing.

StephanTLavavej:
If this works, it will also have the advantage of auto-disabling x64 tests running on an ARM64 host.

BillyONeal:
> If this works, it will also have the advantage of auto-disabling x64 tests running on an ARM64 host.

No, because ARM64 hosts can run the x64 tests :)

BillyONeal:
I confirmed on my Snapdragon X2 machine following the instructions in the README that `TESTS_BUILD_ONLY` got automatically set to OFF by inspecting `CMakeCache.txt` -> removing Draft.

I think the README shouldn't have the separate ARM64 section as all target platforms are now different but that's not really my table so I won't go there. (Build system stuff *is* my table which is where this PR came from :) )

BillyONeal:
> I think the README shouldn't have the separate ARM64 section as all target platforms are now different but that's not really my table so I won't go there. (Build system stuff _is_ my table which is where this PR came from :) )

Actually I changed my mind since I needed to touch the README now that https://github.com/microsoft/STL/pull/6383 merged I needed to remove the instructions to change this anyway.

PR #6385, opened 2026-07-28
Toolset update: MSVC Compiler 19.52.36615, and update LLVM

StephanTLavavej:
* Cycle locations.
* PowerShell 7.6.4.
* MSVC Compiler 19.52.36615 (now required).
  + Picks up MSVC-PR-757497 "feat(cmath, 36 of N): loosen cli flag incompatibilities".
* New pools.
* Update llvm-project.
* Remove tests deleted by llvm/llvm-project#203338.
* llvm/llvm-project#182390 was fixed.
* llvm/llvm-project#90196 was fixed.
* llvm/llvm-project#138375 was fixed.
* Categorize failing tests.
* Avoid putting too much pressure on Australia East.
* New pools, again.
* Group x64 truncation warnings, mark them as SKIPPED.
* Update llvm-project again.
* llvm/llvm-project#212433 added `ADDITIONAL_COMPILE_FLAGS(cl-style-warnings)`.
* llvm/llvm-project#212435 fixed the `_BitInt` failures.
* llvm/llvm-project#210075 implemented more constexpr for cmath (that works for Clang).
* Update libcxx/expected_results.txt for the MSVC-internal test harness.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,8 +6,8 @@ cmake_minimum_required(VERSION 4.3.1)
 set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
 project(msvc_standard_libraries LANGUAGES CXX)
 
-if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "19.52.36530")
-    message(FATAL_ERROR "The STL must be built with MSVC Compiler 19.52.36530 or later. Follow the README instructions.")
+if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "19.52.36615")
+    message(FATAL_ERROR "The STL must be built with MSVC Compiler 19.52.36615 or later. Follow the README instructions.")
 endif()
 
 include(CheckCXXSourceCompiles)
--- a/tests/libcxx/expected_results.txt
+++ b/tests/libcxx/expected_results.txt
@@ -241,6 +241,18 @@ std/strings/c.strings/cuchar.compile.pass.cpp FAIL
 std/language.support/support.limits/support.limits.general/cmath.version.compile.pass.cpp FAIL
 std/language.support/support.limits/support.limits.general/cstdlib.version.compile.pass.cpp FAIL
 std/numerics/c.math/fpclassify.pass.cpp FAIL
+std/numerics/c.math/isgreater.pass.cpp:0 FAIL
+std/numerics/c.math/isgreater.pass.cpp:1 FAIL
+std/numerics/c.math/isgreaterequal.pass.cpp:0 FAIL
+std/numerics/c.math/isgreaterequal.pass.cpp:1 FAIL
+std/numerics/c.math/isless.pass.cpp:0 FAIL
+std/numerics/c.math/isless.pass.cpp:1 FAIL
+std/numerics/c.math/islessequal.pass.cpp:0 FAIL
+std/numerics/c.math/islessequal.pass.cpp:1 FAIL
+std/numerics/c.math/islessgreater.pass.cpp:0 FAIL
+std/numerics/c.math/islessgreater.pass.cpp:1 FAIL
+std/numerics/c.math/isunordered.pass.cpp:0 FAIL
+std/numerics/c.math/isunordered.pass.cpp:1 FAIL
 
 # P0543R3 Saturation Arithmetic
 # Add 'std-at-least-c++26' to tests/utils/stl/test/tests.py when beginning work on C++26.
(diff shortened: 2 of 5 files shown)


tests/libcxx/expected_results.txt
```diff
@@ -1173,6 +1191,62 @@ std/utilities/tuple/tuple.tuple/tuple.cnstr/convert_const_move.pass.cpp:1 FAIL
...
+# Not analyzed.
+# MSVC warning C4804: '<<': unsafe use of type 'bool' in operation
+std/language.support/support.types/byteops/lshift.assign.pass.cpp:0 FAIL
+std/language.support/support.types/byteops/lshift.assign.pass.cpp:1 FAIL
+std/language.support/support.types/byteops/lshift.pass.cpp:0 FAIL
+std/language.support/support.types/byteops/lshift.pass.cpp:1 FAIL
+std/language.support/support.types/byteops/rshift.assign.pass.cpp:0 FAIL
+std/language.support/support.types/byteops/rshift.assign.pass.cpp:1 FAIL
+std/language.support/support.types/byteops/rshift.pass.cpp:0 FAIL
+std/language.support/support.types/byteops/rshift.pass.cpp:1 FAIL
```
frederick-vs-ja: No change requested here. I rewrote these tests in LLVM-204116 and intentionally covered `operator<<` etc. with `bool` operands (since `bool` satisfies the constraints of these operators). Is such coverage itself considered bad?

StephanTLavavej: The test needs to suppress MSVC warnings if it's going to do something like that.

frederick-vs-ja: Got it. I've just opened LLVM-212433.

tests/libcxx/expected_results.txt
```diff
@@ -1191,6 +1191,62 @@ std/utilities/tuple/tuple.tuple/tuple.cnstr/convert_const_move.pass.cpp:1 FAIL
 # Not analyzed, MSVC truncation warnings, Clang static_assert.
 std/algorithms/alg.nonmodifying/alg.fold/ranges.fold_right_last.pass.cpp FAIL
 
+# Not analyzed, test aborts.
+std/containers/unord/unord.map/unord.map.cnstr/exceptions.pass.cpp FAIL
+
+# Not analyzed.
+# MSVC error C2440: 'initializing': cannot convert from 'const std::shared_ptr<int>::{ctor}::_Fancy_t' to 'const std::shared_ptr<int>::{ctor}::_Raw_t'
+# Clang error: no viable conversion from 'const _Fancy_t' (aka 'const Pointer<int>') to 'const _Raw_t' (aka 'int *const')
+std/utilities/memory/util.smartptr/util.smartptr.shared/pointer_destruction.pass.cpp FAIL
```
frederick-vs-ja, line 1303: This one is a bit tricky. The conversion function from `Pointer<T>` to `T*` in that test file lacks `const`, while our implementation strategy requires implicit conversion from a const lvalue.

https://github.com/microsoft/STL/blob/f8ac180e7f87e8db2ad1d10e4f53881c123cadb2/stl/inc/memory#L1692-L1710

In other words, the constrains apparently require `is_convertible_v<_Fancy_t, _Raw_t>` ([[util.smartptr.shared.const]/28](https://eel.is/c++draft/util.smartptr.shared.const#28)) while our logic additionally requires `is_convertible_v<const _Fancy_t&, _Raw_t>`.

cpplearner: So we should remove `const` and use `std::move`?

tests/libcxx/expected_results.txt
StephanTLavavej: Thanks for the quick fix, I've picked that up.

tests/libcxx/expected_results.txt
StephanTLavavej: It does seem that we should be using a modifiable rvalue, yes.

StephanTLavavej:

CMakeLists.txt
```diff
@@ -6,8 +6,8 @@ cmake_minimum_required(VERSION 4.3.1)
 set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
 project(msvc_standard_libraries LANGUAGES CXX)
 
-if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "19.52.36530")
-    message(FATAL_ERROR "The STL must be built with MSVC Compiler 19.52.36530 or later. Follow the README instructions.")
+if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "19.52.36615")
+    message(FATAL_ERROR "The STL must be built with MSVC Compiler 19.52.36615 or later. Follow the README instructions.")
```
davidmrdavid, line 10: non-blocking / subjective: we could consider placing the version in a variable, using that variable in the `if`-check, and just string interpolation to place it in the error message.

Just something to consider, may be nice.

davidmrdavid (approved):

CMakeLists.txt
StephanTLavavej: Thanks, I've recorded a todo to try extracting it in the next toolset update. I agree that having fewer places to update is better.

Issue #1532, opened 2020-12-14
<xmemory>: Investigate constexpr approach between _Default_allocator_traits::allocate and _Allocate


mnatsuhara:
In GH-1369, we will be implementing `is_constant_evaluated()` pathways in both `_Default_allocator_traits::allocate` and the global `_Allocate` function that is called from that function if not evaluated in a `constexpr` context.  

`_Default_allocator_traits::allocate`
```c++
    _NODISCARD static _CONSTEXPR20_DYNALLOC __declspec(allocator) pointer
        allocate(_Alloc& _Al, _CRT_GUARDOVERFLOW const size_type _Count) {
#ifdef __cpp_lib_is_constant_evaluated
        if (_STD is_constant_evaluated()) {
            return _Al.allocate(_Count);
        } else
#endif // __cpp_lib_is_constant_evaluated
        {
            (void) _Al;
            return static_cast<pointer>(
                _Allocate<_New_alignof<value_type>>(_Get_size_of_n<sizeof(value_type)>(_Count)));
        }
    }
```

`_Allocate`: 
```c++
template <size_t _Align, class _Traits = _Default_allocate_traits,
    enable_if_t<(_Align > __STDCPP_DEFAULT_NEW_ALIGNMENT__), int> = 0>
__declspec(allocator) _CONSTEXPR20_DYNALLOC void* _Allocate(const size_t _Bytes) {
    // allocate _Bytes when __cpp_aligned_new && _Align > __STDCPP_DEFAULT_NEW_ALIGNMENT__
    if (_Bytes == 0) {
        return nullptr;
    }

#ifdef __cpp_lib_is_constant_evaluated
    if (_STD is_constant_evaluated()) {
        return _Traits::_Allocate(_Bytes);
    } else
#endif // __cpp_lib_is_constant_evaluated
...
```

NOTE: Once GH-1369 is checked in, we can update the code blurbs above to be links to the actual lines of code - I wasn't able to get links to lines changed in a PR.  

Currently, MSVC cannot handle calling the `_Allocate` function from a `constexpr` context (`_Allocate` calls `::operator new` which MSVC does not allow in a `constexpr` function, though Clang does if it is reached transitively from an evaluation of `std::allocator::allocate`), which is why there is the `is_constant_evaluated()` block in `_Default_allocator_traits::allocate` that instead dispatches to `std::allocator::allocate` (which is permitted by MSVC).  

[...]

Issue #169, opened 2019-10-11
Consider releasing the abi-breaking-changes branch


jonwil:
There are a number of issues out there (both listed here on the STL repository and also things like https://developercommunity.visualstudio.com/content/problem/193041/the-visual-c-2017-condition-variable-implementatio.html listed on the Visual C++ bug tracker) that are listed as being fixable but only in whatever the next release is that breaks ABI.
There are likely people who want those fixes and would be quite happy using their own STL build that broke the ABI if it meant getting those fixes sooner.

I suggest you consider releasing the ABI breaking changes in a way that those who know what they are doing and want something that breaks ABI can get them. (but where anyone who doesn't explicitly download the ABI-breaking-changes branch wont have to worry) Or if its not possible to publish this stuff (either because there are also ABI changes in vcruntime, vcstartup, concrt or other non-open bits or because vcruntime, vcstartup, concrt or other non-open bits need recompiling with the new ABI-breaking code to pick up the new ABI) then at least consider publishing a comprenensive accurate list of the things that have/can be fixed but only by breaking ABI (so people know where the STL has bugs (or differences with the C++ standards and why they can't be fixed at this time)

MikeGitb:
I think the best thing (and I believe they already plan to do it) would be to put the code into a separate branch here on github, but not to ship it in any official form. 

The worst possible outcome would be both version being shipped and supported for quite some time in parallel and people to start #ifdef ing their code based on which version the code is built against and incompatible binaries floating around. Imo that would be a disservice to the eco system. 

All that aside: I guess one of the problems may be that those patches no longer apply cleanly to the current master version of the code.

jonwil:
Releasing it as a separate branch is the answer for sure. People who want to ship it can grab that branch and get a set of binaries out of it knowing full well those dlls are ABI-breaking compared to the official binaries.

Of course it may not be possible to publish the code at all if the ABI-breaking branch would require changes to the compiler or changes or recompilation of other non-open binaries like vcruntime, vcstartup or concrt (that said, it looks like concrt is the only runtime library component shipped with the compiler that actually links to and pulls in the STL so its entirely possible the answer to any problems with an ABI incompatible STL solved by making concrt open source, something that would be nice to see anyway given the circular dependency between concrt and STL)

sylveon:
ABI breaking STL could be released on vcpkg. To avoid conflicts with the built-in STL, the headers could be located under another folder, like so: `#include <stl/string>`

[...]

PR #6391, opened 2026-08-01
`<xmemory>`: Document the iterator debugging invariants

### LINKED ISSUE #2084: `<xmemory>`: Document the iterator debugging invariants
Currently, the iterator debugging scheme involving a proxy object and an intrusive singly-linked list of iterators isn't well-commented in the codebase. The maintainers should capture their knowledge in a detailed comment next to the machinery:

https://github.com/microsoft/STL/blob/bd7adb4a932725f60ba096580c415616486ab64c/stl/inc/xmemory#L1053-L1059

Thanks to @mnatsuhara for the suggestion! Uncommented invariants make it difficult to understand/maintain code.

prakriti31:
Closes #2084.

The iterator debugging (IDL) proxy/iterator machinery in `<xmemory>`
(`_Container_proxy`, `_Container_base12`, `_Iterator_base12`) wasn't
commented at all, even though its invariants are non-obvious (intrusive
linked list of iterators, mutual container/proxy pointers, locking rules).
@StephanTLavavej explained these invariants in the issue thread (originally
via a linked Discord screenshot from 2021), so this PR just captures that
explanation as a comment next to the code it describes, so it doesn't only
live in a screenshot linked from a 4-year-old issue comment.

I verified the described invariants (mutual _Myproxy/_Mycont pointers, the
_Myfirstiter/_Mynextiter intrusive list, the _Lockit(_LOCK_DEBUG) locking
rule) still match the current code exactly before writing this up.

No behavior change -- comment-only.

### DIFF
--- a/stl/inc/xmemory
+++ b/stl/inc/xmemory
@@ -1208,6 +1208,22 @@ struct _Iterator_base0 {
     static constexpr bool _Unwrap_when_unverified = true;
 };
 
+// The iterator debugging library (IDL) below lets a container invalidate ("orphan") its iterators without either
+// side needing to know about the other directly, by routing everything through a shared proxy object. Invariants:
+//
+// * Every container owns a dynamically allocated _Container_proxy at all times, including in its
+//   default-constructed and moved-from states.
+// * A container and its proxy always point to each other (_Container_base12::_Myproxy and
+//   _Container_proxy::_Mycont, respectively), regardless of whether IDL is enabled; if a proxy exists, this holds.
+// * Every valid iterator holds a non-owning pointer to its parent container's proxy (_Iterator_base12::_Myproxy).
+//   An orphaned iterator has a null _Myproxy.
+// * The proxy's _Myfirstiter, together with each iterator's _Mynextiter, forms an intrusive singly linked list of
+//   iterators rooted at the proxy. Every valid iterator belonging to a container is reachable through this list;
+//   there are no valid "free-floating" iterators.
+// * Whenever the proxies and the intrusive list are manipulated, the debug lock (_Lockit(_LOCK_DEBUG)) is held.
+//   The only things we do outside of that lock are things like iterator compatibility checks that compare proxy
+//   pointers: those pointers don't change even if the containers are being swapped concurrently (only the
+//   proxies' data members change, not their addresses).
 struct _Container_base12;
 struct _Container_proxy { // store head of iterator chain and back pointer
     _CONSTEXPR20 _Container_proxy() noexcept = default;
--- a/stl/inc/xmemory
+++ b/stl/inc/xmemory
@@ -1208,11 +1208,15 @@ struct _Iterator_base0 {
     static constexpr bool _Unwrap_when_unverified = true;
 };
 
-// The iterator debugging library (IDL) below lets a container invalidate ("orphan") its iterators without either
-// side needing to know about the other directly, by routing everything through a shared proxy object. Invariants:
+// The machinery below implements iterator debugging (informally "IDL", after the _ITERATOR_DEBUG_LEVEL macro,
+// where "L" stands for "level"). It lets a container invalidate ("orphan") its iterators without either side
+// needing to know about the other directly, by routing everything through a shared proxy object. Invariants:
 //
 // * Every container owns a dynamically allocated _Container_proxy at all times, including in its
-//   default-constructed and moved-from states.
+//   default-constructed and moved-from states. This is TRANSITION, ABI: allocating the proxy separately (instead
+//   of, say, storing it inline) is the major reason many containers' allocator-extended move constructors and
+//   move assignment operators aren't unconditionally noexcept -- reloading the proxy when allocators compare
+//   unequal can throw. We intend to revisit this strategy in vNext (see #169).
 // * A container and its proxy always point to each other (_Container_base12::_Myproxy and
 //   _Container_proxy::_Mycont, respectively), regardless of whether IDL is enabled; if a proxy exists, this holds.
 // * Every valid iterator holds a non-owning pointer to its parent container's proxy (_Iterator_base12::_Myproxy).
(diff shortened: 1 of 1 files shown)


prakriti31:

prakriti31:
@StephanTLavavej What else do we need to merge it. Can you help me? I will push the changes asap.

StephanTLavavej:
Someone who is deeply familiar with the invariants (i.e. me) needs to review this and think about whether the description is correct and complete.

I have a high-priority task to get through (`constexpr <cmath>`) and then I'm going on vacation, but I'll get to this when I return and start dealing with the PR backlog. Apologies for the delay.

prakriti31:
Sure!!

stl/inc/xmemory
```diff
@@ -1208,6 +1208,22 @@ struct _Iterator_base0 {
     static constexpr bool _Unwrap_when_unverified = true;
 };
 
+// The iterator debugging library (IDL) below lets a container invalidate ("orphan") its iterators without either
```
frederick-vs-ja: I think the letter "L" in "IDL" usually stands for "level". Also, perhaps we shouldn't say the iterator debugging mechanisms form a library.

stl/inc/xmemory
```diff
@@ -1208,6 +1208,22 @@ struct _Iterator_base0 {
     static constexpr bool _Unwrap_when_unverified = true;
 };
 
+// The iterator debugging library (IDL) below lets a container invalidate ("orphan") its iterators without either
+// side needing to know about the other directly, by routing everything through a shared proxy object. Invariants:
+//
+// * Every container owns a dynamically allocated _Container_proxy at all times, including in its
+//   default-constructed and moved-from states.
```
frederick-vs-ja: In additional to Copilot, I think it's better to explain that this is the major reason why many containers in MSVC STL don't have `noexcept` move constructors while they ought to have.

Also, this is somehow "TRANSITION, ABI". Perhaps we should say we will change this strategy in vNext (see #169).

stl/inc/xmemory
```diff
@@ -1208,6 +1208,22 @@ struct _Iterator_base0 {
...
+//   _Container_proxy::_Mycont, respectively), regardless of whether IDL is enabled; if a proxy exists, this holds.
+// * Every valid iterator holds a non-owning pointer to its parent container's proxy (_Iterator_base12::_Myproxy).
+//   An orphaned iterator has a null _Myproxy.
+// * The proxy's _Myfirstiter, together with each iterator's _Mynextiter, forms an intrusive singly linked list of
+//   iterators rooted at the proxy. Every valid iterator belonging to a container is reachable through this list;
+//   there are no valid "free-floating" iterators.
+// * Whenever the proxies and the intrusive list are manipulated, the debug lock (_Lockit(_LOCK_DEBUG)) is held.
+//   The only things we do outside of that lock are things like iterator compatibility checks that compare proxy
+//   pointers: those pointers don't change even if the containers are being swapped concurrently (only the
+//   proxies' data members change, not their addresses).
```
frederick-vs-ja: I guess it's meaningful to say that we don't hold the lock during constant evaluation.

stl/inc/xmemory
prakriti31: Hi, I have made the changes. Thankyou for your input, matters a lot. 
Would appreciate if you can verify the changes once.

stl/inc/xmemory
prakriti31: Done in 7c1b71a0. The bullet now explains that allocating the proxy separately is the major reason several containers' allocator-extended move constructors and move assignment operators aren't unconditionally `noexcept` -- reloading the proxy when allocators compare unequal can throw -- tags it `TRANSITION, ABI`, and points at #169 for the vNext plan.

One follow-up: I'll change `#169` to `GH-169` in the next push, to match the convention used elsewhere in the headers (`GH-1532`, `GH-4634`, etc.).

stl/inc/xmemory
prakriti31: Done in 7c1b71a0 -- the bullet now notes that we skip the lock during constant evaluation and take the unlocked paths instead, since constant evaluation is single-threaded and `_Lockit` isn't usable there.

While re-checking this, though, I noticed the bullet is still overstated in the other direction, so I'll tighten it in the next push:

- At `_ITERATOR_DEBUG_LEVEL == 1`, `_Swap_proxy_and_iterators` calls `_Swap_proxy_and_iterators_unlocked` directly, with no lock at all.
- `_Alloc_proxy` and `_Reload_proxy` mutate `_Myproxy`/`_Mycont` without the debug lock at any level.

The neighboring `_Myfirstiter` bullet has the same problem: at IDL 1, `_Adopt` only copies `_Myproxy` and never links the iterator into the list, and `_Orphan_all` is empty, so `_Myfirstiter` stays null. "Every valid iterator is reachable through this list" only holds at IDL 2. I'll scope both bullets to IDL 2 explicitly, along with Copilot's point about IDL 0 having no proxy at all.

stl/inc/xmemory
prakriti31: Following up on this: I had to walk back part of what I wrote here, and I'd value your take.

I documented the reload-can-throw rationale as you suggested, but going back through `vector` I couldn't make it line up with the code:

- `vector(vector&&)` is unconditionally `noexcept` and still calls `_Alloc_proxy` (vector:761-767).
- `vector(vector&&, const Alloc&)` allocates its proxy before branching, on the `is_always_equal` path too (vector:770-777); the path that can throw is the unequal-allocator one that moves elements individually.
- The `_Reload_proxy` in `operator=(vector&&)` is inside the `_Propagate_allocators` branch, which is noexcept, and already carries "intentionally slams into noexcept on OOM, TRANSITION, VSO-466800" (vector:817-824).

So the separate allocation seems to force *termination* on OOM rather than weaken the `noexcept` specs -- those look driven by allocator equality and POCMA instead. I've rewritten the bullet to say that, in 09431279.

If you meant a different set of containers, or the ABI constraint in a broader sense, please say so and I'll correct it -- I only traced `vector` and `deque` directly.

prakriti31:
Hi [Stephan T. Lavavej (@StephanTLavavej)](https://github.com/StephanTLavavej)
I have tried my best to iteratively review the changes. I am hoping that you would not find any voids or corrections, but I would request you to review the changes whenever you find time and get back to me.

Issue #2279, opened 2021-10-16
`<functional>`: Why does function support calling convention


AlexGuteniev:
```
#include <functional>
int main()
{
   std::function<int __stdcall()> f;
}
```

This appears to be deliberate by spawning specializations with macro.

The calling convention does not have an effect: it does not propagate to `function::operator()`, and of course function accepts function pointers and functors with any calling convention.

Is there a rationale to support this?

I'm asking because I don't want to generate code with preprocessor or scripts for `move_only_function`, partially because #348 says to avoid macros, but if there's a need to have every calling convention, I'll have to generate some code.

AlexGuteniev:
@StephanTLavavej:
> It has occasionally occurred that someone forms a function type from a function with an explicit calling convention, e.g. `remove_pointer_t<decltype(&SomeWinAPIFunction)>`

> We might not need to extend this weirdness to move_only_function, since it does have a throughput and complexity cost
It can be added later if there's a strong need for it, but it's certainly not a conformance issue, and the scenario is quite rare
Probably merits a comment at most, `// intentionally not providing specializations for other calling conventions at this time` or something

PR #6401, opened 2026-08-09
Document code-emitting function-like macros

### LINKED ISSUE #348: <xstddef>: document code-emitting function-like macros
At the end of `<xstddef>`: https://github.com/microsoft/STL/blob/1d39dfac9929c0d3c405edf6bc7eff26bbe2a26b/stl/inc/xstddef#L293-L422  we define a small system of macros that are used to stamp out small variations of code for particular cv qualifiers, or calling conventions, etc. We go out of our way to avoid these macros in new code, but there are nevertheless some existing uses. It would be nice to document these macros and their usage via comments in the header.

ABWI-Y:
<type_traits>: Document code-emitting function-like macros

This PR adds comments to document the code-emitting macros that used to live in
<xstddef> and now live in <type_traits>. These macros are used to stamp out
specializations for all the different calling conventions, cv/ref qualifiers,
and noexcept variations that we support.

These macros are kept around only for existing uses (see GH #2279 for why).
New code shouldn't rely on them.

Closes #348.

### DIFF
--- a/stl/inc/type_traits
+++ b/stl/inc/type_traits
@@ -376,58 +376,42 @@ struct _NO_SPECIALIZATIONS_OF_TYPE_TRAITS is_compound : bool_constant<!is_fundam
 _EXPORT_STD template <class _Ty>
 _NO_SPECIALIZATIONS_OF_TYPE_TRAITS constexpr bool is_compound_v = !is_fundamental_v<_Ty>;
 
-// Code-emitting macros for cv-qualified, ref-qualified, and calling-convention
-// variations.  These macros are used to stamp out repetitive specializations
-// for components that need to handle multiple calling conventions, such as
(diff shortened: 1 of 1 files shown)


frederick-vs-ja:
> <type_traits>: Document code-emitting function-like macros

This part should be the PR title.

ABWI-Y:
> > <type_traits>: Document code-emitting function-like macros
> 
> This part should be the PR title.

Thanks for pointing that out! I've updated the PR title accordingly.

StephanTLavavej:
You updated the PR description, not the PR title.

ABWI-Y:
Oh,I'm sorry,I have updated it.

ABWI-Y:
Are there any other issues or feedback that need to be addressed? If not, could you please help merge this? Thanks!

PR #6406, opened 2026-08-18
`<__msvc_int128.hpp>`: optimize `_Signed128::operator*` with a branchless multiply

### LINKED ISSUE #6405: `<__msvc_int128.hpp>`: The multiply operation function has performance issue
## Summary

`std::_Signed128::operator*` in `<__msvc_int128.hpp>` perform sign-normalizes before multiplying, but it is unnecessary and will cause performance issue.

## Detail

The current implementation sign-normalizes both operands before multiplying, and on MSVC x64 /O2 the whole operator* function compiles to roughly 40+ instructions.

I've tried to implement a version without sign handling, the new version has same behavior and will not break any ABI. The new version without sign-normalizing is more faster than current version in `<__msvc_int128.hpp>`.

Here shows the differ: https://godbolt.org/z/Y4x8Pq7Ka, the version without sign handling only contains 7 `mov` + 1 `mul` + 2 `imul` and 2 `add` instructions, which are more faster.

The new version:
```c++
_NODISCARD friend constexpr _Signed128 operator*(_Signed128 _Left, _Signed128 _Right) noexcept {
        // The low 128 bits of a two's-complement product equal the unsigned product, so the sign normalization in the old
        // implementation was unnecessary.
        _Signed128 _Result;
        _Result._Word[0] = _UMul128(_Left._Word[0], _Right._Word[0], _Result._Word[1]);
        _Result._Word[1] += _Left._Word[1] * _Right._Word[0];
        _Result._Word[1] += _Left._Word[0] * _Right._Word[1];
        return _Result;
    }
```

## Measurements

Benchmark on MSVC x64, Release, walking a 1,000,000-pair vector of random operands:

- std::_Signed128 (current): 14.5 ns per multiply
- new 128-bit multiply: 2.37 ns per multiply

The new 128-bit multiply is implemented as `eirin::ext::int128` in my [Eirin Math](https://github.com/Koishi-Satori/EirinMath/blob/feature/algo/include/eirin/ext/builtin_ints.hpp#L184) Library, which has the same behavior as the version without sign handling mentioned before. And I believe this could be a strong evidence that the new version is faster than the old one.

## Notes

- Same semantics, no ABI change; the multiplication core is identical to _Base128::_Multiply.
- I can attach the benchmark (including a faithful replica of the old implementation) if useful.

Koishi-Satori:
Resolves #6405.

`std::_Signed128::operator*` in `<__msvc_int128.hpp>` perform sign-normalizes before multiplying, but it is unnecessary and will cause performance issue.

The low 128 bits of a two's-complement product equal the unsigned product, so the sign handling is unnecessary. I replaced it with the same computation `_Base128::_Multiply` already performs.

I've tried to implement a version without sign handling, the new version has same behavior and will not break any ABI. The new version without sign-normalizing is more faster than current version in `<__msvc_int128.hpp>`.

This PR also adds a signed128_mul benchmark comparing the new implementation against a faithful replica of the old one.

### Benchmark

  On MSVC x64, Release (`/O2`), measured on a GitHub Actions hosted runner (shared 4-vCPU VM; absolute numbers vary between runs):

  | Benchmark | Old (sign-normalizing) | New (branchless) | Speedup |
  |---|---|---|---|
  | `bm_signed128_mul` (independent multiplies) | 9.66 ns | 6.94 ns | ~1.39x |
  | `bm_signed128_mul_horner` (5 dependent multiplies) | 80.9 ns | 9.15 ns | ~8.8x |

  The first case measures multiplication throughput. The second evaluates a degree-5 polynomial with a Horner chain, where every multiply depends on the previous result; this is the same computation shape used by fixed-point math functions such as `log2` and `sin`. Removing the sign normalization shortens the serial dependency chain of each multiply (and eliminates its branches), so latency-bound code benefits far more than the throughput microbenchmarksuggests.


Koishi-Satori:

stl/inc/__msvc_int128.hpp
```diff
@@ -1370,13 +1370,12 @@ struct _Signed128 : _Base128 {
...
-        _Left._Strip_negative(_Negative);
-        _Right._Strip_negative(_Negative);
-        _Signed128 _Result{_Base128::_Multiply(_Left, _Right)};
-        if (_Negative) {
-            _Result = -_Result;
-        }
+        // The low 128 bits of a two's-complement product equal the unsigned product, so the sign normalization in the old
+        // implementation was unnecessary.
+        _Signed128 _Result;
+        _Result._Word[0] = _UMul128(_Left._Word[0], _Right._Word[0], _Result._Word[1]);
```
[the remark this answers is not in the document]
frederick-vs-ja: This is wrong. `_UMul128` is unconditionally `constexpr`.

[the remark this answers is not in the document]
frederick-vs-ja: This is wrong. `_UMul128` is unconditionally provided.

benchmarks/src/signed128_mul.cpp
```diff
@@ -0,0 +1,72 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <benchmark/benchmark.h>
+
+#include <__msvc_int128.hpp>
+#include <array>
+#include <cstdint>
+#include <random>
+#include <utility>
```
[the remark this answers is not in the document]
frederick-vs-ja: I'll also suggest including `<vector>`. But one should be also aware that the name `_Signed128` is MSVC STL-specific if the name is already in the sight.

stl/inc/__msvc_int128.hpp
```diff
@@ -1370,13 +1370,12 @@ struct _Signed128 : _Base128 {
...
     _NODISCARD friend constexpr _Signed128 operator*(_Signed128 _Left, _Signed128 _Right) noexcept {
-        bool _Negative = false;
-        _Left._Strip_negative(_Negative);
-        _Right._Strip_negative(_Negative);
-        _Signed128 _Result{_Base128::_Multiply(_Left, _Right)};
-        if (_Negative) {
-            _Result = -_Result;
-        }
+        // The low 128 bits of a two's-complement product equal the unsigned product, so the sign normalization in the old
+        // implementation was unnecessary.
```
Sir-Walrus: I'm not a huge fan of comments about old versions of the code. That's what commit messages and git blame is for.

stl/inc/__msvc_int128.hpp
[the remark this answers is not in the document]
Koishi-Satori: > This is wrong. `_UMul128` is unconditionally provided.

Thanks, it seems like `_UMul128` has been unconditionally available since #6356

stl/inc/__msvc_int128.hpp
Koishi-Satori: I've removed this comment

benchmarks/src/signed128_mul.cpp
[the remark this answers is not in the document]
Koishi-Satori: acknowledged- and done.

benchmarks/src/signed128_mul.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+using namespace std;
+
+_NODISCARD constexpr _Signed128 _old_mul128(_Signed128 _Left, _Signed128 _Right) {
+    bool _negative = false;
+    _Left._Strip_negative(_negative);
+    _Right._Strip_negative(_negative);
```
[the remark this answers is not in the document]
Koishi-Satori: this is for simulate the old version of multiply operator.

stl/inc/__msvc_int128.hpp
```diff
@@ -1370,13 +1370,10 @@ struct _Signed128 : _Base128 {
...
-        _Right._Strip_negative(_Negative);
-        _Signed128 _Result{_Base128::_Multiply(_Left, _Right)};
-        if (_Negative) {
-            _Result = -_Result;
-        }
+        _Signed128 _Result;
+        _Result._Word[0] = _UMul128(_Left._Word[0], _Right._Word[0], _Result._Word[1]);
+        _Result._Word[1] += _Left._Word[1] * _Right._Word[0];
+        _Result._Word[1] += _Left._Word[0] * _Right._Word[1];
         return _Result;
```
statementreply:
```suggestion
        return _Signed128{_Base128::_Multiply(_Left, _Right)};
```
Since the low 128 bits for signed and unsigned multiplication are the same, could we reuse the existing uint128 code? [`_Base128::_Multiply`](https://github.com/microsoft/STL/blob/48570c4ed800347f749b01c360f4e944cf1232da/stl/inc/__msvc_int128.hpp#L422-L428) is basically the same as your implementation, except for the swapped addition order.

Koishi-Satori: That’s right; we can simplify this code to a single `return _Signed128{_Base128::_Multiply(_Left, _Right)};` statement.

benchmarks/src/signed128_mul.cpp
```diff
@@ -0,0 +1,73 @@
...
+    _Signed128 _Result{_Base128::_Multiply(_Left, _Right)};
+    if (_negative) {
+        _Result = -_Result;
+    }
+    return _Result;
+}
+
+vector<pair<_Signed128, _Signed128>> init_test_set() {
+    vector<pair<_Signed128, _Signed128>> vec(1'000'000);
+    mt19937_64 mt64;
```
AlexGuteniev: I suggest making the default construction more explicit: `mt19937_64 mt64{};`

benchmarks/src/signed128_mul.cpp
```diff
@@ -0,0 +1,73 @@
...
+    const auto vec = init_test_set();
+
+    auto it        = vec.begin();
+    _Signed128 res = 0;
+
+    for (auto _ : state) {
+        res = it->first * it->second;
+
+        benchmark::DoNotOptimize(res);
+        if (++it == vec.end()) {
```
AlexGuteniev: Please use separate statement for pre-increment.

We optimize for easier reading, not for denser code, so we avoid having prefix and postfix increments in unexpected places. (Prefix increment is expected as standalone statement, or as for loop expression, postfix increment is generally not expected).

Ditto below.

benchmarks/src/signed128_mul.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include <__msvc_int128.hpp>
+#include <array>
+#include <cstdint>
+#include <random>
+#include <utility>
+#include <vector>
+
+using namespace std;
+
+_NODISCARD constexpr _Signed128 _old_mul128(_Signed128 _Left, _Signed128 _Right) {
```
AlexGuteniev: We don't use `_ugly` and `_Ugly` for identifiers in benchmarks, and especially avoid mixing them with `snake_case` in the same file.

AlexGuteniev (changes requested):
Please provide your benchmark numbers.

See [Benchmarking the STL](https://github.com/microsoft/STL/wiki/Benchmarking-the-STL#calculating-speedups) on the Wiki or follow other PR precedents.

Koishi-Satori:
> Please provide your benchmark numbers.
> 
> See [Benchmarking the STL](https://github.com/microsoft/STL/wiki/Benchmarking-the-STL#calculating-speedups) on the Wiki or follow other PR precedents.

I've added benchmark results in PR description.

AlexGuteniev (approved):

StephanTLavavej:
You should change your PR description from `Resolves issue #6405.` to `Resolves #6405.` in order to properly invoke [GitHub's close/fix/resolve syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

Koishi-Satori:
> You should change your PR description from `Resolves issue #6405.` to `Resolves #6405.` in order to properly invoke [GitHub's close/fix/resolve syntax](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

Okay, and I've changed the description.

benchmarks/src/signed128_mul.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include <__msvc_int128.hpp>
+#include <array>
+#include <cstdint>
+#include <random>
+#include <utility>
+#include <vector>
+
+using namespace std;
+
+_NODISCARD constexpr _Signed128 old_mul128(_Signed128 _Left, _Signed128 _Right) {
```
StephanTLavavej: Thanks, this made it easy to confirm the speedup. We generally don't preserve old implementations in the benchmarks, though.

benchmarks/src/signed128_mul.cpp
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <benchmark/benchmark.h>
+
+#include <__msvc_int128.hpp>
+#include <array>
+#include <cstdint>
```
StephanTLavavej: `<array>` and `<cstdint>` are unused.

StephanTLavavej (approved):
Thanks! :heart_eyes_cat: I agree that this speedup is worth doing. It looks like there's just enough correctness coverage in `tests/std/tests/P1522R1_difference_type/test.cpp` that we don't need more.

PR #6408, opened 2026-08-20
LWG-3417: Deprecate volatile atomic inc/dec and wait/notify operations

### LINKED ISSUE #6312: LWG-3417 Missing `volatile` atomic deprecations
LWG-3417 Missing `volatile` atomic deprecations

Followup to WG21-P1831R1 #558 implemented by #634.

TPPPP72:
Resolves [LWG 3417 (Missing volatile atomic deprecations)](https://cplusplus.github.io/LWG/issue3417) adding deprecation checks to `volatile` atomic increment/decrement operators as well as `wait`/`notify` operations when `is_always_lock_free` is false.

Close #6312

### DIFF
--- a/stl/inc/atomic
+++ b/stl/inc/atomic
@@ -1662,23 +1662,23 @@ struct _Atomic_integral_facade : _Atomic_integral<_Ty> {
 
     using _Base::operator++;
     _Ty operator++(int) volatile noexcept {
-        (void) _Deprecate_non_lock_free_volatile<_Ty>;
+        static_assert(_Deprecate_non_lock_free_volatile<_Ty>, "Never fails");
         return const_cast<_Atomic_integral_facade*>(this)->_Base::operator++(0);
     }
 
     _Ty operator++() volatile noexcept {
-        (void) _Deprecate_non_lock_free_volatile<_Ty>;
+        static_assert(_Deprecate_non_lock_free_volatile<_Ty>, "Never fails");
         return const_cast<_Atomic_integral_facade*>(this)->_Base::operator++();
     }
 
     using _Base::operator--;
     _Ty operator--(int) volatile noexcept {
-        (void) _Deprecate_non_lock_free_volatile<_Ty>;
+        static_assert(_Deprecate_non_lock_free_volatile<_Ty>, "Never fails");
         return const_cast<_Atomic_integral_facade*>(this)->_Base::operator--(0);
     }
 
     _Ty operator--() volatile noexcept {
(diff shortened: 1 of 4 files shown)


TPPPP72:

stl/inc/atomic
```diff
@@ -1662,19 +1662,23 @@ struct _Atomic_integral_facade : _Atomic_integral<_Ty> {
 
     using _Base::operator++;
     _Ty operator++(int) volatile noexcept {
+        (void) _Deprecate_non_lock_free_volatile<_Ty>;
```
frederick-vs-ja: The existing approach is using `static_assert`.
```suggestion
        static_assert(_Deprecate_non_lock_free_volatile<_Ty>, "Never fails");
```

----

I wonder whether it would be better to put `[[deprecated]]` out of the function body, which can make the warnings emitted before instantiation.

tests/std/tests/P1831R1_deprecated_volatile_atomic/test.cpp
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Microsoft Corporation.
```
frederick-vs-ja, line 1: I strongly doubt whether this test file makes any sense. At least, it certainly isn't verifying deprecation at this moment.

IIUC, MSVC STL hasn't been verifying deprecation messages in its test suite.

tests/std/tests/P1831R1_deprecated_volatile_atomic/test.cpp
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Microsoft Corporation.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <atomic>
+#include <cassert>
+
+using namespace std;
+
+#pragma warning(push)
+#pragma warning(disable : 4996)
```
frederick-vs-ja: If we want to test that some operations are still well-formed, it's conventional to use the following style.
```suggestion
#define _SILENCE_CXX20_VOLATILE_DEPRECATION_WARNING

#include <atomic>
#include <cassert>

using namespace std;
```

tests/std/tests/P1831R1_deprecated_volatile_atomic/test.cpp
AlexGuteniev: It tests deprecation silencing, i.e. that these functions can be made back available as for now.

AlexGuteniev (approved):

PR #6415, opened 2026-08-24
Test: Remove broken `TEST_EVERYTHING` branch from `P0896R4_views_elements`

### LINKED ISSUE #6237: `P0896R4_views_elements`: The test isn't compatible with `TEST_EVERYTHING`
Currently, `P0896R4_views_elements/test.cpp` doesn't compile with `TEST_EVERYTHING` defined.

The reason is `elements_view` requires the reference type of the underlying view (which is possibly a proxy reference) is tuple-like, but the proxy references used in our test suite aren't tuple-like. Even in C++20 where _`has-tuple-element`_ was a bit more relaxed, our proxy references failed to satisfy that concept.

tycket033-tech:
Fixes #6237.

This removes the `TEST_EVERYTHING` branch from `P0896R4_views_elements`.

That branch tries to instantiate proxy-reference ranges, but those references aren't tuple-like and can't satisfy `elements_view`'s constraints. The existing explicit matrix already covers the relevant non-proxy combinations, so it now runs regardless of whether `TEST_EVERYTHING` is defined.

Validation:
- Built and ran `P0896R4_views_elements` normally with `/W4 /WX`.
- Built and ran it again with `/DTEST_EVERYTHING /W4 /WX`.
- Both configurations passed.

AI disclosure: OpenAI Codex with GPT-5.6 Sol was used to inspect the issue and apply this four-line test-only deletion. I reviewed the final diff and test results before submission.

### DIFF
--- a/tests/std/tests/P0896R4_views_elements/test.cpp
+++ b/tests/std/tests/P0896R4_views_elements/test.cpp
@@ -365,9 +365,6 @@ using test_range =
         test::ProxyRef::no>;
 
 constexpr void instantiation_test() {
-#ifdef TEST_EVERYTHING
-    test_in<instantiator, const P>();
-#else // ^^^ test all input range permutations / test only "interesting" permutations vvv
     // The view is sensitive to category, commonality, size, and differencing, but cannot handle proxies.
     using test::Common, test::Sized;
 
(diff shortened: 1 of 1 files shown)


StephanTLavavej:
Thanks. This PR should have been created normally, instead of in draft status (which for us, doesn't run checks; not that it makes a difference for first-time contributor PRs which can't automatically run checks).

tycket033-tech:
> Thanks. This PR should have been created normally, instead of in draft status (which for us, doesn't run checks; not that it makes a difference for first-time contributor PRs which can't automatically run checks).

Thanks for the clarification. I'll open future PRs as ready for review.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thanks for clearing out this broken code and congrats on your first microsoft/STL commit! :heart_eyes_cat:

PR #6419, opened 2026-08-26
Toolset update: MSVC Compiler 19.52.36629

StephanTLavavej:
* libcxx/usual_matrix.lst: Add `/Zc:preprocessor` to ASan, missed in #4052.
* libcxx/expected_results.txt: Enable passing tests.
  + VSO-2338829 "constexpr error 'subtracting pointers to elements of different arrays' in `_String_const_iterator::_Verify_offset()`" was fixed by MSVC-PR-760585 on 2026-07-24 and a regression test was added by MSVC-PR-773972 on 2026-08-25.
* Cycle locations.
* Python 3.14.7.
* PowerShell 7.6.5.
* MSVC Compiler 19.52.36629, centralize version.
  + As requested by @davidmrdavid in https://github.com/microsoft/STL/pull/6385#discussion_r3677862513.
* Work around DevCom-11142709 "MSVC Build Tools Preview omits vcvars batch files unless Latest is installed".
* New pools.

[STL-ASan-CI passed.](https://dev.azure.com/vclibs/STL/_build/results?buildId=20992&view=results)

### DIFF
--- a/azure-devops/config.yml
+++ b/azure-devops/config.yml
@@ -5,16 +5,16 @@
 
 variables:
 - name: x64SlowPoolName
-  value: 'Stl-2026-07-28T0154-x64-Fasv6-Pool'
+  value: 'Stl-2026-08-25T2113-x64-Fasv6-Pool'
   readonly: true
 - name: x64MediumPoolName
-  value: 'Stl-2026-07-28T0154-x64-Fasv7-Pool'
+  value: 'Stl-2026-08-25T2113-x64-Fasv7-Pool'
   readonly: true
 - name: x64FastPoolName
-  value: 'Stl-2026-07-28T0154-x64-Fadsv7-Pool'
+  value: 'Stl-2026-08-25T2113-x64-Fadsv7-Pool'
   readonly: true
 - name: arm64PoolName
-  value: 'Stl-2026-07-28T0154-arm64-Dpdsv6-Pool'
+  value: 'Stl-2026-08-25T2113-arm64-Dpdsv6-Pool'
   readonly: true
 - name: poolDemands
   value: 'EnableSpotVM -equals false'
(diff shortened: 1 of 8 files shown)


azure-devops/config.yml
```diff
@@ -5,16 +5,16 @@
...
+  value: 'Stl-2026-08-25T2113-x64-Fasv7-Pool'
   readonly: true
 - name: x64FastPoolName
-  value: 'Stl-2026-07-28T0154-x64-Fadsv7-Pool'
+  value: 'Stl-2026-08-25T2113-x64-Fadsv7-Pool'
   readonly: true
 - name: arm64PoolName
-  value: 'Stl-2026-07-28T0154-arm64-Dpdsv6-Pool'
+  value: 'Stl-2026-08-25T2113-arm64-Dpdsv6-Pool'
   readonly: true
```
davidmrdavid, line 18: non-blocking / subjective: you _could_ probably create a common variable for the common prefix, to keep future diffs minimal, like so:

```suggestion
variables:
- name: poolNamePrefix
  value: 'Stl-2026-08-25T2113'
  readonly: true
- name: x64SlowPoolName
  value: '$(poolNamePrefix)-x64-Fasv6-Pool'
  readonly: true
- name: x64MediumPoolName
  value: '$(poolNamePrefix)-x64-Fasv7-Pool'
  readonly: true
- name: x64FastPoolName
  value: '$(poolNamePrefix)-x64-Fadsv7-Pool'
  readonly: true
- name: arm64PoolName
  value: '$(poolNamePrefix)-arm64-Dpdsv6-Pool'
  readonly: true
```

davidmrdavid (approved):
LGTM, left a non-blocking suggestion for the future.

azure-devops/config.yml
StephanTLavavej: In this case sure, but in general it is not guaranteed that each pool was slammed out during the very same minute (e.g. if we have to retry the generation of one pool), so this would not be helpful long-term.

PR #6407, opened 2026-08-18
Add SVE implementations of `remove`, `remove_copy`

hazzlim:
This PR adds an SVE implementation of `remove`.

Benchmark results (sizes/benchmarks for which the code path is unchanged are omitted):
#### Measured speedups on a Neoverse N2 Machine (Cobalt 100)

benchmark | speedup
--- | ---
`r<alg_type::std_fn, std::uint32_t>/7` | 0.999x
`r<alg_type::std_fn, std::uint32_t>/15` | 0.997x
`r<alg_type::std_fn, std::uint32_t>/31` | 1.169x
`r<alg_type::std_fn, std::uint32_t>/47` | 1.173x
`r<alg_type::std_fn, std::uint32_t>/63` | 1.243x
`r<alg_type::std_fn, std::uint32_t>/95` | 1.343x
`r<alg_type::std_fn, std::uint32_t>/127` | 1.378x
`r<alg_type::std_fn, std::uint32_t>/255` | 1.488x
`r<alg_type::std_fn, std::uint32_t>/511` | 1.525x
`r<alg_type::std_fn, std::uint32_t>/1023` | 1.61x
`r<alg_type::std_fn, std::uint32_t>/2047` | 1.607x
`r<alg_type::std_fn, std::uint32_t>/4095` | 1.6x
`r<alg_type::std_fn, std::uint32_t>/3230` | 1.608x
`r<alg_type::std_fn, std::uint64_t>/7` | 1.007x
`r<alg_type::std_fn, std::uint64_t>/15` | 1.002x
`r<alg_type::std_fn, std::uint64_t>/31` | 1.019x
`r<alg_type::std_fn, std::uint64_t>/47` | 1.017x
`r<alg_type::std_fn, std::uint64_t>/63` | 1.129x
`r<alg_type::std_fn, std::uint64_t>/95` | 1.197x
`r<alg_type::std_fn, std::uint64_t>/127` | 1.174x
`r<alg_type::std_fn, std::uint64_t>/255` | 1.216x
`r<alg_type::std_fn, std::uint64_t>/511` | 1.186x
`r<alg_type::std_fn, std::uint64_t>/1023` | 1.209x
`r<alg_type::std_fn, std::uint64_t>/2047` | 1.209x
`r<alg_type::std_fn, std::uint64_t>/4095` | 1.312x
`r<alg_type::std_fn, std::uint64_t>/3230` | 1.307x
`r<alg_type::rng, std::uint32_t>/7` | 1.007x
[...]

PR #6420, opened 2026-08-27
Add SVE implementations of `bitset_meow_string`

hazzlim:
This PR adds SVE implementations of `bitset_meow_string`.

As discussed on the STL discord, the ACLE does not expose explicit predicate load/store intrinsics, so this implementation uses explicit casts to/from `svbool_t*`.

### Neoverse N2 (Cobalt 100)

benchmark | speedup
--- | ---
`BM_bitset_to_string_large_single<512, char>` |	1.016
`BM_bitset_to_string_large_single<2048, char>` |	1.254
`BM_bitset_to_string_large_single<512, wchar_t>` |	1.152
`BM_bitset_to_string_large_single<2048, wchar_t>` |	1.433
`bitset_from_string<length_type::char_count, 16, char>` |	1.087
`bitset_from_string<length_type::char_count, 36, char>` |	1.163
`bitset_from_string<length_type::char_count, 64, char>` |	1.408
`bitset_from_string<length_type::char_count, 512, char>` |	1.674
`bitset_from_string<length_type::char_count, 2048, char>` |	1.629
`bitset_from_string<length_type::char_count, 16, wchar_t>` |	1.111
`bitset_from_string<length_type::char_count, 36, wchar_t>` |	1.12
`bitset_from_string<length_type::char_count, 64, wchar_t>` |	1.209
`bitset_from_string<length_type::char_count, 512, wchar_t>` |	1.42
`bitset_from_string<length_type::char_count, 2048, wchar_t>` |	1.452
`bitset_from_string<length_type::null_term, 16, char>` |	1.148
`bitset_from_string<length_type::null_term, 36, char>` |	1.241
`bitset_from_string<length_type::null_term, 64, char>` |	1.327
`bitset_from_string<length_type::null_term, 512, char>` |	1.333
`bitset_from_string<length_type::null_term, 2048, char>` |	1.356
`bitset_from_string<length_type::null_term, 16, wchar_t>` |	1.125
`bitset_from_string<length_type::null_term, 36, wchar_t>` |	1.122
`bitset_from_string<length_type::null_term, 64, wchar_t>` |	1.122
`bitset_from_string<length_type::null_term, 512, wchar_t>` |	1.309
`bitset_from_string<length_type::null_term, 2048, wchar_t>` |	1.264
`bitset_from_stream<16, char>` |	1
`bitset_from_stream<36, char>` |	1.025
`bitset_from_stream<64, char>` |	1.045
`bitset_from_stream<512, char>` |	1
`bitset_from_stream<2048, char>` |	1
`bitset_from_stream<16, wchar_t>` |	1.023
`bitset_from_stream<36, wchar_t>` |	1.023
`bitset_from_stream<64, wchar_t>` |	1.022
`bitset_from_stream<512, wchar_t>` |	1
`bitset_from_stream<2048, wchar_t>` |	1.045

### DIFF
--- a/stl/src/vector_algorithms.cpp
+++ b/stl/src/vector_algorithms.cpp
@@ -59,6 +59,12 @@ namespace {
         constexpr auto _Mask_sve = 1ull << _Idx_sve;
         return (__processor_features_0_63 & _Mask_sve) != 0;
     }
+
+    // IMPORTANT: __declspec(noinline) is necessary because any use of SVE intrinsics
+    // will generate an SVE prologue outside of branches like `if (_Use_FEAT_SVE())`.
+    __declspec(noinline) size_t _Sve_vl() noexcept {
+        return svcntb();
+    }
 #endif // ^^^ defined(_M_ARM64) ^^^
 
     size_t _Byte_length(const void* const _First, const void* const _Last) noexcept {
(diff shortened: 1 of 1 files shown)


stl/src/vector_algorithms.cpp
```diff
@@ -11577,6 +11583,67 @@ namespace {
             _Traits::_Exit_vectorized(); // TRANSITION, DevCom-10331414
         }
 
+#if defined(_M_ARM64) // not ARM64EC, which lacks SVE
+        __forceinline svbool_t _Load_predicate(const void* const _Src) noexcept {
+            return *reinterpret_cast<const svbool_t*>(_Src);
+        }
+
+        __declspec(noinline) void __stdcall _Impl_sve_1(char* const _Dest, const void* _Src, const size_t _Size_bits,
```
StephanTLavavej: We should comment every occurrence of `__declspec(noinline)`. (I am very scarred from that investigation :joy_cat:)

StephanTLavavej: `__stdcall` is appearing only on the to-string impls, not the from-string. I think we should just drop it from ARM64-only internal helpers, as it has no effect for ARM64.

stl/src/vector_algorithms.cpp
```diff
@@ -11602,6 +11669,12 @@ __declspec(noalias) void __stdcall __std_bitset_to_string_1(
     char* const _Dest, const void* const _Src, const size_t _Size_bits, const char _Elem0, const char _Elem1) noexcept {
     using namespace _Bitset_to_string;
 #if defined(_M_ARM64) || defined(_M_ARM64EC)
+#if defined(_M_ARM64) // not ARM64EC, which lacks SVE
```
StephanTLavavej, line 11676: As in #6407, I believe it would be slightly clearer to avoid preprocessor nesting for these SVE early returns.

stl/src/vector_algorithms.cpp
```diff
@@ -11888,6 +11967,111 @@ namespace {
...
+            const size_t _Size_bits, const size_t _Size_chars, const wchar_t _Elem0, const wchar_t _Elem1) noexcept {
+            const size_t _Size_convert = (_Size_chars <= _Size_bits) ? _Size_chars : _Size_bits;
+            const size_t _Step_in      = svcntb();
+            const size_t _Step_out     = _Step_in / 8;
+            const size_t _Half_step    = _Step_in / 2;
+            const wchar_t* _Src_end    = _Src + _Size_convert;
+            uint8_t* _Dest_bytes       = static_cast<uint8_t*>(_Dest);
+            uint8_t* const _Dest_end   = _Dest_bytes + _Size_bytes;
+
+            const auto _Pg  = svptrue_b16();
```
StephanTLavavej: Why is this called `_Pg` when `_Impl_sve_1` has `const auto _True = svptrue_b8();`? I don't know what `_Pg` means.

Astra noticed that Arm developer docs mention Pg as being a "governing predicate". I think naming this one `_True` would make more sense here.

stl/src/vector_algorithms.cpp
```diff
@@ -11888,6 +11967,111 @@ namespace {
...
+
+                _Dest[_Ix >> 3] |= static_cast<uint8_t>(_Cur == _Elem1) << (_Ix & 0x7);
+            }
+
+            return true;
+        }
+
+        __declspec(noinline) bool _Impl_sve_1(void* const _Dest, const char* const _Src, const size_t _Size_bytes,
+            const size_t _Size_bits, const size_t _Size_chars, const char _Elem0, const char _Elem1) noexcept {
+            const size_t _Size_convert = (_Size_chars <= _Size_bits) ? _Size_chars : _Size_bits;
```
StephanTLavavej: `__std_bitset_from_string_1` is already computing `_Size_convert`, which can't be reused because this has to be `__declspec(noinline)`. We should extract this computation.

stl/src/vector_algorithms.cpp
```diff
@@ -11943,6 +12127,12 @@ __declspec(noalias) bool __stdcall __std_bitset_from_string_1(void* const _Dest,
     using namespace _Bitset_from_string;
 
 #if defined(_M_ARM64) || defined(_M_ARM64EC)
+#if defined(_M_ARM64) // not ARM64EC, which lacks SVE
+    const size_t _Size_convert = (_Size_chars <= _Size_bits) ? _Size_chars : _Size_bits;
+    if (_Use_FEAT_SVE() && _Size_convert >= _Sve_vl()) {
```
StephanTLavavej, line 12100: This SVE threshold looks at `_Size_convert`. I noticed that the AVX2 threshold looked at `_Size_bits`, but after thinking about it, the old code was confused, and your use of `_Size_convert` is correct, so I'll change the old code to match.

stl/src/vector_algorithms.cpp
```diff
@@ -11888,6 +11967,111 @@ namespace {
             return true;
         }
 
+#if defined(_M_ARM64) // not ARM64EC, which lacks SVE
+        __forceinline void _Store_predicate(void* const _Dest, const svbool_t _Pred) noexcept {
+            *reinterpret_cast<svbool_t*>(_Dest) = _Pred;
+        }
+
+        template <class _Elem>
+        bool _Finish_sve(uint8_t* const _Dest, uint8_t* const _Dest_end, const _Elem* const _Src,
```
StephanTLavavej: `_Finish_sve` effectively duplicates `_Fallback`. It took me a while to untangle (the difference is that previously, `_Fallback` worked on completely un-processed input, while `_Finish_sve` works on partially processed input), but I figured out how to decompose `_Fallback` such that we can re-use it while avoiding any extra work for the SVE codepath.

StephanTLavavej:
Thanks! :heart_eyes_cat: I pushed fairly significant refactorings to avoid code duplication, reusing the old `_Fallback` after splitting up its "extra" validation.

I kept the old `_Fallback`'s practice of calling `memset` even when it might be zero-length (we could branch if it helps performance). Also I pushed one behavioral change for x64 (which I'm sure you don't care about :joy_cat:); your SVE threshold inspecting `_Size_convert >= _Sve_vl()` made me realize that the AVX2 threshold should be inspecting `_Size_convert` as well.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
:yarn: :heart_eyes_cat: :thread:
===

PR #6423, opened 2026-08-29
`<regex>`: Unwrap iterators in `match_results` overloads

muellerj3:
When `match_results` are requested by users, the matcher in `<regex>` doesn't actually perform any unwrapping or rewrapping of iterators. Instead, matching is performed using wrapped iterators, unnecessarily degrading performance.

On the other hand, iterators are properly unwrapped if no `match_results` are requested, but at the cost of unnecessarily instantiating `match_results` with the internal unwrapped iterator type.

There had been an attempt to represent wrapped and unwrapped iterators using distinct `_BidIt` and `_It` template parameters of the `_Matcher` class template, but this attempt was dysfunctional and was removed in the recent months. Instead, all of the internal machinery dependent on the unwrapped and wrapped iterator type has been centralized in the `_Copy_captures` member function so that only this member function is now potentially dependent on wrapped and unwrapped iterator type simultaneously.

This PR splits the internal `_Regex_match2` and `_Regex_search3` functions each into two overloads, one returning `match_results` and one that does not. The ones that do not return `match_results` take unwrapped iterators, construct the matcher and then perform a match or search. This avoids an unnecessary instantiation of `match_results`. Those returning `match_results` take wrapped iterators, unwrap them to construct the matcher, perform a match or search and then copy the captures to the `match_results` using the matcher's `_Copy_captures2` member function. Thus, the matcher's iterator template parameter is always an unwrapped iterator type now.

The `_Copy_captures2` member takes a reference to the `match_results` object and a wrapped iterator to the end of the input range. Since the iterators in the `match_results` initially do not point into the input range, we first have to assign them a wrapped iterator pointing to the input range before they can be seeked to the position of the unwrapped iterator. For this purpose, `_Copy_captures2` calls the new `_Assign_wrapped` function that handles this assignment of unwrapped iterators to out-of-range wrapped iterators. (If preferred, we can move this function from `<regex>` next to `_Seek_wrapped` in `<xutility>`.)

The common search logic to the `_Regex_search2` overloads has been moved to a new `_Search` member function of the `_Matcher3` class template. Because this new member function can access and mutate the matcher's internal state, the member functions `_Clearf` and `_Match2` with an iterator argument become unnecessary and the `_Skip` member function can be made private and no longer has to handle a null `_Node` argument (at the cost of renaming it for ABI compatibility).

The added test is a small smoke test checking all overloads of `regex_match` and `regex_search`. It seems we didn't have anything testing that these overloads even compile in a single place. For most overloads, we appear to have relied on some individual tests that really intended to test something else but happened to exert these overloads by chance.

### DIFF
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -2644,7 +2644,7 @@ _OutIt _Format_sed(const match_results<_BidIt, _Alloc>& _Match, _OutIt _Out, _In
 
 template <class _It, class _Elem, class _RxTraits>
 bool _Regex_match2(_It _First, _It _Last, const basic_regex<_Elem, _RxTraits>& _Re,
-    regex_constants::match_flag_type _Flgs) { // try to match regular expression to target text
+    const regex_constants::match_flag_type _Flgs) { // try to match regular expression to target text
     static_assert(_Is_ranges_bidi_iter_v<_It>,
         "regex_match requires bidirectional iterators or stronger. See N5014 [re.alg.match]/1.");
 
@@ -2662,7 +2662,7 @@ bool _Regex_match2(_It _First, _It _Last, const basic_regex<_Elem, _RxTraits>& _
 
 template <class _BidIt, class _Alloc, class _Elem, class _RxTraits>
 bool _Regex_match2(_BidIt _First, _BidIt _Last, match_results<_BidIt, _Alloc>& _Matches,
-    const basic_regex<_Elem, _RxTraits>& _Re, regex_constants::match_flag_type _Flgs) {
+    const basic_regex<_Elem, _RxTraits>& _Re, const regex_constants::match_flag_type _Flgs) {
     // try to match regular expression to target text
     static_assert(_Is_ranges_bidi_iter_v<_BidIt>,
         "regex_match requires bidirectional iterators or stronger. See N5014 [re.alg.match]/1.");
--- a/stl/inc/regex
+++ b/stl/inc/regex
@@ -2122,15 +2122,6 @@ public:
         _Mflags |= _Mf;
     }
 
-    void _Clearf(regex_constants::match_flag_type _Mf) { // clear specified flags
-        _Mflags &= ~_Mf;
-    }
-
-    bool _Match2(_It _Pfirst) { // try to match
-        _Begin = _Pfirst;
-        return _Match2();
-    }
-
     bool _Match2() { // try to match
         _Tgt_state._Cur = _Begin;
         _Frames_count   = 0;
@@ -2139,10 +2130,9 @@ public:
         return _Match_pat(_Start) || _Matched;
     }
 
+    bool _Search();
     template <class _BidIt, class _Alsubmatch>
-    void _Copy_captures(match_results<_BidIt, _Alsubmatch>& _Matches);
-
-    _It _Skip(_It _First, _It _Last, const _Node_base* _Node_arg = nullptr, unsigned int _Recursion_depth = 0U);
+    void _Copy_captures2(match_results<_BidIt, _Alsubmatch>& _Matches, _BidIt _Orig_last);
 
 private:
     long long _Complexity_limit;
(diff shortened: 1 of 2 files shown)


tests/std/tests/VSO_0000000_regex_use/test.cpp
```diff
@@ -2726,6 +2726,174 @@ void test_gh_6359() {
...
 }
 
+void test_gh_6423() {
+    // GH-6423: Unwrap iterators in `match_results` overloads
+    // Test that all regex_match and regex_search overloads still compile and work in principle
+    const regex re("a((bc)*d)e");
+    {
+        cmatch cm;
+        const char* input = "abcbcde";
+        const char* last  = input + char_traits<char>::length(input);
```
StephanTLavavej: These could be top-level const.

stl/inc/regex
```diff
@@ -2651,18 +2642,12 @@ _OutIt _Format_sed(const match_results<_BidIt, _Alloc>& _Match, _OutIt _Out, _In
     return _Out;
 }
 
-template <class _BidIt, class _Alloc, class _Elem, class _RxTraits, class _It>
-bool _Regex_match2(const _It _First, const _It _Last, match_results<_BidIt, _Alloc>* const _Matches,
-    const basic_regex<_Elem, _RxTraits>& _Re, const regex_constants::match_flag_type _Flgs) {
-    // try to match regular expression to target text
+template <class _It, class _Elem, class _RxTraits>
+bool _Regex_match2(_It _First, _It _Last, const basic_regex<_Elem, _RxTraits>& _Re,
+    regex_constants::match_flag_type _Flgs) { // try to match regular expression to target text
```
StephanTLavavej: The match overloads are dropping const on the flags which could easily be retained.

stl/inc/regex
```diff
@@ -2672,17 +2657,38 @@ bool _Regex_match2(const _It _First, const _It _Last, match_results<_BidIt, _All
...
+    return _Mx._Match2();
+}
+
+template <class _BidIt, class _Alloc, class _Elem, class _RxTraits>
+bool _Regex_match2(_BidIt _First, _BidIt _Last, match_results<_BidIt, _Alloc>& _Matches,
+    const basic_regex<_Elem, _RxTraits>& _Re, regex_constants::match_flag_type _Flgs) {
+    // try to match regular expression to target text
+    static_assert(_Is_ranges_bidi_iter_v<_BidIt>,
+        "regex_match requires bidirectional iterators or stronger. See N5014 [re.alg.match]/1.");
+    using _It = _Remove_cvref_t<decltype(_STD _Get_unwrapped(_First))>;
```
StephanTLavavej: This alias could be moved down until we need it.

stl/inc/regex
```diff
@@ -2651,18 +2642,12 @@ _OutIt _Format_sed(const match_results<_BidIt, _Alloc>& _Match, _OutIt _Out, _In
     return _Out;
 }
 
-template <class _BidIt, class _Alloc, class _Elem, class _RxTraits, class _It>
-bool _Regex_match2(const _It _First, const _It _Last, match_results<_BidIt, _Alloc>* const _Matches,
-    const basic_regex<_Elem, _RxTraits>& _Re, const regex_constants::match_flag_type _Flgs) {
-    // try to match regular expression to target text
+template <class _It, class _Elem, class _RxTraits>
+bool _Regex_match2(_It _First, _It _Last, const basic_regex<_Elem, _RxTraits>& _Re,
```
StephanTLavavej: The signatures are changing so we don't *need* to rename them for ABI, but given the dramatic change here, I think it would be clearest to revise these function numbers.

For example, #6262 renamed `_Regex_search2` to `_Regex_search3` when it dropped an `_It _Org` parameter.

StephanTLavavej (approved):

StephanTLavavej:

StephanTLavavej:
Thank you again, Regex Lord! :heart_eyes_cat: :crown: :prince:

PR #6425, opened 2026-08-31
`<cmath>`: Support `/Zc:cmath` downlevel

StephanTLavavej:
Followup to #6413. @Codiferous noticed that while MSVC supports `/Zc:cmath` downlevel (i.e. in C++14/17/20 before C++23) and rewrites `<math.h>` to `<__msvc_math.hpp>` which always uses builtins, our `<cmath>` used `/Zc:cmath` builtins only in C++23 mode. In downlevel modes, it fell back to intrinsics, which created conflicts between libc math and UCRT math.

The fix is to generalize `<cmath>` to use builtins when they're available under `/Zc:cmath` regardless of Standard mode. Marking our functions as `constexpr` is still controlled by the Standard mode.

This PR then significantly enhances test coverage to cover downlevel modes. This is the first time that our test suite explicitly acknowledges the difference between `/std:c++23preview` and `/std:c++latest`, which I will be introducing at a larger scale later.


frederick-vs-ja:
Will this resolve the performance issue in #4998?

StephanTLavavej:
Yes, I think so. However, I haven't looked at the codegen.

StephanTLavavej:

davidmrdavid (approved):

davidmrdavid (commented):
Before approving, I have some questions about downstream effects. I'll confirm some thing internally :-]

davidmrdavid (approved):

PR #6430, opened 2026-09-02
Forward declare cmath rounding functions so they can be used as intrinsics

vitknobloch:
To prevent `C2169: intrinsic function, cannot be defined` when using `/Zc:cmath` and its constexpr math additions, we have to forward declare the functions that have intrinsic equivalents and mark them with `#pragma function(name)`. 

This PR does that for `nearbyint`, `rint`, `round`, and `trunc` funtions. Until intrinsic versions exist this is just a forward declaration, compiler would warn about using `#pragma function(non_intrinsic_function)`, but that is disabled with `#pragma warning(disable : 4163)`.

### DIFF
--- a/stl/inc/__msvc_math.hpp
+++ b/stl/inc/__msvc_math.hpp
@@ -287,14 +287,26 @@ _NODISCARD _CONSTEXPR_CMATH26 double __cdecl log(double) noexcept;
 _NODISCARD inline long __cdecl lrintf(float) noexcept;
 _NODISCARD inline long __cdecl lrint(double) noexcept;
 _NODISCARD inline long __cdecl lrintl(long double) noexcept;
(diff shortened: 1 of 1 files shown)


vitknobloch:
@microsoft-github-policy-service agree company="Microsoft"

StephanTLavavej:
Thanks! You should actually link your GitHub account to Microsoft, see https://docs.opensource.microsoft.com/github/opensource/accounts/linking/#linking-your-microsoft-alias-to-your-github-account , but that isn't necessary for me to merge this PR.

StephanTLavavej (approved):

StephanTLavavej:
This is a reverse mirror of MSVC-PR-776391 which I've just approved.

StephanTLavavej:
Thanks for improving the backend's handling of cmath! :heart_eyes_cat:

PR #6435, opened 2026-09-04
`<flat_map>`: Unify `_Emplace_hint()` with `<flat_set>`

### LINKED ISSUE #6069: `<flat_map>`, `<flat_set>`: Unify `_Emplace_hint()`

tycket033-tech:
`<flat_map>` and `<flat_set>` had very different implementations of `_Emplace_hint()`; this PR unifies flat_map's version to follow flat_set's pattern.

The hint is first corrected towards `lower_bound(_Key_val)` (for unique containers) or the closest valid insertion position (for non-unique containers), then the equivalent-element case is handled (overwriting when `_OverwriteIfExists`), and the insertion goes through `_Emplace_exact()` as before. This replaces the previous `weak_ordering` classification logic and removes a duplicated `_IsUnique` special-casing for the overwrite path, as suggested in the issue.

There are no behavior changes.

Fixes #6069.

### DIFF
--- a/stl/inc/flat_map
+++ b/stl/inc/flat_map
@@ -920,56 +920,55 @@ protected:
 
         const const_iterator _Begin = cbegin();
         const const_iterator _End   = cend();
-
-        const weak_ordering _Hint_order = [&] {
-            if constexpr (_IsUnique) {
-                if (_Position == _End || _Compare_keys(_Key_val, *_Position._Key_it)) {
-                    if (_Position == _Begin || _Compare_keys(*(_Position._Key_it - 1), _Key_val)) {
-                        return weak_ordering::equivalent;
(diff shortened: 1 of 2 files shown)


tycket033-tech:
Hi @StephanTLavavej, just following up on my recent PRs. #6435, #6436, and #6437 have all been in Initial Review for about two weeks.

When you have some time, I’d appreciate a review of these. Thanks!

PR #6436, opened 2026-09-04
`<ranges>`: Fix C2382 when mixing `#include <ranges>` with `import std;`

### LINKED ISSUE #6121: `<ranges>`: _Zip_iterator_sentinel_equal redefinition error (C2382) when mixing #include <ranges> with import std;
# Describe the bug
When `#include <ranges>` is used together with `import std;` in the same translation
unit, MSVC emits **C2382** ("redefinition; different exception specifications") on
the internal helper function `std::ranges::_Zip_iterator_sentinel_equal` 
(defined around line 7144 of `<ranges>`).

Using either `#include <ranges>` alone or `import std;` alone compiles successfully.

# Steps to reproduce
```cpp
#include <ranges>

import std;

int main() {
    std::vector<int> a{1, 2, 3};
    std::vector<int> b{4, 5, 6};

    for (const auto& [x, y] : std::views::zip(a, b)) {
        std::println("{}, {}", x, y);
    }
}
```

# Expected behavior
The program compiles and prints:

```
1, 4
2, 5
3, 6
```

# Actual behavior
```
ranges(7144,31): error C2382: "std::ranges::_Zip_iterator_sentinel_equal":
    redefinition; different exception specifications
```

# Analysis

The `noexcept` specification of `_Zip_iterator_sentinel_equal` contains a 
complex fold expression:

```cpp
noexcept((noexcept(_STD declval<const _LHSTupleTypes&>()
    == _STD declval<const _RHSTupleTypes&>()) && ...))
```

When the compiler encounters both a textually-included declaration (from
`#include <ranges>`) and a module-imported declaration (from `import std;`),
it must verify that the two declarations are equivalent. The equivalence
check appears to fail on this nested `noexcept(fold-expression)` — the 
compiler builds two independent AST trees and cannot confirm they represent
the same exception specification.

The same `noexcept` expression also appears on the internal lambda inside
the function body(but this does not cause any error).

# A way to fix

Extract the `noexcept` fold expression into a named variable template. 
This way, the function signature only references a simple name + template
arguments, which the compiler can trivially match across the header and
module paths.

```cpp
// Add before _Zip_iterator_sentinel_equal:

template <class _LHSTuple, class _RHSTuple>
inline constexpr bool _Zip_sentinel_equal_is_nothrow = false;

template <class... _LHSTupleTypes, class... _RHSTupleTypes>
inline constexpr bool _Zip_sentinel_equal_is_nothrow<
    tuple<_LHSTupleTypes...>, tuple<_RHSTupleTypes...>> =
    (noexcept(_STD declval<const _LHSTupleTypes&>()
        == _STD declval<const _RHSTupleTypes&>()) && ...);

// Then update the function:

template <class... _LHSTupleTypes, class... _RHSTupleTypes>
    requires (sizeof...(_LHSTupleTypes) == sizeof...(_RHSTupleTypes))
_NODISCARD constexpr bool _Zip_iterator_sentinel_equal(
    const tuple<_LHSTupleTypes...>& _Lhs_tuple,
    const tuple<_RHSTupleTypes...>& _Rhs_tuple)
    noexcept(_Zip_sentinel_equal_is_nothrow<
        tuple<_LHSTupleTypes...>, tuple<_RHSTupleTypes...>>) {

    const auto _Evaluate_equality_closure =
        [&_Lhs_tuple, &_Rhs_tuple]<size_t... _Indices>(
            index_sequence<_Indices...>)
            noexcept(_Zip_sentinel_equal_is_nothrow<
                tuple<_LHSTupleTypes...>, tuple<_RHSTupleTypes...>>) {
            return ((_STD get<_Indices>(_Lhs_tuple)
                == _STD get<_Indices>(_Rhs_tuple)) || ... || false);
        };

    return _Evaluate_equality_closure(
        index_sequence_for<_LHSTupleTypes...>{});
}
```

# Additional context
I understand that mixing `#include` and `import std;` is generally
discouraged. However, this scenario arises naturally when a third-party
library `#include`s standard headers internally, and the consuming
translation unit uses `import std;`(I think nobody wants to `#define _RANGES_`). It would be helpful if the STL
headers were resilient to this usage pattern where feasible.

# STL version

- MSVC Compiler version: 19.50.35725
- _MSVC_STL_UPDATE: 202508
```

### PR DESCRIPTION
**Pull request by tycket033-tech**
Fixes #6121

`_Zip_iterator_sentinel_equal` has a fold expression in its
`noexcept`-specifier. When a translation unit both `#include <ranges>`
and does `import std;`, the compiler fails to match the two declarations
as equivalent (C2382: redefinition; different exception specifications).

This extracts the fold expression into a variable template
`_Zip_iterator_sentinel_equal_is_nothrow`; both declarations then
reference a simple name + template arguments, which is trivially matched.

### DIFF
--- a/stl/inc/ranges
+++ b/stl/inc/ranges
@@ -7140,14 +7140,23 @@ namespace ranges {
         return _Get_smallest_distance_closure(index_sequence_for<_LHSTupleTypes...>{});
     }
 
+    // Work around C2382 when mixing #include <ranges> with import std;: a noexcept fold expression prevents
+    // the compiler from matching the textual and module declarations, while a variable template is trivially matched.
+    template <class _LHSTupleType, class _RHSTupleType>
+    inline constexpr bool _Zip_iterator_sentinel_equal_is_nothrow = false;
+
+    template <class... _LHSTupleTypes, class... _RHSTupleTypes>
+    inline constexpr bool _Zip_iterator_sentinel_equal_is_nothrow<tuple<_LHSTupleTypes...>, tuple<_RHSTupleTypes...>> =
+        (noexcept(_STD declval<const _LHSTupleTypes&>() == _STD declval<const _RHSTupleTypes&>()) && ...);
+
     template <class... _LHSTupleTypes, class... _RHSTupleTypes>
         requires (sizeof...(_LHSTupleTypes) == sizeof...(_RHSTupleTypes))
     _NODISCARD constexpr bool _Zip_iterator_sentinel_equal(
         const tuple<_LHSTupleTypes...>& _Lhs_tuple, const tuple<_RHSTupleTypes...>& _Rhs_tuple)
-        noexcept((noexcept(_STD declval<const _LHSTupleTypes&>() == _STD declval<const _RHSTupleTypes&>()) && ...)) {
+        noexcept(_Zip_iterator_sentinel_equal_is_nothrow<tuple<_LHSTupleTypes...>, tuple<_RHSTupleTypes...>>) {
         const auto _Evaluate_equality_closure =
             [&_Lhs_tuple, &_Rhs_tuple]<size_t... _Indices>(index_sequence<_Indices...>) noexcept(
-                (noexcept(_STD declval<const _LHSTupleTypes&>() == _STD declval<const _RHSTupleTypes&>()) &&...)) {
+                _Zip_iterator_sentinel_equal_is_nothrow<tuple<_LHSTupleTypes...>, tuple<_RHSTupleTypes...>>) {
                 return ((_STD get<_Indices>(_Lhs_tuple) == _STD get<_Indices>(_Rhs_tuple)) || ... || false);
             };
 

### CODE REVIEW

File: stl/inc/ranges [hunk 1]
```diff
@@ -7140,14 +7140,23 @@ namespace ranges {
         return _Get_smallest_distance_closure(index_sequence_for<_LHSTupleTypes...>{});
     }
 
+    // Work around C2382 when mixing #include <ranges> with import std;: a noexcept fold expression prevents
+    // the compiler from matching the textual and module declarations, while a variable template is trivially matched.
+    template <class _LHSTupleType, class _RHSTupleType>
+    inline constexpr bool _Zip_iterator_sentinel_equal_is_nothrow = false;
```
StephanTLavavej: `constexpr` variable templates don't need to be marked `inline` anymore, and we now avoid doing so. Occurs below.

stl/inc/ranges
```diff
@@ -7140,14 +7140,23 @@ namespace ranges {
         return _Get_smallest_distance_closure(index_sequence_for<_LHSTupleTypes...>{});
     }
 
+    // Work around C2382 when mixing #include <ranges> with import std;: a noexcept fold expression prevents
```
StephanTLavavej: When working around a compiler bug, we must actually report it to the compiler team, then cite the reported bug number (e.g. `TRANSITION, DevCom-nnnnnn`). This gives the compiler team a chance to fix their bugs and for us to ultimately remove our workarounds, which is better than accumulating workarounds forever.

Ideally, this should be reported with a self-contained, library-free repro, focusing on just the fold expression that confuses the compiler.

StephanTLavavej (changes requested):

PR #6437, opened 2026-09-04
Implement LWG-4293: span::subspan/first/last chooses wrong constructor when T is const-qualified bool

### LINKED ISSUE #6239: LWG-4293 `span::subspan`/`first`/`last` chooses wrong constructor when `T` is `const`-qualified `bool`
LWG-4293 `span::subspan`/`first`/`last` chooses wrong constructor when `T` is `const`-qualified `bool`

This LWG issue resolution (accepted at the 2025-11 meeting) has no functional effect after WG21-P2447R6 "Constructing `span<const T>` From `initializer_list<T>`" was reverted by WG21-P4144R1 "Remove `span`'s `initializer_list` Constructor" at the 2026-03 meeting (see #4173).

However, we have a general principle of "Do What The Standard Says", which should be worth following here.

tycket033-tech:
Fixes #6239

LWG-4293 changes the Effects of `span::first`/`last`/`subspan` to
direct-initialize the returned span (`(ptr, count)`) instead of brace-initializing,
so an `initializer_list` constructor can never hijack these calls.

Since P2447R6 was reverted and MSVC STL never implemented it anyway, this has
zero functional impact — just a wording-sync per "Do What The Standard Says".

### DIFF
--- a/stl/inc/span
+++ b/stl/inc/span
@@ -364,8 +364,6 @@ public:
     }
 
     // [span.sub] Subviews
-    // LWG-4293: The returned spans are directly-initialized, so an initializer_list
-    // constructor can never hijack these calls for const-qualified bool element types.
     template <size_t _Count>
     _NODISCARD constexpr auto first() const noexcept /* strengthened */ {
         if constexpr (_Extent != dynamic_extent) {
--- a/stl/inc/span
+++ b/stl/inc/span
@@ -364,6 +364,8 @@ public:
     }
 
     // [span.sub] Subviews
+    // LWG-4293: The returned spans are directly-initialized, so an initializer_list
+    // constructor can never hijack these calls for const-qualified bool element types.
     template <size_t _Count>
     _NODISCARD constexpr auto first() const noexcept /* strengthened */ {
         if constexpr (_Extent != dynamic_extent) {
@@ -375,7 +377,7 @@ public:
         }
 #endif // _MSVC_STL_HARDENING_SPAN || _ITERATOR_DEBUG_LEVEL != 0
 
-        return span<element_type, _Count>{_Mydata, _Count};
+        return span<element_type, _Count>(_Mydata, _Count);
     }
 
     _NODISCARD constexpr auto first(const size_type _Count) const noexcept /* strengthened */ {
@@ -383,7 +385,7 @@ public:
         _STL_VERIFY(_Count <= _Mysize, "Count out of range in span::first(count)");
 #endif
 
-        return span<element_type, dynamic_extent>{_Mydata, _Count};
+        return span<element_type, dynamic_extent>(_Mydata, _Count);
     }
 
     template <size_t _Count>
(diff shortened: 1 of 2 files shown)


stl/inc/span
```diff
@@ -364,6 +364,8 @@ public:
     }
 
     // [span.sub] Subviews
+    // LWG-4293: The returned spans are directly-initialized, so an initializer_list
+    // constructor can never hijack these calls for const-qualified bool element types.
```
frederick-vs-ja: Would you mend to remove these comments? We conventionally don't keep comments for an LWG issue when it's adopted and implemented.

```suggestion
```

tests/std/tests/P0122R7_span/test.cpp
```diff
@@ -996,6 +996,66 @@ constexpr bool test() {
...
+
+        const span<const bool> sp_dyn(sequence);
+        const span<const bool, 9> sp_nine(sequence);
+
+        auto first_1 = sp_dyn.first(4);
+        auto first_2 = sp_nine.first<4>();
+        static_assert(noexcept(sp_dyn.first(4))); // strengthened
+        static_assert(noexcept(sp_nine.first<4>())); // strengthened
+        static_assert(is_same_v<decltype(first_1), span<const bool>>);
+        static_assert(is_same_v<decltype(first_2), span<const bool, 4>>);
```
frederick-vs-ja: Instead of `static_assert(is_same_v<decltype(var), T>);`, I think it's a bit cleaner to use `same_as<T> decltype(auto) var = ...;`.

Although I don't think we need to add test cases.

tycket033-tech: Done — the type assertions now use `same_as<T> decltype(auto) var = ...;`.

I ended up keeping the test block: it pins down the exact return types and
values of the subviews for `span<const bool>`, which seemed worth locking in
even though the change itself is behavior-neutral. Happy to drop it if you'd
rather keep the PR minimal, though.

stl/inc/span
tycket033-tech: Removed, thanks for the pointer!

tests/std/tests/P0122R7_span/test.cpp
```diff
@@ -996,6 +997,56 @@ constexpr bool test() {
...
+        same_as<span<const bool>> decltype(auto) last_1    = sp_dyn.last(4);
+        same_as<span<const bool, 4>> decltype(auto) last_2 = sp_nine.last<4>();
+        static_assert(noexcept(sp_dyn.last(4))); // strengthened
+        static_assert(noexcept(sp_nine.last<4>())); // strengthened
+        assert(last_1.data() == begin(sequence) + 5);
+        assert(last_2.data() == begin(sequence) + 5);
+        assert(last_1.size() == 4);
+        assert(last_2.size() == 4);
+        assert(last_1[0] == false);
+        assert(last_2[3] == true);
```
StephanTLavavej: Nitpick: It's inconsistent to examine different elements here.

StephanTLavavej (approved):

PR #6446, opened 2026-09-10
`<__msvc_math.hpp>`: Restore and test `_USE_MATH_DEFINES` support

StephanTLavavej:
Followup to #6413.

While validating `/Zc:cmath` (which enables LLVM libc generally, and C++23/26 `constexpr <cmath>` specifically) against our Real World Code test suite of popular open-source projects, @Codiferous found that usage of our non-Standard math macros is extremely widespread, and this is a major blocking issue for `/Zc:cmath` adoption.

As a general rule we still want to avoid providing non-Standard legacy math machinery in our new replacement header, but in this case we need to compromise for the sake of practicality. The good news is that this stuff is opt-in via an `_Ugly` macro, so users won't get it if they don't specifically ask for it.

### DIFF
--- a/tests/std/tests/P0533R9_constexpr_for_cmath_and_cstdlib/test.cpp
+++ b/tests/std/tests/P0533R9_constexpr_for_cmath_and_cstdlib/test.cpp
@@ -690,3 +690,9 @@ int main() {
 #endif // ^^^ _HAS_CXX26 && defined(_MSVC_LIBC_MATH) ^^^
 #endif // ^^^ defined(__cpp_lib_constexpr_cmath) ^^^
 }
+
+#ifdef _USE_MATH_DEFINES
+static_assert(static_cast<int>(M_PI * 100.0) == 314, "Unexpected value for M_PI, reboot universe and try again.");
+#elif defined(M_PI)
+static_assert(false, "When _USE_MATH_DEFINES is not defined, M_PI should not be defined.");
+#endif // ^^^ defined(M_PI) ^^^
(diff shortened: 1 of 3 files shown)


StephanTLavavej:

tests/std/tests/P0533R9_constexpr_for_cmath_and_cstdlib/test.cpp
```diff
@@ -690,3 +690,9 @@ int main() {
 #endif // ^^^ _HAS_CXX26 && defined(_MSVC_LIBC_MATH) ^^^
 #endif // ^^^ defined(__cpp_lib_constexpr_cmath) ^^^
 }
+
+#ifdef _USE_MATH_DEFINES
+static_assert(static_cast<int>(M_PI * 100.0) == 314, "Unexpected value for M_PI, reboot universe and try again.");
+#elif defined(M_PI)
```
davidmrdavid, line 696: for my own learning: is there a reason why we didn't add the usual:
```suggestion
#elif defined(M_PI) // ^^^ defined(_USE_MATH_DEFINES)
```

? Perhaps it's just that the diff is so small that it doesn't really help the reader that much.

davidmrdavid (approved):

tests/std/tests/P0533R9_constexpr_for_cmath_and_cstdlib/test.cpp
StephanTLavavej: Yeah, if the regions are very short (1-2 lines) and not nested, we can omit them.

Issue #6156, opened 2026-03-12
`P0088R3_variant`: Investigate possible Clang bugs


StephanTLavavej:
In #4713, I disabled these parts of `P0088R3_variant`, but never investigated:

https://github.com/microsoft/STL/blob/26b7ca58ee3151c81736088e554e5797cafca641/tests/std/tests/P0088R3_variant/test.cpp#L1477-L1489
https://github.com/microsoft/STL/blob/26b7ca58ee3151c81736088e554e5797cafca641/tests/std/tests/P0088R3_variant/test.cpp#L4781-L4788

I commented that these appeared to be mirrored in `libcxx/expected_results.txt`, which are also unsolved mysteries:

https://github.com/microsoft/STL/blob/26b7ca58ee3151c81736088e554e5797cafca641/tests/libcxx/expected_results.txt#L843-L844
https://github.com/microsoft/STL/blob/26b7ca58ee3151c81736088e554e5797cafca641/tests/libcxx/expected_results.txt#L1058-L1059

frederick-vs-ja:
For the `HasFullOrdering` case, I think Clang is emitting correct (but undesired) warnings. `HasFullOrdering::operator<=>` is actually inline but not defined, and it makes Clang warn in the `noexcept` specification of our `operator<=>` for `variant`. The `constexpr` also has some effects for Clang in instantiation.

We reuse `compare_three_way`, which has deduced return type and triggers more instantiations, for that `operator<=>`. Contrarily, libc++ uses an internal lambda with non-deduced trailing return type, which makes less things instantiated. I guess it's a possible improvement to instantiate a bit less things by not reusing `compare_three_way`.

Also, I think `constexpr` should be dropped from these undefined operators. The warning also disappears with `constexpr` dropped.

----

For the `BoomOnAnything`, the error disappears with `constexpr` dropped, which seems to be due to the side effect of `constexpr` function and brace-init-list, see also WG21-P0859R0. If there's a bug around `constexpr`, it might be tracked in LLVM-93364 - but Clang seems to be correct on this.

WG21-P3606R1 is possibly related. Clang seemingly performs overload resolution in copy/move construction in a more annoying (but possibly conforming) way.

We may also consider adding some workaround like llvm/llvm-project@194b337f3327691ee88007061c71225c934b3af0.

Edit: WG21-P3146 also seems related.

----

In all, it seems to me that
- Clang isn't really buggy for these cases, and
[...]

PR #6447, opened 2026-09-11
`<variant>`: Use a simpler three-way comparator for `variant`

frederick-vs-ja:
Towards #6156.

Some occurrences of `constexpr` in the related libcxx test are bogus, and I opened LLVM-222837 to remove them. But it's revealed that MSVC STL's use of `compare_three_way` (using deduced return type in its `operator()`) can cause more things to be instantiated and detected. It seems better to reduce instantiations a bit (at least in the exception specification).

This PR introduces a simplified version of `compare_three_way` using explicit return type, which should be able to stop undesired instantiation and reduce constraint checks.

----

Now LLVM-222837 has been merged, so this PR also mirrors it to MSVC STL's test file. I think it's still valuable to use the new simpler comparator even when there's no bogus `constexpr` or `inline`.

### DIFF
--- a/stl/inc/variant
+++ b/stl/inc/variant
@@ -1376,15 +1376,26 @@ _NODISCARD constexpr bool operator>=(const variant<_Types...>& _Left, const vari
 }
 
 #if _HAS_CXX20
+// This is equivalent to `compare_three_way`,
+// except for no support for non-const comparison, simpler constraints, and explicit return type.
+struct _Simpler_compare_three_way {
+    template <class _Ty1, class _Ty2>
+    _NODISCARD static constexpr auto operator()(const _Ty1& _Left, const _Ty2& _Right) //
+        noexcept(noexcept(_Left <=> _Right)) //
+        /*-*/ -> decltype(_Left <=> _Right) {
+        /*------*/ return _Left <=> _Right;
+    }
+};
+
 _EXPORT_STD template <class... _Types>
     requires (three_way_comparable<_Types> && ...)
 _NODISCARD constexpr common_comparison_category_t<compare_three_way_result_t<_Types>...> operator<=>(
     const variant<_Types...>& _Left, const variant<_Types...>& _Right)
     noexcept(conjunction_v<is_nothrow_invocable_r<common_comparison_category_t<compare_three_way_result_t<_Types>...>,
-            compare_three_way, const _Types&, const _Types&>...>) /* strengthened */ {
+            _Simpler_compare_three_way, const _Types&, const _Types&>...>) /* strengthened */ {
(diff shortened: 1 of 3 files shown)


tests/std/tests/P0088R3_variant/test.cpp
```diff
@@ -1474,20 +1474,11 @@ struct HasOnlySpaceship {
...
 };
 
-#ifdef __clang__ // TRANSITION, not yet investigated
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wundefined-inline"
-#endif // ^^^ workaround ^^^
-
 struct HasFullOrdering {
   constexpr bool operator==(const HasFullOrdering&) const;
   constexpr std::weak_ordering operator<=>(const HasFullOrdering&) const;
```
frederick-vs-ja: If LLVM-222837 is merged soon enough, I think I (or someone else?) should also remove mirrored bogus `constexpr` here in this PR. But I'd like to keep `_Simpler_compare_three_way` even if there's no bogus `constexpr` or `inline`.

stl/inc/variant
```diff
@@ -1376,15 +1376,26 @@ _NODISCARD constexpr bool operator>=(const variant<_Types...>& _Left, const vari
...
 
 #if _HAS_CXX20
+// This is equivalent to `compare_three_way`,
+// except for no support for non-const comparison, simpler constraints, and explicit return type.
+struct _Simpler_compare_three_way {
+    template <class _Ty1, class _Ty2>
+    _NODISCARD static constexpr auto operator()(const _Ty1& _Left, const _Ty2& _Right) //
+        noexcept(noexcept(_Left <=> _Right)) //
+        /*-*/ -> decltype(_Left <=> _Right) {
+        /*------*/ return _Left <=> _Right;
```
StephanTLavavej: I think this will be sufficiently clear without going through heroics to get clang-format to produce this wrapping.

stl/inc/variant
```diff
@@ -1376,15 +1376,26 @@ _NODISCARD constexpr bool operator>=(const variant<_Types...>& _Left, const vari
 }
 
 #if _HAS_CXX20
+// This is equivalent to `compare_three_way`,
+// except for no support for non-const comparison, simpler constraints, and explicit return type.
```
StephanTLavavej: The initial negation here is slightly confusing ("no X, Y, and Z" can be parsed as "no X, no Y, and no Z"). Reordering this would be clearer.

StephanTLavavej (approved):

StephanTLavavej:
Thank you, this is awesome! :heart_eyes_cat: I love getting rid of unexplained weirdness in the test suite. Extra thanks for getting upstream fixed up!

PR #6448, opened 2026-09-13
Unicode 18 UCD tables

clin1234:
Regenerate UCD tables for `<format>` and Grapheme Cluster breaks

Getting closer to 18's final release on the 16th: https://www.unicode.org/versions/beta-18.0.0.html

Claude was only used to discover that the Chisoi block was unassigned; all subsequent effort consisted of manually pulling updated UCD, regenerating relevant files, and manually-inspecting them for correctness.


amlel-el-mahrouss:
https://github.com/microsoft/STL/pull/6448/changes#diff-22a9ee8d91747677ab8a2f6b7f20c44793c0955c1c0de2952578ac401112ed94R155-R631
Why is this change happening?

I'm not sure to follow?

clin1234:
> https://github.com/microsoft/STL/pull/6448/changes#diff-22a9ee8d91747677ab8a2f6b7f20c44793c0955c1c0de2952578ac401112ed94R155-R631 Why is this change happening?
> 
> I'm not sure to follow?

This regenerates grapheme-related properties to match UCDs corresponding to the soon-to-be-official Unicode 18.0.0 standard

amlel-el-mahrouss:
Okay, that explains the remapping

amlel-el-mahrouss (commented):
LGTM

tools/unicode_properties_parse/grapheme_break_test_data_gen.py
```diff
@@ -92,7 +92,7 @@ def line_to_cpp_data_line_utf8(line: BreakTestItem) -> str:
 
 """
 Generate test data from "GraphemeBreakTest.txt"
-This file can be downloaded from: https://www.unicode.org/Public/UCD/latest/ucd/auxiliary/GraphemeBreakTest.txt
+This file can be downloaded from: https://www.unicode.org/Public/18.0.0/ucd/auxiliary/GraphemeBreakTest.txt
```
frederick-vs-ja, line 95: Do we want to refer to a fixed version instead of latest? Ditto below.

clin1234: I think it's more preferable to point to the specific version corresponding to the generated files

clin1234:
@StephanTLavavej Unicode 18 is officially out: https://blog.unicode.org/2026/09/announcing-unicode-standard-version-180.html

tests/std/tests/P0645R10_text_formatting_grapheme_clusterization/test.cpp
```diff
@@ -12,22 +12,19 @@
 
 using namespace std;
 
-// Generator: tools/unicode_properties_parse/grapheme_break_test_data_gen.py
-// Beginning of generated data - DO NOT EDIT manually!
```
StephanTLavavej, line 16: These comments about data being generated are **extremely important** and must not be dropped.

The "// End of generated data - DO NOT EDIT manually!" comment is correctly being kept.

stl/inc/format
```diff
@@ -345,22 +345,20 @@ public:
     }
 };
 
-// Implements a DFA matching the regex of the rule GB9c. The DFA is:
+// Implements a DFA matching the regex on the left side of rule GB9c. The DFA is:
```
StephanTLavavej, line 348: Can you please explain why this state machine is being updated?

tools/unicode_properties_parse/grapheme_break_test_data_gen.py
StephanTLavavej: I think that changing the generator scripts to cite specific versions is undesirable; this adds more places that need to be manually updated when regenerating. The _generated_ data already captures the versioned filenames.

StephanTLavavej (changes requested):

stl/inc/format
clin1234: GB9c has been revised: https://www.unicode.org/reports/tr29/tr29-49.html#GB9c

Now it's defined like this: \p{InCB=Linker} \p{InCB=Extend}* | × | \p{InCB=Consonant}

clin1234: As for only implementing the LHS, for simplicity

stl/inc/format
cpplearner:
```suggestion
                _Left_InCB   = _Right_InCB,
```

stl/inc/format
cpplearner: Presumably:

1. to reflect the change of rule GB9c in Unicode 18, and
2. to match `_GB11_LeftHand_regex` which (as its name suggests) only matches the left part of the rule.

